PR #14, opened 2018-01-11
CI: Fix static checks script for callers

### LINKED ISSUE #13: Fix static checks script
The static checks script is not written in a way that allows calls to use it correctly.

jodh-intel:
The static checks script was incorrectly assuming it was being run in
the tests repository. This "works", but then only runs the checks on the
`tests` repository, *not* those of the caller (which might actually be the
`tests` repository of course). Crucially, callers need to be able to
invoke the static check script without `cd`-ing to the `tests`
repository directory.

Fixes #13.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -10,8 +10,12 @@
 
 set -e
 
+# Since this script is called from another repositories directory,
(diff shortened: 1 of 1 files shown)


chavafg:
lgtm

PR #17, opened 2018-01-11
CI: Add central 'go test' script

### LINKED ISSUE #16: Create a central go test script
Following the pattern of previous PRs, we can create a central `go test` script that can be used by all the repositories.

jodh-intel:
Add a generic 'go test' unit testing script that can be called
by the other repositories.

Fixes #16.

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -1,0 +1,121 @@
+#!/bin/bash -e
+#
+# Copyright (c) 2017-2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Set default test run timeout value.
+#
+# KATA_GO_TEST_TIMEOUT can be set to any value accepted by
+# "go test -timeout X"
+timeout_value=${KATA_GO_TEST_TIMEOUT:-10s}
+
+# KATA_GO_TEST_FLAGS can be set to change the flags passed to "go test".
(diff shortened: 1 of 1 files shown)


jodh-intel:
Script is based on https://github.com/clearcontainers/runtime/blob/master/.ci/go-test.sh.

jodh-intel:
Hi @chavafg, @bergwolf - could you take a look please?

chavafg:
lgtm

.ci/go-test.sh
```diff
@@ -0,0 +1,109 @@
...
+	echo "INFO: Currently running as user '$(id -un)'"
+
+	for pkg in $test_packages; do
+		for user in $users; do
+			printf "INFO: Running 'go test' as %s user on package '%s' with flags '%s'\n" \
+				"$user" "$test_packages" "$go_test_flags"
+
+			run_as_user "$user" go test $go_test_flags -covermode=atomic -coverprofile="$tmp_coverage_file" "$pkg"
+
+			if [ -f "${tmp_coverage_file}" ]; then
```
grahamwhaley: This line feels oddly redundant, but benign, to me - maybe the mix of `tmp` and `test` prefixes is confusing me - won't this file always exist from the line above it? I suspect I'm missing something subtle?, in which case, do we think we can elaborate a touch in the script?

jodh-intel: That's a good question, but the test is in fact required. I've added a comment explaining why (and also caught and fixed a bug in the `printf` call above by doing that).

PR #19, opened 2018-01-12
CI: Only run go linters for golang projects

### LINKED ISSUE #18: Don't call gometalinter unless the calling project contains go code
Currently, `.ci/static-checks.sh` unconditionally calls `gometalinter`. This is actually fine since `gometalinter` is clever enough to detect if there are any go packages to check and it will exit cleanly if not.

However, it's pointless installing `gometalinter` for those calling repositories that don't contain golang code, so add a check to speed up the CI in those scenarios.

jodh-intel:
Stop `.ci/static-checks.sh` from calling `gometalinter` unconditionally.
That tool is clever enough to not error if there are no golang files to
check, but by having the script detect that scenario, the static checks
can been speeded up.

Fixes #18.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -21,6 +21,9 @@ checkcommits \
 	--need-sign-offs \
 	--verbose
 
+go_packages=$(go list ./... 2>/dev/null || true)
+[ -z "$go_packages" ] && exit 0
+
 # Run golang checks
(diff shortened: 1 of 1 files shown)


jodh-intel:
ping @chavafg.

chavafg:
lgtm 👍

PR #21, opened 2018-01-12
CI: Make linters really ignore vendored code

### LINKED ISSUE #20: Really disable checking vendor code with linters
`gometalinter` is not ignoring vendor code since `--vendor` seems to not work as advertised.

jodh-intel:
`gometalinter` has a `--vendor` flag, but it doesn't appear to work so
exclude vendored code instead.

Fixes #20.


chavafg:
lgtm

PR #25, opened 2018-01-12
CI: Support KATA_DEV_MODE to enable lenient linters

### LINKED ISSUE #24: Add developer variable for lenient linters
The static checks script currently only runs a subset of linters under the CI, but all linters by default otherwise.

Add a `KATA_DEV_MODE` variable that, if set, will run the same subset. This is useful and crucially does not require the developer to pretend to be running under a CI (which could be dangerous).

jodh-intel:
The static checks script currently only runs a subset of linters under
the CI, but all linters by default otherwise.

Added a `KATA_DEV_MODE` variable that, if set, will run the same subset.
This is useful and crucially does not require the developer to pretend
to be running under a CI (which could be dangerous).

Fixes #24.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -41,9 +41,9 @@ linter_args+=" --tests"
 # run locally, all linters should be run to allow the developer to review any
 # failures (and potentially decide whether we need to explicitly enable a new
 # linter in the CI).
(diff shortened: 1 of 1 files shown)


chavafg:
lgtm

PR #30, opened 2018-01-19
CI: Disable root tests in dev mode

### LINKED ISSUE #29: Don't run unit tests as root in dev mode
`.ci/go-test.sh` should not run tests as `root` if `KATA_DEV_MODE` is set.

jodh-intel:
Don't run the unit tests as root in developer mode (when
`KATA_DEV_MODE`) is set.

Fixes #29.

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -30,6 +30,12 @@ coverage_file_mode=0644
 # Name of HTML format coverage file
 html_report_file="coverage.html"
 
+warn()
+{
+	local msg="$*"
+	echo >&2 "WARNING: $msg"
+}
+
 # Run a command as either root or the current user (which might still be root).
 #
 # If the first argument is "root", run using sudo, else run as the current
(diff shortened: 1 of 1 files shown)


jodh-intel:
Hi @chavafg, @sameo, @sboeuf, @grahamwhaley, @bergwolf - what do you think?

If we decide that developers really do want to run tests as `root` locally, my view is that they should be required to set an extra variable (`KATA_DEV_MODE_RUN_AS_ROOT=true` or something).

But it should be considered **very** unsafe for a developer to set `CI` in their environment since all scripts should assume if that variable is set that they are running on a real CI system (*so can do whatever they like*).

chavafg:
I agree that developers should not run the tests as root. 
Could it be possible that a developer does not know about `KATA_DEV_MODE` flag and sets `CI=TRUE`? In this case, will he run tests as `root` without noticing? 
If so, we would need to document how the a developer should run the tests.

jodh-intel:
@chavafg - I agree. We need a `README`. I'm happy to write one (either on this PR or another).

jodh-intel:
Hi @chavafg - I'm going to add a `README` on #31 (as the content in master already mentions `KATA_DEV_MODE`).

jodh-intel:
See #32.

chavafg:
lgtm, thanks for the `README`

jodh-intel:
Could we get one more review on this PR please?

.ci/go-test.sh
```diff
@@ -64,16 +70,22 @@ test_coverage()
...
+		warn "As a result, only a subset of tests will be run"
+		warn "(run this script as a non-privileged to ensure all tests are run)."
 	else
-		# Run the unit-tests *twice* (since some must run as root and
-		# others must run as non-root), combining the resulting test
-		# coverage files.
-		users="current root"
+		if [ -n "$KATA_DEV_MODE" ]; then
+			warn "Dangerous to set CI with KATA_DEV_MODE set"
+			warn "Not also running tests as root"
```
grahamwhaley: can we rephrase this line - it doesn't quite make sense in my head....

jodh-intel: Agreed - reworded for improved clarity.

PR #33, opened 2018-01-22
Enabling -race flag depending on arch

basheerk:
changes to modify go_test_flags depending on the arch

Fixes: https://github.com/kata-containers/tests/issues/40
Signed-off-by: Basheer K <basheer@linux.vnet.ibm.com>

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -11,8 +11,12 @@
 # "go test -timeout X"
 timeout_value=${KATA_GO_TEST_TIMEOUT:-10s}
 
+# -race flag is supported only on amd64/x86_64 arch , hence 
+# enabling the flag depending on the arch.
+[ $(go env GOARCH) = "amd64" ] && race="-race"
+
 # KATA_GO_TEST_FLAGS can be set to change the flags passed to "go test".
-go_test_flags=${KATA_GO_TEST_FLAGS:-"-v -race -timeout $timeout_value"}
+go_test_flags=${KATA_GO_TEST_FLAGS:-"-v $race -timeout $timeout_value"}
 
 test_packages=$(go list ./... 2>/dev/null || true)
 


.ci/go-test.sh
```diff
@@ -11,8 +11,12 @@
 # "go test -timeout X"
 timeout_value=${KATA_GO_TEST_TIMEOUT:-10s}
 
+# -race flag is supported only on amd64/x86_64 arch , hence
```
jodh-intel: Out of interest, why is this flag only supported on certain architectures? I wonder if there is an open bug on this we could reference here?

.ci/go-test.sh
```diff
@@ -11,8 +11,12 @@
 # "go test -timeout X"
 timeout_value=${KATA_GO_TEST_TIMEOUT:-10s}
 
+# -race flag is supported only on amd64/x86_64 arch , hence 
+# enabling the flag depending on the arch.
+[ $(uname -m) = "x86_64" -o $(uname -m) = "amd64" ] && race="-race"
```
jodh-intel: Neat. You could make this even simpler I think with:

```
[ $(go env GOARCH) = amd64 ] && race="-race"
```

.ci/go-test.sh
basheerk: I am not aware of the reason why it is not supported on the other arch expect x86_64, but the work is currently going on to support it on the other platforms as well. Refer the below open issue from the golang github

https://github.com/golang/go/issues/19273

jodh-intel: Thanks for the reference!

jodh-intel:
Replaces https://github.com/kata-containers/proxy/pull/35.

Travis is failing due to a missing subsystem (see https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format). Could you:

- Raise an issue for this work.
- Change the commit to something like (changing `XXX` for your issue number):
  ```
  CI: Enable '-race' flag depending on arch.

  Set the `go test` `-race` flag based on the architecture of the host system.
  This indirectly enables support for ppc64le.

  Fixes #XXX.
  ```

basheerk:
@jodh-intel updated the changes and commit

jodh-intel:
Thanks @basheerk!

lgtm

chavafg:
lgtm

GabyCT:
lgtm

PR #71, opened 2018-02-01
ci: Add installation scripts

### LINKED ISSUE #70: ci: Add installation scripts
Install runtime, shim, agent, proxy, qemu, and all dependencies to run kata containers

GabyCT:
This will install runtime, shim, agent, proxy, qemu, and all dependencies
to run kata-containers.

Fixes #70

### DIFF
--- a/.ci/install_kata_kernel.sh
+++ b/.ci/install_kata_kernel.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Currently we will use this repository until this issue is solved
+# See https://github.com/kata-containers/packaging/issues/1
+
+set -e
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+repo_owner="clearcontainers"
+repo_name="linux"
+
+linux_releases_url="https://github.com/${repo_owner}/${repo_name}/releases"
+#fake repository dir to query kernel version from remote
+fake_repo_dir=$(mktemp -t -d kata-kernel.XXXX)
+
+function cleanup {
+	rm  -rf "${fake_repo_dir}"
+}
+
+trap cleanup EXIT
+
+function usage() {
+	cat << EOT
+Usage: $0 <version>
+Install the containers clear kernel image <version> from "${repo_owner}/${repo_name}".
+
+version: Use 'latest' to pull latest kernel or a version from "${cc_linux_releases_url}"
+EOT
+
+	exit 1
+}
+
+#Get latest version by checking remote tags
+#We dont ask to github api directly because force a user to provide a GITHUB token
+function get_latest_version {
+	pushd "${fake_repo_dir}" >> /dev/null
+	git init -q
+	git remote add origin  https://github.com/"${repo_owner}/${repo_name}".git
+
+	cc_release=$(git ls-remote --tags 2>/dev/null \
+		| grep -oP '\-\d+\.container'  \
+			| grep -oP '\d+' \
+				| sort -n | \
+					tail -1 )
+
+	tag=$(git ls-remote --tags 2>/dev/null \
+		| grep -oP "v\d+\.\d+\.\d+\-${cc_release}.container" \
+			| tail -1)
+
+	popd >> /dev/null
+	echo "${tag}"
+}
+
+function download_kernel() {
+	local version="$1"
+	arch=$(arch)
+	[ -n "${version}" ] || die "version not provided"
+	[ "${version}" == "latest" ] && version=$(get_latest_version)
+	echo "version to install ${version}"
+	local binaries_dir="${version}-binaries"
+	local binaries_tarball="${binaries_dir}.tar.gz"
+	local shasum_file="SHA512SUMS"
+	if [ "$arch" = x86_64 ]; then
+		curl -OL "${linux_releases_url}/download/${version}/${binaries_tarball}"
+		curl -OL "${linux_releases_url}/download/${version}/${shasum_file}"
+		sha512sum -c "${shasum_file}"
+		tar xf "${binaries_tarball}"
+	else
+        	die "Unsupported architecture: $arch"
+	fi
+
+	pushd "${binaries_dir}"
+	sudo make install
+	popd
+}
+
+cc_kernel_version="$1"
+
+[ -z "${cc_kernel_version}" ] && usage
+download_kernel "${cc_kernel_version}"
+
+# Make symbolic link to kata-containers
+# FIXME: see https://github.com/kata-containers/packaging/issues/1
+sudo ln -s /usr/share/clear-containers/vmlinux.container /usr/share/kata-containers/
+sudo ln -s /usr/share/clear-containers/vmlinuz.container /usr/share/kata-containers/
--- a/.ci/install_qemu_lite.sh
+++ b/.ci/install_qemu_lite.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+arch=$(arch)
+
+if [ "$#" -ne 3 ]; then
+    echo "Usage: $0 <CLEAR_RELEASE> <QEMU_LITE_VERSION> <DISTRO>"
+    echo "       Install the QEMU_LITE_VERSION from clear CLEAR_RELEASE."
+    exit 1
+fi
+
+clear_release="$1"
+qemu_lite_version="$2"
+distro="$3"
+qemu_lite_bin="qemu-lite-bin-${qemu_lite_version}.${arch}.rpm"
+qemu_lite_data="qemu-lite-data-${qemu_lite_version}.${arch}.rpm"
+
+echo -e "Install qemu-lite ${qemu_lite_version}"
+
+# download packages
+curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/${arch}/os/Packages/${qemu_lite_bin}"
+curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/${arch}/os/Packages/${qemu_lite_data}"
+
+# install packages
+if [ "$distro" == "ubuntu" ];  then
+	sudo alien -i "./${qemu_lite_bin}"
+	sudo alien -i "./${qemu_lite_data}"
+fi
+
+# cleanup
+rm -f "./${qemu_lite_bin}"
+rm -f "./${qemu_lite_data}"
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+cidir=$(dirname "$0")
+
+source "${cidir}/lib.sh"
+
+# Modify the runtimes build-time defaults
+
+# enable verbose build
+export V=1
+
+# tell the runtime build to use sane defaults
+export SYSTEM_BUILD_TYPE=kata
+
+# The runtimes config file should live here
+export SYSCONFDIR=/etc
+
+# Artifacts (kernel + image) live below here
+export SHAREDIR=/usr/share
+
+runtime_config_path="${SYSCONFDIR}/kata-containers/configuration.toml"
+
+PKGDEFAULTSDIR="${SHAREDIR}/defaults/kata-containers"
+NEW_RUNTIME_CONFIG="${PKGDEFAULTSDIR}/configuration.toml"
+# Note: This will also install the config file.
+clone_build_and_install "github.com/kata-containers/runtime"
+
+# Check system supports running Kata Containers
+kata-runtime kata-check
+
+if [ -e "${NEW_RUNTIME_CONFIG}" ]; then
+	# Remove the legacy config file
+	sudo rm -f "${runtime_config_path}"
+
+	# Use the new path
+	runtime_config_path="${NEW_RUNTIME_CONFIG}"
+fi
+
+echo "Enabling all debug options in file ${runtime_config_path}"
+sudo sed -i -e 's/^#\(enable_debug\).*=.*$/\1 = true/g' "${runtime_config_path}"
+
+echo "Add runtime as a new/default Docker runtime. Docker version \"$(docker --version)\" could change according to updates."
+docker_options="-D --add-runtime kata-runtime=/usr/local/bin/kata-runtime"
+
+echo "Add kata-runtime as a new/default Docker runtime."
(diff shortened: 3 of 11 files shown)


.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,62 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+cidir=$(dirname "$0")
+source "/etc/os-release"
+source "${cidir}/lib.sh"
+get_cc_versions
+
+if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
```
jodh-intel: This is `x86_64`-specific, so could you add something like the following around this block:

```
source "${cidir}/lib.sh"

arch=$(arch)

if [ "$arch" = x86_64 ]; then
    ...
else
fi
```

Note that we'll need to define `die` in `lib.sh` too.

.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,62 @@
...
+
+echo "Install Kata Containers Kernel"
+"${cidir}/install_kata_kernel.sh" "latest"
+
+echo -e "Install CRI-O dependencies available for Ubuntu $VERSION_ID"
+sudo -E apt install -y libdevmapper-dev btrfs-tools util-linux
+
+if [ "$VERSION_ID" == "16.04" ]; then
+	echo "Install os-tree"
+	sudo -E add-apt-repository ppa:alexlarsson/flatpak -y
```
jodh-intel: I just checked and this is available for all arches ;)

.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,62 @@
...
+	sudo -E apt update
+fi
+
+sudo -E apt install -y libostree-dev
+
+if ! command -v docker > /dev/null; then
+	"${cidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker install
```
jodh-intel: This tool needs to be imported into kata, but note that a fix for the following is going to be required:

- https://github.com/clearcontainers/tests/issues/881

.ci/install_kata_kernel.sh
```diff
@@ -0,0 +1,79 @@
...
+	local version=$1
+	[ -n "${version}" ] || die "version not provided"
+	if [ "${version}" == "latest" ]; then
+		version=$(get_latest_version)
+	fi
+	echo "version to install ${version}"
+	local binaries_dir="${version}-binaries"
+	local binaries_tarball="${binaries_dir}.tar.gz"
+	local shasum_file="SHA512SUMS"
+	curl -OL "${cc_linux_releases_url}/download/${version}/${binaries_tarball}"
```
jodh-intel: This is architecture-specific and will break if arch != `x86_64`.

@jcvenegas - could you raise a bug and update your tooling so that the binaries file includes `x86_64` in the name please?

.ci/install_qemu_lite.sh
```diff
@@ -0,0 +1,34 @@
...
+    echo "Usage: $0 <CLEAR_RELEASE> <QEMU_LITE_VERSION> <DISTRO>"
+    echo "       Install the QEMU_LITE_VERSION from clear CLEAR_RELEASE."
+    exit 1
+fi
+
+clear_release="$1"
+qemu_lite_version="$2"
+distro="$3"
+qemu_lite_bin="qemu-lite-bin-${qemu_lite_version}.x86_64.rpm"
+qemu_lite_data="qemu-lite-data-${qemu_lite_version}.x86_64.rpm"
```
jodh-intel: Again, use `$arch` here.

.ci/install_qemu_lite.sh
```diff
@@ -0,0 +1,34 @@
...
+
+echo -e "Install qemu-lite ${qemu_lite_version}"
+
+# download packages
+curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/x86_64/os/Packages/${qemu_lite_bin}"
```
jodh-intel: Two more places to use `$arch`.

GabyCT:
@jodh-intel changes applied

jodh-intel:
@jcvenegas, @grahamwhaley - could you take a look at this please?

.ci/install_kata_kernel.sh
```diff
@@ -0,0 +1,79 @@
...
+	tag=$(git ls-remote --tags 2>/dev/null \
+		| grep -oP "v\d+\.\d+\.\d+\-${cc_release}.container" \
+			| tail -1)
+
+	popd >> /dev/null
+	echo "${tag}"
+}
+
+function download_kernel() {
+	local version="$1"
```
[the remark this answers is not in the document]
jodh-intel: Alternatively, make this function check `arch` (or maybe `go env GOARCH`) and download the correct architecture version.

See:

- https://github.com/clearcontainers/packaging/issues/264
- https://github.com/clearcontainers/tests/issues/889

/cc @jcvenegas.

.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+if [ "$VERSION_ID" == "16.04" ]; then
+	echo "Install os-tree"
+	sudo -E add-apt-repository ppa:alexlarsson/flatpak -y
+	sudo -E apt update
+fi
+
+sudo -E apt install -y libostree-dev
+
+if ! command -v docker > /dev/null; then
+	# This will be fixed by https://github.com/clearcontainers/tests/issues/881
```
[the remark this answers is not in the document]
grahamwhaley: So I'll ask the question - should or do we need to have `kata-runtime` as the default runtime?
Most/many of our tests use the `RUNTIME` env varialble to determine which runtime to pass to `--runtime=xxx` on the docker command line - and, if they don't, they should, to make it easier for us to have multiple runtimes installed for checking etc., without having to have them as the default runtime.

I suspect if we change this right now maybe a couple of tests fail - but that's probably the only way we'll find such tests. I propose we *do not* make our runtime the default runtime, and ensure our test frameworks can cope with being told which runtime they should invoke during testing.

FYI, afaik all the metrics tests pick up `RUNTIME` from the env. (but I do suspect there might be one or two we don't run often or under automation that still need updating).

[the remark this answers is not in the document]
jodh-intel: Agreed. And in fact if we rely on `RUNTIME`, we could always run the entire suite of tests multiples times without having to reconfigure Docker/k8s:

```
for runtime in runc kata-runtime-cc kata-runtime-runv
do
    # run tests...
done
```

Any further thoughts on this @chavafg ?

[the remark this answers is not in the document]
chavafg: Agree with this approach.

.ci/install_runtime.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+echo "Enabling global logging for runtime in file ${runtime_config_path}"
+sudo sed -i -e 's/^#\(\[runtime\]\|global_log_path =\)/\1/g' "${runtime_config_path}"
+
+echo "Enabling all debug options in file ${runtime_config_path}"
+sudo sed -i -e 's/^#\(enable_debug\).*=.*$/\1 = true/g' "${runtime_config_path}"
+
+echo "Add runtime as a new/default Docker runtime. Docker version \"$(docker --version)\" could change according to updates."
+docker_options="-D --add-runtime kata-runtime=/usr/local/bin/kata-runtime --default-runtime=kata-runtime"
```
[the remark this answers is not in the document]
chavafg: As discussed here: https://github.com/kata-containers/tests/pull/71#discussion_r166335563 
We can remove the `--default-runtime` flag

.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+arch=$(arch)
+
+if [ "$arch" = x86_64 ]; then
+	if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
+		echo "enable Nested Virtualization"
+		sudo modprobe -r kvm_intel
+		sudo modprobe kvm_intel nested=1
+	fi
+else
+	die "Unsupported architecture: $arch"
```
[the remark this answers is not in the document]
GabyCT: @jodh-intel, I already did it

.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash -e
```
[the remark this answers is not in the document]
GabyCT: @jodh-intel yes that is the idea :)

GabyCT:
@jodh-intel changes applied

.ci/setup_env_ubuntu.sh
GabyCT: @jodh-intel , I removed the comment about this issue

.ci/install_kata_kernel.sh
```diff
@@ -0,0 +1,80 @@
...
+repo_name="linux"
+
+linux_releases_url="https://github.com/${repo_owner}/${repo_name}/releases"
+#fake repository dir to query kernel version from remote
+fake_repo_dir=$(mktemp -t -d kata-kernel.XXXX)
+
+function cleanup {
+	rm  -rf "${fake_repo_dir}"
+}
+trap cleanup EXIT
```
jodh-intel: Nit: missing newline.

.ci/install_kata_kernel.sh
```diff
@@ -0,0 +1,80 @@
...
+function usage() {
+	cat << EOT
+Usage: $0 <version>
+Install the containers clear kernel image <version> from "${repo_owner}"/"${repo_name}".
```
jodh-intel: Nit: You could remove some double-quotes here:

```
... from "${repo_owner}/${repo_name}".
```

.ci/install_kata_kernel.sh
```diff
@@ -0,0 +1,80 @@
...
+
+	exit 1
+}
+
+#Get latest version by checking remote tags
+#We dont ask to github api directly because force a user to provide a GITHUB token
+function get_latest_version {
+	pushd "${fake_repo_dir}" >> /dev/null
+	git init -q
+	git remote add origin  https://github.com/clearcontainers/linux.git
```
jodh-intel: This needs to use `$repo_owner` and `$repo_name`.

.ci/install_kata_kernel.sh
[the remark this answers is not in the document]
jodh-intel: This issue is still outstanding.

.ci/install_runtime.sh
```diff
@@ -0,0 +1,69 @@
...
+config_path="/etc/systemd/system/docker.service.d/"
+sudo mkdir -p ${config_path}
+
+# Check if the system has set http[s] proxy
+if [ ! -z "$http_proxy" ] && [ ! -z "$https_proxy" ] ;then
+	docker_http_proxy="HTTP_PROXY=$http_proxy"
+	docker_https_proxy="HTTPS_PROXY=$https_proxy"
+fi
+
+cat << EOF | sudo tee ${config_path}/clear-containers.conf
```
jodh-intel: s/clear-containers/kata-containers/.

But shouldn't we be using https://github.com/clearcontainers/tests/blob/master/cmd/container-manager/manage_ctr_mgr.sh here instead (in other words import that cmd on this PR too)?

.ci/setup_env_ubuntu.sh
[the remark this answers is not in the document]
jodh-intel: Ack.

.ci/setup_env_ubuntu.sh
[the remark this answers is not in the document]
jodh-intel: Great, thanks.

GabyCT:
@jodh-intel changes applied

.ci/install_kata_kernel.sh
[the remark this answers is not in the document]
GabyCT: @jodh-intel done

.ci/setup_env_ubuntu.sh
jodh-intel: I think this is the last issue then as this won't work currently afaics?

GabyCT:
@jodh-intel I remove the last lines as they did not work (as you mentioned) thanks :)

jodh-intel:
Hi @GabyCT - I think you just need to add the code for `manage_ctr_mgr.sh` and remove the manual configuration you're doing for Docker.

GabyCT:
@jodh-intel the code for `manage_ctr_mgr.sh` should be part of the .ci scripts or should I put it in another place? also I see that this code is doing a configuration for swarm should this be removed?

jodh-intel:
Hi @GabyCT - I think you just need to copy the files in https://github.com/clearcontainers/tests/tree/master/cmd/container-manager into this PR (under the same location [`cmd/container-manager/]). You could do that on a separate commit but as the CI scripts are using the tool, it could be on the single commit you already have I think.

@chavafg - does that sound ok with you?

chavafg:
@GabyCT @jodh-intel yes, sounds good, also we would need to open an issue here for the arm support, and close in cc test repo?

GabyCT:
@chavafg and @jodh-intel changes applied

jodh-intel:
Hmm, pullapprove isn't seeing my ack. Let's try again...

lgtm

.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,64 @@
...
```
chavafg: While testing this PR found that the `-e` didn't had any effect when we call the script using:
 `bash -f ${cidir}/setup_env_ubuntu.sh"` , when errors appeared, the setup continued without failing. 
So I would recommend that we explicitly add  `set -e` to every script.

.ci/setup_env_ubuntu.sh
```diff
@@ -0,0 +1,64 @@
...
+sudo -E apt install -y moreutils
+
+echo "Install kata containers dependencies"
+chronic sudo -E apt install -y libtool automake autotools-dev autoconf bc alien libpixman-1-dev coreutils
+
+echo "Install qemu-lite binary"
+"${cidir}/install_qemu_lite.sh" "${qemu_lite_clear_release}" "${qemu_lite_sha}" "$ID"
+
+echo "Install kata-containers image"
+"${cidir}/install_kata_image.sh"
```
chavafg: When we run this script, we already need docker, and if it is not installed, this will fail. We should install docker using the `manage_ctr_mgr.sh` script before calling `install_kata_image.sh`.

cmd/container-manager/manage_ctr_mgr.sh
```diff
@@ -0,0 +1,321 @@
+#!/bin/bash -e
```
chavafg: Add execution permissions to this script

.ci/setup_env_ubuntu.sh
grahamwhaley: One thing I've considered doing to the metrics scripts is invoking all (sub)scripts with `bash -$- ....`, which then passes on the parent script settings to the child.  That way if you set, say, `-x` on the top level script for debug, it will trickle through to all sub-scripts. Maybe that will work well for your `-e` situation ?

GabyCT:
@chavafg changes applied

PR #95, opened 2018-02-26
CI: Fix proxy log selector in teardown script

### LINKED ISSUE #94: CI: fix proxy log selector in teardown script
`.ci/teardown.sh` currently tries to select on the proxy systemd unit if not asked to copy logs. However, the proxy is no longer a system service so this is invalid.

jodh-intel:
The proxy is no longer a system service so don't select by systemd unit
if no log copy location is passed to `.ci/teardown.sh`.

Fixes #94.


chavafg:
lgtm

PR #103, opened 2018-03-01
Jenkins call static checks

### LINKED ISSUE #102: Update jenkins script to call checkcommits
At some point, we'll be dropping Travis and switching entirely to Jenkins so add back the logic from the Clear Containers jenkins script to call [checkcommits](https://github.com/kata-containers/tests/tree/master/cmd/checkcommits).

jodh-intel:
Ensure Jenkins jobs call the static checks before running setup and the
tests.

Fixes #102.

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -62,8 +62,21 @@ else
 	git fetch origin && git checkout master && git reset --hard origin/master
 fi
 
+# Run the static analysis tools
+.ci/static-checks.sh
+
 # Setup Kata Containers Environment
 .ci/setup.sh
+
+if [ -n "$pr_number" ]
+then
+	# Now that checkcommits has run, move the PR commits into the master
+	# branch before running the tests. Having the commits in "master" is
+	# required to ensure coveralls works.
+	git checkout master
+	git reset --hard "$pr_branch"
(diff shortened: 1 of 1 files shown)


jodh-intel:
What's not immediately obvious about this change is that once it lands the main repos will be calling the static checks *twice*:

- once under Travis.
- once under Jenkins.

Once this lands, I don't see why we can't go around removing the static check call from all the `.travis.yml` files.

wdyt @chavafg?

chavafg:
Well, I think we still need to add the unit tests execution on the `.ci/run.sh` scripts, right? and `codecov.io` also?

chavafg:
lgtm

grahamwhaley:
Right, so, just so we are clear - we now have a single unified static check method (hooray!), as compared to say https://github.com/clearcontainers/runtime/blob/master/Makefile#L506 where we had a slightly different method and call coded into each individual repo...

And thus, we are unifying the single static call into the CI main level script - yes?

jodh-intel:
@grahamwhaley - that's right. Currently the kata components all call that static script themselves (via Travis), but the ideal is for them to just call a single script (the jenkins setup script) and have that call the static checks so that we centralise the CI logic in this repo.

PR #98, opened 2018-02-28
cmd: Add log parser

jodh-intel:
Add a log parsing tool (`kata-log-parser`) that reads multiple logfmt
[*] logfiles and writes their output in time order showing time
differences between log entries.

The tool can also check logs for validity and convert the logs to
various alternative formats.

For full details:

```
$ kata-log-parser --help
```

Fixes #93.

[*] - https://brandur.org/logfmt

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #106, opened 2018-03-02
CI: Set CI variable for Jenkins

### LINKED ISSUE #105: Set CI=true for Jenkins jobs
Now that #103 has been merged, CI jobs are failing as (I believe) the `CI=true` variable is not being set in the environment.

Related: https://github.com/kata-containers/ci/issues/7.

jodh-intel:
The scripts called by the Jenkins script need to know they are running
in a CI environment.

Fixes #105.

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -7,6 +7,9 @@
 
 set -e
 
+# Signify to all scripts that they are running in a CI environment
+[ -z "${KATA_DEV_MODE}" ] && export CI=true
+
 # Need the repo to know which tests to run.
 kata_repo="$1"
 


jodh-intel:
Hi @grahamwhaley - I don't have access to the Jenkins config but I think what must be happening in the config is the scripts are using the `master` version of this repo to start `.ci/jenkins_job_build.sh`. But that will fail since `CI` won't be set in them (yet).

If I'm right, I think we'll need to force-merge this and see if, for example, the CI then passes for #98.

jodh-intel:
The reason I think Jenkins is only considering the `master` branch is that I've run `.ci/jenkins_job_build.sh` locally using this branch, and the CI static tests pass as expected.

grahamwhaley:
@jodh-intel I presume you used the very latest HEAD version of gometalinter and all the other tools - as I think that is what will happen on the Azure CI bots - they are 'clean machines', so always install/run with the HEAD of the gometalinter and related tools afaict.
Those CI errors look like genuine errors, and if you have done this locally already, then a tools version difference could explain it?

grahamwhaley:
If you are stuck, prod me, and I can run locally on the PR branch and see what I get - I have a machine with a pretty recent version of all the tools

grahamwhaley:
@jodh-intel and, yes, the 'using the master branch of the scripts' I think you might be right - I have run into this before - we pull the tests repo to get the jenkins script to run the jenkins - so any mods to the script itself don't show up as the Jenkins shell already has the script open and running, even when we rebase it to the PR branch.

let's see what @chavafg things - I suspect we merge this and then fix if we are wrong. That is what we did last time we updated.

For reference, the Jenkins script is as simple as:
```
set -e

export ghprbPullId
export ghprbTargetBranch

cd $HOME
git clone https://github.com/kata-containers/tests.git
cd tests

.ci/jenkins_job_build.sh "github.com/kata-containers/tests"
```

jodh-intel:
@grahamwhaley - I'm running with the latest version of `gometalinter` here. The point is that `.ci/static-checks.sh` only runs a small set of linters (that we know the code passes with) when `CI` is set. But if you don't have that variable set, it runs all linters. If you look at the `gometalinter` output, it is indicating issues reported by linter tools that should not be running under the `CI`.

jodh-intel:
Personally, I don't see how this PR can "hurt" as that variable really should be set under any CI's ;)

I've updated the branch to add a check so that `CI` is only set if we're not running on a developers system.

grahamwhaley:
nice update. Let's watch the CI's fail, and then as we agreed, force merge so we can check what happens 'in reality'.

jodh-intel:
Ok - it's failed again (as expected) on the static checks.

grahamwhaley:
forced it...

jodh-intel:
Ta.

jodh-intel:
The previous CI run on `master` failed running the static tests:

- http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-tests-ubuntu-16-04-master/13/console

... and the currently running `master` build has got beyond that:

- http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-tests-ubuntu-16-04-master/14/console

... so that proves we're using the wrong branch at Jenkins startup I think.

jodh-intel:
Err, strike that. `master` should be tested with `master`, but the PR CI should be using the PR branches for setting up Jenkins.

grahamwhaley:
Yes, that would be the key - to use the PR branch to execute the `jenkins_job_build.sh` itself - *but*, I will note, afaik this *only* effects changes to the `jenkins_job_build.sh` itself - as once that script is running it will [rebase](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L62) to the PR before [running the tests](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L69).

So, it is only for that short window of 'the top' of `jenkins_job_build.sh` that we see this odd effect.

Issue #126, opened 2018-03-15
CI: Implement depends-on: functionality


chavafg:
Depends-on will allow the CI to test a patch together with another pull request from other `kata-containers` repositories.

An example of why this is needed is that https://github.com/kata-containers/agent/pull/163 needed the changes from https://github.com/kata-containers/runtime/pull/55 to make the CI pass.

PR #127, opened 2018-03-15
CI: Implements Depends-on

chavafg:
Depends-on will allow the CI to test a patch together with
another one from other `kata-containers` repository.
This will be useful when compatibility is broken because
changes need to be done in different repos.

Fixes: #126.

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -59,3 +59,42 @@ function get_cc_versions(){
 	[ ! -f "$_versions_file" ] && { echo >&2 "ERROR: cannot find $_versions_file"; exit 1; }
 	source "$_versions_file"
 }
+
+
+function apply_depends_on() {
+	pushd "${GOPATH}/src/${kata_repo}"
+	label_lines=$(git log --format=%s%b master.. | grep "Depends-on:" || true)
+	if [ "${label_lines}" == "" ]; then
+		popd
+		return 0
+	fi
+
+	nb_lines=$(echo "${label_lines}" | wc -l)
+
+	repos_found=()
+	for i in $(seq 1 "${nb_lines}")
+	do
+		label_line=$(echo "${label_lines}" | sed "${i}q;d")
+		label_str=$(echo "${label_line}" | awk '{print $2}')
+		repo=$(echo "${label_str}" | cut -d'#' -f1)
+		if [[ "${repos_found[@]}" =~ "${repo}" ]]; then
+			echo "Repository $repo was already defined in a 'Depends-on:' tag."
+			echo "Only one repository per tag is allowed."
+			return 1
+		fi
+		repos_found+=("$repo")
+		pr_id=$(echo "${label_str}" | cut -d'#' -f2)
+
+		echo "This PR depends on repository: ${repo} and pull request: ${pr_id}"
+		if [ ! -d "${GOPATH}/src/${repo}" ]; then
+			go get -d "$repo" || true
+		fi
+
+		pushd "${GOPATH}/src/${repo}"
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -56,6 +56,34 @@ else
...
+
+	Fixes: #nnn
+
+	Depends-on:github.com/kata-containers/runtime#999
+
+	Signed-off-by: <contributor@foo.com>
+
+```
+
+In this example, we tell the CI to fetch the pull request 999 from the
```
jodh-intel: I'd be tempted to explain this further. Something like:

> In this example, we tell the CI to fetch pull request 999 from the `runtime` repository and use that rather than the `master` branch when testing the changes contained in this PR.

.ci/lib.sh
```diff
@@ -59,3 +59,34 @@ function get_cc_versions(){
...
+	if [ "${label_lines}" == "" ]; then
+		return 0
+	fi
+
+	nb_lines=$(echo "${label_lines}" | wc -l)
+
+	for i in $(seq 1 "${nb_lines}")
+	do
+		label_line=$(echo "${label_lines}" | sed "${i}q;d")
+		label_str=$(echo "${label_line}" | cut -d' ' -f2)
```
jodh-intel: This line assumes that the commit line starts with `Depends-on:` - any leading whitespace will break that. I'd either document this in the `README.md` or use `awk` instead of `cut` maybe.

I'd actually be tempted to remove `Depends-on:` from the string as soon as you have a match, then remove any leading/trailing whitespace and only then start trying to pick apart the repo+pr elements.

.ci/lib.sh
```diff
@@ -59,3 +59,34 @@ function get_cc_versions(){
 	[ ! -f "$_versions_file" ] && { echo >&2 "ERROR: cannot find $_versions_file"; exit 1; }
 	source "$_versions_file"
 }
+
+
+function apply_depends_on() {
+	pushd "${GOPATH}/src/${kata_repo}"
+	label_lines=$(git log --format=%s%b -n 1 | grep "Depends-on:" || true)
```
jodh-intel: I wonder if we really need this restriction? I've been caught out with this before as my `Depends-on:` wasn't the top commit. Couldn't we just consider all the commits:

```
git log --format=%s%b master..
```

If we do that, we should probably check to ensure we only find a single `Depends-on:` though and error if we find more.

README.md
chavafg: agree, fixed.

.ci/lib.sh
chavafg: Fixed using `awk`

.ci/lib.sh
chavafg: @jodh-intel 
If we consider all commits, we may find `Depends-on:` from older PRs.

For finding a single `Depends-on:` I think there could be a case when we need to add more than one.
e.g.

```
Depends-on: github.com/kata-containers/agent#111
Depends-on: github.com/kata-containers/tests#222
```

jodh-intel: Sure, but if we use that `git log` command above, it will only look at the commits that are in the current (PR) branch but which are not in `master`, so it will be "constrained" to only look at relevant commits.

Yes, I guess if we do add a check, it should just check that there aren't `>1` `Depends-on:` for any one repository.

chavafg: Ohh, true, I tried your command without the dots, sorry about that. Agree, modifying...

chavafg: fixed.

jodh-intel:
lgtm

@klynnrif - could you take a look at the doc changes please?

GabyCT:
lgtm

PR #131, opened 2018-03-16
CI: Enable "set -e" harder

### LINKED ISSUE #130: Use safer "set -e" in go test script
`.ci/go-test.sh` sets `-e` at the top of the script. This isn't safe since the option can be defeated trivially and accidentally by calling the script as:

```
bash .ci/go-test.sh
```

jodh-intel:
Change the shebang `-e` to an explicit `set -e` to avoid the nasty shell
issue whereby the option can be defeated if the script is invoked by
calling it as:

```
  # "set -e" no longer applies!
  bash .ci/go-test.sh
```

Fixes #130.

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -1,9 +1,11 @@
-#!/bin/bash -e
+#!/bin/bash
 #
 # Copyright (c) 2017-2018 Intel Corporation
 #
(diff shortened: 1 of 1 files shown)


sboeuf:
LGTM

GabyCT:
lgtm

PR #153, opened 2018-03-23
CI: move static checks after setup is executed

### LINKED ISSUE #152: CI: move statick-checks.sh to be executed after setup.
Taken from: http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-runtime-ubuntu-16-04-master/24/console 

the static checks are failing because some generated files are not present in the system when the checks are running. So we should move this script to be executed after builing the kata-containers components

chavafg:
running static checks, including gometalinter fails because
it requires generated files to be present in the system.
We need to run .ci/static-checks.sh after the .ci/setup.sh
is executed, since it builds the kata containers components.

Fixes #152.

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -38,7 +38,6 @@ popd
 
 echo "Installing CRI-O"
 make clean
-make install.tools
 make
 make test-binaries
 sudo -E PATH=$PATH sh -c "make install"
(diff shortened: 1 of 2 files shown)


chavafg:
And now I am not sure why maligned is not recognized

devimc:
lgtm

Issue #150, opened 2018-03-22
ci: kernel: should force overwrite of symlinks


grahamwhaley:
At https://github.com/kata-containers/tests/blob/master/.ci/install_kata_kernel.sh#L91-L92
we do:
```
sudo ln -s /usr/share/clear-containers/vmlinux.container /usr/share/kata-containers/
sudo ln -s /usr/share/clear-containers/vmlinuz.container /usr/share/kata-containers/
```

which fails if the files are already there, as they can be on a bare-metal machine. We don't see an issue with the CI at present as it always runs on a clean installed cloud instance.

Change to `ln -sf` to force the over-write.

grahamwhaley:
@chavafg @jodh-intel wdyt?

chavafg:
Agree. Also, image creation has the same issue. 

https://github.com/kata-containers/tests/blob/master/.ci/install_kata_image.sh#L38

grahamwhaley:
right, I'll PR those (as I think they are impacting my ci testing right now ;-) )

jodh-intel:
I guess so. It worries me slightly that we have "stuff" left over on baremetal system as we can't be assured (easily) that we have a fully vanilla environment.

grahamwhaley:
Well, 'left over' is more 'we have stuff installed from a previous run'.
The only ways to ensure we don't would be:
- completely re-install the system each time (for each build) - not very practical
- have a script that removes all evidence that any previous kata was installed or ever ran - that is potentially quite a bit of cleanup
- figure out how to run docker in some sort of sandbox (a bit like [dind](https://github.com/jpetazzo/dind) and --priv --net=host) - but, if we then nest our storage or nework on other underlying layers, that will affect our results.

We do need to find a solution mid-term, and I suspect it might be the 'write a script to clean up the environment' - but, we are not there yet...

grahamwhaley:
@chavafg - the one in the image install does a force remove (`rm -f`) just before the `ln`, so I think that one is safe? (doing a forced `rm` is also an option in the kernel script - but `ln -sf` I guess saves an extra command). Maybe I'll change the crio one to remove the `rm` and add the `-f`...

What do you think about the one in https://github.com/kata-containers/tests/blob/master/.ci/install_crio.sh#L27 as well?

grahamwhaley:
proposed code [here](https://github.com/grahamwhaley/kata-containers-tests/commit/eef5c33de8557ddfa9e4bca3d9b2fb9a6d8ed48e)

PR #176, opened 2018-03-27
ci: installers: add force install to symlinks

grahamwhaley:
Some symlink installs did not use the `-f` force flag, which meant
they failed if we were installing over an already installed system.

Add the `-f` flag, and in the instance of the image code, replace
an explicit `rm -f` with the `-f` on the following `ln`.

Fixes: #150

Signed-off-by: Graham whaley <graham.whaley@intel.com>

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -23,7 +23,7 @@ if [ ! -d "$GOBIN" ]
 then
         mkdir -p "$GOBIN"
 fi
-ln -s $(command -v go-md2man) "$GOBIN"
+ln -sf $(command -v go-md2man) "$GOBIN"
 
 echo "Get CRI Tools"
 critools_repo="github.com/kubernetes-incubator/cri-tools"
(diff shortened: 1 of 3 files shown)


jodh-intel:
lgtm

.ci/install_crio.sh
```diff
@@ -23,7 +23,7 @@ if [ ! -d "$GOBIN" ]
 then
         mkdir -p "$GOBIN"
 fi
-ln -s $(command -v go-md2man) "$GOBIN"
+ln -sf $(command -v go-md2man) "$GOBIN"
```
grahamwhaley: I've since seen a different issue here - if we are running on top of a dirty tree then the `go-md2man` file found from the `command` is already the on in the `$GOBIN`, and then ln complain and fails with 'source file is the same as dest file`.
I've got a local workaround - one of 3 options I think:
- if the dest already exists, skip the `ln`
- if the dest already exists, force remove it before the `ln`
- ignore the error from the `ln` with a `|| true`

@chavafg - any thoughts or preferences from you on this?

(I'm happy to rework the PR btw, but we then run the risk of bumping into the Travis monster, so maybe better as a new PR to update the code)

chavafg: Hi @grahamwhaley, 
Either the 1st or second option would be ok for me. 
I'll merge this one and you can rework on a new one to not hit the travis bug.

chavafg:
lgtm

Issue #177, opened 2018-03-27
checkcommits: use TRAVIS_PULL_REQUEST_SHA


bergwolf:
We've got quite a few checkcommits failures like the following:
```
Detected TravisCI Environment
ERROR: Failed to run command /usr/bin/git rev-list --no-merges --reverse master..8e1f6d1b64742dfad80b6132fb4daad0ea4282ee: exit status 128 (stdout: , stderr: fatal: Invalid revision range master..8e1f6d1b64742dfad80b6132fb4daad0ea4282ee
)
ERROR: checkcommits failed. See the document below for help on formatting
```
Add some debug log to properly identify the cause.

grahamwhaley:
Hi @bergwolf - I think @jodh-intel can give you the full details and the pointer to the open Travis bug report for this - we have tried to track down and fix this before for CC, and could not fix it :-(  Hence, we are rapidly trying to move everything over to Jenkins (which is what we did for CC).

bergwolf:
@grahamwhaley my guess is that the commit `8e1f6d1b64742dfad80b6132fb4daad0ea4282ee` came from nowhere. The same command fails in my local environment as well. Therefore I created the PR to confirm it. If we are sure that it is the case, maybe we should stop relying on `TRAVIS_COMMIT` to get the testing commit.

jodh-intel:
Hi @bergwolf - yes, the problem we believe is https://github.com/travis-ci/travis-ci/issues/7830.

I'd very much like to switch off Travis and use Jenkins entirely due to that bug (and the lack of VT-x). See https://github.com/kata-containers/tests/issues/132 for the last "blocker" to move away from Travis).

In summary, printing env vars won't help debug this particular problem. However, I do think there is a lot we can to do make debugging CI failures clearer generally, including dumping the full environment using my `procenv` tool and using the Jenkins pipeline plugin:

- https://github.com/clearcontainers/jenkins/issues/5
- https://github.com/clearcontainers/jenkins/issues/7

jodh-intel:
Those last 2 issue re-raised in Kata as:

- https://github.com/kata-containers/ci/issues/14
- https://github.com/kata-containers/ci/issues/15

bergwolf:
@jodh-intel Thanks for the links. We should be able to get the testing commit head on our own rather than relying on travis environment variables. Then the issue would go away.

jodh-intel:
[...]

PR #178, opened 2018-03-27
checkcommits: use TRAVIS_PULL_REQUEST_SHA to get testing branch HEAD

bergwolf:
`TRAVIS_COMMIT` is not reliable in travis. Let's try `TRAVIS_PULL_REQUEST_SHA`.

 Fixes: #177

### DIFF
--- a/cmd/checkcommits/checkcommits.go
+++ b/cmd/checkcommits/checkcommits.go
@@ -443,7 +443,7 @@ func detectCIEnvironment() (commit, dstBranch, srcBranch string) {
 	if os.Getenv("TRAVIS") != "" {
 		name = "TravisCI"
 
-		commit = os.Getenv("TRAVIS_COMMIT")
+		commit = os.Getenv("TRAVIS_PULL_REQUEST_SHA")
 
 		srcBranch = os.Getenv("TRAVIS_PULL_REQUEST_BRANCH")
 		dstBranch = os.Getenv("TRAVIS_BRANCH")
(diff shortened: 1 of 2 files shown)


devimc:
LGTM

@jodh-intel ptal

grahamwhaley:
We should probably check the logs to ensure that the PR'd checkcommits was built, installed and used in the travis CI here - unless somebody can verify for sure?

jodh-intel:
Fine with me.

lgtm

bergwolf:
@grahamwhaley I believe travis installs it from scratch. Otherwise everything would be messed up, right?

grahamwhaley:
@bergwolf yes, it will have to install it (as we run on a bare VM instance), but it is a matter more of did it install it from the master branch or the PR branch. I expect we are safe and it uses the PR branch - but there is a short window at the start of running the jenkins scripts where it is running from the master, and then effectively flips over to the PR branch iirc. this only happens for the 'test repo', and may only really be an issue up to [this point](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L63) in the jenkins script though (it has bitten us once or twice before ;-) )

grahamwhaley:
oh, and now I remember, even after the `merge`, it is still running the jenkins script from the master, as that script file is open and loaded into `bash` already. Just to note - so any changes to the whole of that script or that happen before that merge line are at risk.

chavafg:
lgtm

PR #179, opened 2018-03-27
integration: crio: Enable more tests from ctr.bats

### LINKED ISSUE #171: integration: crio: Evaluate ctr.bats tests
I have run CRI-O bats `ctr.bats` with `overlay` and `devicemapper` storage drivers and here is the result:

- Passing tests:
  - `ctr log max`
  - `ctr partial line logging`
  - `ctrs status for a pod`
  - `ctr stop idempotent`
  - `ctr caps drop`
  - `ctr with image with Config.Volumes`
  - `ctr /etc/resolv.conf rw/ro mode`
  - `ctr correctly setup working directory`
  - `ctr execsync conflicting with conmon env`

- Failing tests:
  - `ctr hostname env`: Needs a patch on `kata-containers/runtime` to support executing a process on a container `CREATED` but not yet `STARTED`/`RUNNING`.
  - `ctr oom`: OOM never happened, needs further investigations.
  - `ctr create with non-existent command`: `cleanup_test` failed because we need to properly unmount what we have previously mounted, even in case of failure (which is the case of this test).
  - `ctr create with non-existent command [tty]`: Same reason than `ctr create with non-existent command`
  - `ctr update resources`: We don't have the support of `kata-runtime update` yet.
  - `ctr resources`: We don't have the support of `cpuset`.

sboeuf:
Kata Containers actually pass more tests than what is tested right
now. This commit removes those tests from the skip list.

Fixes #171

### DIFF
--- a/integration/cri-o/crio_skip_tests.sh
+++ b/integration/cri-o/crio_skip_tests.sh
@@ -9,18 +9,10 @@
 
 declare -a skipCRIOTests=(
 'test "ctr hostname env"'
-'test "ctr execsync failure"'
-'test "ctr execsync exit code"'
-'test "ctr execsync std{out,err}"'
-'test "ctr stop idempotent"'
-'test "ctr caps drop"'
-'test "run ctr with image with Config.Volumes"'
 'test "ctr oom"'
+'test "ctr \/etc\/resolv.conf rw\/ro mode"'
 'test "ctr create with non-existent command"'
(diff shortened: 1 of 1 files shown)


jodh-intel:
Nice! :smile:

lgtm

chavafg:
lgtm

devimc:
lgtm

PR #181, opened 2018-03-27
CI: Check if yamllint is available before calling

### LINKED ISSUE #180: Check if yamllint is installed
https://github.com/kata-containers/tests/pull/164 added a requirements for `yamllint`. That PR installs it for the Jenkins CI, but not for Travis, which is now failing. For example, see:

https://github.com/kata-containers/runtime/pull/112

Since we want to move away from Travis, just add a check to only call `yamllint` if available.

jodh-intel:
Don't call `yamllint` if it isn't installed.

Fixes #180.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -168,7 +168,8 @@ check_versions()
 
 	[ ! -e "$db" ] && return
 
-	yamllint "$db"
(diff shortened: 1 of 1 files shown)


chavafg:
lgtm

PR #183, opened 2018-03-27
CI: Do not fail if yamllint is not installed.

### LINKED ISSUE #180: Check if yamllint is installed

chavafg:
If yamllint is not installed, the static-checks.sh
fails. Change the way we verify it is installed.

Fixes #180

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -169,7 +169,9 @@ check_versions()
 	[ ! -e "$db" ] && return
 
 	cmd="yamllint"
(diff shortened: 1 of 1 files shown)


sboeuf:
This will unblock the CI, merging....

PR #194, opened 2018-04-03
Run collect script under ci

### LINKED ISSUE #193: Run data collection script under CI
The CI should run the data collection script under the CI:

- To ensure it actually runs and doesn't fail.

   The utility is designed for users to run to help us, so we need to ensure it works.

- To capture details of the environment on CI failure.

jodh-intel:
- Run the data collection script to see if it works.

- If the CI run fails, run the data collection script to help provide further clues on the failure.

Fixes #193.

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -71,3 +71,6 @@ sudo -E PATH="$PATH" bash -c "make test"
 
 echo "INFO: Checking log files"
 check_log_files
+
+echo "INFO: Running data collection script"
+sudo -E PATH="$PATH" chronic kata-collect-data.sh
--- a/.ci/teardown.sh
+++ b/.ci/teardown.sh
@@ -60,12 +60,16 @@ if [ "${log_copy_dest}" ]; then
 else
 	echo "Kata Containers Runtime Log:"
 	journalctl --no-pager -t kata-runtime
+
 	echo "Kata Containers Proxy Log:"
 	journalctl --no-pager -t kata-proxy
+
 	echo "Kata Containers Shim Log:"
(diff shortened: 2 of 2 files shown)


.ci/run.sh
```diff
@@ -71,3 +71,6 @@ sudo -E PATH="$PATH" bash -c "make test"
 
 echo "INFO: Checking log files"
 check_log_files
+
+echo "INFO: Running data collection script"
+sudo -E PATH="$PATH" chronic kata-collect-data.sh
```
grahamwhaley: I think you know what I'm going to say ... ;-)
Does this currently come out in the main CI log? Any chance you can redirect the output (at least stdout) to a known log file location - and then we can use that output in the teardown script, and save:
- injecting the rather large collect script output into the main console log
- having to run the script more than once?

jodh-intel: Chronic discards all output *unless* the command fails so although there are 2 calls, this one won't bloat the console logs unless the script itself is broken (which we would want to know about).

grahamwhaley: ah, indeed, I didn't spot the `chronic` - mea culpa.

jodh-intel: I know what you're saying, but the two calls are for different reasons - this one basically checks the script "works" (very simplistically of course), whereas the other call captures the script output if a PR fails (implicitly expecting the script to work).

That teardown script is a bit fragile imho as it has to cater for two separate scenarios so I'd be a little bit leery of adding more complexity to it by getting it to find the output of this script and potentially chop it into bits, etc.

chavafg:
lgtm

PR #219, opened 2018-04-05
CI: run virtcontainers unit tests only as root

### LINKED ISSUE #218: Run virtcontainers unit tests only with root user
because CI  is failing on virtcontainers unit tests:
http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-runtime-ubuntu-17-10-PR/7/console

lets see if running only as root helps to resolve this issue.

chavafg:
Virtcontainers unit tests only work if they are executed
using the root user.

Fixes #218.

Depends-on: github.com/kata-containers/runtime#193

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -102,9 +102,14 @@ test_coverage()
 	fi
 
 	echo "INFO: Currently running as user '$(id -un)'"
-
 	for pkg in $test_packages; do
 		for user in $users; do
+			# Run virtcontainers tests only if user is root.
+			if [[ "$user" != "root" ]] && [[ "$pkg" = *"virtcontainers"* ]]; then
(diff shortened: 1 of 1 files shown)


.ci/go-test.sh
```diff
@@ -108,6 +108,12 @@ test_coverage()
 			printf "INFO: Running 'go test' as %s user on package '%s' with flags '%s'\n" \
 				"$user" "$pkg" "$go_test_flags"
 
+			# Run virtcontainers tests only if user is root.
```
grahamwhaley: Which is a little sad, but iirc it is because the vc tests need to install the cni plugins etc., which has to be done as root?
I wonder if the better fix is to ensure we `sudo` all the bits in the vc test setup so we can still run the tests?

sboeuf:
LGTM

Issue #233, opened 2018-04-13
Install go version in CI using versions.yaml


jcvenegas:
As proposed in [1] versions.yaml define a sections for languages versions to be used. Add an script that read this file an install the the lastest supported golang version.

[1] https://github.com/kata-containers/kata-containers/issues/15

PR #234, opened 2018-04-13
ci: Install go defined in versions.yaml

jcvenegas:
This commit adds an script to be used by ci to run
the latest supported golange version.

Fixes: #233

### DIFF
--- a/.ci/install_go.sh
+++ b/.ci/install_go.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+tmp_dir=$(mktemp -d -t install-go-tmp.XXXXXXXXXX)
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+script_name="$(basename "${BASH_SOURCE[0]}")"
+USE_VERSIONS_FILE=""
+PROJECT="Kata Containers"
+
+source "${script_dir}/lib.sh"
+
+install_dest="/usr/local/"
+
+finish() {
+	rm -rf "$tmp_dir"
+}
+
+die() {
+	echo >&2 "ERROR: $*"
+	exit 1
+}
+
+info() {
+	echo "INFO: $*"
+}
+
+usage(){
+	exit_code="$1"
+	cat <<EOT
+Usage:
+
+${script_name} [options] <args>
+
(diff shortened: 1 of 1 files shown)


.ci/install_go.sh
```diff
@@ -0,0 +1,101 @@
...
+		;;
+	"*")
+		die "Arch $(arch) not supported"
+		;;
+esac
+
+info "Download go version ${go_version}"
+curl -OL "https://storage.googleapis.com/golang/go${go_version}.linux-${goarch}.tar.gz"
+info "Remove old go installation"
+sudo rm -rf /usr/local/go/
```
devimc: woot, this is a little dangerous, isn't it?

devimc: wouldn't be better to have an option ins the script to specify the installation path ?

jcvenegas: @devimc  I was thinking that, but because it is a non standard installation path I think ok. I preferred to delete an specific  path rather than a variable with a any value.  , but I can change even the place to installed. Agree I can a --dest-dir option.

jcvenegas:
@devimc  added destination flag, removed the rm command.

.ci/install_go.sh
```diff
@@ -0,0 +1,97 @@
...
+Example:
+${script_name} 1.10
+
+Options
+-d <path> : destination path, path where go will be installed.
+-h        : Show this help
+-p        : Install go defined in ${PROJECT} versions file.
+
+EOT
+exit "$exit_code"
```
chavafg: missing tab?

.ci/install_go.sh
```diff
@@ -0,0 +1,97 @@
...
+		h)	usage 0 ;;
+		p)	USE_VERSIONS_FILE="true" ;;
+	esac
+done
+
+shift $(( $OPTIND - 1 ))
+
+go_version="${1:-""}"
+if [ -z "$go_version" ];then
+	usage 0
```
chavafg: If I run `./test.sh -p`, it fails because I am not providing a version, but it shouldn't as I want to install version from the `versions.yaml`, right?

.ci/install_go.sh
jcvenegas: updated

.ci/install_go.sh
jcvenegas: you are right thx.

jcvenegas:
@GabyCT  @jodh-intel  @devimc @chavafg  ping ready to merge.

jodh-intel:
Love it!

lgtm :)

chavafg:
lgtm

PR #237, opened 2018-04-16
ci: Minor CI fixes.

jcvenegas:
- Build image in a tmp directory 
- clone need qemu sub-modules.

### DIFF
--- a/.ci/install_kata_image.sh
+++ b/.ci/install_kata_image.sh
@@ -5,9 +5,22 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-set -e
+set -o errexit
+set -o nounset
+set -o pipefail
 
 cidir=$(dirname "$0")
+
+tmp_dir=$(mktemp -d -t kata-image-install.XXXXXXXXXX)
+readonly ROOTFS_DIR="${tmp_dir}/rootfs"
+export ROOTFS_DIR
+
+finish() {
+  [ -d "${ROOTFS_DIR}" ] && [[ "${ROOTFS_DIR}" = *"rootfs"* ]] && sudo rm -rf "${ROOTFS_DIR}"
+  rm -rf "$tmp_dir"
+}
+
+trap finish EXIT
 
 OSBUILDER_DISTRO=${OSBUILDER_DISTRO:-clearlinux}
 AGENT_INIT=${AGENT_INIT:-no}
@@ -17,7 +30,6 @@ TEST_INITRD=${TEST_INITRD:-no}
 bash -f ${cidir}/install_agent.sh
 
 osbuilder_repo="github.com/kata-containers/osbuilder"
-export ROOTFS_DIR="${GOPATH}/src/${osbuilder_repo}/rootfs-builder/rootfs"
 
 # Clone os-builder repository
 go get -d ${osbuilder_repo} || true
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -8,23 +8,24 @@
 set -e
 
 QEMU_REPO="github.com/qemu/qemu"
-QEMU_DIR="qemu"
 KATA_QEMU_BRANCH="stable-2.11"
 PACKAGING_REPO="github.com/kata-containers/packaging"
-QEMU_CONFIG_SCRIPT="configure-hypervisor.sh"
+QEMU_CONFIG_SCRIPT="${GOPATH}/src/${PACKAGING_REPO}/scripts/configure-hypervisor.sh"
 
-git clone "https://${QEMU_REPO}"
+go get -d "${QEMU_REPO}" || true
 
 # Get qemu configuration script and copy to
 # the qemu repository
 go get -d "$PACKAGING_REPO" || true
-cp "${GOPATH}/src/${PACKAGING_REPO}/scripts/${QEMU_CONFIG_SCRIPT}" "${QEMU_DIR}"
 
-pushd "$QEMU_DIR"
+pushd "${GOPATH}/src/${QEMU_REPO}"
+git fetch
 git checkout "$KATA_QEMU_BRANCH"
+[ -d "capstone" ] || git clone https://github.com/qemu/capstone.git capstone
+[ -d "ui/keycodemapdb" ] || git clone  https://github.com/qemu/keycodemapdb.git ui/keycodemapdb
 
 echo "Build Qemu"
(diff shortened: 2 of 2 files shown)


.ci/install_kata_image.sh
```diff
@@ -9,6 +9,17 @@ set -e
 
 cidir=$(dirname "$0")
 
+tmp_dir=$(mktemp -d -t kata-image-install.XXXXXXXXXX)
+readonly ROOTFS_DIR="${tmp_dir}/rootfs"
+export ROOTFS_DIR
+
+finish() {
+  sudo rm -rf "${ROOTFS_DIR}"
```
jodh-intel: This works, but it's worth pointing out that the *ordering* of the script is critical here.

Consider if the script looked like this (where the variables are set *after* the `trap`):

```
set -e

finish() {
  sudo rm -rf "${ROOTFS_DIR}"
}

trap finish EXIT

tmp_dir=$(mktemp -d -t kata-image-install.XXXXXXXXXX)
readonly ROOTFS_DIR="${tmp_dir}/rootfs"
export ROOTFS_DIR
```

Now, if `mktemp` fails, `finish()` *still* gets called on exit (even with `set -e`) and `ROOTFS_DIR=""` ! It might be prudent to check `ROOTFS_DIR` *just in case*:

```
finish() {
    [ -d "${ROOTFS_DIR}" ] && sudo rm -rf "${ROOTFS_DIR}"
}
```

/cc @grahamwhaley, @icarus-sparry, @chavafg.

grahamwhaley: I suspect you are correct @jodh-intel - does look like a potential issue, and adding that check would be the correct thing do do imho

jcvenegas: @grahamwhaley @jodh-intel I already tested on bash does not not fail having an empty variable but I can add that extra validation, and because is removing with `sudo` (needed because the rootfs is owned by root) I can also double check that $ROOTFS contains `rootfs` just to make sure is not pointing to somewhere else.

.ci/install_kata_image.sh
```diff
@@ -5,10 +5,23 @@
...
+set -o pipefail
 
 cidir=$(dirname "$0")
 
+tmp_dir=$(mktemp -d -t kata-image-install.XXXXXXXXXX)
+readonly ROOTFS_DIR="${tmp_dir}/rootfs"
+export ROOTFS_DIR
+
+finish() {
+  [ -d "${ROOTFS_DIR}" ] && [[ "${ROOTFS_DIR}" = *"rootfs"* ]] && sudo rm -rf "${ROOTFS_DIR}"
```
jcvenegas: @jodh-intel @grahamwhaley  does it looks better?

jodh-intel:
lgtm

.ci/install_kata_image.sh
grahamwhaley: What we'd call 'belt and braces' :-)
Wacky use of *'s and those [['s
presuming you tested it (which I strongly suspect you did), then
lgtm

.ci/install_qemu.sh
```diff
@@ -8,23 +8,24 @@
 set -e
 
 QEMU_REPO="github.com/qemu/qemu"
-QEMU_DIR="qemu"
 KATA_QEMU_BRANCH="stable-2.11"
 PACKAGING_REPO="github.com/kata-containers/packaging"
-QEMU_CONFIG_SCRIPT="configure-hypervisor.sh"
+QEMU_CONFIG_SCRIPT="${GOPATH}/src/${PACKAGING_REPO}/scripts/configure-hypervisor.sh"
```
icarus-sparry: Remember always that GOPATH is potentially a colon separated list of directories, or may not even exist (in which case the go tooling uses $HOME/go).

I suggest

      MYGOREPO=${GOPATH:-"$HOME/go"}
      MYGOREPO=${MYGOREPO%%:*}
      QEMU_CONFIG_SCRIPT="${MYGOREPO}/src/....

.ci/install_qemu.sh
```diff
@@ -8,23 +8,24 @@
...
-pushd "$QEMU_DIR"
+pushd "${GOPATH}/src/${QEMU_REPO}"
+git fetch
 git checkout "$KATA_QEMU_BRANCH"
+[ -d "capstone" ] || git clone https://github.com/qemu/capstone.git capstone
+[ -d "ui/keycodemapdb" ] || git clone  https://github.com/qemu/keycodemapdb.git ui/keycodemapdb
 
 echo "Build Qemu"
-eval "./${QEMU_CONFIG_SCRIPT}" "qemu" | xargs ./configure
+eval "${QEMU_CONFIG_SCRIPT}" "qemu" | xargs ./configure
```
icarus-sparry: Not sure why you have `eval` here. As long as QEMU_CONFIG_SCRIPT has no `interesting` characters in it it adds nothing, and if it does have them it probably makes the script do the wrong thing.

.ci/install_kata_image.sh
icarus-sparry: Historically things are even worse, when a pathname that was empty could be used to refer to the current directory. You can still see vestiges of this today, e.g.

     cd /   # make sure you are somewhere known 
     cd "" # empty name
     pwd  # should print /
     cd
     pwd  # should be your home directory

So yes, this is something to be concerned about. Essentially you need to write things in the correct order, and in particular don't set up traps which do things until after they are safe to be sprung. Just adding the test to see if ROOTFS_DIR is a directory isn't good enough, as the variable could be inherited from the environment, and you wouldn't like someone to pass say ROOTFS=/usr !

Of course `sudo` takes care to clean the environment to help for exactly this sort of reason.

The other point I would like to make is that traps are dynamic, you can set different things to do on exit (or error) at different points in the program. Of course if you take this too far then you have a spaghetti ball of unmaintainable code.

.ci/install_qemu.sh
grahamwhaley: That's an interesting point now they have changed the go path rules indeed. I think we are currently 'safe' here as this only runs (at present) in our CI, which always sets the GOPATH to a single path - but, good points.

.ci/install_kata_image.sh
icarus-sparry: @grahamwhaley one day you will get used to them 👍

.ci/install_qemu.sh
jcvenegas: @icarus-sparry  thx, I removed the eval call, for GOPATH we are still safe, but we may want to rework all our ci scripts that relies in this variable

Issue #239, opened 2018-04-17
metrics: launch_times: only uses default runtime for scaling


grahamwhaley:
When the `launch_times.sh` test is used in scaling mode, it should use the `RUNTIME` variable when [running the background containers](https://github.com/kata-containers/tests/blob/master/metrics/time/launch_times.sh#L142).
At present it uses the default docker runtime, which may be fine if the runtime you are testing is the default, but that is not always the case (as was at my desk).

Add the `--runtime=${RUNTIME}` to the SCALING run command to fix.

PR #241, opened 2018-04-17
metrics: launch_times: set runtime for background containers

grahamwhaley:
We were not using the defined RUNTIME when launching the background
(scaling test) containers. This meant the scaling test only really
measured what we thought when the runtime we were testing was the
docker default one.

Explicitly state which runtime to use for the background containers
so we test what we expect for any setting of RUNTIME.

Fixes: #239

Signed-off-by: Graham whaley <graham.whaley@intel.com>


jodh-intel:
lgtm

GabyCT:
lgtm

PR #245, opened 2018-04-18
metrics: Add blogbench storage test.

### LINKED ISSUE #244: metrics: Add blogbench test
Add blogbench test as part of kata-containers storage tests.

GabyCT:
This test is designed to test storage for kata-runtime by performing
random reads and writes.

Fixes #244

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -92,6 +92,24 @@ check_images()
 	done
 }
 
+# This function performs a docker build on the image names
+# passed in, to ensure that we have the latest changes from
+# the dockerfiles
+check_dockerfiles_images()
+{
+	local image="$1"
+	local dockerfile_path="$2"
+	if [ -z "$image" ] || [ -z "$dockerfile_path" ]; then
+		die "Missing image or dockerfile path variable"
+		exit 1;
+	fi
+	echo "docker building $image"
+	if ! docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"; then
+		die "Failed to docker build image $image"
+		exit 1;
+	fi
+}
+
 # A one time (per uber test cycle) init that tries to get the
 # system to a 'known state' as much as possible
 metrics_onetime_init()
--- a/metrics/storage/blogbench.sh
+++ b/metrics/storage/blogbench.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Description of the test:
+# This test runs the 'blogbench', and extracts the 'scores' for reads
+# and writes
+# Note - the scores are *not* normalised for the number of iterations run,
+# they are total scores for all iterations (this is the blogbench default output)
+
+set -e
+
+# General env
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+TEST_NAME="blogbench"
+IMAGE="local-blogbench"
+DOCKERFILE="${SCRIPT_PATH}/blogbench_dockerfile/Dockerfile"
+
+# Number of iterations for blogbench to run - note, results are not
+# scaled to iterations - more iterations results in bigger results
+ITERATIONS="${ITERATIONS:-30}"
+
+# Directory to run the test on
+# This is run inside of the container
+TESTDIR="${TESTDIR:-/tmp}"
+CMD="blogbench -i ${ITERATIONS} -d ${TESTDIR}"
+
+function main() {
+	# Check tools/commands dependencies
+	cmds=("awk" "docker")
+
+	init_env
+	check_cmds "${cmds[@]}"
+	check_dockerfiles_images "$IMAGE" "$DOCKERFILE"
+
+	metrics_json_init
+
+	local output=$(docker run --rm --runtime=$RUNTIME $IMAGE $CMD)
+
+	# Save configuration
+	metrics_json_start_array
+
+	local frequency=$(echo "$output" | grep "Frequency" | cut -d "=" -f2 | cut -d ' ' -f2)
+	local iterations=$(echo "$output" | grep -w "iterations" | cut -d ' ' -f3)
+	local spawing_writers=$(echo "$output" | grep -w "writers" | cut -d ' ' -f2)
+	local spawing_rewriters=$(echo "$output" | grep -w "rewriters" | cut -d ' ' -f2)
+	local spawing_commenters=$(echo "$output" | grep -w "commenters" | cut -d ' ' -f2)
+	local spawing_readers=$(echo "$output" | grep -w "readers" | cut -d ' ' -f2)
+
+	local json="$(cat << EOF
+	{
+		"Frequency" : $frequency,
+		"Iterations" : $iterations,
+		"Number of spawing writers" : $spawing_writers,
+		"Number of spawing rewriters" : $spawing_rewriters,
+		"Number of spawing commenters" : $spawing_commenters,
+		"Number of spawing readers" : $spawing_readers
+	}
+EOF
+)"
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Config"
+
+	# Save results
+	metrics_json_start_array
+
+	local writes=$(tail -2 <<< "$output" | head -1 | awk '{print $5}')
+	local reads=$(tail -1 <<< "$output" | awk '{print $6}')
+
+	# Obtaining other Blogbench results
+	local -r data=$(echo "$output" | tail -n +12 | head -n -3)
+	local nb_blogs=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $1} ' | tr '\t' ',' | sed '$ s/.$//')
+	local r_articles=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $2} ' | tr '\t' ',' | sed '$ s/.$//')
+	local w_articles=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $3} ' | tr '\t' ',' | sed '$ s/.$//')
+	local r_pictures=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $4} ' | tr '\t' ',' | sed '$ s/.$//')
+	local w_pictures=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $5} ' | tr '\t' ',' | sed '$ s/.$//')
+	local r_comments=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $6} ' | tr '\t' ',' | sed '$ s/.$//')
+	local w_comments=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $7} ' | tr '\t' ',' | sed '$ s/.$//')
+
+	local json="$(cat << EOF
+	{
+		"write": {
+			"Result" : $writes,
+			"Units"  : "items"
+		},
+		"read": {
+			"Result" : $reads,
+			"Units"  : "items"
+		},
+		"Nb blogs": {
+			"Result" : "$nb_blogs"
+		},
+		"R articles": {
+			"Result" : "$r_articles"
+		},
+		"W articles": {
+			"Result" : "$w_articles"
+		},
+		"R pictures": {
+			"Result" : "$r_pictures"
+		},
+		"W pictures": {
+			"Result" : "$w_pictures"
+		},
+		"R comments": {
+			"Result" : "$r_comments"
+		},
+		"W comments": {
+			"Result" : "$w_comments"
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+}
+
+main "$@"
--- a/metrics/storage/blogbench_dockerfile/Dockerfile
+++ b/metrics/storage/blogbench_dockerfile/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Set up an Ubuntu image with 'blogbench' installed
+
+# Usage: FROM [image name]
+FROM ubuntu
+
+# URL for blogbench test and blogbench version
+ENV BLOGBENCH_URL "https://download.pureftpd.org/pub/blogbench"
+ENV BLOGBENCH_VERSION 1.1
+
+RUN apt-get update && \
+	apt-get install -y build-essential curl && \
+	curl -OkL ${BLOGBENCH_URL}/blogbench-${BLOGBENCH_VERSION}.tar.gz && \
+	tar xzf blogbench-${BLOGBENCH_VERSION}.tar.gz && \
+	cd blogbench-${BLOGBENCH_VERSION} && \
+	./configure && \
+	make && \
+	make install-strip
+
+CMD ["/bin/bash"]
--- a/.ci/run_metrics_PR_ci.sh
+++ b/.ci/run_metrics_PR_ci.sh
@@ -89,6 +89,9 @@ run() {
 	# (so set a token 5s wait)
 	bash density/docker_memory_usage.sh 20 5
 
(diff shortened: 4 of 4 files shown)


metrics/storage/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# Set up an Ubuntu image with 'blogbench' installed
+
+# Usage: FROM [image name]
+FROM ubuntu
+
+RUN apt-get update && \
+	apt-get install -y build-essential wget && \
+	wget https://download.pureftpd.org/pub/blogbench/blogbench-1.1.tar.gz && \
+	tar xvzf blogbench-1.1.tar.gz && \
```
jodh-intel: Version hard-coded multiple times. The ideal would be to have this version in a `versions.txt` or `versions.yaml` and create this `Dockerfile` from `Dockerfile.in`. Alternatively, I'd create a variable in this file to avoid having to duplicate it.

metrics/storage/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2018 Intel Corporation
```
grahamwhaley: I wonder if this is the best place to put the Dockerfile, or maybe the best name to use.
Given we are likely to gain more storage tests, unless we intend to add support for all of those tests to this one Dockerfile, then I suspect we will want to move or rename this one before we merge.
Also, iirc, `docker build` incorporates all the files in the same directory as the dockerfile into the image? (https://docs.docker.com/engine/reference/commandline/build/#options), so this might not be what we want here...

Over on CC we had started to hold all the dockerfiles in a separate tree a little higher up in the dir hierarchy - I think it will actually be better to keep the dockerfiles much closer to the associated tests if we can (which I think will only become a problem if we start sharing images across tests across trees...). Thus, I suggest we make a `blogbench` subdir here and place the dockerfile in that.

What do you think @GabyCT @chavafg ?

metrics/storage/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Set up an Ubuntu image with 'blogbench' installed
+
+# Usage: FROM [image name]
+FROM ubuntu
```
grahamwhaley: @GabyCT , can you investigate for me if/how we might bury a version in the Dockerfile, and then adjust our `docker build` check/invocation so we can spot if the dockerfile has been updated, and rebuild the image. Right now the (rather simple) code I wrote only does a naive check to see if the image already exists, and does not have any capability to update the image. I think one day that will come to bite us....

metrics/storage/Dockerfile
grahamwhaley: Valid point @jodh-intel I will note that the blogbench test has not been modified since 2015 though... ;-)
We may as well add the env var and do the edits though @GabyCT to remove the duplication.

metrics/storage/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN apt-get update && \
+	apt-get install -y build-essential wget && \
+	wget https://download.pureftpd.org/pub/blogbench/blogbench-1.1.tar.gz && \
+	tar xvzf blogbench-1.1.tar.gz && \
+	cd blogbench-1.1 && \
+	./configure && \
+	make && \
+	make install-strip
+
```
grahamwhaley: I just noticed yesterday that, if you are editing this file, we could also delete the .tar.gz file from the image (or untar it in place) now the binary is installed. and/or the same for the source tree once the binary is in place. Just a thought.

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,75 @@
...
+ITERATIONS="${ITERATIONS:-30}"
+
+# Directory to run the test on
+TESTDIR="${TESTDIR:-/tmp}"
+CMD="blogbench -i ${ITERATIONS} -d ${TESTDIR}"
+
+# Check if we have our local image installed, and if not, install it
+# from the dockerfile
+function local_check_images() {
+	local exists=$(docker image inspect $IMAGE > /dev/null; echo $?)
```
grahamwhaley: As per other note - place consider if we can make this a little smarter wrt versioning.
And, if you do, also maybe we should move this out to the generic lib funcs for use by other tests - I predict we will use this same idea in the network tests for instance ;-)

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,75 @@
...
+	check_cmds "${cmds[@]}"
+	local_check_images "$IMAGE"
+
+	metrics_json_init
+
+	local output=$(docker run --rm --runtime=$RUNTIME $IMAGE $CMD)
+
+	metrics_json_start_array
+
+	local writes=$(tail -2 <<< "$output" | head -1 | awk '{print $5}')
```
grahamwhaley: `blogbench` outputs a bunch more info we are throwing away here. The code here was used for the `csv` file output, as that format was somewhat limited in what it could store - but our new shiny JSON format is much much more flexible.
Can you see if you can parse out some more info from the blogbench and store it in the JSON for us? Here (for others to see) is an example of a blogbench run:

```
root@80fbf6a05ff8:/# blogbench -i 5 -d /tmp

Frequency = 10 secs
Scratch dir = [/tmp]
Spawning 3 writers...
Spawning 1 rewriters...
Spawning 5 commenters...
Spawning 100 readers...
Benchmarking for 5 iterations.
The test will run during 0 minutes.

  Nb blogs   R articles    W articles    R pictures    W pictures    R comments    W comments
        57        61008          2995         44577          3313         13619          3584
        77        58250          1138         43798          1045         27073          3948
        95        55041          1279         39257           911         30923          3791
       109        52750          1188         37391           948         30166          3769
       126        52620          1152         37169           945         30842          3668

Final score for writes:           126
Final score for reads :          6021
```

So, I'd quite like to store those per-second statistics, as well as the config info.
I know, it would be nice if `blogbench` had a json output option, or output some other standard parseable format - but sadly I don't think it does.... :-(

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
grahamwhaley: We should consider adding a call to this script in https://github.com/kata-containers/tests/blob/master/.ci/run_metrics_PR_ci.sh as well I believe.

metrics/lib/common.bash
```diff
@@ -92,6 +92,20 @@ check_images()
...
+# passed in, to ensure that we have the latest changes from
+# the dockerfiles
+check_dockerfiles_images()
+{
+	local image="$1"
+	local dockerfile_path="$2"
+	local exists=$(docker image inspect $image > /dev/null; echo $?)
+	if [ $exists -eq 0 ] || [ $exists -ne 0 ]; then
+		echo "Building Dockerfile Image : $image"
+		docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"
```
jcvenegas: Hope we build this with runc as default runtime (or kata not using 9pfs)

metrics/lib/common.bash
```diff
@@ -92,6 +92,20 @@ check_images()
 	done
 }
 
+# This function performs a docker build on the image names
+# passed in, to ensure that we have the latest changes from
+# the dockerfiles
+check_dockerfiles_images()
+{
+	local image="$1"
+	local dockerfile_path="$2"
```
jcvenegas: Missing check if variable is valid.

metrics/lib/common.bash
```diff
@@ -92,6 +92,20 @@ check_images()
...
+	local exists=$(docker image inspect $image > /dev/null; echo $?)
```
jcvenegas: Is this function called with `errexit` flag (that probably would be good to do)?. If yes the echo may not be executed?

metrics/lib/common.bash
```diff
@@ -92,6 +92,20 @@ check_images()
...
+	if [ $exists -eq 0 ] || [ $exists -ne 0 ]; then
```
jcvenegas: this statement will be always true.

metrics/lib/common.bash
jcvenegas: I've seen that docker builds does not repete the steps already done. So if you change the dockerfile will execute the dockerfile. Probably is better always build the image.

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,68 @@
...
+# and writes
+# Note - the scores are *not* normalised for the number of iterations run,
+# they are total scores for all iterations (this is the blogbench default output)
+
+set -e
+
+# General env
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+TEST_ARGS="$@"
```
jcvenegas: Could you document the usage of this variable?

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,68 @@
...
+TEST_NAME="blogbench"
+IMAGE="local-blogbench"
+DOCKERFILE="${SCRIPT_PATH}/blogbench_dockerfile/Dockerfile"
+
+# Number of iterations for blogbench to run - note, results are not
+# scaled to iterations - more iterations results in bigger results
+ITERATIONS="${ITERATIONS:-30}"
+
+# Directory to run the test on
+TESTDIR="${TESTDIR:-/tmp}"
```
jcvenegas: could you use mktemp?

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,68 @@
...
+			"Units"  : "items"
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
```
jcvenegas: Does this function remove the created tmpfiles?

GabyCT:
@jcvenegas thanks for the feedback, I am working on this :)

metrics/storage/blogbench.sh
GabyCT: @grahamwhaley, I just added it

metrics/storage/blogbench.sh
GabyCT: @grahamwhaley this is done too

metrics/storage/blogbench.sh
GabyCT: @jcvenegas the tmpfile is created in the container

metrics/lib/common.bash
GabyCT: @jcvenegas this is done too

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,123 @@
...
+	check_dockerfiles_images "$IMAGE" "$DOCKERFILE"
+
+	metrics_json_init
+
+	local output=$(docker run --rm --runtime=$RUNTIME $IMAGE $CMD)
+
+	# Save configuration
+	metrics_json_start_array
+
+	local frequency=$(echo "$output" | grep "Frequency" | cut -d "=" -f2)
```
grahamwhaley: Looks like this line (and some others maybe) don't generate valid JSON (it generates '10 secs', which is not a number, so will need double-quoting for instance).
Other errors I see are the final members of comma separated lists having commas after them. You might have to do a little scripty fun like https://github.com/kata-containers/tests/blob/master/metrics/lib/json.bash#L90-L96 to handle the no-comma final case etc.

My `vim` hightlights these bits of invalid JSON for me, but I believe you can use `jq . blogbench.json` on the results file to pick out invalid JSON syntax.

grahamwhaley: Can we add the `ITERATIONS` value in here as well pls, why not? :-)

metrics/lib/common.bash
```diff
@@ -92,6 +92,24 @@ check_images()
...
+check_dockerfiles_images()
+{
+	local image="$1"
+	local dockerfile_path="$2"
+	if [ -z "$image" ] || [ -z "$dockerfile_path" ]; then
+		die "Missing image or dockerfile path variable"
+		exit 1;
+	fi
+	echo "docker building $image"
+	if ! docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"; then
```
grahamwhaley: This will build every time I think, which is fine - it is better than only ever building once.
Ideally I'd have liked to have buried a 'version' in the Dockerfile itself and then done a dynamic check here to see if we already had the latest version built and installed or not - which saves us a relatively lengthy docker image build every time. But, I think that is a little tricker and maybe for another day. @GabyCT can you open an Issue noting that we should make the metrics automatic docker image builds smarter to reduce un-necessary rebuilds pls.

metrics/storage/blogbench.sh
```diff
@@ -0,0 +1,124 @@
...
+	local reads=$(tail -1 <<< "$output" | awk '{print $6}')
+
+	# Obtaining other Blogbench results
+	local nb_blogs=$(echo "$output" | tail -n +12 | head -n -3 | awk ' BEGIN {ORS="\t"} {print $1} ' | tr '\t' ',' | sed '$ s/.$//')
+	local r_articles=$(echo "$output" | tail -n +12 | head -n -3 | awk ' BEGIN {ORS="\t"} {print $2} ' | tr '\t' ',' | sed '$ s/.$//')
+	local w_articles=$(echo "$output" | tail -n +12 | head -n -3 | awk ' BEGIN {ORS="\t"} {print $3} ' | tr '\t' ',' | sed '$ s/.$//')
+	local r_pictures=$(echo "$output" | tail -n +12 | head -n -3 | awk ' BEGIN {ORS="\t"} {print $4} ' | tr '\t' ',' | sed '$ s/.$//')
+	local w_pictures=$(echo "$output" | tail -n +12 | head -n -3 | awk ' BEGIN {ORS="\t"} {print $5} ' | tr '\t' ',' | sed '$ s/.$//')
+	local r_comments=$(echo "$output" | tail -n +12 | head -n -3 | awk ' BEGIN {ORS="\t"} {print $6} ' | tr '\t' ',' | sed '$ s/.$//')
+	local w_comments=$(echo "$output" | tail -n +12 | head -n -3 | awk ' BEGIN {ORS="\t"} {print $7} ' | tr '\t' ',' | sed '$ s/.$//')
```
jodh-intel: Please can you either use variables for all these magic numbers or (better) create a new variable so you can say something like:

```
local -r data=$(echo "$output" | tail -n +12 | head -n -3)

local nb_blogs=$(echo "$data" | awk ' BEGIN {ORS="\t"} {print $1} ' | tr '\t' ',' | sed '$ s/.$//') )
local r_articles=(echo "$data" | awk ' BEGIN {ORS="\t"} {print $2} ' | tr '\t' ',' | sed '$ s/.$//') )
  :
```

In fact, I wonder if you can simplify this even further to avoid each variable also calling `tr` and `sed`?

metrics/storage/blogbench_dockerfile/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+FROM ubuntu
+
+# URL for blogbench test and blogbench version
+ENV BLOGBENCH_URL "https://download.pureftpd.org/pub/blogbench"
+ENV BLOGBENCH_VERSION 1.1
+
+RUN apt-get update && \
+	apt-get install -y build-essential wget && \
+	wget ${BLOGBENCH_URL}/blogbench-${BLOGBENCH_VERSION}.tar.gz && \
+	tar xvzf blogbench-${BLOGBENCH_VERSION}.tar.gz && \
```
jodh-intel: Nit: You probably (?) don't need to specify the `v` option here?

metrics/storage/blogbench_dockerfile/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Usage: FROM [image name]
+FROM ubuntu
+
+# URL for blogbench test and blogbench version
+ENV BLOGBENCH_URL "https://download.pureftpd.org/pub/blogbench"
+ENV BLOGBENCH_VERSION 1.1
+
+RUN apt-get update && \
+	apt-get install -y build-essential wget && \
```
jodh-intel: Nit: We appear to be (implicitly) standardising on using `curl` rather than `wget`.

GabyCT:
@jodh-intel thanks for the feedback, changes applied :)

jodh-intel:
Thanks @GabyCT.

lgtm

Issue #246, opened 2018-04-19
make get_version function does not depend on go


jcvenegas:
get_version function is used  by the install_go.sh script, but this function uses go to know the versions. Remove that dependency to install go if go is not installed.

PR #247, opened 2018-04-19
ci: lib: get_version does not depend on go

jcvenegas:
get_version function is used by the install_go.sh script, but this
function uses go to know the versions. Remove that the go dependency.

Fixes: #246

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -48,15 +48,53 @@ function clone_build_and_install() {
 	popd
 }
 
+function install_yq() {
+	GOPATH=${GOPATH:-${HOME}/go}
+	local yq_path="${GOPATH}/bin/yq"
+	local yq_pkg="github.com/mikefarah/yq"
+	[ -x  "${GOPATH}/bin/yq" ] && return
+
+	case "$(arch)" in
+	"aarch64")
+		goarch=arm64
+		;;
+
+	"x86_64")
+		goarch=amd64
+		;;
+	"*")
+		echo "Arch $(arch) not supported"
+		exit
+		;;
+	esac
+
+	mkdir -p "${GOPATH}/bin"
+
+	# Workaround to get latest release from github (to not use github token).
+	# Get the redirection to latest release on github.
+	yq_latest_url=$(curl -Ls -o /dev/null -w %{url_effective} "https://${yq_pkg}/releases/latest")
+	# The redirected url should include the latest release version
+	# https://github.com/mikefarah/yq/releases/tag/<VERSION-HERE>
+	yq_version=$(basename "${yq_latest_url}")
+
+
+	local yq_url="https://${yq_pkg}/releases/download/${yq_version}/yq_linux_${goarch}"
+	curl -o "${yq_path}" -L ${yq_url}
+	chmod +x ${yq_path}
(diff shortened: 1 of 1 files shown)


jcvenegas:
Needed by https://github.com/kata-containers/ci/issues/23

.ci/lib.sh
```diff
@@ -48,15 +48,45 @@ function clone_build_and_install() {
 	popd
 }
 
+function install_yq() {
+	GOPATH=${GOPATH:-${HOME}/go}
+	mkdir -p "${GOPATH}/bin"
```
grahamwhaley: niggle: maybe do the arch check before the mkdir?

jcvenegas: to avoid create the dir if the arch is not supported?

jcvenegas:
@grahamwhaley yeah! install yq from distro packages was my first option but I  see it available for any distro :/ .

.ci/lib.sh
```diff
@@ -48,15 +48,45 @@ function clone_build_and_install() {
...
+	"x86_64")
+		goarch=amd64
+		;;
+	"*")
+		echo "Arch $(arch) not supported"
+		exit
+		;;
+	esac
+
+	local yq_url="https://github.com/mikefarah/yq/releases/download/1.14.0/yq_linux_${goarch}"
```
jodh-intel: You can use this API URL to download the latest version without hard-coding:

```
curl -s https://api.github.com/repos/mikefarah/yq/releases/latest|grep browser_download_url|awk '{print $2}'|tr -d '"'|grep "${goarch}"
```

jcvenegas: I could, but doing request to the github api may need a token at some point if is used too many times. I remember had issues in CI metrics before for that.

.ci/lib.sh
grahamwhaley: yeah, that's all - if the arch is not supported, no point making the dir in the first place :-)

.ci/lib.sh
jcvenegas: @jodh-intel  I wonder if something like this may work:

```
version_url=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/mikefarah/yq/releases/latest)
version=$(basename  $version_url)

```

jodh-intel: Neat!

.ci/lib.sh
jcvenegas: fair enough, moved.

.ci/lib.sh
jcvenegas: Changed to proposed  solution in this comment.

jodh-intel:
lgtm

jcvenegas:
CI filing by issue: 
https://github.com/kata-containers/tests/issues/250

chavafg:
lgtm

jodh-intel:
The `jenkins-ci-ubuntu-16-04 ` CI job *did* pass, so some sort of comms issue between jenkins+github it seems:

- http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-tests-ubuntu-16-04-PR/345/consoleFull

Merging...

PR #261, opened 2018-04-25
tests: Add test to check if a block device is passed as volume.

### LINKED ISSUE #191: Add test to check if a block device passed as a volume is passed to VM using a block driver instead of 9p.
Test needs to be added after https://github.com/kata-containers/runtime/pull/138 lands.

GabyCT:
This will check if a block device passed as a volume is passed to VM using a block driver as well as it adds a tests that checks if a text file under /dev is passed to a container and check that contents are the same.

Fixes #191

### DIFF
--- a/integration/docker/volume_test.go
+++ b/integration/docker/volume_test.go
@@ -5,10 +5,12 @@
 package docker
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 
+	"github.com/kata-containers/tests"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
@@ -23,6 +25,9 @@ var _ = Describe("docker volume", func() {
 		fileTest      string = "hello"
 		exitCode      int
 		stdout        string
+		loopFile      string
+		err           error
+		diskFile      string
 	)
 
 	Context("create volume", func() {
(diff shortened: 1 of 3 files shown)


jodh-intel:
Hi @GabyCT - please can you add more detail to the commit explaining why this (frankly seemingly bizarre) test is required? :smile:

grahamwhaley:
:-) I did wonder if passing a dummy dev made with a `mknod` was more appropriate - so, yes, some more details would help clarify that.

integration/docker/volume_test.go
```diff
@@ -90,4 +90,20 @@ var _ = Describe("docker volume", func() {
 			Expect(stdout).NotTo(ContainSubstring(testFile))
 		})
 	})
+
+	Context("creating a text file under /dev", func() {
+		It("should be passed the content to the container", func() {
+			fileName := "/dev/foo"
```
devimc: you should remove it once the test finish

integration/docker/volume_test.go
```diff
@@ -90,4 +96,58 @@ var _ = Describe("docker volume", func() {
...
 	})
+
+	Context("creating a text file under /dev", func() {
+		It("should be passed the content to the container", func() {
+			if os.Getuid() != 0 {
+				Skip("only root user can create files under /dev")
+			}
+			fileName := "/dev/foo"
+			textContent := "hello"
+			err := ioutil.WriteFile(fileName, []byte(textContent), 0644)
```
devimc: `err` was declared in the `var` block

integration/docker/volume_test.go
```diff
@@ -90,4 +96,58 @@ var _ = Describe("docker volume", func() {
...
+		})
+	})
+
+	Context("passing a block device", func() {
+		It("should be mounted", func() {
+			if os.Getuid() != 0 {
+				Skip("only root user can create loop devices")
+			}
+
+			for i := 0; i < loopDevices; i++ {
```
devimc: `loopDevices` is 1, you don't need a loop to create 1 loopdevice

integration/docker/volume_test.go
```diff
@@ -90,4 +96,58 @@ var _ = Describe("docker volume", func() {
...
+
+			for _, lf := range loopFiles {
+				err = deleteLoopDevice(lf)
+				Expect(err).ToNot(HaveOccurred())
+			}
+
+			for _, df := range diskFiles {
+				err = os.Remove(df)
+				Expect(err).ToNot(HaveOccurred())
+			}
```
devimc: you don't need two arrays (`loopFiles` and `diskFiles`) for 1 loop device

integration/docker/volume_test.go
```diff
@@ -23,6 +23,12 @@ var _ = Describe("docker volume", func() {
 		fileTest      string = "hello"
 		exitCode      int
 		stdout        string
+		loopDevices   int = 1
```
devimc: you don't need this variable

integration/docker/volume_test.go
devimc: `err = deleteLoopDevice(loopFile)` and `err = os.Remove(diskFile)`  should be enough

integration/docker/volume_test.go
```diff
@@ -90,4 +93,50 @@ var _ = Describe("docker volume", func() {
...
+	Context("passing a block device", func() {
+		It("should be mounted", func() {
+			if os.Getuid() != 0 {
+				Skip("only root user can create loop devices")
+			}
+
+			diskFile, loopFile, err = createLoopDevice()
+			Expect(err).ToNot(HaveOccurred())
+
+			args = []string{"--name", id, "--cap-add=SYS_ADMIN", "-v", loopFile + ":" + loopFile, DebianImage, "mount", loopFile, "/mnt"}
```
devimc: to have `mount` working you have to format the loop device before mounting, so you have to do something like this

```go
loopFileP1 = fmt.Sprintf("%sp1", loopFile)
mkfsCmd := tests.NewCommand("mkfs.ext4", loopFileP1)
if _, stderr, exitCode := mkfsCmd.Run(); exitCode != 0 {
	return "", "", fmt.Errorf("%s", stderr)
}
args = []string{"--name", id, "--cap-add=SYS_ADMIN", "-v", loopFileP1 + ":" + loopFileP1, DebianImage, "mount", loopFileP1, "/mnt"}
```

GabyCT: @devimc thanks for the feedback

devimc: updated example code

GabyCT:
@jodh-intel thanks for the feedback, I made the changes

chavafg:
lgtm

PR #266, opened 2018-04-26
CI: setup: Make kata installation optional

jcvenegas:
This PR add a logic to make kata installation optional, this would help to test if kata was installed with packages.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -27,6 +27,10 @@ else
 	exit 1
 fi
 
+if ! command -v docker > /dev/null; then
+        "${cidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker install
+fi
+
 if [ "$arch" = x86_64 ]; then
 	if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
 		echo "enable Nested Virtualization"
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -37,23 +37,8 @@ else
 	die "Unsupported architecture: $arch"
 fi
 
-echo "Install Kata Containers image"
-bash -f "${cidir}/install_kata_image.sh"
-
-echo "Install Kata Containers Kernel"
-"${cidir}/install_kata_kernel.sh"
-
-echo "Install Qemu"
-bash -f "${cidir}/install_qemu.sh"
-
-echo "Install Kata Containers shim"
-bash -f "${cidir}/install_shim.sh"
-
-echo "Install Kata Containers proxy"
-bash -f "${cidir}/install_proxy.sh"
-
-echo "Install Kata Containers runtime"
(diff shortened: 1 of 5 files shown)


.ci/setup.sh
```diff
@@ -37,17 +38,17 @@ else
...
+if [ "${INSTALL_KATA}" == "TRUE" ];then
+	echo "Install Kata from sources"
+	bash -f ${cidir}/install_kata_from_source.sh
+fi
 
-echo "Install runtime"
-bash -f ${cidir}/install_runtime.sh
+# Install docker after install kata image
+# install_kata_from_source.sh: Creates the image from source
+# and requires runc as default runtime
```
chavafg: To create the image, we need docker, right? Then we need to install docker first and then call `install_kata_from_source.sh` ?

jcvenegas: -_- ups.

.ci/setup.sh
```diff
@@ -40,6 +40,13 @@ fi
 echo "Install Kata from sources"
 bash -f ${cidir}/install_kata_from_source.sh
 
+# Install docker after install kata image
```
grahamwhaley: So I think maybe this is not so much 'install docker', as 'install kata into docker', is it? As we now know that docker will always already be installed, as the `install_kata_from_source.sh` would not have worked otherwise?

.ci/setup.sh
```diff
@@ -14,6 +14,7 @@ source "${cidir}/lib.sh"
 apply_depends_on
 
 arch=$(arch)
+INSTALL_KATA="${INSTALL_KATA:-TRUE}"
```
grahamwhaley: So, I think these patches allow us to do two things...
- Install kata from sources, without doing any of the other package installs etc. (allows us to check a system has all the dependencies it needs?)
- Allows us to skip installing kata from source (this patch), which then allows us to run all the tests on an already installed system (such as if we are testing packages coming out of OBS?).

If I'm right - maybe it is also time we have a small overview document of what scripts are useful for what situation, and how to configure/invoke them?
The .ci directory is getting quite busy and complex - probably time we started trying to document it a bit.

.ci/install_kata_from_source.sh
```diff
@@ -0,0 +1,32 @@
...
+
+cidir=$(dirname "$0")
+source /etc/os-release
+source "${cidir}/lib.sh"
+
+echo "Install kata-containers image"
+"${cidir}/install_kata_image.sh"
+
+echo "Install Kata Containers Kernel"
+"${cidir}/install_kata_kernel.sh" "latest"
```
devimc: "latest" is no more needed

sboeuf:
@jcvenegas you need to rebase this PR.

chavafg:
lgtm

Issue #271, opened 2018-05-02
Modify CI setup to be able to execute virtio-blk


chavafg:
We need to support `virtio-blk` in addition to `virtio-scsi` in the CI.

GabyCT:
Closing this issue as we already have support for virtio-blk

PR #280, opened 2018-05-04
CI: add support for adding virtio-blk job

chavafg:
In addition to the current CI jobs, which currently test
using virtio-scsi driver, we need to also test virtio-blk
as we also support this driver but using qemu-lite 2.7

Fixes: #271.

### DIFF
--- a/.ci/install_qemu_lite.sh
+++ b/.ci/install_qemu_lite.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+arch=$(arch)
+
+if [ "$#" -ne 3 ]; then
+    echo "Usage: $0 <CLEAR_RELEASE> <QEMU_LITE_VERSION> <DISTRO>"
+    echo "       Install the QEMU_LITE_VERSION from clear CLEAR_RELEASE."
+    exit 1
+fi
+
+clear_release="$1"
+qemu_lite_version="$2"
+distro="$3"
+qemu_lite_bin="qemu-lite-bin-${qemu_lite_version}.${arch}.rpm"
+qemu_lite_data="qemu-lite-data-${qemu_lite_version}.${arch}.rpm"
+
+echo -e "Install qemu-lite ${qemu_lite_version}"
+
+# download packages
+curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/${arch}/os/Packages/${qemu_lite_bin}"
+curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/${arch}/os/Packages/${qemu_lite_data}"
+
+# install packages
+if [ "$distro" == "ubuntu" ];  then
+	sudo alien -i "./${qemu_lite_bin}"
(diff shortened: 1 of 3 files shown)


.ci/install_qemu_lite.sh
```diff
@@ -0,0 +1,41 @@
...
+clear_release="$1"
+qemu_lite_version="$2"
+distro="$3"
+qemu_lite_bin="qemu-lite-bin-${qemu_lite_version}.${arch}.rpm"
+qemu_lite_data="qemu-lite-data-${qemu_lite_version}.${arch}.rpm"
+
+echo -e "Install qemu-lite ${qemu_lite_version}"
+
+# download packages
+curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/${arch}/os/Packages/${qemu_lite_bin}"
```
jodh-intel: At the risk of stating the obvious, because we're re-using CLR binaries, this is only going to work for x86_64. As such, I'd add a test below the `arch=$(arch)` line to die if any other architecture is detected for the time being.

/cc @bergwolf.

.ci/install_qemu_lite.sh
```diff
@@ -0,0 +1,41 @@
...
+curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/${arch}/os/Packages/${qemu_lite_data}"
+
+# install packages
+if [ "$distro" == "ubuntu" ];  then
+	sudo alien -i "./${qemu_lite_bin}"
```
jodh-intel: I appreciate this is what we did for Clear Containers too, but this still worries me rather as Clear Linux is pretty much always going to be using the latest versions of all libraries, compilers and linkers which might result in trouble when we try to install these binaries on distros using older versions of those items.

.ci/setup.sh
```diff
@@ -37,8 +37,15 @@ else
 	die "Unsupported architecture: $arch"
 fi
 
+
+# Use qemu-lite 2.7 for virtio-blk as there is
+# a bug using qemu>=2.9
 echo "Install Qemu"
-bash -f ${cidir}/install_qemu.sh
+if [ "$USE_VIRTIO_BLK" == true ]; then
+	bash -f ${cidir}/install_qemu_lite.sh "22280" "741f430a960b5b67745670e8270db91aeb083c5f-31" "$ID"
```
jodh-intel: Why (only) release `22280`? Would `current` work too?

katabuilder:
testComment

katabuilder:
testComment02

katabuilder:
This is a test (disregard these numbers)
PSS Measurement:
Qemu: 145937 KB
Proxy: 6806 KB
Shim: 10899 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2014628 KB

katabuilder:
PSS Measurement:
Qemu:  
Proxy:  
Shim:  

Memory inside container:
Total Memory:  
Free Memory:

jodh-intel:
Hi @chavafg - this PR is almost antique :) Do we still need it? If so, please could you update or close.

grahamwhaley:
Just to check, is this still something we plan to get running @chavafg ?

chavafg:
I think we can close this, open for discussion if we still plan to add it.

PR #284, opened 2018-05-07
teardown: collect logs using sudo

### LINKED ISSUE #283: teardown.sh: collect log files using sudo
We need to collect docker, kubelet and kata logs using sudo to get logs. If not, we can get this error:

```
Hint: You are currently not seeing messages from other users and the system.
      Users in the 'systemd-journal' group can see all messages. Pass -q to
      turn off this notice.
```

chavafg:
Add sudo to the journalctl commands that collect the
logs from kata components, docker and kubelet.

Fixes #283.


jcvenegas:
lgtm

PR #319, opened 2018-05-18
CI: Sanitise "yq" output

### LINKED ISSUE #318: Sanitise output of yq command
The `yq` command used by `.ci/lib.sh` (github.com/mikefarah/yq) returns `null` if the specified key does not exist.

Change it to return the empty string.

jodh-intel:
The `yq` command used by `.ci/lib.sh` returns `null` if the specified key does not exist.

Change it to return the empty string for a saner experience.

Fixes #318.


chavafg:
lgtm

PR #328, opened 2018-05-21
cmd: Add kata-manager script to install from packages

### LINKED ISSUE #327: Add a command to install+configure the packaged version of Kata
We can now generate scripts from documents using https://github.com/kata-containers/tests/blob/master/.ci/kata-doc-to-script.sh.

We need a command that can install the packaged version of Kata, which we currently generate using the OBS scripts in https://github.com/kata-containers/packaging. This command will use `kata-doc-to-script.sh` to handle the installation meaning we'll be testing our install documentation by "executing" it.

Once we have this we will be able to tackle testing the images and initrd's that osbuilder generates (https://github.com/kata-containers/osbuilder/issues/97) by running this script which will read the install doc and execute it to install the packaged version of kata. All osbuilder will then need to do is iterate through all the distros it supports calling this script and ensuring each image+initrd can be used to book a container.

jodh-intel:
Created a script that will install the packaged version of Kata and can
also configure the runtime for an initrd/image, toggle debug and also
uninstall the packages.

Fixes #327.


jodh-intel:
Hi @chavafg, @jcvenegas - as @chavafg said earlier, it may make sense to merge this with `cmd/container-manager/manage_ctr_mgr.sh`?

I've only tested on Ubuntu so far. It would be great if you could test it with CentOS too (https://github.com/kata-containers/documentation/pull/111).

chavafg:
Sure, I can give it a try and post back.

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,258 @@
...
+
+  -c <file> : Specify full path to configuration file
+              (default: '$default_config_file').
+  -h        : Display this help.
+  -v        : Verbose output.
+
+Commands:
+
+  enable-debug   : Enable full debug for all system components.
+  install        : Install the packaged version of Kata Containers.
```
chavafg: this should be `install-packages` as `install` fails:
```
[fuentess@centos-kata-manager kata-manager]$ ./kata-manager.sh install
ERROR: invalid command: 'install'
```

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,258 @@
...
+		die "invalid distro: '$distro'"
+	fi
+}
+
+setup()
+{
+	source /etc/os-release
+
+	# lower-case distribution name
+	distro=$(echo "$NAME"|tr '[A-Z]' '[a-z]')
```
chavafg: I think we should better use `$ID` instead of `$NAME`.
In CentOS:
`NAME="CentOS Linux"`

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,258 @@
...
+  -h        : Display this help.
+  -v        : Verbose output.
+
+Commands:
+
+  enable-debug   : Enable full debug for all system components.
+  install        : Install the packaged version of Kata Containers.
+  configure-image  : Configure the runtime to use the specified image.
+  configure-initrd : Configure the runtime to use the specified initial ramdisk.
+
```
chavafg: add `remove-package` command?

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,258 @@
...
+				;;
+		esac
+	done
+
+	shift $[$OPTIND-1] || true
+	cmd="$1"
+
+	shift || true
+
+	[ -z "$cmd" ] && die "need command"
```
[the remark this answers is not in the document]
jodh-intel: Good catch - fixed.

cmd/kata-manager/kata-manager.sh
jodh-intel: Fixed.

cmd/kata-manager/kata-manager.sh
jodh-intel: Another good catch - fixed.

cmd/kata-manager/kata-manager.sh
jodh-intel: Fixed.

jodh-intel:
Tested on:

- Centos 7 (7.5.1804)
- Fedora 27
- Ubuntu 17.10

jodh-intel:
Hi @chavafg - I've just updated the script so that it now only modifies the `/etc/kata-containers/configuration.toml` file. This provides a couple of important advantages:

- It makes the script safe for developers to run.
- It allowed me to add a `reset-config` command to undo any config changes.

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,318 @@
...
+
+	local dir=$(dirname "$local_config_file")
+	sudo rmdir "$dir" 2>/dev/null || true
+}
+
+setup()
+{
+	source /etc/os-release
+
+	# lower-case distribution name
```
chavafg: this is no longer changing to lower-case here. modify or maybe delete comment?

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,318 @@
...
+		ubuntu)
+			packages=$(dpkg-query -W -f='${Package}\n'|egrep "${packages_regex}" || true)
+			;;
+
+		*)
+			die "invalid distro: '$distro'"
+			;;
+	esac
+
+	[ -z "$packages" ] && { echo "packages not installed" && exit 1; }
```
chavafg: we could use `die` here

jodh-intel: Good point - fixed.

cmd/kata-manager/kata-manager.sh
jodh-intel: Comment zapped.

jodh-intel:
Hi @devimc, @jcvenegas, @grahamwhaley - ptal.

grahamwhaley:
@jcvenegas , would appreciate your install/packaging eyeballs on this one pls? :-)

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,325 @@
...
+
+	enable_initrd "$file"
+	disable_image
+}
+
+# Install the packaged version of Kata by executing the commands
+# specified in the installation guide document.
+cmd_install_packages()
+{
+	[ -z "${GOPATH}" ] && die "need GOPATH"
```
grahamwhaley: `GOPATH` is now not a requirement to be set for `go` - maybe you can use a `go env GOPATH` here? - after maybe checking `go` is actually installed? ;-)

jodh-intel: Fixed.

cmd/kata-manager/kata-manager.sh
```diff
@@ -0,0 +1,328 @@
...
+	sudo sed -i 's/^\(initrd *=.*\)/# \1/g' "$config_file"
+}
+
+enable_image()
+{
+	local file="$1"
+
+	config_checks
+
+	sudo sed -i "s!^#*.*image *=.*\$!image = \"$file\"!g" "$config_file"
```
jcvenegas: The image will be modified if is commented , should the regex also replace if the image is not commented?

jodh-intel: It actually will since the first part of the regex is `#*` - aka "zero or more hashes", so the hash can be absent.

Issue #341, opened 2018-05-24
CRI-O: use block device when using devicemapper driver on tests


chavafg:
CRI-O is deprecating the use of loopback devices when running with devicemapper driver. Then we need to be able to specify a device when we run our tests in our CI.

PR #342, opened 2018-05-25
CRI-O: Use block device for devicemapper driver

chavafg:
Now that cri-o is deprecating the use of loop back
devices when using devicemapper driver, we need to
use a block device on the Azure VMs from our CI.

This fix assumes that the Azure VM will have an
/dev/sdb device and it will only be executed when
CI=true, to prevent from being executed on a
development machine.

Fixes: #341

### DIFF
--- a/integration/cri-o/cri-o.sh
+++ b/integration/cri-o/cri-o.sh
@@ -56,7 +56,27 @@ fi
 # This will allow us to run tests with at least 2 different
 # storage drivers.
 if [ "$ID" == "ubuntu" ] && [ "$MAJOR" -ge 17 ]; then
-	export CRIO_STORAGE_DRIVER_OPTS="--storage-driver overlay"
+	export STORAGE_OPTIONS="--storage-driver overlay"
+fi
+
+
+# If we are testing a CRI-O PR in a CI environment,
+# use devicemapper driver using a block device.
+# This is needed since support for loopback devices
+# is being deprecated.
+# More info: https://github.com/kubernetes-incubator/cri-o/pull/1574
+#            https://github.com/containers/storage/pull/80/commits/da1e7e5d28da7d0a1e6ac0d4a9e647bee72282e2
+if [ "$ghprbGhRepository" == "${crio_repository/github.com\/}" ] && [ "$CI" == true ] && [ -z "${KATA_DEV_MODE}" ] ;then
+	# block device attached to the Azure VM where we run the CI
+	# if the block device has a partition, cri-o will not be able to use it.
+	block_device=/dev/sdb
+	if sudo fdisk -l "$block_device" | grep "${block_device}[1-9]"; then
+		die "detected partitions on block device: ${block_device}. Will not continue"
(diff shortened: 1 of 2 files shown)


chavafg:
@sboeuf this should solve the problem with the CRI-O PR https://github.com/kubernetes-incubator/cri-o/pull/1574 
But we also need some changes in how the cri-o manages the tests: https://github.com/kubernetes-incubator/cri-o/issues/1576 

Also, I am assuming that this tests would have an `/dev/sdb` device that is available for use. So it might be dangerous to run on a development environment. I am adding a condition to only run these tests when `CI=true`, but not sure if this is enough.

integration/cri-o/cri-o.sh
```diff
@@ -56,7 +56,27 @@ fi
...
+# If we are testing a CRI-O PR in a CI environment,
+# use devicemapper driver using a block device.
+# This is needed since support for loopback devices
+# is being deprecated.
+# More info: https://github.com/kubernetes-incubator/cri-o/pull/1574
+#            https://github.com/containers/storage/pull/80/commits/da1e7e5d28da7d0a1e6ac0d4a9e647bee72282e2
+if [ "$ghprbGhRepository" == "${crio_repository/github.com\/}" ] && [ "$CI" == true ];then
+	# block device attached to the Azure VM where we run
+	# the kata-containers CI
+	block_device=/dev/sdb
```
sboeuf: To go further into the protection of a developer machine, we should start at `/dev/sde` and increment with `/dev/sdf`, `/dev/sdg`, ... until we find a path/device that does not exist. This way, we make sure we don't override an existing device.

chavafg: @sboeuf `/dev/sdb` is the device that will be available on the Azure machines. I don't think we can manage to go until `/dev/sde` unless we attach more disks to the VM. 
What I was thinking is that I can wipe the partitions of `/dev/sdb` in the VM startup script of the Jenkins setup. This way we do not mess with developer machines.

jodh-intel: +10000000 for lots more paranoia in this scenario.

Can you also add a check so that if `KATA_DEV_MODE` is set (even if `CI` is set), we do *NOT* do this to be more paranoid about dev mode:

- https://github.com/kata-containers/tests#developer-mode

For extra safety, we could potentially only run this if we're in a VM. I'm not sure how we check this reliably though since the following only seems to work on `x86_64`:

```
$ egrep -m1 "flags.*:.*\<hypervisor\>" /proc/cpuinfo
```

chavafg: @jodh-intel I can also check that `sdb` does not have a partition before running the tests. I'll add the  `KATA_DEV_MODE` check to not run the tests if it is set

chavafg:
@sboeuf, @jodh-intel updated to address your comments.

jodh-intel:
oh - no more disk reformatting required? :)

lgtm

PR #344, opened 2018-05-25
 CI: Install kernel from packages if up-to-date 

chavafg:
Instead of building the kernel from sources, we can
use the packaged version if there hasn't been any
changes. This will allow us to save time in the CI
jobs.

In addition, this PR contains two other changes:
- Install OBS repositories (needed to install kata packages).
- Move the kata image installation to the main setup script instead 
   calling it on every distro setup.

### DIFF
--- a/.ci/install_kata_kernel.sh
+++ b/.ci/install_kata_kernel.sh
@@ -12,6 +12,7 @@ set -e
 
 cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
+source "/etc/os-release"
 
 repo_name="packaging"
 repo_owner="kata-containers"
@@ -20,6 +21,7 @@ kernel_arch="$(arch)"
 get_kernel_url="https://cdn.kernel.org/pub/linux/kernel"
 tmp_dir="$(mktemp -d)"
 hypervisor="kvm"
+packaged_kernel="kata-linux-container"
 
 download_repo() {
 	pushd ${tmp_dir}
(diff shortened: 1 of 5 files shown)


.ci/install_kata_kernel.sh
```diff
@@ -72,16 +85,33 @@ build_and_install_kernel() {
...
 
 main() {
-	download_repo
-	kernel_version=$(get_kernel_version)
-	download_kernel ${kernel_version}
-	build_and_install_kernel ${kernel_version}
-	cleanup
+	current_kernel_version=$(get_current_kernel_version)
+	packaged_kernel_version=$(get_packaged_kernel_version)
+	if [ "$packaged_kernel_version" == "$current_kernel_version" ] && [ "$kernel_arch" == "x86_64" ]; then
```
jcvenegas: correct me if I am wrong but I think this wont guarantee that the kernel has the same configuration and patches. 

For this we will need a way to know the status/version of the kernel configuration (config and patches). 

One way (what we did in the past is to have a file that we need to increment every time we update the kernel and config. We could add that number as part of the package version. And append it to the extra kernel version (to verify when uname -a is used).

jcvenegas:
```
packaging/
--> kernel/
     ----> kernel.yaml
```
```
patchset_version=1
```
If a patch/config is added or modified we bump patch-set version.

If check if the packaged kernel version is (kernel_version-patchset-version-kernel_pkg-version) if not build from source

chavafg:
@jcvenegas I have updated my PR to address your comment above, still need to merge https://github.com/kata-containers/packaging/pull/44 and fix https://github.com/kata-containers/packaging/pull/45

.ci/install_kata_kernel.sh
jcvenegas: @chavafg  https://github.com/kata-containers/packaging/pull/46 now adds an extra number to the kernel package version.

chavafg:
this is ready for review

jodh-intel:
lgtm

PR #348, opened 2018-05-28
tests: Fix aerospike test

### LINKED ISSUE #347: Fix aerospike test for popular docker hub images
Fix aerospike test for popular docker hub images testing

GabyCT:
Fix aerospike test for popular docker hub images as there were changes at the Docker Hub image.

Fixes #347

### DIFF
--- a/integration/popular_docker_hub_images/popular_docker_images.bats
+++ b/integration/popular_docker_hub_images/popular_docker_images.bats
@@ -13,11 +13,21 @@ setup() {
 	clean_env
 }
 
-@test "[volume mkdir] create a directory in an aerospike container" {
-	tmpdir=$(mktemp -d)
-	docker run --rm --runtime=$RUNTIME -d -v $tmpdir:/opt/aerospike/data aerospike/aerospike-server timeout 5 /entrypoint.sh asd
-	[ -n ${tmpdir} ] || false
-	rm -rf $tmpdir
+@test "[insert data] insert data in an aerospike container" {
+	docker run --runtime=$RUNTIME -d --name aerospike aerospike/aerospike-server
+	status=1
+	set +e
+	for i in $(seq 1 5); do
+		docker run --rm --runtime=$RUNTIME -i aerospike/aerospike-tools aql -h $(docker inspect -f '{{.NetworkSettings.IPAddress}}' aerospike) -c "insert into test.foo (PK, foo) values ('123','any'); select * from test.foo"
+		if [ $? == 0 ]; then
+			status=0
+			break
+		fi
+		sleep 1
+	done
+	set -e
+	docker rm -f aerospike
+	return $status
 }
 
 @test "[display text] hello world in an alpine container" {


integration/popular_docker_hub_images/popular_docker_images.bats
```diff
@@ -13,11 +13,8 @@ setup() {
 	clean_env
 }
 
-@test "[volume mkdir] create a directory in an aerospike container" {
-	tmpdir=$(mktemp -d)
-	docker run --rm --runtime=$RUNTIME -d -v $tmpdir:/opt/aerospike/data aerospike/aerospike-server timeout 5 /entrypoint.sh asd
-	[ -n ${tmpdir} ] || false
-	rm -rf $tmpdir
+@test "[display configuration] display port in configuration file in an aerospike container" {
+	docker run --rm --runtime=$RUNTIME aerospike cat /etc/aerospike/aerospike.conf | grep port
```
devimc: I think that would be better to have a test where data is inserted in a aerospike server

```
docker run --rm --runtime=$RUNTIME  -d --name aerospike aerospike/aerospike-server
sleep 10
docker run --rm --runtime=$RUNTIME -it aerospike/aerospike-tools aql -h  $(docker inspect -f '{{.NetworkSettings.IPAddress }}' aerospike) -c "insert into test.foo (PK, foo) values ('123','any') ; select * from test.foo"
docker rm -f aerospike
```

@GabyCT what do you think?

GabyCT:
@devimc changes applied

integration/popular_docker_hub_images/popular_docker_images.bats
```diff
@@ -13,11 +13,11 @@ setup() {
...
 }
 
-@test "[volume mkdir] create a directory in an aerospike container" {
-	tmpdir=$(mktemp -d)
-	docker run --rm --runtime=$RUNTIME -d -v $tmpdir:/opt/aerospike/data aerospike/aerospike-server timeout 5 /entrypoint.sh asd
-	[ -n ${tmpdir} ] || false
-	rm -rf $tmpdir
+@test "[insert data] insert data in an aerospike container" {
+	docker run --runtime=$RUNTIME -d --name aerospike aerospike/aerospike-server
+	sleep 10
```
jodh-intel: Is this waiting for aerospike to start up? How do we know this value is going to be reliable?

jodh-intel:
I really don't understand this. Please can you document in the commit why this change was required to fix aerospike?

integration/popular_docker_hub_images/popular_docker_images.bats
devimc: this `sleep` is to wait for `aerospike-server` initialization

I guess something like this can replace the `sleep`

```sh
set +e
for i in $(seq 1 5); do
    docker run --rm --runtime=$RUNTIME -i aerospike/aerospike-tools aql -h $(docker inspect -f '{{.NetworkSettings.IPAddress}}' aerospike) -c "insert into test.foo (PK, foo) values ('123','any'); select * from test.foo"
    if [ $? == 0 ]; then
        set -e
        return 0
    fi
done
set -e
return 1
```

jodh-intel:
lgtm

PR #351, opened 2018-05-29
integration/docker: read default number of vCPUs from config file

### LINKED ISSUE #350: read runtime configuration file to get the default number of vCPUs
instead of run a container with `nproc` as workload to get the default number of vCPUs, we should read the configuration file since it is more reliable

devimc:
Read the default number of vCPUs from config file is more reliable
than runnning a container with nproc as workload

fixes #350

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,11 @@
 
 
 [[projects]]
+  name = "github.com/BurntSushi/toml"
+  packages = ["."]
+  revision = "a368813c5e648fee92e5f6c30e3944ff9d5e8895"
+
+[[projects]]
   name = "github.com/davecgh/go-spew"
   packages = ["spew"]
   revision = "346938d642f2ec3594ed81d874461961cd0faa76"
@@ -91,6 +96,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "ae48206f96c44f95169510ceb62c21ca20643a08bb6411387a6e423a2a3c19b5"
+  inputs-digest = "bb7757ba33c6529f87e8a29631748c7d73459fdf397b8d55bdcd1c0f583b2049"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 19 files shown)


Gopkg.lock
```diff
@@ -1,6 +1,11 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
+  name = "github.com/BurntSushi/toml"
+  packages = ["."]
+  revision = "b26d9c308763d68093482582cea63d69be07a0f0"
```
grahamwhaley: Any chance we could move this commit, maybe to the head commit. This is tied to the 0.3.0 release I believe, but I'd like to pick up [this commit](https://github.com/BurntSushi/toml/commit/4678bf20d0945128c23c31e748cde56476e8fd91) In particular to get the updated license for that package :-)

integration/docker/main_test.go
```diff
@@ -17,11 +19,33 @@ const (
...
+
 func randomDockerName() string {
 	return RandID(30)
 }
 
 func TestIntegration(t *testing.T) {
+	var err error
+	runtimeConfigPath := DefaultRuntimeConfigPath
+
+	args := []string{"--kata-show-default-config-paths"}
```
jodh-intel: Nice

integration/docker/main_test.go
```diff
@@ -17,11 +19,33 @@ const (
...
+	cmd := NewCommand(Runtime, args...)
+	stdout, _, exitCode := cmd.Run()
+	if exitCode == 0 && stdout != "" {
+		for _, c := range strings.Split(stdout, "\n") {
+			if _, err = os.Stat(c); os.IsExist(err) {
```
jodh-intel: This doesn't look right to me - I don't think `IsExist()` does what we want here as `err` is expected to be `nil` when `os.Stat()` succeeds.

Gopkg.lock
devimc: done, thanks

devimc:
@jodh-intel @grahamwhaley  changes applied, thanks

jodh-intel:
lgtm

chavafg:
lgtm

PR #355, opened 2018-05-29
CI: specify runc version that we use for crio

### LINKED ISSUE #354: specify runc version that we use for cri-o
In `install_crio.sh` we clone runc and checkout to a `runc_version`, but there is no value assigned to that variable, then we are not checking out to a specific version and we end up building master.
We need to get the version from the runtime `versions.yaml`

chavafg:
We use a `runc_version` variable for building an specific
runc version, but there is no value assigned to that variable.
This fixes it by getting the value from versions.yaml in the
runtime repository.

Fixes #354.

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -74,6 +74,7 @@ sudo install -m0444 test/policy.json "$containers_config_path"
 popd
 
 echo "Install runc for CRI-O"
+runc_version=$(get_version "externals.runc.version")
 go get -d github.com/opencontainers/runc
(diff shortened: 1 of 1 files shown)


.ci/install_crio.sh
```diff
@@ -74,6 +74,7 @@ sudo install -m0444 test/policy.json "$containers_config_path"
 popd
 
 echo "Install runc for CRI-O"
+runc_version=$(get_version "externals.runc.version")
```
jodh-intel: Nice! :smile:

PR #364, opened 2018-05-31
CI: Install qemu-lite from OBS if it is updated.

### LINKED ISSUE #363: CI: Install qemu from OBS if it is available and updated
To reduce CI time, we can install `qemu-lite` from OBS if the packaged version is the same as in `version.yaml` from the runtime repository.

chavafg:
If the qemu-lite package version on OBS is the same as the
one in the 'versions.yaml', we should install the package
from OBS instead of builing from sources.
This allows to save CI time.

Fixes #363.

### DIFF
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -9,41 +9,79 @@ set -e
 
 cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
+source /etc/os-release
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm.

I still wonder if we could encapsulate all this sort of logic in https://github.com/kata-containers/tests/blob/master/cmd/kata-manager/kata-manager.sh though (by adding commands like `install-hypervisor`).

GabyCT:
lgtm

PR #366, opened 2018-05-31
metrics: Add cpu information test

### LINKED ISSUE #365: metrics: Add cpu information test
This will retrieve cpu information like instructions, cycles and instructions per cycle using `perf` while running an `iperf3` network benchmark.

GabyCT:
This will retrieve cpu information like instructions,
cycles and instructions per cycle using perf while
running an iperf3 network benchmark.

Fixes #365

### DIFF
--- a/.ci/run_metrics_PR_ci.sh
+++ b/.ci/run_metrics_PR_ci.sh
@@ -95,6 +95,9 @@ run() {
 	# Run the density test inside the container
 	bash density/memory_usage_inside_container.sh
 
+	# Run the cpu statistics test
+	bash network/cpu_statistics_iperf.sh
+
 	popd
 }
 
--- a/metrics/network/cpu_statistics_iperf.sh
+++ b/metrics/network/cpu_statistics_iperf.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Description of the test:
+# This test runs the 'iperf' network benchmark and
+# we retrieve cpu information while running the networking
+# test with 'perf'. CPU utilization provides more stable
+# results that help to detect performace regressions compared
+# with the actual throughput.
+
+set -e
+
+# General env
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+# Test name
+TEST_NAME="${TEST_NAME:-cpu information}"
+# Image name
+IMAGE="${IMAGE:-local-iperf}"
+# Dockerfile
+DOCKERFILE="${SCRIPT_PATH}/iperf_dockerfile/Dockerfile"
+# Client name
+CLIENT_NAME="${CLIENT_NAME:-client_iperf}"
+# Server name
+SERVER_NAME="${SERVER_NAME:-server_iperf}"
+# Target bandwidth
+BANDWIDTH="${BANDWIDTH:-1G}"
+# Time in seconds to transmit
+TIMEOUT="${TIMEOUT:-160}"
+# Sample period in seconds to measure
+SAMPLE_PERIOD="${SAMPLE_PERIOD:-120}"
+# File to save perf results
+TMP_FILE=$(mktemp cpuinfo.XXXXXXXXXX)
+# Time in seconds is required to let the benchmark settle down
+SETTLE_TIME="${SETTLE_TIME:-20}"
+
+function remove_tmp_file() {
+	rm -rf $TMP_FILE
+}
+
+trap remove_tmp_file EXIT
+
+function main() {
+	# Check tools/commands dependencies
+	cmds=("awk" "docker" "perf")
+
+	init_env
+	check_cmds "${cmds[@]}"
+	check_dockerfiles_images "$IMAGE" "$DOCKERFILE"
+
+	# Start iperf server configuration
+	# Set the TMPDIR to an existing tmpfs mount to avoid a 9p unlink error
+	local init_cmds="export TMPDIR=/dev/shm"
+	local server_cmd="$init_cmds && iperf -s"
+	docker run --runtime $RUNTIME --name $SERVER_NAME -d $IMAGE bash -c "$server_cmd"
+	local server_address=$(docker inspect --format "{{.NetworkSettings.IPAddress}}" $SERVER_NAME)
+
+	# Start iperf client
+	local client_cmd="$init_cmds && iperf -c $server_address -b $BANDWIDTH -t $TIMEOUT"
+	docker run --runtime $RUNTIME --name $CLIENT_NAME -d $IMAGE bash -c "$client_cmd"
+
+ 	# This time in seconds is required to let the benchmark settle down
+ 	sleep "$SETTLE_TIME"
+
+ 	# Retrieve qemu information
+ 	PIDS="$(pgrep -d ',' qemu)"
+
+ 	metrics_json_init
+
+ 	# Start collecting cpu information
+ 	sudo perf stat -a -o $TMP_FILE -e cycles -e instructions -p ${PIDS} sleep ${SAMPLE_PERIOD}
+
+ 	# Save configuration
+ 	metrics_json_start_array
+
+	local instructions="$(cat $TMP_FILE | grep -w instructions | awk '{print $1}' | tr ',' ' ' | tr -d ' ')"
+
+	local cycles="$(cat $TMP_FILE | grep -w cycles | awk '{print $1}' | tr ',' ' ' | tr -d ' ')"
+
+	local instructions_per_cycle="$(cat $TMP_FILE | grep -w 'per cycle' | cut -d '#' -f2 | awk '{print $1}')"
+
+	local units_instructions_per_cycle="$(cat $TMP_FILE |  grep -w 'per cycle' | cut -d '#' -f2 |  awk '{print $2}')"
+
+	local json="$(cat << EOF
+	{
+		"instructions per cycle": {
+			"Result" : $instructions_per_cycle,
+			"Units"  : "$units_instructions_per_cycle per cycle"
+		},
+		"cycles": {
+			"Result" : $cycles,
+			"Units"  : "cycles"
+		},
+		"instructions": {
+			"Result" : $instructions,
+			"Units"  : "instructions"
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+}
+
+main "$@"
--- a/metrics/network/iperf_dockerfile/Dockerfile
+++ b/metrics/network/iperf_dockerfile/Dockerfile
@@ -1,0 +1,16 @@
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Usage: FROM [image name]
+FROM ubuntu:16.04
+
+# Version of the Dockerfile
+LABEL DOCKERFILE_VERSION="1.0"
+
+# Install iperf
+RUN apt-get update && \
+    apt-get remove -y unattended-upgrades && \
+    apt-get install -y iperf
+
+CMD ["/bin/bash"]


.ci/run_metrics_PR_ci.sh
```diff
@@ -95,6 +95,9 @@ run() {
 	# Run the density test inside the container
 	bash density/memory_usage_inside_container.sh
 
+	# Run the cpu stadistics test
+	bash network/cpu_stadistics_iperf.sh
```
jodh-intel: typo: "statistics".

metrics/network/cpu_stadistics_iperf.sh
```diff
@@ -0,0 +1,96 @@
...
+set -e
+
+# General env
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+TEST_NAME="cpu information"
+
+# Image name
+image="gabyct/network"
```
jodh-intel: Do we have to use a custom image? If so, please explain why and add a link to a github issue for the problem.

metrics/network/cpu_stadistics_iperf.sh
```diff
@@ -0,0 +1,96 @@
...
+	# Start collecting cpu information
+	sudo perf stat -o $TMP_FILE -e cycles -e instructions -p ${PIDS} sleep ${sample_period}
+
+	# Save configuration
+	metrics_json_start_array
+
+	local result=$(cat results)
+	local instructions_per_cycle=$(echo "$result" | grep -w "insn per cycle" | cut -d '#' -f2 | awk '{print $1}')
+	local cycles=$(echo "$result" | grep -w "cycles" | awk '{print $1}' | tr ',' ' ' | awk '{for(i=1;i<=NF;i++) if($i>maxval) maxval=$i;}; END { print maxval;}')
+	local instructions=$(echo "$result" | grep -w "instructions" | awk '{print $1}' | tr ',' ' ' | awk '{for(i=1;i<=NF;i++) if($i>maxval) maxval=$i;}; END { print maxval;}')
```
jodh-intel: Some of these lines are rather long so you might want to fold them on the pipes:

```
local instructions=$(echo "$result" |\
    grep -w "instructions" |\
    awk '{print $1}' |\
    ...
```

metrics/network/cpu_stadistics_iperf.sh
```diff
@@ -0,0 +1,96 @@
...
+
+	local json="$(cat << EOF
+	{
+		"instructions per cycle": {
+			"Result" : $instructions_per_cycle,
+			"Units"  : "ins per cycle"
```
jodh-intel: I'm not really clear why the units always matches the object name?!?

metrics/network/cpu_stadistics_iperf.sh
```diff
@@ -0,0 +1,96 @@
...
+# Server name
+server_name="server_iperf"
+# Target bandwidth
+bandwidth="1G"
+# Time in seconds to transmit
+timeout="160"
+# Sample period in seconds to measure
+sample_period="120"
+# File to save perf results
+TMP_FILE=$(mktemp cpuinfo.XXXXXXXXXX || true)
```
jodh-intel: I don't think we want this to silently succeed so I'd drop the `|| true` bit.

GabyCT:
@jodh-intel , changes applied and regarding about the units name like it is shown in perf
```
Performance counter stats for process id 'xxx':

        x,x,x      cycles
         x,x,x      instructions              #    xxx  insn per cycle
```

chavafg:
lgtm

@grahamwhaley ptal.

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Description of the test:
+# This test runs the 'iperf3' network benchmark and
+# we retrieve cpu information while running the networking
+# test with 'perf'.
```
grahamwhaley: can we add a note here explaining why we are measuring the CPU info - noting that (we are hoping ;-) that cpu utilisation is a more stable value to measure to detect changes than actual throughput. Also, discuss how we are not running a 'flat out' test here, as othewise we fully expect that to hit 100% utilisation on a local pure-software network stack.

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+set -e
+
+# General env
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+TEST_NAME="cpu information"
+
+# Image name
+image="networkstatic/iperf3"
```
grahamwhaley: Ooh, that is a different image than we have used for all our other previous tests (which was gabyct/networking iirc).
That's great, that it is a smaller image (just has iperf3 in it) - but... what I think I'd really like is for us to have all the test images we can (where practical) embedded into our source tree - so, if we could add a local Dockerfile and build method like we have over in the [blogbench](https://github.com/kata-containers/tests/tree/master/metrics/storage) test, that'd be great.
As jodh would say - you get 'bonus points' if you can make that Dockerfile 'versioned', so we don't have to rebuild it every time if we detect the correct version of the image is already installed in Docker...

Extra notes on that then - adding Dockerfiles increases our CI time, as we then have to build them all from sources. And as our CIs run on 'clean instances', we have to do that for every run - we don't (presently) cache the images.
/cc @chavafg

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+# Client name
+client_name="client_iperf"
+# Server name
+server_name="server_iperf"
+# Target bandwidth
+bandwidth="1G"
```
grahamwhaley: Can we make all these setup type vars over-rideable, so the invoker can tweak the test if necessary. Over in other tests we use the shell idiom like:
```
ITERATIONS="${ITERATIONS:-30}"
```

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+# Time in seconds to transmit
+timeout="160"
+# Sample period in seconds to measure
+sample_period="120"
+# File to save perf results
+TMP_FILE=$(mktemp cpuinfo.XXXXXXXXXX)
```
grahamwhaley: I suspect we could not use a tmpfile at all if we capture the output of the `perf` into a shell var - but, if you stick with the tmpfile, can you add a trap exit catch to remove it (I had some failed runs, and it left tmp files laying around my dir).
There is an example of the trap over in the [pr run script](https://github.com/kata-containers/tests/blob/master/.ci/run_metrics_PR_ci.sh#L74)

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+function main() {
+	# Check tools/commands dependencies
+	cmds=("awk" "docker" "perf")
+
+	init_env
+	check_cmds "${cmds[@]}"
+	check_images "$image"
+
+	# Start iperf3 server
+	local server_cmd="export TMPDIR=/dev/shm; iperf3 -s"
```
grahamwhaley: Maybe add a comment noting the /dev/shm is a workaround for a 9p unlink issue - and linking to the relevant Issue (if we have one in Kata - I think we have one in CC if not)

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+	docker run --runtime $RUNTIME --name $server_name -d $image bash -c "$server_cmd"
+	local server_address=$(docker inspect --format "{{.NetworkSettings.IPAddress}}" $server_name)
+
+	# Start iperf3 client
+	local client_cmd="export TMPDIR=/dev/shm; iperf3 -c $server_address -b $bandwidth -t $timeout"
+	docker run --runtime $RUNTIME --name $client_name -d $image bash -c "$client_name"
+
+	# This time in seconds is required to let the benchmark settle down
+	sleep 20
```
grahamwhaley: Maybe lift this into an over-rideable constant at the top to go with the other consts.

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+
+	# Retrieve qemu information
+	PIDS="$(pgrep -d ',' qemu)"
+
+	metrics_json_init
+
+	# Start collecting cpu information
+	sudo perf stat -o $TMP_FILE -e cycles -e instructions -p ${PIDS} sleep ${sample_period}
```
grahamwhaley: if we are sat here waiting for this to complete (that is, we don't background it, which is fine), then maybe we can just capture the output directly, and skip the TMPFILE usage?

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+	# Check tools/commands dependencies
+	cmds=("awk" "docker" "perf")
+
+	init_env
+	check_cmds "${cmds[@]}"
+	check_images "$image"
+
+	# Start iperf3 server
+	local server_cmd="export TMPDIR=/dev/shm; iperf3 -s"
+	docker run --runtime $RUNTIME --name $server_name -d $image bash -c "$server_cmd"
```
grahamwhaley: Can you stick a `--rm` on your docker commands - I don't seem to be able to run the test twice in a row as I get a name clash trying to create the second set of containers ;-)

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,98 @@
...
+	local server_address=$(docker inspect --format "{{.NetworkSettings.IPAddress}}" $server_name)
+
+	# Start iperf3 client
+	local client_cmd="export TMPDIR=/dev/shm; iperf3 -c $server_address -b $bandwidth -t $timeout"
+	docker run --runtime $RUNTIME --name $client_name -d $image bash -c "$client_name"
```
grahamwhaley: Err, should that final element be `$client_cmd`, not `$client_name`. As it is, the test does not run for me or produce any results - as the client command I suspect quits instantly, so the subsequent `pgrep` finds no QEMUs running.

grahamwhaley:
Marked as DNM, as the big green button was temptingly still active....

grahamwhaley:
/me notes to @chavafg - yep, looks like the CI kicked off when I added review comments and a label - we probably need to tweak Jenkins to only build on commit changes etc.

metrics/network/cpu_statistics_iperf.sh
grahamwhaley: aha, /me notes #404 - so seems we can move to using that as well ;-)

chavafg:
yes, sorry about that, I already modified the github webhooks. I'll keep an eye on it to see it behaves correctly.

GabyCT:
@grahamwhaley changes applied, this PR depends on https://github.com/kata-containers/tests/pull/404. Here it is the information about the results that I obtained running this test

Instructions Measurement - Mean 932 - Max 962 - Min 904 - SD 25.35 - Cov 2.718
Instructions per cycle Measurement - Mean .762 - Max .79 - Min .74 - SD 0.023 - Cov 3.038
Cycles Measurement - Mean 719 - Max 880 - Min 664 - SD 120.63 - Cov 16.77

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,108 @@
...
+
+	metrics_json_init
+
+	# Start collecting cpu information
+	sudo perf stat -o $TMP_FILE -e cycles -e instructions -p ${PIDS} sleep ${SAMPLE_PERIOD}
+
+	# Save configuration
+	metrics_json_start_array
+
+	local result=$(cat results)
```
grahamwhaley: There is no file `results` - this caused a `cat: results not found` error for me. I believe this is meant to be `$TMP_FILE`... but see other feedback...

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,108 @@
...
+	# This time in seconds is required to let the benchmark settle down
+	sleep "$SETTLE_TIME"
+
+	# Retrieve qemu information
+	PIDS="$(pgrep -d ',' qemu)"
+
+	metrics_json_init
+
+	# Start collecting cpu information
+	sudo perf stat -o $TMP_FILE -e cycles -e instructions -p ${PIDS} sleep ${SAMPLE_PERIOD}
```
grahamwhaley: I see we stuck with the `TMP_FILE`, rather than capture the output with a `results=$(cmd)` sequence.
That's OK, but, you didn't add in the TRAP rm catch code I suggested. When I hit the `cat` error elsewhere, I got tmpfiles left around my disk...

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,108 @@
...
+
+	# Save configuration
+	metrics_json_start_array
+
+	local result=$(cat results)
+	local instructions_per_cycle="$(echo "$result" | grep -w "insn per cycle" | cut -d '#' -f2 | awk '{print $1}')"
```
grahamwhaley: I didn't get any instructions_per_cycle results in my .json result file (after I'd tweaked the script to run). Looking at the output of `perf` on my system, I believe you need to s/insn/insns/ in that grep line. This could be a difference between your and my systems - can you check pls?

GabyCT:
@grahamwhaley changes applied, thanks

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,108 @@
...
+# Time in seconds to transmit
+TIMEOUT="${TIMEOUT:-160}"
+# Sample period in seconds to measure
+SAMPLE_PERIOD="${SAMPLE_PERIOD:-120}"
+# File to save perf results
+TMP_FILE=$(mktemp cpuinfo.XXXXXXXXXX)
+# Time in seconds is required to let the benchmark settle down
+SETTLE_TIME="${SETTLE_TIME:-20}"
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
```
grahamwhaley: Setting the `RUNTIME` is redundant I think - we do that in the `lib/common.bash` I believe.

metrics/network/iperf_dockerfile/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Usage: FROM [image name]
+FROM ubuntu:16:04
```
grahamwhaley: I believe this should be `ubuntu:16.04` - that is s/:/./ - it did not build for me until I made that edit.

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+	rm -rf $TMP_FILE
```
grahamwhaley: As requested, can you put the rm of the TMP_FILE into a bash TRAP please, so we don't litter tmp files around if we crash out of the test. Something similar to https://github.com/kata-containers/tests/blob/master/.ci/run_metrics_PR_ci.sh#L74
And to be really safe, you might want to do that as the first command in the script, as we call `mktemp` very early via the env var setup.
thx

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,108 @@
...
+
+	# Start collecting cpu information
+	sudo perf stat -o $TMP_FILE -e cycles -e instructions -p ${PIDS} sleep ${SAMPLE_PERIOD}
+
+	# Save configuration
+	metrics_json_start_array
+
+	local result=$(cat $TMP_FILE)
+	local instructions_per_cycle="$(echo "$result" | grep -w "insns per cycle" | cut -d '#' -f2 | awk '{print $1}')"
+	local cycles="$(echo "$result" | grep -w "cycles" | awk '{print $1}' | tr ',' ' ' | \
```
grahamwhaley: I've a feeling this 'find the max' calculation is all wrong. Here is the example `pef` output I get from a local run:
```
   216,511,365,940      cycles
   373,206,617,528      instructions              #    1.72  insns per cycle
```

It looks like you are parsing off the largest 3 digit number from that comma separated sequence - **but** - that is not a sequence of numbers, that is **one number** - one very large number :-). So, that is telling you we did 373 *billion* instructions in 216 *billion* cycles.
So, no max parsing necessary - we probably need to remove the `,`'s from the results (and maybe `perf` even has an option for that?).

Please fix, re-run the test and re-post your results to check reproducibility.

kenthelm: So, looking at this entire thread, it seems like a lot of time is being spent in review that shouldn't have been necessary.  I think more due diligence prior to submitting code (like testing) would be good.

GabyCT:
@grahamwhaley changes applied, here are the results
```
cycles  mean 537.105  max 946.000 min 316.000  sd 291.855 cov 54.339
instructions per cycle mean 1.689 max 1.730 min 1.650 sd  0.023 cov  1.351
instructions mean 869.105 max 996.000 min 764.000 sd 80.524 cov 9.265
```

grahamwhaley:
I'm confused. I still see all that 'max' searching code in the diff, and when I do a test run locally, my tmpfile looks like:
```
# started on Thu Jun 28 10:33:01 2018

 Performance counter stats for process id '4463,4644':

   218,957,037,351      cycles                                                        (100.00%)
   378,646,182,579      instructions              #    1.73  insns per cycle

     120.000838706 seconds time elapsed
```

but my json results look like:

```json
  "Results": [
                        {
                "instructions per cycle": {
                        "Result" : ,
                        "Units"  : "ins per cycle"
                },
                "cycles": {
                        "Result" : 957,
                        "Units"  : "cycles"
                },
                "instructions": {
                        "Result" : 646,
                        "Units"  : "instructions"
                }
        }
```

So:
- my instructions per cycle is empty?
- both cycles and instructions are still picking out the largest sub-part of the result, and not just taking the result itself (that is, showing hundreds, not billions...).

I suspect something wrong with your commit/push @GabyCT ? Can you check and try again ?

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+	trap remove_tmp_file EXIT
```
grahamwhaley: This trap needs to be very very early - basically as soon as you've created the tmpfile, which probably means it has to be the first command in main(). pls read up on trap in the bash manual.

GabyCT: @grahamwhaley, first I see that in your system is showing `insns per cycle` while mine is showing `ins per cycle` that is the reason why is empty there. I will change the `grep` for that.  Now as we discussed we need to select the highest `instruction` value with its correspondent `cycle` value. This is what is doing is locating the highest `instruction` and then match with it is `cycle`. Or what is the result that you want to see?

grahamwhaley: argh, `ins` vs `insns` - I suspect that is going to be a perf version difference thing - you might just have to regexp the code to handle both of those.

For the instruction and cycle - those strings of comma separated numbers I believe are **not** strings of comma separated numbers - they are a whole single number (in the billions), printed in the American? comma separated format :-(  Sooo, what I expect to see is that number (such as from the above example - '378646182579' for the cycles.

We don't extract a max in this test. We don't currently run the test multiple times either - and if we did, I'd expect us to store all the individual results, and then let the CI system work out how it checked that (for instance, if it looked for the max or the mean etc.).

OK?

GabyCT:
@grahamwhaley changes applied. Here are the results
```
cycles mean 447383590555.400  max 449735209918.000 min 445547432128.000 sd 1135733153.227 cov 0.254
instructions per cycle mean 1.702  max 1.730  min 1.670  sd 0.018  cov 1.065
instructions  mean 761774536699.450  max 773674271741.000  min 743100495867.000 
sd 8467614779.885 cov 1.112
```

grahamwhaley:
@chavafg - the 'Can one of the admins verify this patch?' I believe is coming from a Jenkins with the ghprb plugin - I think it is the lack of the 'build allways without asking (dangerous)' tickbox being ticked. Can you check any recently added new Jenkins jobs to see if we missed one of those? thx. (/cc @jodh-intel )

metrics/network/cpu_statistics_iperf.sh
```diff
@@ -0,0 +1,112 @@
...
+
+ 	metrics_json_init
+
+ 	# Start collecting cpu information
+ 	sudo perf stat -a -o $TMP_FILE -e cycles -e instructions -p ${PIDS} sleep ${SAMPLE_PERIOD}
+
+ 	# Save configuration
+ 	metrics_json_start_array
+
+	local instructions="$(cat $TMP_FILE | grep -w instructions | awk '{print $1}' | tr ',' ' ' | tr -d ' ')"
```
grahamwhaley: nit - I guess you can do this with a single `tr -d ','` ?

chavafg:
@grahamwhaley oohh true, its from the new initrd job. thanks, I was worried about that messages

PR #375, opened 2018-06-05
integration/docker: add test to ensure cgroup filesystem is read-only

### LINKED ISSUE #373: Add test to ensure all cgroup inside the container are read-only (security)
Inside the container, iterate over all directories in `/sys/fs/cgroup` and write anything in each file, write operation MUST fail, for example

```
echo 100 > /sys/fs/cgroup/cpu/cpu.shares
bash: /sys/fs/cgroup/cpu/cpu.shares: Read-only file system
```

devimc:
this is a security test, to ensure the container can't update it's constraints
and get access to more resources.

fixes #373

### DIFF
--- a/integration/docker/run_test.go
+++ b/integration/docker/run_test.go
@@ -206,3 +206,28 @@ var _ = Describe("run nonexistent command", func() {
 		})
 	})
 })
+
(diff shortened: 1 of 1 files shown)


GabyCT:
lgtm

PR #385, opened 2018-06-06
tests: Add k8s test to check cpu constraints.

### LINKED ISSUE #309: kubernetes: add test to check CPU constrains
Next yaml can be used as base for this test, but it needs to be improved to check the constrains

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cpu-demo
  namespace: sandbox
spec:
  containers:
  - name: cpu0
    image: debian
    resources:
      limits:
        cpu: "1"
    args:
    - "bash"
    tty: true
    stdin: true
  - name: cpu1
    image: debian
    resources:
      limits:
        cpu: "3"
    args:
    - "bash"
    - "-c"
    - "sleep 5; nproc"
    tty: true
    stdin: true
```

GabyCT:
This will verify cpu constraints like limits and requests.

Fixes #309


integration/kubernetes/k8s-cpu-ns.bats
```diff
@@ -0,0 +1,45 @@
...
+
+	# Create the pod
+	sudo -E kubectl create -f pod-cpu.yaml
+
+	# Check pod creation
+	pod_status_cmd="sudo -E kubectl get pods -a | grep $pod_name | grep Running"
+	waitForProcess "$wait_time" "$sleep_time" "$pod_status_cmd"
+
+	# Check cpu limits
+	cpu_limits_information=$(sudo -E kubectl describe pods $pod_name | grep -A 1 "Limits" | grep cpu | cut -d ':' -f2)
```
chavafg: I think that in addition, we should check inside the container that the cpu limits are correct. 
wdyt @devimc

devimc: yes we should check the constraint inside the container

GabyCT:
@devimc and @chavafg changes applied

PR #386, opened 2018-06-07
scripts: Add kata arch script

### LINKED ISSUE #382: Create an "arch" script that is cleverer than arch(1)
`arch(1)` (or `uname -m`) are useful but certain parts of the system require the golang `GOARCH` value instead:

https://golang.org/doc/install/source#environment

Create a script (`kata-arch.sh` ?) in this repo that other repos can call that:

- does not require `golang` to be installed.
- calls `arch(1)` and maps the output values like `x86_64` or `aarch64` to their `GOARCH` equivalents (`amd64`, `arm64`).

This shouldn't be a huge maintenance burden, particularly if there is only a single centralised utility to handle the mapping.

Related:

- https://github.com/kata-containers/documentation/issues/140
- https://github.com/kata-containers/ci/issues/35
- https://github.com/kata-containers/documentation/issues/128

/cc @grahamwhaley, @Pennyzct, @nitkon, @chavafg.

jodh-intel:
The standard `arch(1)` utility is rather limited in that it only shows
the host architecture in a single format.

Various repos need to know the architecture in either "golang" format
or using the name the Linux kernel config system uses.

Created a new script that provides this information for the currently
supported architectures that can be used by repos without having to
hard-code the logic themselves (or download `golang`).

Fixes #382.

### DIFF
--- a/.ci/kata-arch.sh
+++ b/.ci/kata-arch.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+typeset -r script_name=${0##*/}
+
+typeset -r cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+usage()
+{
+cat <<EOT
+Description: Display host architecture name in various formats.
+
+Usage: $script_name [options]
+
+Options:
+
+ -d, --default         : Show arch(1) architecture (this is the default).
+ -g, --golang          : Show architecture name using golang naming.
+ -h, --help            : Show this help.
+ -k, --kernel          : Show architecture name compatible with Linux* build system.
+
+EOT
+}
+
+# Convert architecture to the name used by golang
+arch_to_golang()
+{
+	local -r arch="$1"
+
+	case "$arch" in
+		arm64|ppc64el) echo "$arch";;
+		x86_64) echo "amd64";;
+		*) die "unsupported architecture: $arch";;
+	esac
+}
+
+# Convert architecture to the name used by the Linux kernel build system
+arch_to_kernel()
+{
+	local -r arch="$1"
+
+	case "$arch" in
+		arm64|x86_64) echo "$arch";;
+		ppc64el) echo "powerpc";;
+		*) die "unsupported architecture: $arch";;
+	esac
+}
+
+main()
+{
+	local type="default"
+
+	local args=$(getopt \
+		-n "$script_name" \
+		-a \
+		--options="dghk" \
+		--longoptions="default golang help kernel" \
+		-- "$@")
+
+	eval set -- "$args"
+	[ $? -ne 0 ] && { usage >&2; exit 1; }
+
+	while [ $# -gt 1 ]
+	do
+		case "$1" in
+			-d|--default) ;;
+
+			-g|--golang) type="golang";;
+
+			-h|--help)
+				usage
+				exit 0
+				;;
+
+			-k|--kernel) type="kernel";;
+
+			--)
+				shift
+				break
+				;;
+		esac
+		shift
+	done
+
+	local -r arch=$(arch)
+
+	case "$type" in
+		default) echo "$arch";;
+		golang) arch_to_golang "$arch";;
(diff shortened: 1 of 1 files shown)


scripts/kata-arch.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2018 Intel Corporation
```
grahamwhaley: 2017-2018 - did we clone this from or start writing this last year? (just curious)

scripts/kata-arch.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
grahamwhaley: Just a note on the commit message - 'golang' format could also be called 'distro' or 'package' format as well - which is actually where we see a number of the issues I think - consider adding those to the description maybe.

scripts/kata-arch.sh
```diff
@@ -0,0 +1,105 @@
...
+# Copyright (c) 2017-2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+typeset -r script_name=${0##*/}
+
+die()
```
grahamwhaley: Consider pulling this into a lib.sh like we do in many other places - I know this is the first and only script in this dir, but if we do it now then the changes of somebody just doing a clone later without the split are vastly reduced :-)

scripts/kata-arch.sh
```diff
@@ -0,0 +1,105 @@
...
+
+	local args=$(getopt \
+		-n "$script_name" \
+		-a \
+		--options="dghk" \
+		--longoptions="default golang help kernel" \
+		-- "$@")
+
+	eval set -- "$args"
+	[ $? -ne 0 ] && { usage; exit 1; }
```
grahamwhaley: Yeah, I never quite know if I should use the `die()` here or not - I think using `die()` would actually just complicate things. I'll note that I think all the output in this failure case goes only to stdout though.

scripts/kata-arch.sh
jodh-intel: Fixed.

scripts/kata-arch.sh
jodh-intel: Fixed.

scripts/kata-arch.sh
jodh-intel: Fixed.

scripts/kata-arch.sh
jodh-intel: Fixed.

scripts/kata-arch.sh
jodh-intel: Err - not fixed. I was thinking this too, but I'm not really sure how to make this 100% clear since, for example, `amd64` is used for Ubuntu packaging, but not for Fedora. That's why I stuck with golang for now as that's consistent but if anyone has an ISO standard / RFC spec for these naming schemes, please shout! ;)

jodh-intel:
@grahamwhaley - ha - I learnt a possible new euphemism there! "notes" indeed! ;)

scripts/kata-arch.sh
grahamwhaley: heh - np :-)

scripts/lib.sh
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+die()
```
chavafg: why not use the `die` function in `.ci/lib.sh`?

scripts/kata-arch.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+#
```
devimc: Q. why this script is not in `.ci` ?

scripts/lib.sh
jodh-intel: I've moved the script to `.ci/` now. I still think we might want to split out some of what's in `.ci/` to a `scripts` directory but for the time being let's keep all the scripts together.

scripts/kata-arch.sh
jodh-intel: Fixed ;)

GabyCT:
lgtm

PR #396, opened 2018-06-08
metrics: Disable auto updates in Ubuntu.

### LINKED ISSUE #387: metrics: Ubuntu based metrics probably don't disable auto updates
I think I noticed during a metrics run last week that our Ubuntu based metrics tests (which I think means in particular anything that is constructed through a Dockerfile such as https://github.com/kata-containers/tests/blob/master/metrics/storage/blogbench_dockerfile/Dockerfile#L8, have not disabled the Ubuntu auto updater package. This has two potential effects:
 - if we do any apt actions in the container then that can fail due to a resource clash for the apt lock
 - if we are running metrics (which we no doubt are...), then the auto update is likely in the background consuming resource and time - that will inject noise into our results.

There are a number of ways to disable this - I am currently using:
```
apt remove -y unattended-upgrades
```

for the metrics VMs.
Can we have a scan of the code base to identify any tests that could be suffering from this, maybe verify that this is a real concern, and then apply the package remove to ensure we do not run into this.

GabyCT:
This will disable the auto updates in the ubuntu image.

Fixes #387

### DIFF
--- a/metrics/storage/blogbench_dockerfile/Dockerfile
+++ b/metrics/storage/blogbench_dockerfile/Dockerfile
@@ -13,6 +13,7 @@ ENV BLOGBENCH_VERSION 1.1
 
 RUN apt-get update && \
 	apt-get install -y build-essential curl && \
+	apt-get remove -y unattended-upgrades && \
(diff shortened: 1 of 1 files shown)


chavafg:
lgtm

GabyCT:
@grahamwhaley yes, I confirmed that

PR #398, opened 2018-06-08
CI: Add URL check for documents

### LINKED ISSUE #397: validate document URLs
We need a CI test to validate all URLs in our documents.

jodh-intel:
Added a new CI test that checks for invalid URLs in our documents.

Fixes #397.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -262,14 +262,80 @@ EOT
 # Perform basic checks on documentation files
 check_docs()
 {
+	local cmd="xurls"
+
+	if [ ! "$(command -v $cmd)" ]
+	then
+		echo "Install $cmd utility"
+		go get -u "mvdan.cc/xurls/cmd/$cmd"
+	fi
+
 	echo "INFO: Checking documentation"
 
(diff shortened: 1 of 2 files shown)


grahamwhaley:
Silly question (sorry, easier to ask than to try and understand what the code is doing...) - does this take into account the pending PR branch changes will not appear on the main repo http on github, but would have to be checked via, I think, the PR originator branch?

btw, I think the notion of checking is good - as long as we can make it work for links that will only be real once the PR under review has landed iyswim

grahamwhaley:
Oh, and it occurs to me that the check might be a little fragile and at the mercy of the vagaries of the external sites where those URLs are pointing to - but, that is unavoidable I think - just thought I should explicitly point it out...

jodh-intel:
Well, this PR just adds an extra check to the static checks script so isn't doing anything new wrt the branch it operates on (hence it should be testing the PR branch for a PR and `master` otherwise).

Yes, we might occasionally get hit by network timeouts *etc*, but I think that's the price we have to pay to ensure the quality of our docs is as high as possible.

chavafg:
lgtm

GabyCT:
lgtm

PR #404, opened 2018-06-11
metrics: Automate docker image builds to reduce unnecessary rebuilds

### LINKED ISSUE #259: metrics: Automate docker image builds to reduce unnecessary rebuilds
We should make the metrics automatic docker image builds smarter to reduce unnecessary rebuilds.

GabyCT:
This will check that the image exists and in case that exists it will
check the dockerfile version as well as the test version in order to
build or avoid unnecesary rebuilds from the image.

Fixes #259

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -95,18 +95,47 @@ check_images()
 # This function performs a docker build on the image names
 # passed in, to ensure that we have the latest changes from
 # the dockerfiles
+build_dockerfile_image()
+{
+	local image="$1"
+	local dockerfile_path="$2"
+
+	echo "docker building $image"
+	if ! docker build --label "$image" --tag "${image}" -< "$dockerfile_path"; then
+		die "Failed to docker build image $image"
+	fi
+}
+
+# This function verifies that the dockerfile version is
+# equal to the test version in order to build the image or
+# just run the test
 check_dockerfiles_images()
 {
 	local image="$1"
 	local dockerfile_path="$2"
+
 	if [ -z "$image" ] || [ -z "$dockerfile_path" ]; then
 		die "Missing image or dockerfile path variable"
-		exit 1;
 	fi
-	echo "docker building $image"
-	if ! docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"; then
-		die "Failed to docker build image $image"
-		exit 1;
+
+	# Verify that dockerfile version is equal to test version
+	check_image=$(docker images "$image" -q)
+	if [ -n "$check_image" ]; then
+		# Check image label
+		check_image_version=$(docker image inspect $image | grep -w DOCKERFILE_VERSION | head -1 | cut -d '"' -f4)
+		if [ -n "$check_image_version" ]; then
+			echo "$image is not updated"
+			build_dockerfile_image "$image" "$dockerfile_path"
+		else
+			# Check dockerfile label
+			dockerfile_version=$(grep DOCKERFILE_VERSION $dockerfile_path | cut -d '"' -f2)
+			if [ "$dockerfile_version" != "$check_image_version" ]; then
+				echo "$dockerfile_version is not equal to $check_image_version"
+				build_dockerfile_image "$image" "$dockerfile_path"
+			fi
+		fi
+	else
+		build_dockerfile_image "$image" "$dockerfile_path"
 	fi
 }
 
--- a/metrics/storage/blogbench_dockerfile/Dockerfile
+++ b/metrics/storage/blogbench_dockerfile/Dockerfile
@@ -7,6 +7,9 @@
 # Usage: FROM [image name]
 FROM ubuntu
 
+# Version of the Dockerfile
+LABEL DOCKERFILE_VERSION="1.0"
+
 # URL for blogbench test and blogbench version
 ENV BLOGBENCH_URL "https://download.pureftpd.org/pub/blogbench"
 ENV BLOGBENCH_VERSION 1.1


metrics/lib/common.bash
```diff
@@ -95,18 +95,42 @@ check_images()
...
 	fi
-	echo "docker building $image"
-	if ! docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"; then
-		die "Failed to docker build image $image"
-		exit 1;
+
+	# Verify that dockerfile version is equal to test version
+	check_image=$(docker images "$image")
+	if [ -z "$check_image" ]; then
+		dockerfile_version=$(cat "$dockerfile_path" | grep "DOCKERFILE_VERSION" | cut -d '=' -f2)
```
grahamwhaley: I was thinking maybe more what we'd do would use a
```
LABEL Version="1.3"
```
in the Dockerfile (that then shows up in `docker inspect` easily I believe, and then just check the `docker inspect image` against the grep of the `Version` from the Dockerfile source. I don't think we need to have TEST_VERSION in the script itself - we are (I think) only interested if the contents of the `Dockerfile` change, and thus we need an image rebuild.

Oh, and 'I c a r u s' would note that you don't need the `cat | grep` - you can just put the filename on the end of the grep command (`grep Version filename`), which saves us spawning an extra process for the `cat` ;-)  (I blame 'I c a r u s' ;-) )

metrics/lib/common.bash
```diff
@@ -95,18 +95,42 @@ check_images()
...
 # the dockerfiles
+build_dockerfile_image()
+{
+	local image="$1"
+	local dockerfile_path="$2"
+
+	echo "docker building $image"
+	if ! docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"; then
+		die "Failed to docker build image $image"
+		exit 1;
```
jodh-intel: The `exit` is redundant as `die` calls `exit`.

metrics/lib/common.bash
```diff
@@ -95,18 +95,42 @@ check_images()
 # This function performs a docker build on the image names
 # passed in, to ensure that we have the latest changes from
 # the dockerfiles
+build_dockerfile_image()
+{
+	local image="$1"
+	local dockerfile_path="$2"
+
+	echo "docker building $image"
+	if ! docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"; then
```
jodh-intel: I don't think this is reliable - what if I called this function with `image=busybox:latest`? The fix would seem to be to only add on `:latest` if the image doesn't contain a colon. `:latest` is implied I think, but it's good to be explicit here to hint that that is what the function is trying to check.

GabyCT:
@grahamwhaley and @jodh-intel changes applied

GabyCT:
@grahamwhaley and @jodh-intel , I made the changes to this PR, however as the CI is currently failing randomly with ubuntu 17.10, I can not make this PR pass

GabyCT:
@grahamwhaley and @jodh-intel could you please review the changes, this PR https://github.com/kata-containers/tests/pull/366 depends on this PR changes

GabyCT:
@chavafg changes applied

metrics/lib/common.bash
```diff
@@ -95,18 +95,40 @@ check_images()
...
-	if ! docker build --label "$image" --tag "${image}:latest" -< "$dockerfile_path"; then
-		die "Failed to docker build image $image"
-		exit 1;
+
+	# Verify that dockerfile version is equal to test version
+	check_image=$(docker images "$image" -q)
+	if [ -n "$check_image" ]; then
+		check_image_version=$(docker image inspect $image | grep -w DOCKERFILE_VERSION | head -1 | cut -d '"' -f4)
+		dockerfile_version=$(grep DOCKERFILE_VERSION $dockerfile_path | cut -d '"' -f2)
+		if [ "$dockerfile_version" -ne "$check_image_version" ]; then
```
grahamwhaley: I ran with a `bash -x ./storage/blogbench.sh`, and found this in the output:

```
+ dockerfile_version=1.0
+ '[' 1.0 -ne '' ']'
/home/gwhaley/gopath/src/github.com/kata-containers/tests/metrics/storage/../lib/common.bash: line 126: [: 1.0: integer expression expected
+ metrics_json_init
```

Can you check that for me - it might be because I probably have an image in my docker already without the `DOCKERFILE_VERSION` in it - maybe we can then improve the script capture/handling of that?

metrics/lib/common.bash
```diff
@@ -95,18 +95,40 @@ check_images()
...
+			echo "$dockerfile_version is not equal to $check_image_version"
+			build_dockerfile_image "$image" "$dockerfile_path"
```
grahamwhaley: and... if you are doing any edits, it occurs to me that if you refactor the tests, then you could pull this build_dockerfile line out to the bottom and only have one call to it (rather than the two we have right now). So, make the tests check if a build is needed, and return if not - and then finally drop through to a single build line at the bottom of the func if it is needed - iyswim?

metrics/lib/common.bash
jodh-intel: `-ne` only works with integer comparison so we should either use `!=` or use `bc(1)` or something cleverer still if we have to handle version numbers containing periods.

GabyCT:
@grahamwhaley and @jodh-intel changes applied

metrics/lib/common.bash
```diff
@@ -95,18 +95,47 @@ check_images()
...
+	if [ -n "$check_image" ]; then
+		# Check image label
+		check_image_version=$(docker image inspect $image | grep -w DOCKERFILE_VERSION | head -1 | cut -d '"' -f4)
+		if [ -n "$check_image_version" ]; then
+			echo "$image is not updated"
+			build_dockerfile_image "$image" "$dockerfile_path"
+		else
+			# Check dockerfile label
+			dockerfile_version=$(grep DOCKERFILE_VERSION $dockerfile_path | cut -d '"' -f2)
+			if [ "$dockerfile_version" -ne "$check_image_version" ]; then
```
jodh-intel: I think this is still comparing two values containing periods? As mentioned in the last review, `-ne` won't handle that:

```
a=2
b=3.0
[ "$a" -ne "$b" ] && echo not equal
[: integer expression expected: 3.0
```

But you can use a string compare if you only need to know if they are not the same:

```
a=2
b=3.0
[ "$a" != "$b" ] && echo not equal
not equal
```

GabyCT:
@jodh-intel sorry it was my mistake, but now I fixed it thanks

metrics/lib/common.bash
```diff
@@ -95,18 +95,47 @@ check_images()
...
-		exit 1;
+
+	# Verify that dockerfile version is equal to test version
+	check_image=$(docker images "$image" -q)
+	if [ -n "$check_image" ]; then
+		# Check image label
+		check_image_version=$(docker image inspect $image | grep -w DOCKERFILE_VERSION | head -1 | cut -d '"' -f4)
+		if [ -n "$check_image_version" ]; then
+			echo "$image is not updated"
+			build_dockerfile_image "$image" "$dockerfile_path"
```
grahamwhaley: Just a note for 'next time'.... I still think you should have inverted the tests and refactored here to pull the `build_dockerfile_image` call out to a single tail-end call in the function - we have it called three times in the same way at present! - but, that needn't be a blocker....

PR #413, opened 2018-06-13
CI: Make URL checker ignore new Jenkins server

### LINKED ISSUE #412: Make URL checker ignore new Jenkins server
We need to update the URL checker to ignore the new Jenkins URL as it requires the client to login, and fails as a result.

jodh-intel:
The new Jenkins server requires the client to authenticate so just
ignore those URLs

Fixes #412.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -308,6 +308,10 @@ check_docs()
 		# This prefix requires the client to be logged in to github, so ignore
 		echo "$url"|grep -q 'https://github.com/pulls' && continue
 
+		# This prefix require the client to be logged into Jenkins, so
+		# ignore
+		echo "$url"|grep -q 'http://199.204.45.34' && continue
+
 		# Sigh.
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -308,6 +308,10 @@ check_docs()
 		# This prefix requires the client to be logged in to github, so ignore
 		echo "$url"|grep -q 'https://github.com/pulls' && continue
 
+		# This prefix require the client to be logged into Jenkins, so
+		# ignore
+		echo "$url"|grep -q 'http://199.204.45.34' && continue
```
grahamwhaley: I didn't realise we did this filtering for some URLs already.
If this list gets any longer, maybe we can pull it into an array we iterate (so, yes, a loop in a loop) - will be easier to maintain then.

WeiZhang555: Let's fix CI first and optimize it later :-)

WeiZhang555:
LGTM

Issue #195, opened 2018-04-03
Import soak test from Clear Containers for mount checks


jodh-intel:
We really need the soak test to be imported into Kata so we can run this as part of a CI run:

- https://github.com/clearcontainers/tests/blob/master/integration/stability/soak_parallel_rm.sh

See https://github.com/kata-containers/runtime/pull/136 (and also https://github.com/kata-containers/agent/pull/196/).

grahamwhaley:
I'm un-assigning this from me, as this is more a QA issue than a metrics issue - I just happen to have written the original when I found the problems when running metrics tests.
See also #215

grahamwhaley:
note to self: I realised I can add a check to count `/run/vc/sbs/*' as well as for `/var/lib/vc/sbs/*` - let me go add that.
I liked the idea we had elsewhere of checking the proxy sock count (iirc), but also seem to remember that was going to maybe be tricky to do. I'll leave that one for now.

grahamwhaley:
I will note though, if we are considering adding this to the default CI runs, that the PR in its current default configuration (5 cycles of 110 nginx containers) takes ~18minutes to run on my desk - a time overhead we are probably unwilling to add to the CIs. We can of course trim that - it is a balance of runtime vs likelyhood of uncovering issues.
It may be better to take the sanity check parts of this script and integrate them into the CI as suggested over in #215 .

grahamwhaley:
Running on master merges might be an option - we care less about how long they take to test. As long as it is obvious to somebody that a master merge has broken (that we get good notifications that is)

jodh-intel:
That's what worries me - we might not notice if `master` broke until we come to make a new release, and then we're back to the situation we're in now: having to spend lots of time debugging / bisecting / etc to resolve the issue (and identify which PR introduced it).

grahamwhaley:
Right. We already have Jenkins CI master jobs - the question is, what happens if one of those fails to build/pass? I'll pick on one just for example - [here](http://jenkins.katacontainers.io/job/kata-containers-agent-ubuntu-16-04-master/) we can see that the agent on ubuntu 16.04 has been failing on master half the time.
@chavafg - currently I guess we have no feedback hooks for the master branch builds? (not blaming btw, I suspect that is how we have always had it for eons). We should consider if we can get master branches to feed back - either to github (but where would it go - maybe ideally onto the last PR to be merged), or via email (which would need an MTA to be available to us somewhere I guess).

[...]

Issue #215, opened 2018-04-05
ci: check system is 'clean' after each test


grahamwhaley:
We should check the system is in a 'clean and expected' state after each test - this should pick up any issues where parts of the runtime are left behind, including any VM/proxy/vc files/mounts/net connections/etc.

See https://github.com/clearcontainers/tests/issues/826
Also see the related #195 - where you can find the code to check many such items

grahamwhaley:
Hi @GabyCT 
This is the Issue I was thinking you might look at.
It basically involves lifting the 'system sanity checks' out of the [rm soak test](https://github.com/kata-containers/tests/blob/master/integration/stability/soak_parallel_rm.sh), adding them to a top level library somewhere/somehow, and then running those checks after every test.

Now, the tricky bit I think is really that we have a number of different types of test - `go test`, ginkgo, bats etc. I don't think we'll manage to add this sanity checking to all of those in one go, in which case maybe a first step is to make a github markdown ticklist here with all the sub-types of tests we have and then we can try to add the sanity test to the subsets one at a time. I do expect each subset of tests may require a different approach to implement/wrapping.

Shout if you need more input, guidance, idea review, RFC etc.

GabyCT:
@grahamwhaley ok :) I'll take a look

GabyCT:
This is a PR related with that https://github.com/kata-containers/tests/pull/726

GabyCT:
This issue has been solved with PRs https://github.com/kata-containers/tests/pull/801 , https://github.com/kata-containers/tests/pull/732, https://github.com/kata-containers/tests/pull/726

PR #414, opened 2018-06-13
integration: soak: add rm soak test

grahamwhaley:
Add an 'rm' soak test. The test was originally written
to capture 'stuck' docker rm's of many containers, but
as it also does a lot of sanity checking of many other
parts of the system (checks for runtime/qemu/proxy/shims
running when they should, and not running when they should
not, and that 'kata-runtime list' matches what we have asked
docker to do, and check we don't leave dangling mounts around
etc.), it has also been useful for general stability checking.

Fixes: #195

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@
 TIMEOUT := 60
 
 # union for 'make test'
-UNION := functional docker crio docker-compose openshift kubernetes swarm cri-containerd
+UNION := functional docker crio docker-compose docker-stability openshift kubernetes swarm cri-containerd
 
 # skipped test suites for docker integration tests
 SKIP :=
@@ -58,6 +58,11 @@ docker-compose:
 	cd integration/docker-compose && \
 	bats docker-compose.bats
 
+docker-stability:
+	systemctl is-active --quiet docker || sudo systemctl start docker
+	cd integration/stability && \
+	export ITERATIONS=2 && export MAX_CONTAINERS=20 && chronic ./soak_parallel_rm.sh
+
 kubernetes:
 	bash -f .ci/install_bats.sh
 	bash -f integration/kubernetes/run_kubernetes_tests.sh
@@ -88,6 +93,7 @@ check: checkcommits log-parser
 	checkcommits \
 	crio \
 	docker-compose \
+	docker-stability \
 	functional \
 	ginkgo \
 	docker \
--- a/integration/stability/soak_parallel_rm.sh
+++ b/integration/stability/soak_parallel_rm.sh
@@ -1,0 +1,212 @@
+#!/bin/bash
+#
+# Copyright (c) 2017-2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test will run a number of parallel containers, and then try to
+# 'rm -f' them all at the same time. It will check after each run and
+# rm that we have the expected number of containers, shims, proxys,
+# qemus and runtimes active
+# The goals are two fold:
+# - spot any stuck or non-started components
+# - catch any hang ups
+
+cidir=$(dirname "$0")
+source "${cidir}/../../metrics/lib/common.bash"
+
+# How many times will we run the test loop...
+ITERATIONS="${ITERATIONS:-5}"
+
+# the system 'free available' level where we stop running the tests, as otherwise
+#  the system can crawl to a halt, and/or start refusing to launch new VMs anyway
+# We choose 2G, as that is one of the default VM sizes for Kata
+MEM_CUTOFF="${MEM_CUTOFF:-(2*1024*1024*1024)}"
+
+# The default container/workload we run for testing
+# nginx has show up issues in the past, for whatever reason, so
+# let's default to that
+PAYLOAD="${PAYLOAD:-nginx}"
+
+# do we need a command argument for this payload?
+COMMAND="${COMMAND:-}"
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
+RUNTIME_PATH=$(command -v $RUNTIME)
+
+# And set the names of the processes we look for
+QEMU_NAME="${QEMU_NAME:-qemu-lite-system-x86_64}"
+RUNTIME_NAME="${RUNTIME_NAME:-kata-runtime}"
+SHIM_NAME="${SHIM_NAME:-kata-shim}"
+PROXY_NAME="${PROXY_NAME:-kata-proxy}"
+
+# The place where virtcontainers keeps its active pod info
+# This is ultimately what 'kata-runtime list' uses to get its info, but
+# we can also check it for sanity directly
+VC_POD_DIR="${VC_POD_DIR:-/var/lib/vc/sbs}"
+
+# let's cap the test. If you want to run until you hit the memory limit
+# then just set this to a very large number
+MAX_CONTAINERS="${MAX_CONTAINERS:-110}"
+
+count_containers() {
+	docker ps -qa | wc -l
+}
+
+check_all_running() {
+	local goterror=0
+
+	echo "Checking ${how_many} containers have all relevant components"
+
+	# check what docker thinks
+	how_many_running=$(count_containers)
+
+	if (( ${how_many_running} != ${how_many} )); then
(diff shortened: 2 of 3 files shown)


grahamwhaley:
Marking RFC/WIP right now, as I'd like to confirm some things with others....
For me, right now, this looks like it hangs up on a `docker rm` of many containers. @chavafg @jodh-intel - could one or either of you give this a spin for me and confirm if you are seeing that?

I have a sneaky feeling this could be related to https://github.com/kata-containers/runtime/issues/396 - just a feeling. It feels like the old 'stuck lock' in vc issues we had a long long time ago - but, irrc, @sboeuf rewrote how all the locking works, so I suspect it has similar symptoms, but is not the same issue.
Note, once things are stuck, a `docker ps` works, but a `sudo kata-runtime list` gets hung up.

/cc @jamiehannaford for completeness.

grahamwhaley:
Oh, once we've worked over this somewhat I'll add a commit to add this into the QA CI scripts so it runs on every PR - if we agree (we should look at how long it takes to run for instance).

Also, this PR is somewhat strongly related to #215 - that is, much of the sanity checking code from this PR can be lifted out into a lib maybe and then we can run that before/after each test.

sboeuf:
@grahamwhaley I confirm that running the soak test triggers the error you're describing. I have tried on a simple Azure VM with 4vCPUs and 16GiB of RAM. I have run the test for 20 containers only.
The full first run went well, but during the second run, it hanged during container removal.

The issue seems to be related to locking but it might not be the root cause. This is worth some investigations.

grahamwhaley:
OK - I've got some debug from the stuck `docker rm`'s - let me open an Issue to put that data on, and ref it back here...

jodh-intel:
As mentioned on this issue, it would be extremely useful to get this landed.

grahamwhaley:
This'll still be mine then. Things to do are:
- [ ] Test against current HEAD repos
- [ ] Check we are checking all the correct paths (in /var, /lib, /run etc.)
- [ ] Add support to check the `vc/vm` directory - although, with the factory, it is not clear how to check when that should or should not be empty etc. - will need a touch of research

grahamwhaley:
OK...
@jodh-intel I got to revisit this. I tested with the HEAD components (it ran the full 110 containers, 5 cycles sweep fine). I fixed one minor thing that has niggled me forerver - an off-by-one check on the number of containers to run.
And then.... I have added it into the 'make tests' rules with a config to make it do a limited length run, so we can run a basic test in the CIs.
A full run of 110 containers for 5 cycles takes ~17minutes on my machine. As configured in the Makefile, for 20 containers for 2 cycles, it takes ~1m7s, which maybe we can accept.

I've dropped the RFC and DNM...

/cc @chavafg @GabyCT

grahamwhaley:
doh - I missed the `.PHONY` - let me do that...

jodh-intel:
@grahamwhaley - nice! Hopefully the CI will be able to crunch through this a bit quicker (I see they are triggered so we'll need to look at the logs once finished). Hence, aside from the `Makefile` conflict...

lgtm

grahamwhaley:
phew, let's try one more time (fingers crossed!). This time:
- rebased to fix Makefile conflict
- added a `chronic` around the call to reduce log noise
- removed `function` keywords from script, as we are moving away from them (and they are not needed).

/cc @bergwolf , as #578 will maybe have similar or benefit from all of those as well :-)

chavafg:
lgtm

chavafg:
I think you'll need to check if docker service is started (and start it if not) before executing the tests.

grahamwhaley:
Ah, I see:
```
cd integration/stability && \
export ITERATIONS=2 && export MAX_CONTAINERS=20 && chronic ./soak_parallel_rm.sh
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Checking Kata runtime kata-runtime
Start iteration 1 of 2
Running...
Checking 0 containers have all relevant components
Run kata-runtime: nginx: 
Checking 1 containers have all relevant components
Wrong number of containers running (0 != 1) - stopping
Got 1 errors, quitting
Makefile:62: recipe for target 'docker-stability' failed
make: *** [docker-stability] Error 255
```

@chavafg - is that the common 'idiom' for our test suites then - they have to check and start the services they need before they run?
Is there a set of shell library helper funcs for that perhaps?, like a `start_docker()` func or something I can invoke?

grahamwhaley:
OK, I see we do something similar for other test suites - sometimes in the Makefile (swarm), sometimes in the scripts themselves (crio). I've added the relevant `systemctl` lines copied from the swarm Makefile case to the docker-stability section of the Makefile and re-pushed. Let's see how that goes with the CIs.

chavafg:
network errors on the jobs :(
Sent a rebuild...
```

  5 99.4M    5 5390k    0     0  5390k      0  0:00:18 --:--:--  0:00:18 12.4M
 47 99.4M   47 46.9M    0     0  46.9M      0  0:00:02  0:00:01  0:00:01 33.4M
 48 99.4M   48 47.9M    0     0  47.9M      0  0:00:02  0:00:01  0:00:01 33.2M
curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104
```

grahamwhaley:
Heh heh. From the F27 CI:
```
systemctl is-active --quiet docker || sudo systemctl start docker
cd integration/stability && \
export ITERATIONS=2 && export MAX_CONTAINERS=20 && chronic ./soak_parallel_rm.sh
sudo: kata-runtime: command not found
Unable to find image 'nginx:latest' locally
latest: Pulling from library/nginx
be8881be8156: Already exists
32d9726baeef: Pulling fs layer
87e5e6f71297: Pulling fs layer
87e5e6f71297: Verifying Checksum
87e5e6f71297: Download complete
32d9726baeef: Verifying Checksum
32d9726baeef: Download complete
32d9726baeef: Pull complete
87e5e6f71297: Pull complete
Digest: sha256:d85914d547a6c92faa39ce7058bd7529baacab7e0cd4255442b04577c4d1f424
Status: Downloaded newer image for nginx:latest
sudo: kata-runtime: command not found
Checking Kata runtime kata-runtime
Start iteration 1 of 2
Running...
Checking 0 containers have all relevant components
Run kata-runtime: nginx: 
caa1e8a411ccc5c12e968c76d9d7bb107c962ab12500bbb88d05de6524c9db7d
Checking 1 containers have all relevant components
Wrong number of qemus running (1 != 0) - stopping
Wrong number of 'runtime list' containers running (1 != 0) - stopping
Got 2 errors, quitting
make: *** [Makefile:63: docker-stability] Error 255
```

and the same for Centos7 it seems. 

I'll check where/what the `sudo: kata-runtime: command not found` really means (maybe on those CI systems `sudo` cannot see `kata-runtime` in its path by default?).

Now, otherwise in a way this is sort of good, as that is exactly the sort of situation the test is mean to pick up.
I think what I should do is add a `dump_diagnostics()` func that will give us a bunch of `pgrep` and `docker ps -qa` type info to help diagnosis if the test fails. wdyt @chavafg @jodh-intel :-)

chavafg:
@grahamwhaley, the `dump_diagnostics` function sounds good. 
And you are right,  sudo cannot access kata-runtime:
```
[fuentess@kata-fedora ~]$ which kata-runtime
/usr/local/bin/kata-runtime
[fuentess@kata-fedora ~]$ sudo -E PATH=$PATH kata-runtime
sudo: kata-runtime: command not found
```

Although not sure what is the best way to solve this. In https://github.com/kata-containers/tests/blob/master/integration/openshift/hello_world.bats, we use:

```
kata_runtime_bin=$(command -v kata-runtime)
sudo -E "$kata_runtime_bin"
```

grahamwhaley:
Added a system info dump function to the common lib. Fixed the sudo RUNTIME invocation. Let's see how the CIs are feeling...

chavafg:
lgtm, 

CI happy. Merging

Issue #416, opened 2018-06-14
Change LVM device used for cri-o tests


chavafg:
Now that we changed the VMs used for the CI, the volume that we will use for CRI-O devicemapper tests is `vdb` instead of `sdb`. Also we should be able to set the device to another value if needed.

PR #417, opened 2018-06-14
CI: change block device used for cri-o

chavafg:
Now that we changed the VMs used for the CI, the block
device assigned for these tests is `/dev/vdb`.
This patch also provides the option to specify another value
if it is needed.

Fixes: #416.


devimc:
lgtm

jodh-intel:
lgtm

Issue #433, opened 2018-06-19
Google Cloud Platform resource URLs fail the URL checker


jon:
In Google Cloud Platform resources are identified by URLs under https://www.googleapis.com. These URLs are real and can be accessed over HTTPS, but in most cases only with a valid GCP OAuth2 token or service account. To prevent URL validation errors for GCP resources, https://www.goopgleapis.com should be added to the whitelist of ignored URLs patterns.

Issue #430, opened 2018-06-19
make url checker ignore https://www.googleapis.com urls


jodh-intel:
- See https://github.com/kata-containers/documentation/pull/154.

- Blocked on https://github.com/kata-containers/tests/pull/424.

jodh-intel:
Replaced by #433.

PR #424, opened 2018-06-15
CI: Only check URL changed by PRs

jodh-intel:
Change the URL checker to only check the URLs inside document files modified
by a PR. This is relatively difficult since if a PR adds a new document and
modifies an existing document that references that new document, the URL for
the new document will not be valid. The checker handles this by constructing
the correct URL for this new document which will be valid once the PR lands.
As long as the new-but-not-yet-valid URL referenced by the modified document
matches what the checker things the URL will be, the URL is accepted.

For the master branch, all doc URLs are checked.

This change necessitated modifying how the static check script is called by
the CI.

Includes refactoring commits.

Fixes #409.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #434, opened 2018-06-19
Add googleapis.com to unchecked URL list.

jon:
Google Cloud Platform resources are identified by URLs under
https://www.googleapis.com that are only accessible with a valid OAuth2
token. This change suppresses validation of Google APIs URLs to allow GCP
resource references in Kata documentation.

Fixes: #433
Signed-off-by: Jon Olson <jonolson@google.com>

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -314,6 +314,9 @@ check_docs()
 
 		# Sigh.
 		echo "$url"|grep -q 'https://example.com' && continue
+		
+		# Google APIs typically require an auth token.
+		echo "$url"|grep -q 'https://www.googleapis.com' && continue
 
 		# Check the URL, saving it if invalid
 		( curl -sLf -o /dev/null "$url" ||\


jon:
Whoops -- didn't realize there was already an open-issue. What's best practice for duplicate handling with Github issues? Should I close mine and update the commit to reference #430?

Also it looks like @jodh-intel wanted to hold off on #430 until #424 was submitted. I'm happy to do whatever folks would prefer here.

grahamwhaley:
For the duplicate, we can just drop the other once this is merged, so you can leave this as is right now.
For the potential wait on #424, I'm not sure exactly why apart from somebody is going to have to rebase most likely :-). I'll mark this one with a WIP label, and we can let @jodh-intel have a peek and decide on best order in the morning.
(thanks for raising this btw @jon  :-) )

jodh-intel:
Hi @jon - thanks for raising!

My bad - as @grahamwhaley says, there is no reason to block on #424 so removing the wip label...

jodh-intel:
lgtm

PR #400, opened 2018-06-11
CI: check for segfaults in logs

jodh-intel:
Check for segfaults in the log files after the tests have run. This is
particularly important when running with the agent as `init` inside
the image.

Fixes #399.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #450, opened 2018-06-27
CI: Move more tests to teardown

### LINKED ISSUE #449: Move more checks into teardown script
Currently, the CI `run.sh` scripts runs some checks. But those checks should be in `teardown.sh` so that they can be used by all the other repos.

jodh-intel:
The CI run script contained a couple of tests that need to be in the
teardown script since they:

- are not part of the test run itself.
- need to be run for all repos, not just `tests`.

Also required the teardown script to be reworked to put the existing code into its own function.

Fixes #449.

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -9,86 +9,6 @@
 
 set -e
 
-check_log_files()
-{
-	make log-parser
-
-	# XXX: Only the CC runtime uses structured logging,
-	# XXX: hence specify by name (rather than using $RUNTIME).
-	for component in \
-		kata-proxy \
-		kata-runtime \
-		kata-shim
-	do
-		file="${component}.log"
-		args="--no-pager -q -o cat -a -t \"${component}\""
-
-		cmd="sudo journalctl ${args} > ${file}"
-		eval "$cmd"
-	done
-
-	logs=$(ls "$(pwd)"/*.log)
-	{ kata-log-parser --debug --check-only --error-if-no-records $logs; ret=$?; } || true
-
-	errors=0
-
-	for log in $logs
-	do
-		# Display *all* errors caused by runtime exceptions and fatal
-		# signals.
-		for pattern in "fatal error" "fatal signal"
-		do
-			# Search for pattern and print all subsequent lines with specified log
-			# level.
-			results=$(sed -ne "/\<${pattern}\>/,\$ p" "$log" || true | grep "level=\"*error\"*")
-			if [ -n "$results" ]
-			then
-				errors=1
-				echo >&2 -e "ERROR: detected ${pattern} in '${log}'\n${results}" || true
-			fi
-		done
-
-		for pattern in "segfault at [0-9]"
-		do
-			results=$(sed -ne "/\<${pattern}\>/,\$ p" "$log" || true)
-
-			if [ -n "$results" ]
-			then
-				errors=1
-				echo >&2 -e "ERROR: detected ${pattern} in '${log}'\n${results}" || true
-			fi
-		done
-	done
-
-	[ "$errors" -ne 0 ] && exit 1
-
-	# Always remove logs since:
-	#
-	# - We don't want to waste disk-space.
-	# - The teardown script will save the full logs anyway.
-	# - the log parser tool shows full details of what went wrong.
-	rm -f $logs
-
-	[ $ret -eq 0 ] && true || false
-}
-
-check_collect_script()
-{
-	local cmd="kata-collect-data.sh"
-	local cmdpath=$(command -v "$cmd" || true)
-
-	local msg="Kata data collection script"
-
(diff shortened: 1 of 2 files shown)


jodh-intel:
I think this PR will solve the mystery of why the CI did not detect https://github.com/kata-containers/agent/issues/281 ...

I added #400 to look for segfaults in the logs. The problem with that being the main tests (only) currently run using an image, not an initrd.

But wait I hear you cry, osbuilder has [an initrd tests](https://github.com/kata-containers/osbuilder/blob/master/tests/image_creation.sh#L346)! Yes, it does, but since the Jenkins job runs the `setup.sh` and `run.sh` from the repo under test (osbuilder) but runs the teardown from the https://github.com/kata-containers/tests repo, the segfault check was never being run.

This PR resolves that by moving the tests into `teardown.sh` so that all repos get to take advantage of these checks.

grahamwhaley:
Can you explain why those funcs were not being run for the other repos? afaik, the other repos chdir and invoke this .ci/run.sh from this repo, which would then do those checks on the tail end, would it not?
So, I agree with the move to the teardown btw, just wondering why the funcs were not being executed for the other repos?

grahamwhaley:
ah, was this osbuilder repo specific then?

jodh-intel:
All the repos have their own `.ci/*` area containing atleast `setup.sh` and `run.sh`. The CI (Jenkins) jobs, then come along and run those two scripts. For example, here's the relevant snippet from the osbuilder Jenkins config file for Ubuntu 16.04:

- https://github.com/kata-containers/ci/blob/master/jenkins/jobs/kata-containers-osbuilder-ubuntu-16-04-PR/config.xml#L115..L117

*But*, we make Jenkins call the central `teardown.sh` script:

- https://github.com/kata-containers/ci/blob/master/jenkins/jobs/kata-containers-osbuilder-ubuntu-16-04-PR/config.xml#L143..L144

That was the problem since the only script that was checking for segfaults was the tests repos `run.sh`:

- https://github.com/kata-containers/tests/blob/master/.ci/run.sh#L101

Whereas most repos make their `.ci/run.sh` call the central version of that script in the tests repo, osbuilder doesn't do that:

- https://github.com/kata-containers/osbuilder/blob/master/.ci/run.sh

Hence, osbuilder was the only repo testing initrd's. The agent has seemingly been segfaulting for a while. But the only thing checking for segfaults was the tests repo's `run.sh` script which is not used by osbuilder.

chavafg:
BTW, currently the teardown step that runs in jenkins is configured to not fail the job, even if the script itself fails. I think I need to change that, so that we notice when a segfault occurs, right?

chavafg:
also, teardown is only executed if the job fails, I also need to change that.

jodh-intel:
@chavafg - oh, good point! Yes, I think that's probably why I put those tests in `run.sh` originally.

Since we want to guarantee that we always store the logfiles (`collect_logs()` function), I'll let that run, then `set -e` for the remaining tests if you agree?

jodh-intel:
Ping :)

chavafg:
@jodh-intel , I was making some tests yesterday with the post-build task plugin, but it seems that there is no reliable way to run the `teardown.sh` script when the build runs successful. 
Seems that there is no way to tell the plugin to run the teardown always
 
I think we have 2 options:
1. do not use the post-build task plugin for the `teardown.sh` and call it directly in the `jenkins_job_build.sh`

2. Add something like `echo "Kata Containers Build has finished successfully"` to the `jenkins_job_build.sh`, so that the post-build task plugin can grep  it and run the `teardown.sh` when this is found.

I'll prefer option 1, since we wouldnt need to continue manage the plugin in jenkins 

wdyt?

jodh-intel:
I agree - options 1 sounds like the best to me. If you do that, could you add a comment explaining why we're doing it that way as it might look odd otherwise ;)

grahamwhaley:
Hmm, do you know under what circumstances the post script does not fire @chavafg ? I have one on my metrics CI, and afaik, that always fires (it shuts down my VMs...). The only difference I can see is that for the 'log expression match' I have `.*`, and you have a blank entry (which I would have presumed did the same thing - but, maybe not?).
There is also the 'run script only if all the previous steps were successful', which I presume is what you were trying out?

I've not tracked closely the exact details here - so, feel free to go with your suggestions above etc.

chavafg:
ha, thanks for the expression `.*`,  that should do the trick. I'll try it

chavafg:
@grahamwhaley your advice worked here, thanks.
@jodh-intel seems that the script is running correctly, but for some reason, the `kata-collect-data.sh` is not found on CentOS and Fedora, maybe a PATH issue when using sudo?

chavafg:
Fixed the PATH issue by adding `/usr/local/bin`

PR #455, opened 2018-06-28
CI: Fix cleanup in install_kata_image.sh

### LINKED ISSUE #454: Fix cleanup in install_kata_image.sh
Verify that when running the cleanup in this script, the `IMG_MOUNT_DIR` variable might be null, for example in the case of initrd, we do not use this variable and the cleanup is failing. 
Add a verification that the variable has a value.

chavafg:
Verify that `IMG_MOUNT_DIR` variable has a value before
trying to umount it.

Fixes #454.


GabyCT:
lgtm

Issue #466, opened 2018-07-02
Disable systemd-journald rate limit


chavafg:
Disable to capture full logs and to ensure we can debug failed CI runs.

PR #467, opened 2018-07-02
CI: Disable systemd-journald rate limit

chavafg:
Disable the systemd-journald rate limit in the CI systems
to capture full logs from Kata Containers components.
This will help us ensure we can debug failed CI jobs.

Fixes: #466.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -67,6 +67,11 @@ bash -f "${cidir}/install_kubernetes.sh"
 echo "Install Openshift"
 bash -f "${cidir}/install_openshift.sh"
 
+echo "Disable systemd-journald rate limit"
+sudo crudini --set /etc/systemd/journald.conf Journal RateLimitInterval 0s
+sudo crudini --set /etc/systemd/journald.conf Journal RateLimitBurst 0
+sudo systemctl restart systemd-journald
+
 echo "Drop caches"
 sync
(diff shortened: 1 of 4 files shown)


.ci/setup.sh
```diff
@@ -67,6 +67,11 @@ bash -f "${cidir}/install_kubernetes.sh"
 echo "Install Openshift"
 bash -f "${cidir}/install_openshift.sh"
 
+echo "Disable systemd-journald rate limit"
+sudo bash -c "echo RateLimitInterval=0s >> /etc/systemd/journald.conf"
+sudo bash -c "echo RateLimitBurst=0 >> /etc/systemd/journald.conf"
```
sboeuf: looks good, but would this work if some predefined entries were already there ? Would this override them because they're located at the end ?

chavafg: ohhh, good question.
I've tried and the service at least is restarted successfully, but I'll better check to see if the entry is already there.

.ci/setup.sh
```diff
@@ -67,6 +67,11 @@ bash -f "${cidir}/install_kubernetes.sh"
 echo "Install Openshift"
 bash -f "${cidir}/install_openshift.sh"
 
+echo "Disable systemd-journald rate limit"
+sudo crudini --set /etc/systemd/journald.conf Journal RateLimitInterval 0s
```
jodh-intel: Huh - never heard of it until now ;)

jodh-intel:
lgtm

PR #464, opened 2018-06-29
ci: Disabling TODO/FIXME checker

GabyCT:
As currently we are having issues in the CI with several PRs,
the tool should not run in order to have a healthy CI.

Fixes #463

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

PR #471, opened 2018-07-03
CI: Add options to static-checks script

### LINKED ISSUE #470: Add CLI options to static-checks script
It would be useful to be able to run an individual check function rather than all the tests by specifying an option on the command-line.

jodh-intel:
Add command-line options to all individual test functions in the `static-checks.sh` script to be run, rather than all of the tests.

Also includes commits for improved messages, and an explicit "skip" for the `TODO`/`FIXME` checker that was disabled on #464.

Fixes #470.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -523,7 +523,7 @@ main()
 
 			info "Auto-detected repo as $repo"
 		else
-			usage && exit 1
+			echo >&2 "ERROR: need repo" && usage && exit 1
 		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -20,19 +20,60 @@ script_name=${0##*/}
...
+)
+
 usage()
 {
 	cat <<EOT
 
 Usage: $script_name help
-       $script_name repo-name [true]
+       $script_name [options] repo-name [true]
+
```
grahamwhaley: Somewhere, but I don't know where, we should probably try and note in the help that if you do not supply any options, then the default is 'all'. I don't think that is listed anywhere in the help?

jodh-intel: Done.

Issue #481, opened 2018-07-06
CI: use alpine:3.7 docker image


chavafg:
latest version of alpine image does not work correctly, instead use 3.7 which is the last known version. 

```
$ sudo docker pull alpine
Using default tag: latest
latest: Pulling from library/alpine
no matching manifest for linux/amd64 in the manifest list entries
```

PR #482, opened 2018-07-06
CI: do not use latest version of alpine docker image

chavafg:
latest version of alpine image is broken, use 3.7 which
is the last know working version.

Fixes: #481.

### DIFF
--- a/integration/docker/docker.go
+++ b/integration/docker/docker.go
@@ -24,7 +24,7 @@ const (
 	Image = "busybox"
 
 	// AlpineImage is the alpine image
-	AlpineImage = "alpine"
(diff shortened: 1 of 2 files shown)


GabyCT:
lgtm

chavafg:
yes, we can think of doing that in another PR, in the meantime to unblock the CI, we can just have this. 
There is already a record of an alpine version used for initrd: https://github.com/kata-containers/runtime/blob/master/versions.yaml#L90
maybe we can use it.

jodh-intel:
Yep - would be good in `versions.yaml` later but for now,

lgtm

devimc:
lgtm

Issue #479, opened 2018-07-06
CI: distro installs: use more chronic


grahamwhaley:
A whole bunch of our apt installs etc. in the CI scripts are filtered via [chronic](https://joeyh.name/code/moreutils/) so they are silent unless they fail - but we only seem to have that on the first half of the script, and its use sort of stops around:
https://github.com/kata-containers/tests/blob/master/.ci/setup_env_ubuntu.sh#L44

Add more `chronic`s as necessary to calm the log output down...

grahamwhaley:
@chavafg - just to check - is there any reason we don't do this at present, or has it just been oversight as we've added things to the scripts maybe?

jodh-intel:
+1 for less apt output :)

chavafg:
Found some more places where we can add chronic.

PR #498, opened 2018-07-16
CI: add chronic to more apt/dnf/yum commands

chavafg:
Add chronic to reduce the log size of our builds.

Fixes: #479.


GabyCT:
lgtm

PR #510, opened 2018-07-18
metrics: Iperf3 bandwidth, jitter, multiqueue and bidirectional tests

### LINKED ISSUE #509: metrics: Iperf3 bandwidth, jitter, multiqueue and bidirectional tests
This test measures bandwidth,  jitter, multiqueue and bidirectional bandwidth using iperf3

GabyCT:
This test measures bandwidth, jitter, multiqueue and bidirectional
bandwidth using iperf3.

Fixes #509

### DIFF
--- a/metrics/network/network-metrics-iperf3.sh
+++ b/metrics/network/network-metrics-iperf3.sh
@@ -1,0 +1,591 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test measures the following network essentials:
+# - bandwith simplex
+# - bandwith duplex
+# - jitter
+#
+# These metrics/results will be got from the interconnection between
+# a client and a server using iperf3 tool.
+# The following cases are covered:
+#
+# case 1:
+#  container-server <----> container-client
+#
+# case 2"
+#  container-server <----> host-client
+
+set -e
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/lib/network-common.bash"
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+# Port number where the server will run
+port="${port:-5201}"
+fwd_port="${fwd_port:-$port:$port}"
+# Image name
+# Note: This image will be replaced with a
+# dockerfile once this issue esnet/iperf#153
+# is solved
+image="${image:-gabyct/network}"
+# Measurement time (seconds)
+transmit_timeout="${transmit_timeout:-30}"
+
+# Iperf server configuration
+# Set the TMPDIR to an existing tmpfs mount to avoid a 9p unlink error
+init_cmds="export TMPDIR=/dev/shm"
+server_command="$init_cmds && iperf3 -s"
+
+# Test single direction TCP bandwith
+function iperf3_bandwidth() {
+	local TEST_NAME="network iperf3 bandwidth"
+	local server_address=$(start_server "$image" "$server_command" "$server_extra_args")
+
+	# Verify server IP address
+	if [ -z "$server_address" ];then
+		clean_env
+		die "server: ip address no found"
+	fi
+
+	# Start client
+	local client_command="$init_cmds && iperf3 -c ${server_address} -t ${transmit_timeout}"
+
+	metrics_json_init
+	save_config
+
+	# Start server
+	result=$(start_client "$image" "$client_command" "$client_extra_args")
+
+	metrics_json_start_array
+
+	local result_line=$(echo "$result" | grep -m1 -E '\breceiver\b')
+	local -a results
+	read -a results <<< $result_line
+	local total_bandwidth=${results[6]}
+	local total_bandwidth_units=${results[7]}
+
+	local json="$(cat << EOF
+ 	{
+ 		"bandwidth": {
+			"Result" : $total_bandwidth,
+			"Units"  : "$total_bandwidth_units"
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+}
+
+# Test jitter on single direction UDP
+function iperf3_jitter() {
+	local TEST_NAME="network iperf3 jitter"
+	# Start server
+	local server_address=$(start_server "$image" "$server_command" "$server_extra_args")
+
+	# Verify server IP address
+	if [ -z "$server_address" ];then
+		clean_env
+		die "server: ip address no found"
+	fi
+
+	metrics_json_init
+	save_config
+
+	# Start server
+	local client_command="$init_cmds && iperf3 -c ${server_address} -u -t ${transmit_timeout}"
+	result=$(start_client "$image" "$client_command" "$client_extra_args")
+
+	metrics_json_start_array
+
+	local result_line=$(echo "$result" | grep -m2 -A2 -E '\bJitter\b' | tail -1)
+	local -a results
+	read -a results <<< $result_line
+	local total_jitter=${results[8]}
+	local total_jitter_units=${results[9]}
+
+	local json="$(cat << EOF
+	{
+		"jitter": {
+			"Result" : $total_jitter,
+			"Units"  : "$total_jitter_units"
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+}
+
+# Run bi-directional TCP test, and extract results for both directions
+function iperf3_bidirectional_bandwidth_client_server() {
+	local TEST_NAME="network iperf3 bidirectional bandwidth"
+	# Start server
+	local server_address=$(start_server "$image" "$server_command" "$server_extra_args")
+
+	# Verify server IP address
+	if [ -z "$server_address" ];then
+		clean_env
+		die "server: ip address no found"
+	fi
+
+	metrics_json_init
+	save_config
+
+	# Start client
+	local client_command="$init_cmds && iperf3 -c ${server_address} -d -t ${transmit_timeout}"
+	result=$(start_client "$image" "$client_command" "$client_extra_args")
+
+	metrics_json_start_array
+
+	local client_result=$(echo "$result" | grep -m1 -E '\breceiver\b')
+	local server_result=$(echo "$result" | grep -m1 -E '\bsender\b')
+	local -a client_results
+	read -a client_results <<< ${client_result}
+	read -a server_results <<< ${server_result}
+	local total_bidirectional_client_bandwidth=${client_results[6]}
+	local total_bidirectional_client_bandwidth_units=${client_results[7]}
+	local total_bidirectional_server_bandwidth=${server_results[6]}
+	local total_bidirectional_server_bandwidth_units=${server_results[7]}
+
+	local json="$(cat << EOF
+	{
+		"client to server": {
+			"Result" : $total_bidirectional_client_bandwidth,
+			"Units"  : "$total_bidirectional_client_bandwidth_units"
+		},
+		"server to client": {
+ 			"Result" : $total_bidirectional_server_bandwidth,
+			"Units"  : "$total_bidirectional_server_bandwidth_units"
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+}
+
+# This function checks/verify if the iperf3 server
+# is ready/up for requests.
+function check_iperf3_server() {
+	local retry=6
+	local period=0.5
+	local test_cmd="iperf3 -c "$server_address" -t 1"
+
+	# check tools dependencies
+	local cmds=("netstat")
+	check_cmds "${cmds[@]}"
+
+	while [ 1 ]; do
+		if ! bash -c "$test_cmd" > /dev/null 2>&1; then
+			echo "waiting for server..."
+			count=$((count++))
+			sleep $period
+		else
+			echo "iperf3 server is up!"
+			break;
+		fi
+
+		if [ "$count" -le "$retry" ]; then
+			die "iperf3 server init fails"
+		fi
+	done
+
+	# Check listening port
+	lport="$(netstat -atun | grep "$port" | grep "LISTEN")"
+	if [ -z "$lport" ]; then
+		die "port is not listening"
+	fi
+}
+
+# This function parses the output of iperf3 execution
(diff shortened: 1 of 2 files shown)


GabyCT:
@amshinde and @egernst here it is the first part of the bandwidth iperf3 tests, I will submitted the rest in other PRs

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test measures the following network essentials:
+# - bandwith simplex
+# - bandwith duplex
+# - jitter
+#
+# These metrics/results will be got from the interconnection between
+# a client and a server using iperf tool.
```
grahamwhaley: s/iperf/iperf3/ - as there are more than one variant of iperf out there iirc, so let's clarify.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+source "${SCRIPT_PATH}/lib/network-common.bash"
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+# Port number where the server will run
+port="${port:-5201}"
+fwd_port="${fwd_port:-$port:$port}"
+# Image name
+image="${image:-gabyct/network}"
+# Measurement time (seconds)
+transmit_timeout="${transmit_timeout:-5}"
```
grahamwhaley: My suspicion is that we'll need a larger runtime/timeout for this in order to get stable repeatable results. I'll let you experiment and crunch the numbers to determine its effect.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/lib/network-common.bash"
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+# Port number where the server will run
+port="${port:-5201}"
+fwd_port="${fwd_port:-$port:$port}"
+# Image name
+image="${image:-gabyct/network}"
```
grahamwhaley: Can we translate this into the local Dockerfile idiom please, and use the funcs you added like https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L98, so we hold the Dockerfile locally in the repo :-)

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+	read -a client_results <<< ${client_result}
+	read -a server_results <<< ${server_result}
+	local total_bidirectional_client_bandwidth=${client_results[6]}
+	local total_bidirectional_client_bandwidth_units=${client_results[7]}
+	local total_bidirectional_server_bandwidth=${server_results[6]}
+	local total_bidirectional_server_bandwidth_units=${server_results[7]}
+
+	local json="$(cat << EOF
+	{
+		"network iperf3 bidir bw client to server": {
```
grahamwhaley: The long section name here is probably fine, but it might become tedious to parse out of the JSON, and most of that info is encoded in the TEST_NAME already. I'd suggest just naming those sections 'to server' and 'to client' maybe.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+	metrics_json_end_array "Results"
+	metrics_json_save
+	clean_env
+}
+
+# This function checks/verify if the iperf3 server
+# is ready/up for requests.
+function check_iperf_server() {
+	# timeout of 3 seconds aprox.
+	local time_out=6
```
grahamwhaley: maybe s/time_out/retrys/ to make it clearer.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+	local test_cmd="iperf3 -c "$server_address" -t 1"
+
+	# check tools dependencies
+	local cmds=("netstat")
+	check_cmds "${cmds[@]}"
+
+	while [ 1 ]; do
+		if ! bash -c "$test_cmd" > /dev/null 2>&1; then
+			echo "waiting for server..."
+			count=$((count+1))
```
grahamwhaley: silly bash nit - I think you can do `((count++))` here? :-)

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+			sleep $period
+		else
+			echo "iperf server is up!"
+			break;
+		fi
+
+		if [ "$count" == "$time_out" ]; then
```
grahamwhaley: I think this is a string compare, and nominally you'd be more correct doing a numeric compare with a `-le` or similar?

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+		die "port is not listening"
+	fi
+}
+
+# This function parses the output of iperf execution
+function parse_iperf_bwd() {
+	local TEST_NAME="$1"
+	local result="$2"
+
+	metrics_json_init
```
grahamwhaley: May as well move this to below the death check :-)

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+
+	if [ -z "$result" ]; then
+		die "no result output"
+	fi
+
+	# Save configuration
```
grahamwhaley: iirc, the config save is done by the `metrics_json_init` - so this comment may have stuck to the wrong line.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+
+ 	# Getting results
+	rx_bwd=$(echo "$rx_res" | awk '{print $5}')
+	tx_bwd=$(echo "$tx_res" | awk '{print $5}')
+	rx_uts=$(echo "$rx_res" | awk '{print $6}')
+	tx_uts=$(echo "$tx_res" | awk '{print $6}')
+
+	local json="$(cat << EOF
+	{
+ 		"${TEST_NAME} receiver": {
```
grahamwhaley: The `TEST_NAME` is encoded during the `metrics_json_init` I think, so is somewhat redundant here - I'd drop it from these section names.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+	# Extract results
+	local lost=$(echo "$result" | jq '.end.sum.lost_packets')
+ 	local total=$(echo "$result" | jq '.end.sum.packets')
+	local notlost=$((total-lost))
+	local pps=$((notlost/transmit_timeout))
+
+	local json="$(cat << EOF
+	{
+		"${TEST_NAME} receiver": {
+			"Result" : $pps,
```
grahamwhaley: As we have the total and lost counts already above, and storing them in the JSON is cheap and easy, I'd inject those values in here as well. Better to have too much info than too little.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,540 @@
...
+
+	if [ -z "$result" ]; then
+		die "no result output"
+	fi
+
+	# Save configuration
+	metrics_json_start_array
+
+	# Extract results
+	local lost=$(echo "$result" | jq '.end.sum.lost_packets')
```
grahamwhaley: Given the input we get here is JSON - we can I believe dump that whole input into the JSON results file as well as the data we cook out of it.

metrics/network/network-metrics-iperf3.sh
grahamwhaley: For anything that is configurable or might be good to know/reference/document in the future (such as maybe also the iperf3 versions being used) - can we inject those into a separate JSON section in the results files please. I particular this timeout sprang to mind - but there may be other things like bandwidth and packet size settings.
You can see what I mean with the `save_config` func over in the docker density test for instance: https://github.com/kata-containers/tests/blob/master/metrics/density/docker_memory_usage.sh#L290

metrics/network/network-metrics-iperf3.sh
GabyCT: @grahamwhaley, currently iperf3 is having a lot of issues in the installation https://github.com/esnet/iperf/issues/153 that is the main reason I am not using a Dockerfile. After this issue is solved, I will change the script to build the Dockerfile as we currently we have in the other metrics.

GabyCT:
@grahamwhaley changes applied

egernst:
@grahamwhaley  PTAL?

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,567 @@
...
+			"Result" : $notlost,
+			"Units"  : "PPS"
+		}
+	}
+EOF
+)"
+
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Results"
+	metrics_json_save
```
grahamwhaley: I still think we can add a json save here of the value of `$result`, as I think that is a fully qualified json item - it might have a little JSON wrap around it that we may need to trim in order to embed it. But, seeing as we have a full JSON formatted detailed result for this test, I think we should store it in the results file rather than throw away any details. I grabbed some output. It is quite large, so here is a trimmed example:

```json
{
        "start":        {
                "connected":    [{
                                "socket":       4,
                                "local_host":   "172.17.0.1",
                                "local_port":   59172,
                                "remote_host":  "172.17.0.2",
                                "remote_port":  5201
                        }],
                "version":      "iperf 3.0.11",
                "system_info":  "Linux bignuc 4.4.0-104-generic #127-Ubuntu SMP Mon Dec 11 12:16:42 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux\n",
                "timestamp":    {
                        "time": "Tue, 24 Jul 2018 12:44:50 GMT",
                        "timesecs":     1532436290
                },
                "connecting_to":        {
                        "host": "172.17.0.2",
                        "port": 5201
                },
                "cookie":       "bignuc.1532436290.341903.6630386652f",
                "test_start":   {
                        "protocol":     "UDP",
                        "num_streams":  1,
                        "blksize":      64,
                        "omit": 0,
                        "duration":     30,
                        "bytes":        0,
                        "blocks":       0,
                        "reverse":      1
                }
        },
        "intervals":    [{
                        "streams":      [{
                                        "socket":       4,
                                        "start":        0,
                                        "end":  1.00001,
                                        "seconds":      1.00001,
                                        "bytes":        12268992,
                                        "bits_per_second":      9.8151e+07,
                                        "jitter_ms":    0.00722195,
                                        "lost_packets": 651136,
                                        "packets":      842839,
                                        "lost_percent": 77.2551,
                                        "omitted":      false
                                }],
                        "sum":  {
                                "start":        0,
                                "end":  1.00001,
                                "seconds":      1.00001,
                                "bytes":        12268992,
                                "bits_per_second":      9.8151e+07,
                                "jitter_ms":    0.00722195,
                                "lost_packets": 651136,
                                "packets":      842839,
                                "lost_percent": 77.2551,
                                "omitted":      false
                        }
                }, {
...                }],
        "end":  {
                "streams":      [{
                                "udp":  {
                                        "socket":       4,
                                        "start":        0,
                                        "end":  30,
                                        "seconds":      30,
                                        "bytes":        1601305600,
                                        "bits_per_second":      4.27015e+08,
                                        "jitter_ms":    0.0010719,
                                        "lost_packets": 18732833,
                                        "packets":      25020345,
                                        "lost_percent": 74.8704
                                }
                        }],
                "sum":  {
                        "start":        0,
                        "end":  30,
                        "seconds":      30,
                        "bytes":        1601305600,
                        "bits_per_second":      4.27015e+08,
                        "jitter_ms":    0.0010719,
                        "lost_packets": 18732833,
                        "packets":      25020345,
                        "lost_percent": 74.8704
                },
                "cpu_utilization_percent":      {
                        "host_total":   57.289,
                        "host_user":    11.7294,
                        "host_system":  45.5612,
                        "remote_total": 99.7418,
                        "remote_user":  19.7103,
                        "remote_system":        80.0314
                }
        }
}
```

I'd like us to store that in the JSON results in a new section (maybe called `raw` for example).
Of course, ideally I'd probably now convert all the `iperf3` tests to output json results natively, which we could store - but that's a moderate rewrite. the PPS tests are easy, as they already generate JSON which they then parse.

metrics/network/network-metrics-iperf3.sh
GabyCT: @grahamwhaley , the `iperf_dockerfile` is using `iperf2` which currently is not broken. And the way to reproduce the issue run a container and then download the latest `iperf3` and then try to install it (first you will get an error of libiperf.so that does not exist https://github.com/esnet/iperf/issues/168). Now to solve this issue what I did was the suggested workaround (install lib32z1) and it worked, however, if you want to perform a `iperf3 --version` is broken which is related with the first issue  esnet/iperf#153.

metrics/network/network-metrics-iperf3.sh
GabyCT: @grahamwhaley all the changes were applied, I implemented the function `raw_iperf3_pps_results` to save all the rest of the information

GabyCT:
@grahamwhaley changes applied

GabyCT:
for the documentation, I already create this issue https://github.com/kata-containers/tests/issues/529

metrics/network/network-metrics-iperf3.sh
grahamwhaley: Ah, and, iirc we need to build from source as we need a newer-than-distro-provided iperf3 in order to get the TMPDIR support to work around a 9p issue...
That's just crazy that iperf3 still has bugs open from 2014...
OK, if I had some time I'd have a fiddle, but not right now...

metrics/network/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,591 @@
...
+}
+
+# This test measures the packet-per-second (PPS) between two containers.
+# It uses the smallest (64byte) UDP packet streamed with unlimited bandwidth
+# to obtain the result.
+function iperf3_cnt_cnt_pps() {
+	local TEST_NAME="network pps cnt cnt"
+	local result="$(get_cnt_cnt_pps)"
+	parse_iperf3_pps "$TEST_NAME" "$result"
+	raw_iperf3_pps_results "$TEST_NAME" "$result"
```
grahamwhaley: Oh, OK - I meant to add the raw results into the same .json results file as a new sub-section called 'raw', not into a separate file... so we would keep all the results from a single run together in the one results file.
Well, this will do for now - I'll open an Issue to note that all iperf3 results should probably run in -J json mode and store their raw data in their single results file.

chavafg:
lgtm

PR #487, opened 2018-07-10
ci: support install_qemu in aarch64

Pennyzct:
for now, default hypervisor would be qemu instead of qemu-lite in aarch64. packaged qemu has only be supproted in fedora and ubuntu for aarch64.

Fixes: #472

Signed-off-by: Penny Zheng <penny.zheng@arm.com>
Signed-off-by: Jianyong Wu <jianyong.wu@arm.com>

PR #514, opened 2018-07-19
 ci: add filter scheme for architecture-independent

Pennyzct:
1. Not all test components are supported in all archs, so adding filter scheme to select specific ones for different archs is necessary.
a few test components wouldn't be supported in aarch64 for quite a time(such like: swarm, etc.)
2. For supporting filter scheme in arm64, configuration_aarch64.yaml file should be added to elaborate specific test components in arm64 ci.
for now, this configuration file is just for elaborating whole filter frame.

@jodh-intel @weichen81

### DIFF
--- a/.ci/filter_docker_aarch64.sh
+++ b/.ci/filter_docker_aarch64.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 ARM Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+GOPATH_LOCAL="${GOPATH%%:*}"
+kata_dir="${GOPATH_LOCAL}/src/github.com/kata-containers"
+test_dir="${kata_dir}/tests"
+ci_dir="${test_dir}/.ci"
+test_config_file="${ci_dir}/configuration_aarch64.yaml"
+
+describe_skip_flag="docker.Describe"
+context_skip_flag="docker.Context"
+it_skip_flag="docker.It"
+
+# value for '-skip' in ginkgo
+_skip_options=()
+
+filter_and_build()
+{
+	local dependency="$1"
+	local array_docker=$("${GOPATH_LOCAL}/bin/yq" read "${test_config_file}" "${dependency}")
+	[ "${array_docker}" = "null" ] && return
+	mapfile -t _array_docker <<< "${array_docker}"
+	for entry in "${_array_docker[@]}"
+	do
+		_skip_options+=("${entry#- }|")
+	done
+}
+
+main()
+{
+	# build skip option based on Describe block
+	filter_and_build "${describe_skip_flag}"
+
+	# build skip option based on context block
+	filter_and_build "${context_skip_flag}"
+
+	# build skip option based on it block
+	filter_and_build "${it_skip_flag}"
+
+	skip_options=$(IFS= ; echo "${_skip_options[*]}")
+
+	echo "${skip_options%|}"
+}
+
+main
--- a/.ci/filter_test_aarch64.sh
+++ b/.ci/filter_test_aarch64.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 ARM Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+GOPATH_LOCAL="${GOPATH%%:*}"
+kata_dir="${GOPATH_LOCAL}/src/github.com/kata-containers"
+test_dir="${kata_dir}/tests"
+ci_dir="${test_dir}/.ci"
+test_config_file="${ci_dir}/configuration_aarch64.yaml"
+
+test_filter_flag="test"
+
+_test_union=()
+
+main()
+{
+	local array_test=$("${GOPATH_LOCAL}/bin/yq" read "${test_config_file}" "${test_filter_flag}")
+	[ "${array_test}" = "null" ] && return
+	mapfile -t _array_test <<< "${array_test}"
+	for entry in "${_array_test[@]}"
+	do
+		_test_union+=("${entry#- }")
+	done
+	test_union=$(IFS=" "; echo "${_test_union[*]}")
+	echo "${test_union}"
+}
+
+main
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -13,7 +13,7 @@ source "${cidir}/lib.sh"
 
 apply_depends_on
 
-arch=$(arch)
+arch=$("${cidir}"/kata-arch.sh -d)
 INSTALL_KATA="${INSTALL_KATA:-yes}"
 
 echo "Set up environment"
@@ -40,15 +40,21 @@ if ! sudo docker version | grep -q "$docker_version" && [ "$CI" == true ]; then
 	"${cidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker install -f
 fi
 
-if [ "$arch" = x86_64 ]; then
+case "$arch" in
+	x86_64)
 	if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
 		echo "enable Nested Virtualization"
 		sudo modprobe -r kvm_intel
 		sudo modprobe kvm_intel nested=1
 	fi
-else
+	;;
+	aarch64)
+	info "CI running in bare machine"
+	;;
+	*)
 	die "Unsupported architecture: $arch"
-fi
+	;;
+esac
 
 if [ "${INSTALL_KATA}" == "yes" ];then
 	echo "Install Kata sources"
@@ -58,14 +64,21 @@ fi
 echo "Install CNI plugins"
 bash -f "${cidir}/install_cni_plugins.sh"
 
-echo "Install CRI-O"
-bash -f "${cidir}/install_crio.sh"
+# values indicating whether related intergration tests have been supported
+CRIO="true"
+KUBERNETES="true"
+OPENSHIFT="true"
 
-echo "Install Kubernetes"
-bash -f "${cidir}/install_kubernetes.sh"
+# load arch-specific lib file
+if [ -f "${cidir}/lib_setup_${arch}.sh" ]; then
+        source "${cidir}/lib_setup_${arch}.sh"
+fi
 
-echo "Install Openshift"
-bash -f "${cidir}/install_openshift.sh"
+[ "${CRIO}" = "true" ] && echo "Install CRI-O" && bash -f "${cidir}/install_crio.sh"
+
+[ "${KUBERNETES}" = "true" ] && echo "Install Kubernetes" && bash -f "${cidir}/install_kubernetes.sh"
+
+[ "${OPENSHIFT}" = "true" ] && echo "Install Openshift" && bash -f "${cidir}/install_openshift.sh"
 
 echo "Disable systemd-journald rate limit"
 sudo crudini --set /etc/systemd/journald.conf Journal RateLimitInterval 0s
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ TIMEOUT := 60
 # union for 'make test'
 UNION := functional docker crio docker-compose openshift kubernetes swarm cri-containerd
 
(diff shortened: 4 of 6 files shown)


Weichen81:
@Pennyzct It's good to provide a method for different architecture to configure their test cases : )

Pennyzct:
#516

Makefile
```diff
@@ -4,8 +4,14 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-# The time limit in seconds for each test
-TIMEOUT ?= 60
+# get arch
+ARCH := $(shell bash -c arch)
```
jodh-intel: https://github.com/kata-containers/tests/blob/master/.ci/kata-arch.sh ?

.ci/configuration_aarch64.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+# test components
+test:
+  - functional
+  - docker
+
+# test units for docker integration
+docker:
+  Describe:
+    - docker create
```
jodh-intel: Not knowing much about ginkgo, I don't really understand how this is working - could you explain?

.ci/configuration_aarch64.yaml
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2018 ARM Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# test components
```
jodh-intel: Can you add a comment in this file explaining what it's for please?

x86_64-options.mk
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (c) 2017-2018 Intel Corporation
```
jodh-intel: I think this should probably be copyright ARM as you created it.

I'm in two minds about having this file though. At this stage, I think it would be better if this file did not exist and the variables are defined in `Makefile` for two reasons:

1) By having these variables in the main `Makefile`, all architectures by default will be using the default values.

    If we add support for further architectures and those architectures can use the same values as x86_64, we'll have duplication in `x86_64-options.mk` and `${new_arch}-options.mk`.

2) It introduces a maintenance issue.

    The `UNION` variable lists all the targets that are defined in `Makefile`. I think it would be better to have `UNION` in `Makefile` where the targets are defined to make it easier to remember to update that variable as new test types are added.

aarch64-options.mk
```diff
@@ -0,0 +1,13 @@
+#
```
jodh-intel: For consistency with the runtime, I think it would be better to create this as `arch/aarch64-options.mk`.

x86_64-options.mk
jodh-intel: ... but to be clear we will still need a `aarch64-options.mk` :)

.ci/configuration_aarch64.yaml
jodh-intel: Also, if this config file is skipping tests, we need a github URL for each skipped test (or test category) with details of why the test(s) is/are being skipped.

x86_64-options.mk
Pennyzct: ok. I will relocate them to Makefile as default values for all archs. and reload specific value in aarch64-options.mk to list currently supported integration tests in arm64

.ci/configuration_aarch64.yaml
Pennyzct: What docker.Describe,  docker.Context and docker.It offer is part of the values of ${FOCUS} in regex format. 
Copying from ginkgo documentation:
-focus string
    	If set, ginkgo will only run specs that match this regular expression.
-skip string
    	If set, ginkgo will only run specs that do not match this regular expression.
We only offer '-focus' flag, maybe we should also offer '-skip'.
ginkgo will search 'Describe', 'Context', and 'It' to match this regular expression.
Describe("context need to be matched", func() { ... })
Context("context need to be matched", func() { ... })
It("context need to be matched", func() { ... })

Makefile
Pennyzct: ok. update asap.

.ci/configuration_aarch64.yaml
Pennyzct: OK. maybe "Load specific integration units for aarch64"

.ci/configuration_aarch64.yaml
Pennyzct: For now, configuration_aarch64.yaml is just a demo for displaying how filter scheme works.
specific values are not defined yet.
we are working on verify test cases and integration units in arm arch, but still a long list to do.
in the future, we will add more and more bits into the successful lists(configuration_aarch64.yam).
just like what kaly said in #516.

Makefile
```diff
@@ -24,7 +30,7 @@ else
 	./ginkgo -v functional/ -- -runtime=${RUNTIME} -timeout=${TIMEOUT}
 endif
 
-integration: ginkgo
+docker: ginkgo
```
devimc: I like this change :smile:

jodh-intel:
Hi @Pennyzct - do you think you'll get a chance to update this PR this week?

Pennyzct:
Hi~ @jodh-intel 
so sorry for the delay. I was restructuring the whole filter frame. I divided it into three pieces to respond to three different application. 
I was moving the configuration file into another PR, since it was just a demo file for illustrating. the config file was intended to address what integration test suites are fully supported in arm arch. and it's a long list, we are still working on it.
As you commented last time about removing `aarch64-option.mk `into `arch/` dir, we are considering to create` arch/aarch64` dir to load all aarch64-specific files all across the test project. It may be solved in another issue. For now most lib files were under` .ci/` and @chavafg proposed in #487 to group the lib scripts that are specific for aarch64 in a sub-directory of `.ci/`.  so wdyt, do we just set up one `arch/aarch64` or one for `arch/ `and one for `.ci/`?

Pennyzct:
@jodh-intel @Weichen81 @devimc Updated, ptal.😁

jodh-intel:
Hi @Pennyzct - no problem!

Regarding where to store architecture-dependent information, I personally like the top-level `arch/` approach that we use in the runtime. However, in that repo we have architecture-specific files which *only* set variables. That's the ideal in my opinion.

But if we actually need architecture-specific CI *code*, yes, I agree that that should live in `.ci/$arch/` or `.ci/arch/$arch/`.

.ci/filter_docker_aarch64.sh
```diff
@@ -0,0 +1,51 @@
...
+it_filter_flag="docker.It"
+
+# value for '-focus' in ginkgo
+_focus_options=()
+
+filter_and_build()
+{
+	dependency="$1"
+	array_docker=$("${GOPATH_LOCAL}/bin/yq" read "${test_config_file}" "${dependency}")
+	[ "${array_docker}" = "null" ] && array_docker="" && return
```
jodh-intel: I don't understand the assignment here. Can't it just be the following?

```
[ "${array_docker}" = "null" ] && return
```

Same comment applies to the file below.

.ci/filter_docker_aarch64.sh
```diff
@@ -0,0 +1,51 @@
...
+	# build focus option based on context block
+	filter_and_build "${context_filter_flag}"
+
+	# build focus option based on it block
+	filter_and_build "${it_filter_flag}"
+
+	focus_options=$(IFS= ; echo "${_focus_options[*]}")
+
+	echo "${focus_options%|}"
+
```
jodh-intel: Nit: extraneous blank line.

.ci/filter_test_aarch64.sh
```diff
@@ -0,0 +1,32 @@
...
+_test_union=()
+
+main()
+{
+	array_test=$("${GOPATH_LOCAL}/bin/yq" read "${test_config_file}" "${test_filter_flag}")
+	[ "${array_test}" = "null" ] && array_test="" && return
+	mapfile -t _array_test <<< "${array_test}"
+	for entry in "${_array_test[@]}"
+	do
+		_test_union+=("${entry#- }")
```
jodh-intel: Should this contain a pipe symbol (`|`) like `_focus_options` in `.ci/filter_docker_aarch64.sh`?

Pennyzct:
@jodh-intel so I will move `aarch64-options.mk` under `arch/` and update asap. then I will create another issue to group all lib files under `.ci ` into `.ci/aarch64/`.😊

jodh-intel:
Hi @Pennyzct - well, before you move it, let's get more input from @chavafg as he owns the CI :smile:

.ci/setup.sh
```diff
@@ -40,15 +40,21 @@ if ! sudo docker version | grep -q "$docker_version" && [ "$CI" == true ]; then
...
+	x86_64)
 	if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
 		echo "enable Nested Virtualization"
 		sudo modprobe -r kvm_intel
 		sudo modprobe kvm_intel nested=1
 	fi
-else
+	;;
+	aarch64)
+	info "CI running in bare machine"
```
devimc: Q. Doesn't `arm` support nested ?

Makefile
devimc: please update `.PHONY`, remove `integration` and add `docker`

.ci/setup.sh
grahamwhaley: I think @kalyxin02 said not, or at least not on the slave machine we have for CI.

aarch64-options.mk
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# The time limit in seconds for each test
+TIMEOUT := 120
+
+# union for 'make test'
+UNION := $(shell bash -f .ci/filter_test_aarch64.sh)
+
+# selected test suites for docker integration tests
+FOCUS := $(shell bash -f .ci/filter_docker_aarch64.sh)
```
devimc: I guess you will file some issues to track the tests are not passing on ARM, right?

Weichen81: @devimc We want to create one issue to track the tests are not passing on Arm. In this issue, we will list all nonworking tests, and link PRs to fix them. Create each issue for tests will keep too many flying issues, How do you think about it?

Makefile
Pennyzct: ok. update asap.

.ci/filter_docker_aarch64.sh
Pennyzct: yes, the assignment of "" would be redundant. updated asap.

.ci/filter_test_aarch64.sh
Pennyzct: The symbol `| `is not a pipe symbol, but a OR operator  for regex expression. variable `test_union` will offer value to `UNION` in `Makefile`, which indicates what integration tests are run. default `UNION` in x86_64 is `functional docker crio docker-compose openshift kubernetes swarm cri-containerd`.

.ci/filter_docker_aarch64.sh
Pennyzct: 😆，delete asap.

aarch64-options.mk
devimc: I agree

.ci/setup.sh
devimc: great! thanks for the info @Weichen81

jodh-intel:
Hi @chavafg - ptal.

chavafg:
Hi @Pennyzct, @jodh-intel, yes, I agree with the approach. Please make the changes :)

Makefile
grahamwhaley: Still waiting for that `.PHONY` update s/integration/docker/ change afaict. Are there other pending changes on this PR @Pennyzct ?

devimc: @grahamwhaley  https://github.com/kata-containers/tests/pull/514/files#diff-b67911656ef5d18c4ae36cb6741b7965R93

grahamwhaley: Hmm, oh. I wasn't seeing that in my github file view just now.... thx @devimc

chavafg:
lgtm

as refactor to the `.ci/aarch64` will be handled after this one.

PR #515, opened 2018-07-19
crio test fixes

chavafg:
Changes:
- only export LVM_DEVICE when using devicemapper
- ensure crio service is stopped before running tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ openshift:
 	bash -f .ci/install_bats.sh
 	bash -f integration/openshift/run_openshift_tests.sh
 
-test: functional integration crio docker-compose openshift kubernetes swarm cri-containerd
+test: functional integration crio docker-compose openshift kubernetes swarm
 
 check: checkcommits log-parser
 
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ kubernetes:
 	bash -f integration/kubernetes/run_kubernetes_tests.sh
 
 swarm:
+	systemctl is-active --quiet docker || sudo systemctl start docker
 	bash -f .ci/install_bats.sh
 	cd integration/swarm && \
 	bats swarm.bats
--- a/integration/cri-o/cri-o.sh
+++ b/integration/cri-o/cri-o.sh
@@ -79,6 +79,15 @@ if [ "$ID" == "fedora" ] || [ "$ID" == "centos" ]; then
 	export STORAGE_OPTIONS="$OVERLAY_STORAGE_OPTIONS"
 fi
 
+echo "Ensure crio service is stopped before running the tests"
+if systemctl is-active --quiet crio; then
+	sudo systemctl stop crio
+fi
+
+echo "Ensure docker service is stopped before running the tests"
+if systemctl is-active --quiet docker; then
+	sudo systemctl stop docker
+fi
 
 ./test_runner.sh ctr.bats
 
--- a/integration/cri-o/cri-o.sh
+++ b/integration/cri-o/cri-o.sh
@@ -16,7 +16,7 @@ crio_repository="github.com/kubernetes-incubator/cri-o"
 crio_repository_path="$GOPATH/src/${crio_repository}"
 
 # devicemapper device and options
-export LVM_DEVICE=${LVM_DEVICE:-/dev/vdb}
+LVM_DEVICE=${LVM_DEVICE:-/dev/vdb}
 DM_STORAGE_OPTIONS="--storage-driver devicemapper --storage-opt dm.directlvm_device=${LVM_DEVICE}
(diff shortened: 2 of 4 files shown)


GabyCT:
lgtm

chavafg:
ok, so this didn't fix the issue. I also cannot reproduce it locally...
@sboeuf any idea of what else could be wrong? 
We haven't changed anything related to cri-o versions and yesterday it was working ok.

chavafg:
This already fixes the cri-o tests, but now I am seeing issues with a k8s tests and cri-containerd test.

k8s exec command returning this error:
```
zuul@ubuntu-xenial-vexxhost-vexxhost-ca-ymq-1-0000830605:~/src/github.com/kata-containers/tests/integration/kubernetes$ sudo -E kubectl exec constraints-cpu-test -c first-cpu-container -- nproc
error: unable to upgrade connection: 404 page not found
```

cri-containerd test error:
```
=== RUN   TestImageLoad
E0719 20:06:37.751921   29002 remote_runtime.go:115] StopPodSandbox "7ebcebfcccdc557b46d31343774d6daaf40d993fe5c614ad0675f56a258890b2" from runtime service failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded
E0719 20:06:37.753199   29002 remote_runtime.go:132] RemovePodSandbox "7ebcebfcccdc557b46d31343774d6daaf40d993fe5c614ad0675f56a258890b2" from runtime service failed: rpc error: code = Unknown desc = sandbox container "7ebcebfcccdc557b46d31343774d6daaf40d993fe5c614ad0675f56a258890b2" is not fully stopped
--- FAIL: TestImageLoad (62.01s)
	image_load_test.go:41: docker save image into tarball
	image_load_test.go:53: make sure no such image in cri
	image_load_test.go:60: load image in cri
	image_load_test.go:65: make sure image is loaded
	image_load_test.go:71: create a container with the loaded image
	image_load_test.go:89: make sure container is running
	assertions.go:237: 
                        
	Error Trace:	image_load_test.go:76
		
			image_load_test.go:93
		
	Error:		Received unexpected error rpc error: code = DeadlineExceeded desc = context deadline exceeded
		
	assertions.go:237: 
                        
	Error Trace:	image_load_test.go:77
		
			image_load_test.go:93
		
	Error:		Received unexpected error rpc error: code = Unknown desc = sandbox container "7ebcebfcccdc557b46d31343774d6daaf40d993fe5c614ad0675f56a258890b2" is not fully stopped
		
FAIL
./hack/test-utils.sh: line 53: 28924 Terminated              keepalive "sudo PATH=${PATH} ${ROOT}/_output/containerd ${CONTAINERD_FLAGS}" 
```

Btw, the cri-containerd `DeadlineExceeded` error seems to not be related with  the kata-containers runtime, as I couldnt find that pod id in the runtime or proxy logs.

bergwolf:
@chavafg Could we investigate what has changed in the last two days that caused the CI failures and how that change got in without CI validation?

I am particularly interested in the environment set up part, because assuming we gate everything within kata, the breakage likely comes from CI environment set up, e.g., are we installing latest version of any components (e.g., base OS/golang/k8s/docker/containerd/crio etc.) or do we always stay with certain version of them? Take the docker service conflict for an example, either it is not installed/enabled before, or it does not conflict with cri-o service before. So clearly something has changed about either docker or cri-o in the CI machines.

If we can identify what has changed, we can easily revert/gate that change and then CI can be restored...

chavafg:
Tried to identified if something had changed in our repos, but seems like nothing has changed, so I suspect that something in the vexxhost cloud infra changed. 

All the third-party components that are installed, are fixed to a version that is specified in the https://github.com/kata-containers/runtime/blob/master/versions.yaml and the last changed there was 8 days ago, so I think we can discard a version is affecting our CI.

Also, I have tested in an Azure machine and I am not seeing any of this problems.  

In the meantime, the fastest solution to have a CI working right now is to disable the `k8s-cpu-ns` test and the cri-containerd tests.

bergwolf:
@chavafg Thanks for the details. I thought vexxhost provides us VMs, right? If yes, can we control the base OS image to boot up the VM on vexxhost?

sboeuf:
@bergwolf yes that's exactly what we should do. Still trying to identify what is going wrong with this new environment, but basically, we need (as the rest of the stack) to ensure that our VMs are not moving to a different image/kernel without knowing.

chavafg:
yes, we should control the images, but we currently don't. @mnaser do you think we could be able to have our own images, so if there is an update with a change that could potentially break us, we don't need to worry?

@bergwolf, @sboeuf, @grahamwhaley this is all green now, with cri-containerd tests and one of the k8s tests skipped for the moment.

grahamwhaley:
Green green! - just the non-required f27 jenkins running - heh, I'm going to merge it!

PR #524, opened 2018-07-23
Re-enable cri-containerd and k8s tests

chavafg:
This PR contains 3 commits:

1. Fixes the container-manger.sh to install the correct docker version on centos. 
2. Adds a check to the setup.sh to install correct docker version if the installed one is different.
3. Re-enable cri-containerd and k8s tests


GabyCT:
lgtm

Issue #526, opened 2018-07-23
ci: silently fail if no nesting enabled


grahamwhaley:
If the build machine does not have nested enabled, presently our grep of the settings comes out as an error in the logs - whereas presently it is benign in the [setup](https://github.com/kata-containers/tests/blob/master/.ci/setup.sh#L36-L39).

```
Install cri-containerd dependencies
Install crudini
grep: /sys/module/kvm_intel/parameters/nested: No such file or directory
Install Kata sources
Install kata-containers image
```

Redirect the error to `/dev/null` to keep the logs clean.

grahamwhaley:
We've decided to handle this more strongly - replaced by #586

PR #527, opened 2018-07-23
CI: Die if we do not have nested virtualisation

grahamwhaley:
We grep `/sys/module/kvm_intel/parameters/nested` to check
if we need to load the modules, but if that file does not exist
we get:
  `No such file or directory`
in the logs.

The check failing is currently benign, so ditch the error to
/dev/null

Fixes: #526

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -44,10 +44,13 @@ fi
 
 case "$arch" in
 	x86_64)
-	if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
+	if [ "$CI" == true ] && grep -q "N" /sys/module/kvm_intel/parameters/nested 2> /dev/null; then
 		echo "enable Nested Virtualization"
 		sudo modprobe -r kvm_intel
 		sudo modprobe kvm_intel nested=1
+		if grep -q "N" /sys/module/kvm_intel/parameters/nested 2> /dev/null; then
+			die "Failed to find or enable Nested virtualization"
+		fi
 	fi
 	;;
 	aarch64)


grahamwhaley:
I'm not sure this is the right thing to do though. Must we have nested enabled to run .... probably...?
@chavafg - thoughts?
@WeiZhang555 - you noted seeing this over on https://github.com/kata-containers/runtime/pull/301

If we decide the correct action is to abort/fail at this point - let me know and I'll mod the PR - let's either make this cleanly pass or noisily fail :-)

chavafg:
hmm, now I see:
```
time="2018-07-23T15:56:26Z" level=error msg="open /sys/module/kvm_intel/parameters/nested: no such file or directory" name=kata-runtime pid=6545 source=runtime 
```
when installing the runtime.
Should we also enable the nested virtualization when the file does not exist? I mean in addition to the `grep`?

sboeuf:
@chavafg yeah I guess we should try to enable it if not found. And if still not found after we enabled it, then there's a real issue.

jodh-intel:
CI borkage. Prolly quicker to re-force-push to respin the wheels :)

grahamwhaley:
I've a rebase ready to push - but before I do that, I suspect we have decided we want to fix this differently....
@chavafg @sboeuf - so in the event of not finding the nested entry, do we want to:
- first insmod the driver (err, will that be different for different ARCH's??)
- and if the device/mode is still not existant/enabled, then bork out with a `die()`?

Let me know, and I'll attempt a rewrite..

sboeuf:
>@chavafg @sboeuf - so in the event of not finding the nested entry, do we want to:
>- first insmod the driver (err, will that be different for different ARCH's??)

Yes we would have to load a different driver and not the `kvm_intel` one.

>- and if the device/mode is still not existant/enabled, then bork out with a die()?

Yes, this looks like the easiest approach IMO.

grahamwhaley:
Taking a peek at this, it is going to be better to rebase/write once #514 has landed, as that reflows this piece of code... so, I'll leave this pending for a little bit.

grahamwhaley:
Oh, that was fun....
- made a new issue to cover our changed plans (to die if we don't have nested)
- so had to rewrite the PR title - sorry if that confuses anybody or some tooling...
- and then finally, a question - do we *really* need to have nested virtualisation enabled to run the tests? I think .... only if we are already running in a VM, like the majority of the CIs. But, will this potentially stop people running the tests locally on their bare metal machines? If so, then ideally we would only do this check if we found we were already in a VM - but, iirc, that is quite hard to find out...

thoughts @chavafg @sboeuf @jodh-intel ?

chavafg:
at least for the `x86_64` arch, we know that the CI machines will actually be VMs, so we can have this and in addition check that we have the `CI=true` envar, wdyt?

grahamwhaley:
@chavafg - gating the check on `CI=true` sounds good to me - nice idea! I'll try to add that tomorrow.

grahamwhaley:
Updated with check for `$CI` @chavafg

GabyCT:
lgtm

PR #534, opened 2018-07-24
tests: Add k8s memory constraints tests

### LINKED ISSUE #533: Add k8s memory constraints tests
Add tests when we are trying to launch a pod exceeding memory limit and within memory limit.

GabyCT:
These tests when we are trying to launch a pod exceeding
memory limit and within memory limit.

Fixes #533

### DIFF
--- a/integration/kubernetes/k8s-cpu-ns.bats
+++ b/integration/kubernetes/k8s-cpu-ns.bats
@@ -20,7 +20,7 @@ setup() {
 }
 
 @test "Check CPU constraints" {
-	wait_time=30
+	wait_time=120
 	sleep_time=5
 
 	# Create the pod
--- a/integration/kubernetes/k8s-memory.bats
+++ b/integration/kubernetes/k8s-memory.bats
@@ -16,8 +16,10 @@ setup() {
 	memory_limit_size="50Mi"
 	allocated_size="250M"
 	# Create test .yaml
-	cat pod-memory-limit.yaml | sed -e "s/\${memory_size}/${memory_limit_size}/" | \
-		sed -e "s/\${memory_allocated}/${allocated_size}/" > test_exceed_memory.yaml
+        sed \
+            -e "s/\${memory_size}/${memory_limit_size}/" \
+            -e "s/\${memory_allocated}/${allocated_size}/" \
+            pod-memory-limit.yaml > test_exceed_memory.yaml
 
 	# Create the pod exceeding memory constraints
 	run sudo -E kubectl create -f test_exceed_memory.yaml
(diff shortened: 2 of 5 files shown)


chavafg:
lgtm

integration/kubernetes/k8s-cpu-ns.bats
```diff
@@ -20,7 +20,7 @@ setup() {
 }
 
 @test "Check CPU constraints" {
-	wait_time=30
+	wait_time=120
```
egernst: Should we just wait to see the pods status based on watching instead of sleep?  Or, WDYT?

egernst:
..and, good test - thank. you.

integration/kubernetes/k8s-cpu-ns.bats
grahamwhaley: I think - if we can - yes, we should do sleepy polls across the whole codebase rather than hard wired 'sleep and pray' :-)
If we can make this a poll loop with a sleep and a timeout, then great, please do :-)

chavafg: This is what we do, you can check: 
https://github.com/kata-containers/tests/blob/master/.ci/lib.sh#L145-L158

It polls the status every 5 secs... so the maximum is 120 so that we can fail the test.

integration/kubernetes/k8s-memory.bats
```diff
@@ -0,0 +1,47 @@
...
+	rm -f test_exceed_memory.yaml
+}
+
+@test "Running within memory constraints" {
+	memory_limit_size="200Mi"
+	allocated_size="150M"
+	wait_time=120
+	sleep_time=5
+	# Create test .yaml
+	cat pod-memory-limit.yaml | sed -e "s/\${memory_size}/${memory_limit_size}/" | \
```
devimc: I think you don't need `cat` and two `sed`

```sh
sed \
    -e "s/\${memory_size}/${memory_limit_size}/" \
    -e "s/\${memory_allocated}/${allocated_size}/" \
    pod-memory-limit.yaml > test_within_memory.yaml
```

integration/kubernetes/k8s-cpu-ns.bats
GabyCT: @egernst and @grahamwhaley ^

grahamwhaley: yep, understood - our mistake :-)

GabyCT:
@devimc changes applied

devimc:
ping @egernst

jodh-intel:
@egernst - please can you re-review as your nack is holding this from being merged.

PR #545, opened 2018-07-30
tests: Check that processes are not left behind before running swarm

### LINKED ISSUE #544: Check that processes are not left behind before running swarm
Check that processes like shim or proxy are not left behind before it runs the
integration swarm tests.

GabyCT:
This will check that processes are not left behind before it runs
the integration swarm tests.

Fixes #544

### DIFF
--- a/integration/swarm/swarm.bats
+++ b/integration/swarm/swarm.bats
@@ -8,6 +8,8 @@
 # Swarm testing : This will start swarm as well as it will create and
 # run swarm replicas using Nginx
 
+load "${BATS_TEST_DIRNAME}/../../metrics/lib/common.bash"
+
 # Image for swarm testing
 nginx_image="gabyct/nginx"
 # Name of service to test swarm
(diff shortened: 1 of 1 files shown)


integration/swarm/swarm.bats
```diff
@@ -37,6 +37,9 @@ check_processes() {
 }
 
 setup() {
+	check_processes ${PROXY_PATH}
+	check_processes ${SHIM_PATH}
```
devimc: what about qemu and runtime?

GabyCT:
@devimc changes applied

jodh-intel:
lgtm

integration/swarm/swarm.bats
```diff
@@ -8,6 +8,8 @@
 # Swarm testing : This will start swarm as well as it will create and
 # run swarm replicas using Nginx
 
+load "${BATS_TEST_DIRNAME}/../../metrics/lib/common.bash"
```
grahamwhaley: Sorry, late to the party...
So, if we are going to start using these check functions outside of the metrics (and we should...), then really we should probably lift the code out of the metrics lib and into the .ci/lib.sh more global place. Which, iirc, the metrics then pulls in anyway, so hopefully that code move should be less pain that it may have otherwise been.

integration/swarm/swarm.bats
```diff
@@ -18,25 +20,18 @@ number_of_replicas=1
...
-# SHIM PATH
-SHIM_PATH="${SHIM_PATH:-/usr/libexec/kata-containers/kata-shim}"
-# PROXY PATH
-PROXY_PATH="${PROXY_PATH:-/usr/libexec/kata-containers/kata-proxy}"
 
-# This function checks if active processes were
-# left behind by kata-runtime.
+setup() {
+	# Check that processes are not running
+	run check_processes ${PROXY_PATH}
```
grahamwhaley: What I'd really like is for these sanity checks to be run (be)for(e) every test :-) Have a peek at #215 for more info.
Having a quick look at [bats](https://github.com/sstephenson/bats#load-share-common-code), it seems that the `load` command above is just a `source` underneath... in which case.... I think we could write a `bats_common.sh`, that could source `common.sh`, and then run a set of sanity checks - and make all the bats files `load` that common file. That should then have every bats test run the system sanity check.
Actually, I'd like to add that check to both `setup` and `teardown`, but that is harder to do in a nice way that folks won't forget to do for every new bats file.

I'm open to ideas here - but I think with a little more effort we could have gotten these checks, and probably some others (inspired from the mega death soak test) into all the bats files :-)

PR #552, opened 2018-08-01
metrics: fix proxy version and add kata-env to JSON results

grahamwhaley:
Fix the proxy version extraction for the JSON metrics results data.
Add storage of the complete `kata-env` json output to the JSON metrics results data.


grahamwhaley:
There is more cleanup/work to do here later, for instance to add other runtime specific data (and not store kata data in that instance). That will come in a later series when we add the ability to test other runtimes. At that time the "env" section may disappear completely.
We have to be slighlty careful about forward/backward compatibility, as once we start storing results into some database, then some of them do not support changing the fundamental data layout in the same database.

jodh-intel:
lgtm. One of the Travis CI's seems a bit perturbed though:

```
OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:86: executing setns process caused \"exit status 21\"": unknown
```

grahamwhaley:
Yeah, that would seem to be completely unrelated to this change (that does not even execute on the QA CI afaik)...
Is it @bergwolf who was collecting up the CI failure modes?
I'm going to see if I can nudge a rebuild...

GabyCT:
lgtm

PR #557, opened 2018-08-02
CI: Add procenv output to teardown

### LINKED ISSUE #556: Run procenv in teardown script
Ideally, we'd run this before the build (as it might help diagnose why the build failed).

However, since `procenv` produces a lot of *lines* of output (~4k), I think we might need to settle for adding it to the teardown script for now.

See: https://github.com/kata-containers/ci/issues/14.

jodh-intel:
Run the `procenv(1)` tool (which dumps as much information about it's
environment as possible) twice at the end of the CI: once as the CI
user and once as `root`. This could be useful to diagnose test failures
by comparing its CI output with the output generated on a local system.

Fixes #556.


jodh-intel:
I believe so (since the teardown script is called with a parameter in the Jenkins config files).

.ci/teardown.sh
```diff
@@ -81,13 +91,22 @@ collect_logs()
...
+
 		local prefix
 
 		for prefix in $prefixes
 		do
 			gzip -9 "$prefix"*
 		done
 
+		# The procenv log is tiny so doesn't require chunking
+		gzip -9 "${procenv_log_path}"
```
devimc: I don't see where this variable was defined, I guess you mean `procenv_user_log_path` and/or `procenv_root_log_path`

jodh-intel: Aieeee! Good catch - branch updated ;)

jodh-intel:
Hi @chavafg - I just noticed that we don't have the procenv artifacts attached to the CI results. Is it possible we're doing some regex matching on "`*_[0-9][0-9]*.gz`" files to attach? If so, that would explain it as the files are called:

```
procenv-root.log.gz
procenv-${USER}.log.gz
```

chavafg:
Hi @jodh-intel, yes, need to add them on the jenkins jobs. 
Opened https://github.com/kata-containers/ci/issues/53 to address this.

Issue #561, opened 2018-08-03
Consider changing `docker stop` to `docker kill` in the tests


grahamwhaley:
`docker stop` has some interesting side effects to do with SIGTERM etc. when trying to kill 'PID1' processes and similar. See moby/moby#3766 for a good jumping off point to many discussions on this.
Fundamentally, what happens in some of our tests at present is we fire a `stop`, which sends off a signal, then waits for its 10s default timeout, and then does a `kill` anyway. That means we are taking on avg. about 10.5s to kill off a container once we are finished using it for tests.
It's not a huge problem, but the metrics runs at least 2x20 container tests, which then take at least (10.5x40) == 420s == 7 minutes to shut down. They should take more like 0.5sx20 == 10s to shut down.... I'd be really happy to shave 6m50s off of my metrics runtimes ;-)

`docker stop` can be found in at least:
https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L194
https://github.com/kata-containers/tests/blob/master/metrics/density/docker_memory_usage.sh#L285
https://github.com/kata-containers/tests/blob/master/metrics/density/footprint_data.sh#L157

Whilst there, we should probably ensure that our `docker stop` performance is the same as the default runtime - it is possible we are handling some signals differently or we have different semantics as we are inside a VM (although the workload in the VM is not the real PID1, and is namespaced to PID1 the same as a soft container would be, so I think it should be the same)...

PR #563, opened 2018-08-03
metrics: use docker kill rather than docker stop

grahamwhaley:
For many of our 'used' test containers, issuing a `docker stop`
actually results in the stop timing out (as it cannot signal the
container to stop for any one of a number of potential reasons),
and then docker reverts to a kill anyhow. The cost of this is a
10s delay per `stop`, whilst docker times out and then issues the
kill. That costs us heavily when we are doing multi-container tests
using say 20 or 100 containers, as each one takes a 10s overhead
on stop/kill/removing it.

Move all the metrics test `stops` to be `kills`.

Fixes: #561

### DIFF
--- a/metrics/density/docker_memory_usage.sh
+++ b/metrics/density/docker_memory_usage.sh
@@ -282,7 +282,9 @@ EOF
 	# Or, in fact we can leave the '--rm' on the docker run, and just change this
 	# to a 'docker stop *'.
 	for c in ${containers[@]}; do
-		docker stop $c
+		# Use a kill, as the containers are benign, and most of the time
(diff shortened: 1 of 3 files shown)


jodh-intel:
lgtm

It would be great if we could drop the `sleep 3`'s on the lines below the changes though :)

chavafg:
lgtm

grahamwhaley:
@jodh-intel - yeah - I need to verify that all the yamux type fixes have made it stable enough to drop those - and didn't have time right now. I thought I had an Issue open for that so we don't forget, but cannot see one. Let me go create....

PR #566, opened 2018-08-03
kata-doc-to-script: Remove duplicate warning from help message

nitkon:
Fixes: #565

Signed-off-by: Nitesh Konkar niteshkonkar@in.ibm.com

### DIFF
--- a/.ci/kata-doc-to-script.sh
+++ b/.ci/kata-doc-to-script.sh
@@ -48,8 +48,6 @@ Options:
   -s : strict mode - perform extra checks.
   -v : verbose mode.
 
-${warning}
-
 Example usage:
 
   $ ${script_name} foo.md foo.md.sh


jodh-intel:
That was in fact added on purpose. However, on reflection I think once *should* be enough (famous last words... :)

lgtm

nitkon:
@jodh-intel : So I executed  `kata-doc-to-script`  on Developer-Guide.md file but the results were not as expected.
```
set -e

sudo systemctl daemon-reload
sudo systemctl restart kubelet

# If using CRI-O
sudo kubeadm init --skip-preflight-checks --cri-socket /var/run/crio/crio.sock --pod-network-cidr=10.244.0.0/16

# If using CRI-containerd
sudo kubeadm init --skip-preflight-checks --cri-socket /run/containerd/containerd.sock --pod-network-cidr=10.244.0.0/16

export KUBECONFIG=/etc/kubernetes/admin.conf
```

Shouldn't all the code blocks be a part of the script or am I missing something?

grahamwhaley:
Hi @nitkon 
I believe, by design, the `kata-doc-to-script` only converts sections marked as type `bash`. We do this as some sections are not suitable or cannot be executed in a script. Mostly we use the converter for the installation documents. I don't think it can apply to all of the developer document.
I'll let @jodh-intel confirm/correct that.

jodh-intel:
Hi @nitkon - @grahamwhaley is correct. The `kata-doc-to-script` **only** extracts `bash` code blocks. But also, it is meaningless trying to run it over the developer guide as that is full of alternatives - to work, the document needs to have a non-branching flow, top to bottom. We guarantee that for the install guides (both parts - O/S part and container manager part):

- https://github.com/kata-containers/documentation/tree/master/install
- https://github.com/kata-containers/documentation/tree/master/install/docker

nitkon:
@jodh-intel: Thanks for the explanation. I thought the script would give an option to select initrd/rootfs; which distro to select etc. I wanted to execute the developer guide as a script with options asked/populated, so was exploring it out. Thanks! :-)

jodh-intel:
That would be great indeed and ensure -- like we currently do for the install guides -- that the dev guide is always correct.

However https://github.github.com/gfm/ is intentionally basic so doesn't support all the cool stuff that doc formats like rST and asciidoc do.

However, I think we could conceivably allow *parts* of the dev guide to be executed. For example, we'd like an osbuilder test to ensure the debug console feature explained in the dev guide always works. Assuming we can extract just that section, I think that is possible. See:

- https://github.com/kata-containers/osbuilder/issues/96

PR #569, opened 2018-08-06
ci: using ephemeral port to avoid conflicting

Pennyzct:
Former port "8080" is a frequently used user port, and could cause 'address already in use' err. 
Above error maybe would not occur where CI is set up in vm, but it's possible in bare machine.
So I think maybe using ephemeral port would be a better option.
Furthermore, I have considered that we could pick free port from 1024 to 49151, but it will need `netstat` or `lsof` to check port state, leading to extra dependency. 
sooooooo, wdyt?😅 @jodh-intel

### DIFF
--- a/integration/docker/port_test.go
+++ b/integration/docker/port_test.go
@@ -17,7 +17,7 @@ var _ = Describe("port", func() {
 
 	BeforeEach(func() {
 		id = randomDockerName()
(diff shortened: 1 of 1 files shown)


chavafg:
lgtm

GabyCT:
lgtm

jodh-intel:
Thanks @Pennyzct - looks good based on https://en.wikipedia.org/wiki/Ephemeral_port.

PR #575, opened 2018-08-06
CI: compare only MAJOR.MINOR version of docker

### LINKED ISSUE #574: ci: compare only major.minor version of docker in setup.sh
We check that the docker version installed in the system is the one that we officially support on `versions.yaml`. But we need to only get the `MAJOR.MINOR` of the whole version string. i.e. from `v18.06-ce` we only need `18.06` to make the correct comparison.

chavafg:
In our .ci/setup.sh script, we make a comparsion of
the installed docker version. But we only need the major
and minor versions to make the correct comparison. This
change removes unneeded characters from the version
string.

Fixes #574.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -36,6 +36,8 @@ fi
 # in versions.yaml. If there is a different version installed,
 # install the correct version..
 docker_version=$(get_version "externals.docker.version")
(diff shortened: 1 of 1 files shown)


GabyCT:
lgtm

jodh-intel:
lgtm

Issue #577, opened 2018-08-07
Add vm factory tests


bergwolf:
So that we include vm factory basic functionality tests in our daily CIs.

PR #578, opened 2018-08-07
integration: add vm factory integration test

bergwolf:
To test vm template init and destroy.

Depends-on: kata-containers/runtime#684
Fixes: #577

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@
 TIMEOUT := 60
 
 # union for 'make test'
-UNION := functional docker crio docker-compose docker-stability openshift kubernetes swarm cri-containerd
+UNION := functional docker crio docker-compose docker-stability openshift kubernetes swarm cri-containerd vm-factory
 
 # skipped test suites for docker integration tests
 SKIP :=
@@ -84,21 +84,26 @@ openshift:
 	bash -f .ci/install_bats.sh
 	bash -f integration/openshift/run_openshift_tests.sh
 
+vm-factory:
+	bash -f integration/vm_factory/vm_templating_test.sh
+
 test: ${UNION}
 
 check: checkcommits log-parser
 
+# PHONY in alphabetical order
 .PHONY: \
 	check \
 	checkcommits \
 	crio \
+	docker \
 	docker-compose \
 	docker-stability \
 	functional \
 	ginkgo \
-	docker \
 	kubernetes \
 	log-parser \
 	openshift \
 	swarm \
-	test
+	test \
+	vm-factory
--- a/integration/vm_factory/vm_templating_test.sh
+++ b/integration/vm_factory/vm_templating_test.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 HyperHQ Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test will perform several tests to validate kata containers with
+# vm templating.
+
+set -e -x
+
+cidir=$(dirname "$0")
+
+source "${cidir}/../../metrics/lib/common.bash"
+
+# Environment variables
+IMAGE="${IMAGE:-busybox}"
+CONTAINER_NAME=$(random_name)
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+
+kata_runtime_bin=$(command -v kata-runtime)
+
+template_tmpfs_path="/run/vc/vm/template"
+
+enable_vm_template_config() {
+	echo "enable vm template config ${RUNTIME_CONFIG_PATH}"
+	sudo sed -i -e 's/^#\(enable_template\).*=.*$/\1 = true/g' "${RUNTIME_CONFIG_PATH}"
+}
+
+disable_vm_template_config() {
+	echo "disable vm template config"
+	sudo sed -i -e 's/^\(enable_template\).*=.*$/#\1 = true/g' "${RUNTIME_CONFIG_PATH}"
+}
+
+init_vm_template() {
+	echo "init vm template"
+	sudo -E PATH=$PATH "$kata_runtime_bin" factory init
+}
+
+destroy_vm_template() {
+	echo "destroy vm template"
+	sudo -E PATH=$PATH "$kata_runtime_bin" factory destroy
+	# verify template is destroied
+	res=$(mount | grep template | wc -l)
+	[ $res -eq 0 ] || die "template factory is not cleaned up"
+}
+
+check_vm_template_factory() {
+	echo "checking vm template factory"
+	mount | grep tmpfs | grep ${template_tmpfs_path}
+	res=$(sudo ls ${template_tmpfs_path} | grep memory |wc -l)
+	[ $res -eq 1 ] || die "template factory is not set up, missing memory file"
+	res=$(sudo ls ${template_tmpfs_path} | grep state |wc -l)
+	[ $res -eq 1 ] || die "template factory is not set up, missing state file"
+}
+
+setup() {
+	clean_env
+	extract_kata_env
+}
+
+check_qemu_for_vm_template() {
+	echo "checking qemu command line for vm template arguments"
+	ps aux | grep ${HYPERVISOR_PATH} | grep ${template_tmpfs_path}
+	[ $? -eq 0 ] || die "vm is not backed by vm template"
+}
+
+test_create_container_with_vm_template() {
+	sudo docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+	check_qemu_for_vm_template
+	sudo docker rm -f $CONTAINER_NAME
+}
+
+test_factory_init_destroy() {
+	echo "test kata-runtime factory init destroy"
+	enable_vm_template_config
+	init_vm_template
+	check_vm_template_factory
+	test_create_container_with_vm_template
+	destroy_vm_template
+	disable_vm_template_config
+}
+
+test_docker_create_auto_init_vm_factory() {
+	echo "test docker create auto init vm factory"
+	enable_vm_template_config
+	sudo docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+	check_vm_template_factory
+	check_qemu_for_vm_template
+	sudo docker rm -f $CONTAINER_NAME
+	destroy_vm_template
+	disable_vm_template_config
+}
+
+teardown() {
+	clean_env
+}
+
+if [ -z $INITRD_PATH ]; then
+	echo "Skipping vm templating test as initrd is not set"
+	exit 0
+fi
+
+echo "Starting vm templating test"
+setup
+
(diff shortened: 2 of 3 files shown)


grahamwhaley:
typo in msg s/tempalte/template/ :-)

bergwolf:
@grahamwhaley Good catch, Thanks! That's why I have `wip` there ;) Waiting first round CI results before updating.

grahamwhaley:
Heh heh - maybe I should go review the soak test for `function` ;-) (I have to rebase anyway as we seem to have maybe 3 PRs in flight messing with the same bit of the Makefile!).
re: BATS - most of the tests are written in BATS. I personally am not a huge fan, mostly as it does not seem to have a debug mode, so I find writing them very painful! - but, we should ask @chavafg and @devimc what they think. Yeah, I know the soak test is not BATS - mostly because it was written to find a bug, and then got added as a test.

@chavafg @devimc @jodh-intel - do we have a policy/theory/rule on BATS usage? I'm open to input here (and will re-write the soak test on #414 if we say BATS rules etc.).

thx @bergwolf :-)

grahamwhaley:
/me sees you have a merge conflict in `Makefile` as well - probably need to rebase before the metrics will pass the 'merge'.

jodh-intel:
Weeeeell, personally, bats tests sound good in theory, but they are a royal pita to debug. In fact, I switched the osbuilder bats test to a shell script as bats wasn't actually buying us much, and it allowed me to keep my sanity debugging all the various failure scenarios:

- https://github.com/kata-containers/osbuilder/pull/114/commits (see middle commit)

chavafg:
We don't have a strict policy for adding tests using `bats`, so if you think that debugging would be easier using a `bash` file, it is ok for me.

bergwolf:
Thanks for the inputs guys. I'll keep it in bash then to make it easier for me to debug the script. Will rebase/update soon.

bergwolf:
Still working on it. I'm seeing random proxy connection failure when running this in a loop locally. I'll fix that up before pushing the PR forward so that it won't block CI randomly.

ydjainopensource:
@bergwolf Any updates on this one?

bergwolf:
It looks like the `Depends-on` tag can only be set once. @chavafg could you please take a look?
```
Repository github.com/kata-containers/runtime was already defined in a 'Depends-on:' tag.
Only one repository per tag is allowed.
```

chavafg:
Hi @bergwolf, yes, we only though on adding one tag per repo, but will check how to enhance it. I've opened #719 so I don't forget it.

integration/vm_factory/vm_templating_test.sh
```diff
@@ -0,0 +1,113 @@
...
+}
+
+disable_vm_template_config() {
+	echo "disable vm template config"
+	sudo sed -i -e 's/^\(enable_template\).*=.*$/#\1 = true/g' "${runtime_config_path}"
+}
+
+init_vm_template() {
+	echo "init vm template"
+	sudo -E PATH=$PATH kata-runtime factory init
```
chavafg: centos and fedora do not find `kata-runtime` when using sudo (PATH problem it seems). We have hit this before and we have use the full path of the binary using:
```
kata_runtime_bin=$(command -v kata-runtime)
sudo -E  "$kata_runtime_bin" factory init
```

integration/vm_factory/vm_templating_test.sh
```diff
@@ -0,0 +1,113 @@
...
+	clean_env
+}
+
+check_qemu_for_vm_template() {
+	echo "checking qemu command line for vm template arguments"
+	ps aux | grep qemu | grep ${template_tmpfs_path} | wc -l
+}
+
+test_create_container_with_vm_template() {
+	docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
```
chavafg: use `sudo` for docker in case the tests is run using a user that is not part of the docker group? Same for all other docker calls below.

bergwolf:
CI green. @kata-containers/tests PTAL.

jodh-intel:
Thanks @bergwolf!

lgtm

bergwolf:
While it's ready for reviewing, please do not merge it accidentally since it Depends-on: kata-containers/runtime#684

integration/vm_factory/vm_templating_test.sh
```diff
@@ -0,0 +1,120 @@
...
+IMAGE="${IMAGE:-busybox}"
+CONTAINER_NAME=$(random_name)
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+
+# Timeout is the duration of this test (seconds)
+# We want to stress the agent for a significant
+# time (approximately running for two days)
+timeout=186400
+start_time=$(date +%s)
+end_time=$((start_time+timeout))
```
chavafg: seems like these three variables are not being used.

bergwolf:
kata-containers/runtime#684 has been merged. This is now ready to be merge as well. @grahamwhaley @chavafg ptal.

bergwolf:
rebased to master in case there are any unexpected conflicts ;)

gnawux:
LGTM as kata-containers/runtime#684 was merged

gnawux:
LGTM as kata-containers/runtime#684 was merged

integration/vm_factory/vm_templating_test.sh
```diff
@@ -0,0 +1,117 @@
...
+	echo "init vm template"
+	sudo -E PATH=$PATH "$kata_runtime_bin" factory init
+}
+
+destroy_vm_template() {
+	echo "destroy vm template"
+	sudo -E PATH=$PATH "$kata_runtime_bin" factory destroy
+	# verify template is destroied
+	res=$(mount | grep template | wc -l)
+	test $res -eq 0
```
grahamwhaley: Using `test` is fine (and in a way, makes the code obvious). I'll note though that I don't think this will print an error message in the event of failure will it? Using `[ $res -ne 0 ] || die "a horrid death"` idiom might make it clearer why we died. Not a blocker though, just a thought.

integration/vm_factory/vm_templating_test.sh
```diff
@@ -0,0 +1,117 @@
...
+	sudo docker rm -f $CONTAINER_NAME
+	destroy_vm_template
+	disable_vm_template_config
+}
+
+teardown() {
+	clean_env
+}
+
+if [ x$INITRD_PATH == "x" ]; then
```
grahamwhaley: Heh, old school idiom I think - we tend to use `-z` now I think ;-)

integration/vm_factory/vm_templating_test.sh
```diff
@@ -0,0 +1,117 @@
...
+
+source "${cidir}/../../metrics/lib/common.bash"
+
+# Environment variables
+IMAGE="${IMAGE:-busybox}"
+CONTAINER_NAME=$(random_name)
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+
+if [ -f ${DEFAULTCONFIG} ]; then
+	runtime_config_path=${DEFAULTCONFIG}
```
grahamwhaley: I wonder if we are better finding these via a `kata-runtime kata-env --json | jq blah-blah` or similar `kata-env` based query? Similarly, we could use `kata-env` to find the actual hypervisor path and replace the `ps | grep qemu` elsewhere in the file with a `pgrep $HYPERVISOR_PATH` type thing.

metrics/lib/common.bash
```diff
@@ -24,6 +24,9 @@ KSM_SLEEP_FILE="${KSM_BASE}/sleep_millisecs"
 KSM_AGGRESIVE_PAGES=1000
 KSM_AGGRESIVE_SLEEP=50
 
+PKGDEFAULTCONFIG="${PKGDEFAULTCONFIG:-/usr/share/defaults/kata-containers/configuration.toml}"
```
grahamwhaley: See comment around the usage point of this - I suspect we might/should be using `kata-env` to get this info.

grahamwhaley:
F27 CI hit a network issue - I've nudged a rebuild.

integration/vm_factory/vm_templating_test.sh
bergwolf: It is not just for querying. I actually need to modify the config file and change it back in the test. See `enable/disable_vm_template_config`. Can I do that with `kata-runtime kata-env`?

grahamwhaley: Hi. You cannot modify using `kata-runtime kata-env`, but you can use `kata-env` to find the actual config file path the runtime is using, and then modify that with `sed` like you are already.

In fact, we already have a piece of code in the metrics common.lib that does the extraction https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L34
```
	RUNTIME_CONFIG_PATH=$(awk '/^\[Runtime\]$/ {foundit=1} /^    Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
```
(this was written before we had the json kata-env, which might be slightly cleaner ;-) )
That code is called from the metrics `common_init()`, which you are probably not calling. You could call directly `extract_kata_env` to get RUNTIME_CONFIG_PATH set in the env.
I think using `kata-env` to find the config file the runtime is actually using is probably the right thing to do - having the paths wired into `common.bash` feels wrong.

bergwolf: Thanks @grahamwhaley. I'll update the PR accordingly.

bergwolf:
updated. @grahamwhaley PTAL.

Makefile
```diff
@@ -101,4 +104,5 @@ check: checkcommits log-parser
 	log-parser \
 	openshift \
 	swarm \
+	vm-factory \
```
grahamwhaley: ah, we tend to keep these in alphabetical order ... if you don't mind shuffling your addition to the end?

metrics/lib/common.bash
```diff
@@ -44,6 +44,8 @@ extract_kata_env(){
 	HYPERVISOR_PATH=$(awk '/^\[Hypervisor\]$/ {foundit=1} /^  Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
 	HYPERVISOR_VERSION=$(awk '/^\[Hypervisor\]$/ {foundit=1} /^  Version =/ { if (foundit==1) {$1=$2=""; print $0; foundit=0} } ' <<< "$toml" | sed 's/"//g')
 
+	INITRD_PATH=$(awk '/^\[Initrd\]$/ {foundit=1} /^  Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+
```
grahamwhaley: :-) I was just thinking yesterday I wonder if I have that, as I should add it into the metrics data if I don't... thx

bergwolf:
travis is stuck. otherwise the PR is ready to be merged. ping @grahamwhaley @chavafg @jodh-intel

grahamwhaley:
Poor old Travis.... merging!

PR #599, opened 2018-08-09
integration/docker: save container ID to check kata components status

### LINKED ISSUE #598: ExistDockerContainer must use container ID not container name
`ExistDockerContainer` checks if the kata components (proxy, agent, qemu) are still running, the functions used to check this, are `HypervisorRunning` , `ProxyRunning` and `ShimRunning` https://github.com/kata-containers/tests/blob/master/integration/docker/docker.go#L210-L219, these functions need the container ID to check the commandline of all the running processes at `/proc` https://github.com/kata-containers/tests/blob/master/process.go#L55-L85 ,
the problem is that `ExistDockerContainer` is using the container name not the container ID, hence `ExistDockerContainer` always returns `false`. 

**Note: container's name can't be used since it's not present in the process command line**

devimc:
Use --cidfile to save container ID and use it later to check if kata
components are still running.

Container's name can't be used since it's not present in the process
command line.

fixes #598

### DIFF
--- a/integration/docker/docker.go
+++ b/integration/docker/docker.go
@@ -8,6 +8,9 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


amshinde:
@devimc Can you address @grahamwhaley comments on the issue for this PR?
Name should be unique as well, so I would like to understand the reasoning for this change.

devimc:
@amshinde  done

amshinde:
Thanks @devimc 
lgtm

chavafg:
lgtm

GabyCT:
lgtm

GabyCT:
lgtm

PR #603, opened 2018-08-13
metrics: json: fix invalid and misleading JSON output

grahamwhaley:
Under some conditions we generated invalid JSON for some launch time data.
The runtime 'path' was also the path to the config file, not the runtime itself.
Correct both of those.


jodh-intel:
```
cat <<< "lgtm"
```

... ermmm...
lgtm

GabyCT:
lgtm

PR #605, opened 2018-08-14
metrics: Add network memory PSS consumption

GabyCT:
Measures network PSS memory consumption using smem tool while
running a network workload with nuttcp.

Fixes #604

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

PR #620, opened 2018-08-20
metrics: Add network metric for RSS memory footprint

### LINKED ISSUE #619: Add network metric for RSS memory footprint
Add a test that measures resident set size (RSS) of the hypervisor footprint on the host using smem while running a network bandwidth test

GabyCT:
This measures resident set size (RSS) of the hypervisor footprint on the
host using smem while we are running a nuttcp network bandwidth test.

Fixes #619

### DIFF
--- a/metrics/network/README.md
+++ b/metrics/network/README.md
@@ -27,6 +27,9 @@ are used to test both servers for the maximum amount of throughput.
 - `network-metrics-memory-pss.sh` measures the Proportional Set Size (PSS) of the hypervisor footprint on the host using smem
 while running a transfer of one GB with nuttcp.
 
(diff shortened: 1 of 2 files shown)


grahamwhaley:
I need to go read this - but is it very similar to #605? (which measures PSS, not RSS). Is it using the same tool - basically, could we just have one test that measures 'memory footprint of network test', and stores away all of VSS, PSS, RSS into the JSON, and maybe also it could try to grab 'free' memory reduction or size as well? (if you grab size, you also really need a pre-test reference so you can calculate the reduction for each iteration/instance of the run etc.).

GabyCT:
@grahamwhaley ok, I 'll do the changes thanks

GabyCT:
@grahamwhaley , I'll close this PR and submit another that will have RSS and VSS support

PR #634, opened 2018-08-22
CI: checkout to correct branch on kata-containers repos

chavafg:
When building and installing kata-containers components,
we need to checkout to the specific branch we are testing
instead of doing it with master.

Fixes: #633.

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -52,14 +52,23 @@ mkdir -p $(dirname "${kata_repo_dir}")
 
 pushd "${kata_repo_dir}"
 
+
+# Variables needed when we test a PR.
 pr_number=
+target_branch=
+
+# Variable needed when a merge to a branch needs to be tested.
+branch=
 
 # $ghprbPullId and $ghprbTargetBranch are variables from
 # the Jenkins GithubPullRequestBuilder Plugin
-[ "${ghprbPullId}" ] && [ "${ghprbTargetBranch}" ] && pr_number="${ghprbPullId}"
+[ "${ghprbPullId}" ] && [ "${ghprbTargetBranch}" ] && export pr_number="${ghprbPullId}"
+
+
 
 if [ -n "$pr_number" ]
 then
+	export target_branch="${ghprbTargetBranch:-master}"
 	if [ "${kata_repo}" != "${tests_repo}" ]
 	then
 		# Use the correct branch for testing.
@@ -67,7 +76,7 @@ then
 		# of the kata repository branch where the change is
 		# going to be merged.
 		pushd "${test_repo_dir}"
-		git fetch origin && git checkout "${ghprbTargetBranch}"
+		git fetch origin && git checkout "${target_branch}"
 		popd
 	fi
 
@@ -78,7 +87,7 @@ then
 	# the target branch.
 	git fetch origin "pull/${pr_number}/head:${pr_branch}"
 	git checkout "${pr_branch}"
-	git rebase "origin/${ghprbTargetBranch}"
+	git rebase "origin/${target_branch}"
 
 	# As we currently have CC and runv runtimes as git submodules
 	# we need to update the submodules in order to get
@@ -88,8 +97,8 @@ then
 else
 	# Othewise we test an specific branch
 	# GIT_BRANCH env variable is set by the jenkins Github Plugin.
-	remote="${GIT_BRANCH/\/*/}"
-	branch="${GIT_BRANCH/*\//}"
+	[ -z "${GIT_BRANCH}" ] && echo >&2 "GIT_BRANCH is empty" && exit 1
+	export branch="${GIT_BRANCH/*\//}"
 
 	if [ "${kata_repo}" != "${tests_repo}" ]
 	then
@@ -97,11 +106,11 @@ else
 		# 'tests' repository branch should have the same name
 		# as the kata repository branch that will be tested.
 		pushd "${test_repo_dir}"
-		git fetch "$remote" && git checkout "$branch"
+		git fetch origin && git checkout "$branch"
 		popd
 	fi
 
-	git fetch "$remote" && git checkout "$branch" && git reset --hard "$GIT_BRANCH"
+	git fetch origin && git checkout "$branch" && git reset --hard "$GIT_BRANCH"
 fi
 
 # Install go after repository is cloned and checkout to PR
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -33,6 +33,10 @@ function clone_and_build() {
 
 	pushd ${project_dir}
 
+	# Override branch if we are testing a PR.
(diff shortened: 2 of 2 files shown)


.ci/jenkins_job_build.sh
```diff
@@ -52,22 +52,31 @@ mkdir -p $(dirname "${kata_repo_dir}")
...
 # $ghprbPullId and $ghprbTargetBranch are variables from
 # the Jenkins GithubPullRequestBuilder Plugin
-[ "${ghprbPullId}" ] && [ "${ghprbTargetBranch}" ] && pr_number="${ghprbPullId}"
+[ "${ghprbPullId}" ] && [ "${ghprbTargetBranch}" ] && export pr_number="${ghprbPullId}"
+
+
 
 if [ -n "$pr_number" ]
 then
+	export target_branch="${ghprbTargetBranch}"
```
jcvenegas: export target_branch="${ghprbTargetBranch:-master}" just in case is empty ? or fail if ghprbTargetBranch is not set.

.ci/jenkins_job_build.sh
```diff
@@ -88,20 +97,19 @@ then
 else
 	# Othewise we test an specific branch
 	# GIT_BRANCH env variable is set by the jenkins Github Plugin.
-	remote="${GIT_BRANCH/\/*/}"
-	branch="${GIT_BRANCH/*\//}"
+	export branch="${GIT_BRANCH/*\//}"
```
jcvenegas: do we need to check if branch is not empty ? 
Why we are exporting this variable will be used by other script?

.ci/jenkins_job_build.sh
```diff
@@ -88,20 +97,19 @@ then
...
 
 	if [ "${kata_repo}" != "${tests_repo}" ]
 	then
 		# Use the correct branch for testing.
 		# 'tests' repository branch should have the same name
 		# as the kata repository branch that will be tested.
 		pushd "${test_repo_dir}"
-		git fetch "$remote" && git checkout "$branch"
+		git fetch origin && git checkout "$branch"
```
jcvenegas: I am ok with the this change but could you tell me why remove the remote usage?

.ci/lib.sh
```diff
@@ -33,6 +33,13 @@ function clone_and_build() {
 
 	pushd ${project_dir}
 
+	if [ -n "$pr_number" ]
+	then
+		git fetch origin && git checkout "${target_branch}"
```
jcvenegas: Or:
```bash
# override  branch is a PR testing.
[ -z "$pr_number" ] ||  branch="${target_branch}"
git fetch origin && git checkout "${branch}"
```

jcvenegas:
```
INFO: Download go version 1.10

  2  114M    2 3342k    0     0  8038k      0  0:00:14 --:--:--  0:00:14 8035k
 58  114M   58 66.7M    0     0  57.4M      0  0:00:01  0:00:01 --:--:-- 57.4M
curl: (56) GnuTLS recv error (-54): Error in the pull function.
```
restarting ubuntu job ...

.ci/jenkins_job_build.sh
chavafg: yes, this will be used by the `.ci/lib.sh` script

.ci/jenkins_job_build.sh
chavafg: agree to make master default

.ci/jenkins_job_build.sh
chavafg: As we do not have other remotes other than `origin`, I prefer to change it, wdyt?

chavafg:
@jcvenegas changes applied

chavafg:
@cboylan We are adding 3 new environment variables for testing stable branches, can you help us adding them for the zuul jobs?

we will be using for PRs:
`pr_number`, which will be the number of the PR
`target_branch`, which will be the branch where the PR is going to be merged.

For testing branches:
`branch`, which will be the branch that we will be testing. E.g. `master`, `stable-1.1`

I think here would be the correct place:
https://git.openstack.org/cgit/openstack-infra/openstack-zuul-jobs/tree/playbooks/kata-runsh/run.yaml#n14

Thanks

.ci/jenkins_job_build.sh
jcvenegas: I am fine with it :)

jcvenegas:
Thanks @chavafg  still LGTM :)

fungi:
I'm probably missing some important nuance. It's unclear to me why you have separate target_branch and branch environment variables specified. Based on their descriptions it sounds like you anticipate testing against a different branch than you'll merge the PR under at least some circumstances? If that's the case, how are you signaling which branch you want to test when it's not the branch to which you're going to merge?

More importantly, as corvus mentioned in #openstack-infra, the fetch logic seems completely unnecessary under Zuul so you might just consider wrapping that in a conditional and only worrying about it in Jenkins. Zuul will provide you with a local repository on the job node checked out to the current tip of the target branch for the PR with the PR commits and any of its dependency PRs merged to the branch already.

chavafg:
Hi @fungi, 

`target_branch` is basically used when a PR is going to be tested. You can see that we are assigning the value of `$ghprbTargetBranch` which is taken from the `GitHub pull request builder plugin`.  We cannot use this variable when we are testing a change to a branch, since this variable wont exist (as this is not a PR). 

`branch` variable takes the value of `$GIT_BRANCH`, from the `Github Plugin` (which is a different one from the above). This will let us know which branch will be tested. 

I think we still need to fetch and change to the correct branch in Zuul, because we need to build all repositories using the same branch. Let me illustrate with an example:

Lets say a PR is opened on `runtime` repo and it will be merged on `stable-1.1`. I know that zuul will manage to have `runtime` repository in the correct branch: `stable-1.1`. But in the process of setting up the testing environment, we also need to build and install other kata-components (agent, proxy, shim...). This is done by a function in `.ci/lib.sh`, which is called `clone_and_build` [1]. And what this is gonna do is clone all repos and use master branch, which would be wrong, as we need to test all components using the `stable-1.1` branch.

[1] https://github.com/kata-containers/tests/blob/master/.ci/lib.sh#L23

jodh-intel:
lgtm

fungi:
I don't understand why your multi-project example means you can't use the Git states Zuul prepares for you. When you define or vary a Zuul job you tell it which projects the job uses. Zuul will then prepare the repositories for all involved projects and check them out to identically-named branches for the PR (if they exist) or the master branch (if no matching branch name). There are also ways to instruct Zuul to use particular branches other than the target branch in cases where this default matching behavior is not desired.

On the case of the branch vs target_branch variable, what I'm trying to find out is how you're triggering the jobs you describe which are "changes to a branch" not initiated by a pull request. Without some insight into how you're triggering Zuul to run jobs against something other than a PR, I have no idea how to divine the correct value for the branch variable.

PR #645, opened 2018-08-23
metrics: Remove sleep for the footprint script

### LINKED ISSUE #644: Remove sleep from the footprint script
Now that this issue has been solved kata-containers/agent#263, it is necessary to remove the workaround of sleeping.

GabyCT:
Now that this issue has been solved kata-containers/agent#263,
it is necessary to remove the workaround of sleeping.

Fixes #644

### DIFF
--- a/metrics/density/footprint_data.sh
+++ b/metrics/density/footprint_data.sh
@@ -149,16 +149,8 @@ function cleanup() {
 	# This is what we have for now from the generics.
 	# Expect this to be improved (the generics) soon...
 
-	# FIXME - temporary workaround for https://github.com/kata-containers/runtime/issues/406
-	# When that is fixed, we can go back to a hard 'docker rm -f *' with no sleep.
-	# Or, in fact we can leave the '--rm' on the docker run, and just change this
-	# to a 'docker stop *'.
-	for c in $(docker ps -qa); do
-		# Use a kill, as the containers are benign, and most of the time the stop
-		# request ends up doing a kill anyhow
-		docker kill $c
-		sleep 3
-	done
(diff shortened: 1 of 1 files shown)


jodh-intel:
\o/

lgtm

metrics/density/footprint_data.sh
```diff
@@ -149,15 +149,8 @@ function cleanup() {
...
-	# FIXME - temporary workaround for https://github.com/kata-containers/runtime/issues/406
-	# When that is fixed, we can go back to a hard 'docker rm -f *' with no sleep.
-	# Or, in fact we can leave the '--rm' on the docker run, and just change this
-	# to a 'docker stop *'.
 	for c in $(docker ps -qa); do
-		# Use a kill, as the containers are benign, and most of the time the stop
-		# request ends up doing a kill anyhow
-		docker kill $c
-		sleep 3
+		docker stop $c
```
grahamwhaley: Can we do a 'kill' or 'rm -f' instead please - issuing a 'stop' has a tendency to get the STOP ignored at the container end, and then that hits us with a default 10s timeout penalty, before docker issues a kill anyway.
See https://github.com/moby/moby/issues/3766 and a number of other related Issues for details if interested.

So, this could probably just be a:

```bash
$ docker kill $(docker ps -qa)
```

which is a little harsh, as it will kill all containers even if we did not launch them - but, I think the pre-init scripts have probably ensured that is the case already when running metrics tests.

thx!

GabyCT:
@grahamwhaley changes applied :)

chavafg:
lgtm

PR #654, opened 2018-08-27
metrics: Add more concurrency levels to network nginx test

### LINKED ISSUE #653: Add more concurrency levels to nginx network test
Currently the nginx network test handles only one concurrency level, we need to add more concurrency levels (500 and 1000) to exercise the test

GabyCT:
Currently this test only handles 100 as concurrency level, with these
modifications we will be able to run other concurrency levels.

Fixes #653

### DIFF
--- a/metrics/network/network-metrics-nginx-ab-benchmark.sh
+++ b/metrics/network/network-metrics-nginx-ab-benchmark.sh
@@ -14,8 +14,6 @@ set -e
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/../lib/common.bash"
 
-# Test name
-TEST_NAME="${TEST_NAME:-network nginx ab benchmark}"
 # Ports where it will run
 port="${port:-80:80}"
 # Image name
@@ -29,14 +27,22 @@ requests="${requests:-10000}"
 start_time="${start_time:-2}"
 # File to save ab results
 TMP_FILE=$(mktemp results.XXXXXXXXXX)
-# Concurrency
-concurrency="${concurrency:-100}"
+# Maximum number of seconds to wait before
+# the  socket  times  out
+socket_time="${socket_time:-120}"
 
 function remove_tmp_file {
 	rm -rf $TMP_FILE
 }
 
 trap remove_tmp_file EXIT
+
+function concurrency {
+	local concurrency_value=$1
+	local TEST_NAME="${TEST_NAME:-network nginx ab benchmark with ${concurrency_value} concurrency}"
+	local result="$(nginx_ab_networking)"
+	save_results "$TEST_NAME"
+}
 
 function save_config {
 	metrics_json_start_array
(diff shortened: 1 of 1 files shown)


devimc:
restarting fedora CI

chavafg:
Don't know about the requirements for this script, but shouldn't it be more efficient to pass the concurrency number as an argument? 
Something like: 
```
./network-metrics-nginx-ab-benchmark.sh -n 500

or

./network-metrics-nginx-ab-benchmark.sh -n 1000
```

This way the user can change the number without changing the script.

GabyCT:
@chavafg the main issue with that is that we can hit a limitation with 9p which will not produce a result, we need to select values that we know we currently handle

metrics/network/network-metrics-nginx-ab-benchmark.sh
```diff
@@ -29,15 +27,37 @@ requests="${requests:-10000}"
...
 }
 
 trap remove_tmp_file EXIT
 
+function concurrency_100 {
+	local concurrency=100
+	local TEST_NAME="${TEST_NAME:-network nginx ab benchmark with ${concurrency} concurrency}"
+	local result="$(nginx_ab_networking)"
+	save_results "$TEST_NAME"
+}
```
chavafg: you could reduce to only 1 function here: 
```
function concurrency_n {
	local concurrency=$1
	local TEST_NAME="${TEST_NAME:-network nginx ab benchmark with ${concurrency} concurrency}"
	local result="$(nginx_ab_networking)"
	save_results "$TEST_NAME"
}
```
and then call it:
```
concurrency_n 100
concurrency_n 500
concurrency_n 1000
```

GabyCT:
@chavafg changes applied, thanks for the feedback

grahamwhaley:
Hi @GabyCT Yeah, I think we might have to rework this so the user can also set a concurrency (or multiple concurrencies) on the command line. So, maybe if we could support something like:
```
network-metrics-nginx-ab-benchmark.sh -n 100 -n 500 -n 1000
```

and then collect those numbers up in a bash string or array, and then execute them internally with a:
```
for concurrency in $concurrency_list; do
 run_concurrency $concurrency
done
```

and have an all/default setting of `concurrency_list="100 500 1000"`
then maybe that will satisfy everything:
- we get a default of the set of concurrencies we know we want to test
- the user can over-ride it
- and they can even orver-ride it with multiple entries on the command line.

I'm just hoping the shell arg parsing can handle multiple arguments of the same `-n` flag type - I expect so though...

GabyCT:
@grahamwhaley changes applied

GabyCT:
@grahamwhaley changes applied

grahamwhaley:
I'm OK, with this, but it does feel like we have been around in a large circle and ended up quite near where we started :-)
We've added the `socket_time` environment override, but have moved the `concurrency` environment override to be a command line parameter, and dropped the notion of the default 100/500/1000 parallel defaults.

So, overall, I suspect we've ended up with quite little change?
I can live with it, so will ack.

chavafg:
lgtm

Issue #639, opened 2018-08-23
Depends-on tag doesn't work 


lifupan:
I add Depends-on tag in this commit https://github.com/kata-containers/tests/pull/617/commits/2c6ba70da3dcf29ba0865948c65d9d56242e1006
and send a PR https://github.com/kata-containers/tests/pull/617, all of the CI failed with the following errors:

/tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/src/github.com/kata-containers/tests      /tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/src/github.com/kata-containers/tests
Repository  was already defined in a ‘Depends-on:’ tag.
Only one repository per tag is allowed.

Fupan

chavafg:
reopening as it still has an issue.

PR #657, opened 2018-08-27
CI: Fix Depends-on feature

chavafg:
Depends-on tag was not working when there was a missing
space between the colon and repository name of the tag.

Fixes: #639.


GabyCT:
lgtm

PR #658, opened 2018-08-27
metrics: Fix launch_times script to emit valid json numbers

### LINKED ISSUE #562: metrics: launch_times: `bc` can emit non-JSON valid numbers
We use `bc` in some math in [some of our time calculations](https://github.com/kata-containers/tests/blob/master/metrics/time/launch_times.sh#L107), but if the result is <1 then `bc` does not print a leading-0 (for instance, it would print `.87654` rather than `0.87654`). That sadly does not count as a valid JSON number, and our JSON readers grok when reading the results files.
Sadly there seems to be no built in way to get `bc` to add a leading 0. Work around it using a `printf` (which it appears we have already done elsewhere in the same file).

GabyCT:
Fix the to kernel period time to print valid json numbers.

Fixes #562

### DIFF
--- a/metrics/time/launch_times.sh
+++ b/metrics/time/launch_times.sh
@@ -104,7 +104,7 @@ run_workload() {
 		kernel_period=$(echo $kernel_last_line | awk '{print $2}' | tr -d "]")
 
 		# And we can then work out how much time it took to get to the kernel
-		to_kernel_period=$(bc <<<"scale=3; $(ns_to_s $workload_period) - $kernel_period")
+		to_kernel_period=$(printf "%0f" $(bc <<<"scale=3; $(ns_to_s $workload_period) - $kernel_period"))
 	else
 		kernel_period="0.0"
 		to_kernel_period="0.0"


metrics/time/launch_times.sh
```diff
@@ -104,7 +104,7 @@ run_workload() {
 		kernel_period=$(echo $kernel_last_line | awk '{print $2}' | tr -d "]")
 
 		# And we can then work out how much time it took to get to the kernel
-		to_kernel_period=$(bc <<<"scale=3; $(ns_to_s $workload_period) - $kernel_period")
+		to_kernel_period=$(printf "%.03f" $(bc <<<"scale=3; $(ns_to_s $workload_period) - $kernel_period"))
```
devimc: here you are cutting off the precision to 3 decimals.
correct me if I'm wrong, but I think the problem @grahamwhaley mentions in #562 is not the precision of the floating point, the problem is the leading 0, so I think you should use `%0f`

GabyCT: @devimc , yes I cutting off the precision but is also solving the issue with the leading 0.

GabyCT:
@devimc , I changed it to print all the number instead of cutting the precision, thanks

chavafg:
lgtm

grahamwhaley:
thx. I meant to get to peek at this. I'm gonna trust that `.03f` does what we need - I will note, as noted in the Issue, further down the file there is an exsting `printf "%0f" `, just for reference.
I think I even had a patch somewhere on my machine to fix this, but it has been lost in the mists of time. Thanks for picking it up @GabyCT

PR #659, opened 2018-08-27
CI: KATA_RUNIME default must be kata-runtime

### LINKED ISSUE #600: ci: lib.sh: should `KATA_RUNTIME` default to `kata-runtime`?
I happened to be browsing the code, when I saw this in `.ci/lib.sh`:

https://github.com/kata-containers/tests/blob/master/.ci/lib.sh#L10
```
export KATA_RUNTIME=${KATA_RUNTIME:-cc}
```

I suspect that is wrong, and a hangover from early kata integration effort, for a couple of reasons:
- `cc` implies maybe **C**lear **C**ontainers
- but `cc` is actually the C compiler??

@chavafg @jodh-intel - wdyt - that should probably default to `kata-runtime`, yes?

chavafg:
Instead of making cc the default value of this environment
variable, the default must be kata-runtime.

Fixes #600.

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -7,7 +7,7 @@
 
 set -e
 
-export KATA_RUNTIME=${KATA_RUNTIME:-cc}
+export KATA_RUNTIME=${KATA_RUNTIME:-kata-runtime}
 
 # If we fail for any reason a message will be displayed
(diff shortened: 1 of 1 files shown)


GabyCT:
lgtm

PR #662, opened 2018-08-27
ci: Fix checkout to stable branch.

jcvenegas:
Fix variable name that prevents checkout to stable tests branch.


jodh-intel:
lgtm

Might be worth adding a `set -o nounset` to all our scripts to avoid this type of bug.

PR #667, opened 2018-08-28
ci: Check GOPATH exists

### LINKED ISSUE #535: ci: setup.sh: should quit nicely if GOPATH is not set
The [setup.sh](https://github.com/kata-containers/tests/blob/master/.ci/setup.sh#L14) invokes [lib.sh](https://github.com/kata-containers/tests/blob/eccc14891c1c2d8c3ad823db647be5590eb61753/.ci/lib.sh#L108) functions that expect to have GOPATH set - that currently means they die with an obscure 'dir not found' error. We should check GOPATH is set as a prereq, and quit nicely if it is not.

GabyCT:
With this we will fix that when running the setup.sh script (which is
calling functions from lib.sh).

Fixes #535

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -107,6 +107,12 @@ function get_version(){
 	echo "$result"
 }
 
+function check_gopath() {
+	# Verify GOPATH is set
+	if [ -z "$GOPATH" ]; then
+		export GOPATH=$(go env GOPATH)
+	fi
+}
 
 function apply_depends_on() {
 	pushd "${GOPATH}/src/${kata_repo}"
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -9,6 +9,11 @@
 
 set -e
 
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+check_gopath
+
 export RUNTIME="kata-runtime"
 
 echo "INFO: Running checks"
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -11,6 +11,7 @@ cidir=$(dirname "$0")
 source /etc/os-release
 source "${cidir}/lib.sh"
 
+check_gopath
 apply_depends_on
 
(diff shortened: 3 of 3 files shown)


grahamwhaley:
(marked DNM whilst we discuss)
I think both @caoruidong and @GabyCT are both right here.....
- our scripts use `GOPATH` heavily (I find 82 instances in a quick grep of the tests repo for instance), so we really expect to find it in the ENV.
- a few versions back iirc, golang started to set a default GOPATH if one was not set in the environment (https://github.com/golang/go/wiki/SettingGOPATH#setting-gopath), so theoretically the end user does not have to have set one.

I wonder if we can do something early in the lib.sh such as:
```bash
if [ -z "$GOPATH" ]; then
    export GOPATH=$(go env GOPATH)
fi
```
that would then satisfy all?

GabyCT:
@grahamwhaley, yes I saw that option, however, one of the things that I noticed is that for example if I use `gabyubuntu` as my user and then I do the `go env GOPATH` , it retries `GOPATH="/home/gabyubuntu/go"` which is an existing path, however, if I do the same using `root`, it retries `GOPATH="/root/go"` which is a non existing path and if you want to run the `setup.sh` it will fail.

chavafg:
also, should we source `.ci/lib.sh` into the `.ci/run.sh`, so that the `GOPATH` can also be checked/set there?

GabyCT:
@chavafg yes, that can be done :)

GabyCT:
@chavafg changes applied

chavafg:
lgtm

This approach works for me, @grahamwhaley what do you think?

GabyCT:
ping @grahamwhaley

grahamwhaley:
l g t m
but I think you need to update the commit message, as it still says 'fail gracefully', whereas now we fix it and carry on :-)

GabyCT:
@grahamwhaley commit message updated

Issue #671, opened 2018-08-29
ci: need a few clean-up work for CI environment on bare mental machine


Pennyzct:
We should provide clean environment for each build. So when CI is running on bare mental, a few extra clean-up work such as killing stale process, removing history data, etc, are needed.
We have discussed above issue on [kata-containers/proxy #97](https://github.com/kata-containers/proxy/pull/97). We got a few CI shutting-down which come from dirty CI environment, such as stale docker-containerd-shim process, etc.
Relevant pr will upload ASAP.
@grahamwhaley @kalyxin02

PR #672, opened 2018-08-29
ci: a few clean-up work for CI environment on bare-metal machine

Pennyzct:
We should provide clean environment for each build, when CI running on bare-metal, so a few clean-up work such as killing stale process, removing history data, etc, are needed.

Fixes: #671

Signed-off-by: Penny Zheng penny.zheng@arm.com

### DIFF
--- a/.ci/aarch64/clean_up_aarch64.sh
+++ b/.ci/aarch64/clean_up_aarch64.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 ARM Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+stale_process_union=( "docker-containerd-shim" )
+#docker supports different storage driver, such like overlay2, aufs, etc.
+docker_storage_driver=$(docker info --format='{{.Driver}}')
+stale_docker_mount_point_union=( "/var/lib/docker/containers" "/var/lib/docker/${docker_storage_driver}" )
+stale_docker_dir_union=( "/var/lib/docker" )
+stale_kata_dir_union=( "/var/lib/vc" "/run/vc" )
+
+lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
+source "${lib_script}"
+
+metrics_lib_script="${GOPATH}/src/${tests_repo}/metrics/lib/common.bash"
+source "${metrics_lib_script}"
+
+kill_stale_process()
+{
+	# use function kill_processes_before_start() under $metrics_lib_script to kill stale containers or shim/proxy/hypervisor process
+	kill_processes_before_start
+	for stale_process in "${stale_process_union[@]}"; do
+		result=$(check_processes "${stale_process}")
+		if [[ $result -ne 0 ]]; then
+			sudo killall -9 "${stale_process}" || true
+		fi
+	done <<< "${stale_process_union}"
+}
+
+delete_stale_docker_resource()
+{
+	local docker_status=false
+	# check if docker service is running
+	systemctl is-active --quiet docker
+	if [ $? -eq 0 ]; then
+		docker_status=true
+		sudo systemctl stop docker
+	fi
+	# before removing stale docker dir, you should umount related resource
+	for stale_docker_mount_point in "${stale_docker_mount_point_union[@]}"; do
+		local mount_point_union=$(mount | grep "${stale_docker_mount_point}" | awk '{print $3}')
+		if [ -n "${mount_point_union}" ]; then
+			while IFS='$\n' read mount_point; do
+				sudo umount "${mount_point}"
+			done <<< "${mount_point_union}"
+		fi
+	done
+	# remove stale docker dir
+	for stale_docker_dir in "${stale_docker_dir_union[@]}"; do
+		if [ -d "${stale_docker_dir}" ]; then
+			sudo rm -rf "${stale_docker_dir}"
+		fi
+	done <<< "${stale_docker_dir_union}"
+	[ "${docker_status}" = true ] && sudo systemctl restart docker
+}
+
+delete_stale_kata_resource()
+{
+	for stale_kata_dir in "${stale_kata_dir_union[@]}"; do
+		if [ -d "${stale_kata_dir}" ]; then
+			sudo rm -rf "${stale_kata_dir}"
+		fi
+	done <<< "${stale_kata_dir_union}"
+}
+
+main() {
+	info "kill stale process: ${stale_process_union[@]}"
+	kill_stale_process
+	info "delete stale docker resource under ${stale_docker_dir_union[@]}"
+	delete_stale_docker_resource
+	info "delete stale kata resource under ${stale_kata_dir_union[@]}"
+	delete_stale_kata_resource
+}
+
+main
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -50,8 +50,13 @@ mkdir -p $(dirname "${tests_repo_dir}")
 mkdir -p $(dirname "${kata_repo_dir}")
 [ -d "${kata_repo_dir}" ] || git clone "https://${kata_repo}.git" "${kata_repo_dir}"
 
-pushd "${kata_repo_dir}"
+# If CI running on bare-metal, a few clean-up work before walking into test repo
+if [ "${BAREMETAL}" == true ]; then
+	clean_up_script="${tests_repo_dir}/.ci/${arch}/clean_up_${arch}.sh"
+	[ -f "${clean_up_script}" ] && source "${clean_up_script}"
+fi
 
+pushd "${kata_repo_dir}"
 
 # Variables needed when we test a PR.
 pr_number=


Pennyzct:
@grahamwhaley ptal.;)

.ci/jenkins_job_build.sh
```diff
@@ -12,6 +12,10 @@ source "/etc/os-release"
 # Signify to all scripts that they are running in a CI environment
 [ -z "${KATA_DEV_MODE}" ] && export CI=true
 
+# CI environment on aarch64 only supports bare mental for now
+arch=$(uname -m)
+[ "${arch}" == "aarch64" ] && export BARE=true
```
grahamwhaley: It might be better to `export BARE=true` in the Jenkins job setup scripts on the jenkins master, and/or even use a Jenkins slave label and see if we can get that exported into the environment.
That way this script stays more generic, and later if we do get nested VM ARM Jenkins slaves we don't have to come fix this script again - we  can just set the correct settings inside Jenkins itself.
That is what I do for the [`METRICS_CI`](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L136) and [`JSON_URL`](https://github.com/kata-containers/tests/blob/master/metrics/lib/json.bash#L115) vars for instance.

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 ARM Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+stale_process_union=( "docker-containerd-shim" "qemu-system-aarch64" )
+stale_docker_mount_point_union=( "/var/lib/docker/containers" "/var/lib/docker/overlay2" )
```
jodh-intel: It would be safer to check the output of the following before adding the overlay entry:

```
$ storage_driver=$(docker info|grep "^Storage Driver"|cut -d: -f2-|tr -d ' ')
$ echo "'$storage_driver'"
'overlay2'
```

.ci/jenkins_job_build.sh
```diff
@@ -31,6 +35,9 @@ then
 	exit 1
 fi
 
+# If CI running on bare mental, wipe out stale Jenkins WORKSPACE
+[ "${BARE}" == true ] && [ -d "${WORKSPACE}" ] && sudo rm -rf "${WORKSPACE}"
```
grahamwhaley: You can set up the jenkins jobs to do this. You might have to fiddle the perms to work around the tests generating root owned files (like the `sudo` on this `rm` here is doing I think). Previously I have had for instance on my Jenkins jobs:

```bash
#Hack - our scripts leave some things not owned by us....
# so the jenkins user can later nuke the workspace tree
echo "hacking our file ownerships"
sudo chown -R ${USER} ${WORKSPACE}
sudo chgrp -R $(id -ng) ${WORKSPACE}
```

and then tick the (either/both) delete workspace before/after build in the Jenkins job setup dialog.

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,62 @@
...
+stale_docker_dir="/var/lib/docker"
+stale_kata_dir="/var/lib/vc/sbs"
+
+lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
+source "${lib_script}"
+
+kill_stale_process()
```
jodh-intel: We should probably leverage the existing functions in:

- https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash

They'll need to be moved to a `./lib/common.bash` though probably.

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,62 @@
...
+{
+	for stale_process in "${stale_process_union[@]}"; do
+		local pid_union=$(ps -eo pid,command | grep "${stale_process}" |  grep -v "grep" | awk '{print $1}')
```
grahamwhaley: you might be able to use `pkill` here and do away with much of this logic. Also see `pgrep`

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,62 @@
...
+		local mount_point_union=$(mount | grep "${stale_docker_mount_point}" | awk '{print $3}')
+		if [ -n "${mount_point_union}" ]; then
+			while IFS='$\n' read mount_point; do
+				sudo umount "${mount_point}"
+			done <<< "${mount_point_union}"
+		fi
+	done
+	if [ -d "${stale_docker_dir}" ]; then
+                sudo rm -rf "${stale_docker_dir}"
+		systemctl is-active --quiet docker && systemctl restart docker
```
grahamwhaley: I'd probably ensure `docker` is shut down before you start the unmounts and dir removes, and then `start` it again at the end.

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 ARM Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+stale_process_union=( "docker-containerd-shim" "qemu-system-aarch64" )
```
grahamwhaley: You might want to consider adding the `kata-shim` and `kata-proxy` here as well.
I guess you've added what you have seen 'left around' so far - but we can probably pre-empt some other components that should not be present but might :-)

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,62 @@
...
+stale_docker_mount_point_union=( "/var/lib/docker/containers" "/var/lib/docker/overlay2" )
+stale_docker_dir="/var/lib/docker"
+stale_kata_dir="/var/lib/vc/sbs"
```
grahamwhaley: you may as well nuke the whole of `/var/lib/vc` I think. And, there is also a `/run/vc` dir as well I believe.

.ci/jenkins_job_build.sh
```diff
@@ -50,8 +57,13 @@ mkdir -p $(dirname "${tests_repo_dir}")
 mkdir -p $(dirname "${kata_repo_dir}")
 [ -d "${kata_repo_dir}" ] || git clone "https://${kata_repo}.git" "${kata_repo_dir}"
 
-pushd "${kata_repo_dir}"
+# If CI running on bare mental, a few clean-up work before walking into test repo
+if [ "${BARE}" == true ]; then
+	clean_up_script="${tests_repo_dir}/.ci/${arch}/clean_up_${arch}.sh"
+	source "${clean_up_script}"
```
grahamwhaley: Maybe add a check that the file exists before you try to source it, just in case we add an arch that is BARE but does not have a script in place. Just a safety measure.

.ci/aarch64/clean_up_aarch64.sh
Pennyzct: You're right. Just randomly checking on another machine, the `Storage Driver` for docker would be the old `aufs`.

.ci/jenkins_job_build.sh
Pennyzct: right~~Let's set it in the Jenkins job setup scripts on the jenkins master for  ARM CI. Since I don't have the perm, maybe you'd like to fix it? 😊

chavafg: sure, I can give jenkins access when needed.

.ci/aarch64/clean_up_aarch64.sh
Pennyzct: yes, I will use the existing functions` kill_processes_before_start` to substitute much of the logic here. It dealt with most of the stale processes. The only left part will be the 'docker-containerd-shim' process which i will  use `pgrep` or `pkill` to eliminate them as @grahamwhaley suggested.
If you are considering to move referred files to `./lib/common.bash` lately, maybe I should wait for them landed firstly? or just ping me at that time, I will refine here accordingly. 😊

.ci/aarch64/clean_up_aarch64.sh
Pennyzct: yep, you're right. I will assure that docker is totally shut down at the first step. maybe `systemctl is-active --quiet docker && systemctl stop docker`

Pennyzct: Hi~ @grahamwhaley when I firstly want to delete /var/lib/docker, I encountered the` resource busy` err, then it occurred to me that I should kill and unmount relevant resource. I occasionally had the same issue with docker-modification in `manage_ctr_mgr.sh`. which also need to delete relevant dirs.
```
	dir="/var/lib/docker"
	echo "Removing $dir"
```
do you think it also need to check and remove the stale resource like here? 😊

.ci/jenkins_job_build.sh
Pennyzct: That's fine with me ;).

jodh-intel:
Hi @Pennyzct - When you do your next update, please could you correct the (very funny!) typo: "bare mental" -> "bar metal"? :smile:

Pennyzct:
ahhh.....AWKWARD😭

Pennyzct:
new update. ptal. 😊@jodh-intel @grahamwhaley

Pennyzct:
Hi~ @grahamwhaley We have re-pushed the test pr and `BAREMETAL` has already been added into the `env`.
```
+ echo 'Our ENV looks like:'
Our ENV looks like:
+ env
...
BAREMETAL=true
...
```
thanks!

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,79 @@
...
+}
+
+delete_stale_docker_resource()
+{
+	local docker_status=false
+	# check if docker service is running
+	systemctl is-active --quiet docker
+	if [ $? -eq 0 ]; then
+		docker_status=true
+		systemctl stop docker
```
grahamwhaley: You may need to prefix a `sudo` here. This script may work as-is on your machine already, but some slaves may be running this as the `jenkins` user that has no direct root privs but does have passwordless `sudo`. I see you already do that for `killall`, so I guess adding here might be the right thing to do?

jodh-intel: +1. See https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#code-blocks.

jodh-intel: (although that is strictly a doc requirement we use the same assumptions "real code" too :)

Pennyzct:
Hi~ @grahamwhaley @jodh-intel  my neglect of root permission in `systemctl stop` and `systemctl restart`, already updated. ptal. 😊.

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -0,0 +1,79 @@
...
+		if [ -n "${mount_point_union}" ]; then
+			while IFS='$\n' read mount_point; do
+				sudo umount "${mount_point}"
+			done <<< "${mount_point_union}"
+		fi
+	done
+	# remove stale docker dir
+	for stale_docker_dir in "${stale_docker_dir_union[@]}"; do
+		if [ -d "${stale_docker_dir}" ]; then
+			sudo rm -rf "${stale_docker_dir}"
```
jodh-intel: You could simplify this to:

```
[ -d "${stale_docker_dir}" ] && sudo rm -rf "${stale_docker_dir}"
```

Same comment below.

jodh-intel:
Thanks @Pennyzct.

lgtm

grahamwhaley:
I'm happy with the expanded `if` statements (that is what my head types by default ;-) ), so merging :-)

PR #682, opened 2018-08-30
CI: dont checkout kata_repo to the target branch

chavafg:
We dont want to test the target_branch of the kata_repo
that we are testing as we want to test the changes of the
PR, so don't checkout this repo to the target branch.

Fixes: #681.


jodh-intel:
lgtm

jcvenegas:
lgtm

.ci/lib.sh
```diff
@@ -35,7 +35,9 @@ function clone_and_build() {
 
 	# Override branch if we are testing a PR.
```
devimc: please update this comment

.ci/lib.sh
```diff
@@ -35,7 +35,9 @@ function clone_and_build() {
...
 	[ -z "$pr_number" ] || branch="${target_branch}"
```
devimc: I wonder if this condition should be moved too

chavafg: no, this is ok as we still want to checkout the other repos to the correct branch

.ci/lib.sh
chavafg: not sure what you mean, but comment is still valid

GabyCT:
lgtm

.ci/lib.sh
devimc: if it's valid, it's valid :smile:

PR #687, opened 2018-08-30
metrics: Support individual results for memory

### LINKED ISSUE #686: Save individual results for memory
Store all the memory results for each individual container for qemu, shim, proxy and runc process for docker_memory_usage.sh script.

GabyCT:
This will be able to store all the memory results for each individual
container for qemu, shim and proxy and for runc.

Fixes #686

### DIFF
--- a/metrics/density/docker_memory_usage.sh
+++ b/metrics/density/docker_memory_usage.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Copyright (c) 2017-2018 Intel Corporation
-# 
+#
 # SPDX-License-Identifier: Apache-2.0
 #
 #  Description of the test:
@@ -28,6 +28,14 @@ AUTO_MODE="$3"
 TEST_NAME="memory footprint"
 SMEM_BIN="smem"
 KSM_ENABLE_FILE="/sys/kernel/mm/ksm/run"
+MEM_TMP_FILE=$(mktemp meminfo.XXXXXXXXXX)
+PS_TMP_FILE=$(mktemp psinfo.XXXXXXXXXX)
+
+function remove_tmp_file() {
+	rm -rf $MEM_TMP_FILE $PS_TMP_FILE
+}
+
+trap remove_tmp_file EXIT
 
 # Show help about this script
 help(){
@@ -66,13 +74,15 @@ get_runc_pss_memory(){
         mem_amount=0
         count=0
 
-	#FIXME - it would be nice to spit out this data into
-	# the JSON file as well.
         shim_instances=$(pgrep  -f "^$docker_shim")
         for shim in $shim_instances; do
                 child_pid="$(pgrep -P $shim)"
                 child_mem=$(sudo "$SMEM_BIN" -c "pid pss" | \
                                 grep "$child_pid" | awk '{print $2}')
+
+		# Getting all individual results
+		echo $child_mem >> $MEM_TMP_FILE
+
                 if (( $child_mem > 0 ));then
                         mem_amount=$(( $child_mem + $mem_amount ))
                         (( count++ ))
@@ -87,6 +97,31 @@ get_runc_pss_memory(){
         echo "$avg"
 }
 
+get_runc_individual_memory() {
+	runc_process_result=$(cat $MEM_TMP_FILE | tr "\n" " " | sed -e 's/\s$//g' | sed 's/ /, /g')
+
+	# Verify runc process result
+	if [ -z "$runc_process_result" ];then
+		die "Runc process not found"
+	fi
+
+	read -r -a runc_values <<< "${runc_process_result}"
+
+	metrics_json_start_array
+
+	local json="$(cat << EOF
+	{
+		"runc individual results": [
+			$(for ((i=0;i<${NUM_CONTAINERS[@]};++i)); do
+				printf '%s\n\t\t\t' "${runc_values[i]}"
+			done)
+		]
+	}
+EOF
+)"
+	metrics_json_add_array_element "$json"
+	metrics_json_end_array "Raw results"
+}
 
 # This function measures the PSS average
 # memory of a process.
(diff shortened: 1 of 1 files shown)


metrics/density/docker_memory_usage.sh
```diff
@@ -28,6 +28,8 @@ AUTO_MODE="$3"
 TEST_NAME="memory footprint"
 SMEM_BIN="smem"
 KSM_ENABLE_FILE="/sys/kernel/mm/ksm/run"
+MEM_TMP_FILE=$(mktemp meminfo.XXXXXXXXXX || true)
```
jcvenegas: Could you remove or explain `||true`  is needed here?

metrics/density/docker_memory_usage.sh
```diff
@@ -87,6 +91,23 @@ get_runc_pss_memory(){
         echo "$avg"
 }
 
+get_runc_individual_memory() {
+	runc_process_result=$(cat $MEM_TMP_FILE | tr "\n" " " |   sed -e 's/\s$//g')
```
jcvenegas: can we validate if what we get here is number?

jcvenegas: I am not clear here do we expect a numerical number or multiples numerical numbers?

GabyCT:
@jcvenegas changes applied

metrics/density/docker_memory_usage.sh
GabyCT: @jcvenegas it can be a number or multiple numbers it depends of how manny containers are you running

jcvenegas:
@GabyCT I agree with @grahamwhaley  for someone using the json would be better an arrary int values.

GabyCT:
@jcvenegas and @grahamwhaley , yes the only problem that I found is that I do not have an idea of how to create that with the current setup that we have

grahamwhaley:
Ah, yes @GabyCT , normally I'd say you could use the lib `metrics_json_start_array()` helper funcs, but as you are already inside an array generation sequence (and the json helper funcs don't support nesting at the moment...), I think your only option is to open code the generation of the elements inside your `local json` fragment. A bit like the json helper funcs do themselves:
https://github.com/kata-containers/tests/blob/master/metrics/lib/json.bash#L180-L188

You could either try and store the set of ints in a bash array and steal that snippet directly, or modify it so it handles the multi-entry string I think you have with a `for element in $mem_usage` type loop.
And, yes, the fun bit is trying to get the `,`'s correct in the array ;-)

marcov:
@GabyCT you can populate the array with: `read -r -a values <<< "foo bar baz"`. 
* All elements: `${values[@]}`
* n-th element: `${values[n]}`
* Array size: `${#values[@]}`

GabyCT:
@grahamwhaley and @marcov thanks for the suggestions :), let me try them (fingers crossed)

metrics/density/docker_memory_usage.sh
```diff
@@ -87,6 +98,34 @@ get_runc_pss_memory(){
...
+	fi
+
+	read -r -a runc_values <<< "${runc_process_result}"
+	read -r -a numbers <<< "${individual_numbers}"
+
+	metrics_json_start_array
+
+	local json="$(cat << EOF
+	{
+		"runc individual results": {
```
jcvenegas: I'd prefer not having spaces in json keys.

jcvenegas:
@grahamwhaley I'd like to have a second  review here.

metrics/density/docker_memory_usage.sh
```diff
@@ -87,6 +98,34 @@ get_runc_pss_memory(){
...
+			$(for ((i=0;i<${NUM_CONTAINERS[@]};++i)); do
+				printf '"Result_%s (Kb)": %s\n\t\t\t' "${numbers[i]}" "${runc_values[i]}"
```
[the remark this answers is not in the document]
GabyCT: @jcvenegas , `json` can not handle a generic key name if you have multiple results that is why we need `Result_%s`

metrics/density/docker_memory_usage.sh
GabyCT: @jcvenegas this is the way that we have been doing the `json` for metrics

GabyCT: @jcvenegas all metrics have the same format

metrics/density/docker_memory_usage.sh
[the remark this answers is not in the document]
GabyCT: @jcvenegas if you leave only `Result` your `json` will not be valid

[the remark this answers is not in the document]
grahamwhaley: I think we can handle this with an array of ints though in json, something like:

```json
"qemu-lite-system-x86_64 memory": [
                        128820,
                        134953,
                        137046,
                        137269,
                        138987
                ],
```

That is more what I envisioned - an array of un-tagged/named ints. wdyt @GabyCT I think that will also make it much easier to parse out the array from external json tools like `jq` and `R`.

Whilst I'm here :-), is there an extra set of `{}`'s in the current output at the outer level we can also do away with? I'm not sure we need the `{` between the `[` and the "qemu-lite..." iyswim.
```json
"Raw results": [
                        {
                "qemu-lite-system-x86_64 memory": {
                        "Result_1 (Kb)": 128820,
                        "Result_2 (Kb)": 134953,
                        "Result_3 (Kb)": 137046,
                        "Result_4 (Kb)": 137269,
                        "Result_5 (Kb)": 138987

                },
                "kata-shim memory": {
                        "Result_1 (Kb)": 4996,
                        "Result_2 (Kb)": 5051,
                        "Result_3 (Kb)": 5058,
                        "Result_4 (Kb)": 5059,
                        "Result_5 (Kb)": 5202

                },
                "kata-proxy memory": {
                        "Result_1 (Kb)": 3356,
                        "Result_2 (Kb)": 3358,
                        "Result_3 (Kb)": 3358,
                        "Result_4 (Kb)": 3360,
                        "Result_5 (Kb)": 3370

                }
        }
```

metrics/density/docker_memory_usage.sh
grahamwhaley: I'm not sure if that is strictly true ;-), but I believe we have got some and they are creeping in more. I don't have a distinct problem with spaces in the identifiers - as they are quoted we can parse them out OK I believe. I do suspect we can mostly clearly define the data by using appropriately named sub sections in the JSON structure - so, let's keep questioning ourselves and trying to structure the data so it is clear and makes sense etc.

GabyCT:
@grahamwhaley changes applied

grahamwhaley:
@jcvenegas OK with this now?

jcvenegas:
lets merge it

PR #692, opened 2018-08-31
travis: add linux-ppc64le and osx

chavafg:
Run statick checks on linux-ppc64le and osx.

Fixes: #691.

### DIFF
--- a/cmd/log-parser/Makefile
+++ b/cmd/log-parser/Makefile
@@ -23,7 +23,8 @@ $(TARGET): $(SOURCES) check
 	go build -o "$(TARGET)" -ldflags "-X main.name=${TARGET} -X main.commit=${COMMIT} -X main.version=${VERSION}" .
 
 install: $(TARGET)
-	install -D $(TARGET) $(DESTTARGET)
+	install -d $(shell dirname $(DESTTARGET))
+	install $(TARGET) $(DESTTARGET)
 
 clean:
 	rm -f $(TARGET)
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Copyright (c) 2017-2018 Intel Corporation
 #
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,19 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-dist: trusty
+os:
+  - linux
+  - osx
+  - linux-ppc64le
+
+matrix:
+  include:
+    - os: linux
+      sudo: required
+      dist: trusty
 
 language: go
+go_import_path: github.com/kata-containers/tests
 
 go:
   - "1.10.x"
@@ -15,5 +25,10 @@ go:
 env:
   - target_branch=$TRAVIS_BRANCH
 
-before_script:
-  - ".ci/static-checks.sh github.com/kata-containers/tests"
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get update -qq              ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -y -qq automake ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install bash; fi
+
+script:
+  - bash .ci/static-checks.sh github.com/kata-containers/tests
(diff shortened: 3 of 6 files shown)


chavafg:
getting this error:
```
.ci/static-checks.sh: line 24: typeset: -A: invalid option
```
but that one should not be a problem if installing latest available `bash`, right? 
any idea @jodh-intel @raravena80?

raravena80:
@chavafg right. But brew installs it under `/usr/local/bin/bash` and not `/usr/bin/bash` (which is the system bash)

marcov:
@raravena80 right, so @chavafg should probably add something like:
```
before_script:
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PATH=/usr/local/bin:$PATH; fi
```

chavafg:
ohh right, thank you both, let me add that line.

marcov:
Also you need to change the shebang in `.ci/static_checks.sh` with `#!/usr/bin/env bash`

jodh-intel:
Thanks @chavafg!

lgtm

/cc @raravena80.

chavafg:
Seems like `SIGSTKFLT ` and `SIGPWR` are not available in Darwin, any idea of how can I skip those if running on Darwin, but use them if running on Linux?

```
integration/docker/kill_test.go:135:14:warning: SIGSTKFLT not declared by package syscall (staticcheck)
integration/docker/kill_test.go:135:14:warning: unused variable or constant SIGSTKFLT not declared by package syscall (varcheck)
integration/docker/kill_test.go:149:14:warning: unused variable or constant SIGPWR not declared by package syscall (varcheck)
integration/docker/kill_test.go:135:14:warning: unused struct field SIGSTKFLT not declared by package syscall (structcheck)
integration/docker/kill_test.go:149:14:warning: unused struct field SIGPWR not declared by package syscall (structcheck)
integration/docker/kill_test.go:135:14:warning: SIGSTKFLT not declared by package syscall (unused)
```

caoruidong:
Split them into two files with +build tags?

marcov:
@chavafg you can safely remove SIGSTKFLT (it's unused on Linux and not existing on darwin).

About SIGPWR, is not a POSIX signal and maybe that can be removed too. From signal(7):
```
       SIGPWR (which is not specified in POSIX.1-2001) is typically ignored by default on those other  UNIX  systems  
       where it appears
```

cmd/log-parser/Makefile
```diff
@@ -22,8 +24,14 @@ check:
 $(TARGET): $(SOURCES) check
 	go build -o "$(TARGET)" -ldflags "-X main.name=${TARGET} -X main.commit=${COMMIT} -X main.version=${VERSION}" .
 
+INSTALL_FLAGS :=
+
+ifeq ($(detected_OS),Linux)
+	INSTALL_FLAGS := -D
```
marcov: To keep things cleaner you can avoid checking OS version here and just run:

```Makefile
install: $(TARGET)
	install -d $(shell dirname $(DESTTARGET))
	install $(TARGET) $(DESTTARGET)
```

jodh-intel:
Agreed - we've already removed `SIGSTKFLT` from the shim and proxy:

- https://github.com/kata-containers/shim/pull/87
- https://github.com/kata-containers/proxy/pull/82

We still have `SIGPWR` in the runtime and the default behaviour appears to be "terminate". To handle that for Darwin, you could adopt the approach we're using in the runtime which:

- Creates architecture-specific `.go` files.
- Creates variables and functions prefixed with `arch*`. Those variables and functions then either provide a custom (architecture-specific) implementation, *or* they are set to a function called `generic*` to provide the default behaviour.

Specifically, I think you could:

- Move the signal and bool values in the `withSignal()` calls into a `var genericSignalMap map[syscall.Signal]bool`, but don't add `SIGPWR` to the map.
- Create `integration/docker/kill_linux_test.go` which will:
  - Create a `var archSignalMap = genericSignalMap`.
  - Add a `syscall.SIGPWR` entry to the map (maybe in a `func init()` function).
  - Call the common `Describe()` function in  `integration/docker/kill_test.go`.
- Create `integration/docker/kill_darwin_test.go` which will:
  - Create a `var archSignalMap = genericSignalMap`.
  - Call the common `Describe()` function in  `integration/docker/kill_test.go`.
- Rework the `Describe()` function to iterate the `archSignalMap`, calling `WithSignal()` on the values.

jodh-intel:
Thanks @chavafg!

lgtm

GabyCT:
lgtm

Issue #698, opened 2018-09-03
metrics: CI: Only run time sensitive tests if on a stable machine


grahamwhaley:
Many of our metrics are only really useful if we are running on a 'non-noisy' machine and can repeat the tests (which generally dictates, some sort of dedicated bare metal single occupancy machine).
But, there are a few tests (in particular the density/footprint tests) that are not generally time sensitive, and in theory can be run on transient cloud instances and should be fairly immune to different systems and noisy neighbors etc.

Adapt the [metrics run script](https://github.com/kata-containers/tests/blob/master/.ci/run_metrics_PR_ci.sh) so it can be told to only run the time insensitive tests, hence making it somewhat useful for cloud instance footprint tests.

PR #699, opened 2018-09-03
metrics: ci: enable cloud instance metrics CI

grahamwhaley:
If we are on a known noisy/time variant system then we can still
run some metrics that are nominally time insensitive (like the
footprint/density tests). Modify the scripts so we can set an
env var to notify we are on such a system, and thus adapt the
tests we run, and how we locate a correct checkmetrics file to
use to check the results.

Fixes: #698

### DIFF
--- a/.ci/run_metrics_PR_ci.sh
+++ b/.ci/run_metrics_PR_ci.sh
@@ -12,6 +12,8 @@ SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_DIR}/../metrics/lib/common.bash"
 RESULTS_DIR=${SCRIPT_DIR}/../metrics/results
 CHECKMETRICS_DIR=${SCRIPT_DIR}/../cmd/checkmetrics
+CHECKMETRICS_CONFIG_DIR="/etc/checkmetrics"
+CM_DEFAULT_DENSITY_CONFIG="${CHECKMETRICS_DIR}/baseline/density-CI.toml"
 
 # Set up the initial state
 init() {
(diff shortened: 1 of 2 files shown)


grahamwhaley:
Notes then:
- I expect to have to tweak some of this, but the only real way to find out is to try it...
   - but, tbh, I can probably do that using this branch and over-riding the default 'tests repo' setting for the CI - I'll give that a try whilst we review
- I'm not sure storing the default density checkmetrics file in this repo with the checkmetrics tool is the ideal place. It could also live over in the ci repo - but, maybe having it sat with the tool that uses the file is not the worst place for it.

.ci/run_metrics_PR_ci.sh
```diff
@@ -67,7 +73,30 @@ check() {
...
+		# FIXME - we need to document the whole metrics CI setup, and the ways it
+		# can be configured and adapted. See:
+		# https://github.com/kata-containers/ci/issues/58
+		#
+		# If we are running on a (transient) cloud machine then we cannot
+		# tie the name of the expected results config file to the machine name - but
+		# we can expect the cloud image to have a default named file in the correct
+		# place.
+		if [ -n "${METRICS_CI_CLOUD}" ]; then
+			local CM_BASE_FILE="${CHECKMETRICS_CONFIG_DIR}/checkmetrics-json.toml"
```
jodh-intel: Amusing filenames ftw! :smile:

jodh-intel:
lgtm

.ci/run_metrics_PR_ci.sh
grahamwhaley: Yep - whilst writing this PR, when I ran into that it made me 👀 squint somewhat and scratch my head - and I wrote that file!!! Finally I figured out it is named that way as it is a toml file that indexes a bunch of json files - whereas its predecessor used to index csv files... I'm not going to rename it now unless somebody really wants me to ;-)

PR #701, opened 2018-09-03
tests: Add versions.yaml in order to support nginx working image

### LINKED ISSUE #700: Add versions.yaml in order to support nginx working image
With versions.yaml we will able to specify which nginx version is working

GabyCT:
With versions.yaml we will be able to specify which nginx version is
working.

Fixes #700

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,0 +1,13 @@
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+docker_images:
+  description: "Docker hub images used for testing"
+
+  nginx:
+    description: "Proxy server for HTTP, HTTPS, SMTP, POP3 and IMAP protocols"
+    url: "https://hub.docker.com/_/nginx/"
+    version: "1.14"
--- a/integration/kubernetes/nginx.bats
+++ b/integration/kubernetes/nginx.bats
@@ -8,7 +8,9 @@
 load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
 
 setup() {
-	nginx_image="nginx:1.14"
+	versions_file="${BATS_TEST_DIRNAME}/../../versions.yaml"
+	nginx_version=$("${GOPATH}/bin/yq" read "$versions_file" "docker_images.nginx.version")
+	nginx_image="nginx:$nginx_version"
 	busybox_image="busybox"
 	service_name="nginx-service"
 	export KUBECONFIG=/etc/kubernetes/admin.conf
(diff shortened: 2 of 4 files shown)


grahamwhaley:
Indeed, this is good to have. I think having it in its own separate file in this repo makes sense. As to txt/yaml....
- I agree using yaml like we do for the runtime versions file makes a lot of sense
    - except it will add some complexity to the test code - we will need a `ver=$(yq version.yaml myimage)` type line
- if we stick with a txt file, then I think that should probably be a '.bash' file, seeing as we are sourcing it - it is shell syntax/grammar really.
    - but sourcing that will start to 'pollute' the tests with a lot of image vars, only one of which each test is really interested in.

I think I vote yaml therefore - the `yq` line does not look too bad (my line above is not correct btw, but you can find/copy similar lines that read the runtime version.yaml file from this codebase). I also like the global shell name space pollution reduction.

jodh-intel:
Sounds good to me.

We could potentially move some of the contents of `versions.yaml` into this repo too?

See also: https://github.com/kata-containers/packaging/issues/165.

GabyCT:
@chavafg changes applied

GabyCT:
@grahamwhaley and @jodh-intel , I did these changes :)

versions.yaml
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+  nginx:
+    description: "Proxy server for HTTP, HTTPS< SMTP, POP3 and IMAP protocols"
```
jodh-intel: It looks like there might be a typo here - should that be?:

> HTTPS, SMTP,

versions.yaml
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+  nginx:
```
jodh-intel: It's difficult to know the exact structure, but I would be inclined to have a "higher level" key as this file might grow over time. Similar to https://github.com/kata-containers/runtime/blob/master/versions.yaml, maybe you could do something like:

```
dependencies:
  nginx:
     ...
```

Or if not `dependencies`, maybe `externals` or something. I'd also consider adding in `description` and `format` as we do for the runtime equivalent of this file.

GabyCT:
@jodh-intel , yes sorry it was a typo, fixed

versions.yaml
chavafg: agree with @jodh-intel, maybe call them `docker_images`? we later can add all docker images that we use:

```
docker_images:
  description: "Docker images used for testing"

  nginx:
    ....
```

GabyCT:
@chavafg and @jodh-intel changes applied

jcvenegas:
@grahamwhaley  @jodh-intel  please  check again and merge if you are happy with the current changes.

Issue #703, opened 2018-09-04
increase go test timeout


bergwolf:
Hit more go test timeout recently in CI. Please consider increasing the 10 second timeout.

http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-16-04-PR/676/console
http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-16-04-PR-initrd/428/consoleFull
http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-16-04-PR/677/console

PR #704, opened 2018-09-04
ci: increase go test timeout

bergwolf:
We've hit more go test timeout in CI. CI already cost 10s of minutes,
there is no reason not to tolerant a bit more time just for `go test`.

Fixes: #703

Signed-off-by: Peng Tao <bergwolf@gmail.com>

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -20,7 +20,7 @@ long_options=(
 #
 # KATA_GO_TEST_TIMEOUT can be set to any value accepted by
 # "go test -timeout X"
-timeout_value=${KATA_GO_TEST_TIMEOUT:-10s}
+timeout_value=${KATA_GO_TEST_TIMEOUT:-30s}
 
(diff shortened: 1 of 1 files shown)


sboeuf:
@bergwolf have you run this patch several times to check it fixes the CI issues you're mentioning?
I am asking cause sometimes bumping the timeout will only hide/workaround some real issue.

bergwolf:
@sboeuf I don't have environment to run ci locally right now but as mentioned in #703 there are multiple timeout failures on different PRs without any real failures.

bergwolf:
Not sure why `pullapprove` disagree with my SOB line.
```
"Signed-off-by" needed on 0 commits
```

chavafg:
lgtm

@bergwolf have you tried a repush to see if pullapprove gets unstuck?

jodh-intel:
Pullapprove is now awake but Travis appears to have "gone fishing". Let's just do this...

PR #720, opened 2018-09-05
CI: Fix optional /usr/lib/os-release

### LINKED ISSUE #589: improve robustness of `source /etc/os-release`
We use `source /etc/os-release` a lot across the repos to check which distro we are on for packaging, testing etc.
According to `man os-release` (visible at https://www.freedesktop.org/software/systemd/man/os-release.html):

> The file /etc/os-release takes precedence over /usr/lib/os-release. Applications should check for the former, and exclusively use its data if it exists, and only fall back to /usr/lib/os-release if it is missing. Applications should not read data from both files at the same time. /usr/lib/os-release is the recommended place to store OS release information as part of vendor trees. /etc/os-release should be a relative symlink to /usr/lib/os-release, to provide compatibility with applications only looking at /etc. A relative symlink instead of an absolute symlink is necessary to avoid breaking the link in a chroot or initrd environment such as dracut.

Thus, I think we should try to source from `/etc/os-release`, and if that fails then we should try `/usr/lib/os-release`. Right now we only go to `/etc`.
This is not a problem for the distros we run on today, but may be in the future.

I wonder if:
```sh 
source /etc/os-release || source /usr/lib/os-release
```

would work? If not, we might have to write a little helper func in the global shell lib.

raravena80:
Fixes #589

Low hanging fruit 😄  @grahamwhaley @jodh-intel @jcvenegas PTAL

Adding "|| source /usr/lib/os-release" to all the lines with "source /etc/os-release"

Signed-off-by: Ricardo Aravena <raravena@branch.io>

### DIFF
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -8,11 +8,15 @@
 set -e
 
 cidir=$(dirname "$0")
-source "/etc/os-release"
+source "/etc/os-release" || "source /usr/lib/os-release"
 source "${cidir}/lib.sh"
 
 # Obtain CentOS version
-centos_version=$(grep VERSION_ID /etc/os-release | cut -d '"' -f2)
+if [ -f /etc/os-release ]; then
+  centos_version=$(grep VERSION_ID /etc/os-release | cut -d '"' -f2)
+else
+  centos_version=$(grep VERSION_ID /usr/lib/os-release | cut -d '"' -f2)
+fi
 
 # Check EPEL repository is enabled on CentOS
 if [ -z $(yum repolist | grep "Extra Packages") ]; then
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -19,7 +19,7 @@ source "${script_dir}/lib.sh"
 cri_containerd_version=$(get_version "externals.cri-containerd.version")
 containerd_version=$(get_version "externals.cri-containerd.meta.containerd-version")
 
-source /etc/os-release
+source /etc/os-release || source /usr/lib/os-release
 
 echo "Set up environment"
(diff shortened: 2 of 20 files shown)


jodh-intel:
> Low hanging fruit

Always the juiciest! :smile:

lgtm

jodh-intel:
Centos CI prodded with a pointy stick (network issues...)

grahamwhaley:
I need to refresh myself on what the preferred search order is and check this is the right one - did I put that on the original Issue? (no time to surf and check myself just now...)

jodh-intel:
Good question. Quoting from `os-release(5)`:

> The file /etc/os-release takes precedence over /usr/lib/os-release.

grahamwhaley:
I prodded Centos CI again - network fail again - looks like a 404, so we may just have to wait or choose to ignore that one for now.

jcvenegas:
restarted again centos network issues again.

.ci/setup_env_centos.sh
```diff
@@ -8,11 +8,15 @@
 set -e
 
 cidir=$(dirname "$0")
-source "/etc/os-release"
+source "/etc/os-release" || "source /usr/lib/os-release"
 source "${cidir}/lib.sh"
 
 # Obtain CentOS version
-centos_version=$(grep VERSION_ID /etc/os-release | cut -d '"' -f2)
+if [ -f /etc/osrelease ]; then
```
chavafg: typo `/etc/os-release` preventing have a value on `$centos_version`

raravena80: Pushed the fix @chavafg PTAL Thx.

jodh-intel: Good catch @chavafg! :smile:

PR #724, opened 2018-09-07
integration/docker: modify testcase of memory constraints update

### LINKED ISSUE #723: integration/docker: modify testcase of memory constraints update
In test integration/docker/mem_test.go, we run a docker with 512M memory, and update it with 256MB. But in https://github.com/kata-containers/runtime/pull/624 we don't support memory hot remove now. 
So I think we should change this testcase to update memory constraints with 512MB+128MB=640MB.

clarecch:
In test integration/docker/mem_test.go, we run a docker with
512M memory, and update it with 256MB. But we don't support
memory hot remove in kata-runtime update interface now. So we
should change the memory constraints update.

Fixes #723

Signed-off-by: Clare Chen <clare.chenhui@huawei.com>

### DIFF
--- a/integration/docker/mem_test.go
+++ b/integration/docker/mem_test.go
@@ -142,8 +142,8 @@ var _ = Describe("memory constraints", func() {
 			_, _, exitCode = dockerRun(args...)
 			Expect(exitCode).To(BeZero())
 
-			// 256 MB
-			memSize = fmt.Sprintf("%d", 256*1024*1024)
+			// 640 MB: 512MB + 128MB
+			memSize = fmt.Sprintf("%d", 640*1024*1024)
 
 			args = []string{"--memory", memSize, "--memory-reservation", memSize}
 			if useSwap {


jodh-intel:
Hi @clarecch - I'm a little confused. If this test used to be doing "hot remove", why wasn't it failing before?

I think the test should assert both that hot plug works and that hot remove fails.

/cc @devimc.

clarecch:
> I think the test should assert both that hot plug works and that hot remove fails.

@jodh-intel I also think we should assert both that hot plug works and that hot remove fails. But it should be after https://github.com/kata-containers/runtime/pull/624 merged. So I assert hot plug works first.

jodh-intel:
Thanks @clarecch.

clarecch:
The CI is green now, PTAL

jodh-intel:
lgtm

caoruidong:
lgtm

PR #726, opened 2018-09-07
tests: Verify processes (hypervisor, shim and proxy)

GabyCT:
Verify that processes (hypervisor, shim and proxy) are not running before
and after running tests like swarm, popular docker images, kubernetes,
cri-o and openshift.

Fixes #725

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

PR #732, opened 2018-09-10
metrics: Check processes in network metrics

### LINKED ISSUE #731: Check processes in network metrics
We need to check that processes like shim, proxy and hypervisor are not running before and after running the network metrics.

GabyCT:
We need to check that processes like shim, proxy and hypervisor are not
running before and after running the network metrics.

Fixes #731

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -1,0 +1,67 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# This file contains common functions that
+# are being used by our metrics and integration tests
+
+die(){
+	msg="$*"
+	echo "ERROR: $msg" >&2
+	exit 1
+}
+
+# Gets versions and paths of all the components
+# list in kata-env
+extract_kata_env(){
+	local toml
+
+	toml="$(kata-runtime kata-env)"
+
+	# The runtime path itself, for kata-runtime, will be contained in the `kata-env`
+	# section. For other runtimes we do not know where the runtime Docker is using lives.
+	RUNTIME_CONFIG_PATH=$(awk '/^\[Runtime\]$/ {foundit=1} /^    Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+	RUNTIME_VERSION=$(awk '/^  \[Runtime.Version\]$/ {foundit=1} /^    Semver =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+	RUNTIME_COMMIT=$(awk '/^  \[Runtime.Version\]$/ {foundit=1} /^    Commit =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+
+	SHIM_PATH=$(awk '/^\[Shim\]$/ {foundit=1} /^  Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+	SHIM_VERSION=$(awk '/^\[Shim\]$/ {foundit=1} /^  Version =/ { if (foundit==1) {$1=$2=""; print $0; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+
+	PROXY_PATH=$(awk '/^\[Proxy\]$/ {foundit=1} /^  Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+	PROXY_VERSION=$(awk '/^\[Proxy\]$/ {foundit=1} /^  Version =/ { if (foundit==1) {print $5; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+
+	HYPERVISOR_PATH=$(awk '/^\[Hypervisor\]$/ {foundit=1} /^  Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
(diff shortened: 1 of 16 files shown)


grahamwhaley:
Looking at the [check_processes code](https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L248), I'm not sure that is how you run it, or it will quite do what you need. That function takes an argument of the process name to check, so you'd have to invoke it for each component you are looking for aiui (proxy, shim, runtime, qemu etc. etc.)

What I think we can and should utilise is the code that lives in the 
[check_all_running function](https://github.com/kata-containers/tests/blob/master/integration/stability/soak_parallel_rm.sh#L55)
in the rm soak test, which checks the correct number of components are running for the correct number of containers (including for 0 containers iirc). We should lift that code out to a common place (which may not be the metrics lib, it may have to be the common tests repo lib), and then we can utilise that code both here, and in the rm soak test, and then in other tests as well.

Then you can use `check_all_running 0` here.

GabyCT:
@grahamwhaley , the code https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L249 is already looking for shim, proxy and hypervisor processes. It is not necessary to pass an argument to the function (when you are calling it).

GabyCT:
@grahamwhaley so if we move the `check_processes` to a common place, we need also to modify the lib/common.bash to move all this information https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L34. Are you ok with these changes?

grahamwhaley:
Oh, I think I see what has happened - over on #726 , you added that list into `check_processes`, but maybe missed some updates in common.bash itself - like the three calls in the [kill_processes_before_start](https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L221) function I think. That is what threw me. The comment above `check_processes` also still says it checks $1 ;-)

For moving the kata env and relevant path/process name detection logic out of metrics and into common space - yes, I think if we want to do a full sanity check after every test we run (which we already know will find some 'left over runtime bits' type bugs earlier, then we don't really have a choice.
Hopefully we can pull that whole kata_env function out pretty much as-is and then ensure it gets called just once very early on in our test cycle so we don't keep calling/extracting for each test.

We'll have to check thoroughly during this code rework that we don't break either the QA tests or the metrics tests - it's probably going to take a little time and iteration to find the right balance and setup.

grahamwhaley:
Just for clarity (for future herders), I'm going to stick a DNM label on this until @GabyCT can re-visit and do the rework and updates.

GabyCT:
@grahamwhaley, I'll do the changes, one final question so the `check_processes` function should be located at a new path like `kata-containers/tests/lib/common.bash` so it stays generic for all the integration, stability, etc tests or just to have it at `kata-containers/tests/metrics/lib/common.bash`, thanks.

grahamwhaley:
I think we can and/or will re-use it out in the QA tests to check system sanity, so I think moving it out to `tests/lib/common.bash` is probably the right thing to do @GabyCT thx!

GabyCT:
@grahamwhaley changes applied

lib/common.bash
```diff
@@ -0,0 +1,73 @@
...
+
+	INITRD_PATH=$(awk '/^\[Initrd\]$/ {foundit=1} /^  Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
+}
+
+# Checks that processes are not running
+check_processes() {
+	extract_kata_env
+	# Get system architecture
+	arch=$(go env GOARCH)
+	if [ "$arch" == "arm64" ]; then
```
grahamwhaley: I don't remember seeing this before. Do we know why we do this here? Given we are checking there are no kata processes laying around, I thought this check would be pretty `arch` neutral?

metrics/lib/common.bash
```diff
@@ -195,65 +172,23 @@ init_env()
...
 
-	result=$(check_processes "$HYPERVISOR_PATH")
-	if [[ $result -ne 0 ]]; then
-		warning "Found unexpected hypervisor [${HYPERVISOR_PATH}] processes present"
-		# Sometimes we race and the process has gone by the time we list
-		# it - so make a pgrep fail non-fatal
-		pgrep -a -f "$HYPERVISOR_PATH" || true
-		sudo killall -9 "${HYPERVISOR_PATH##*/}" || true
+	check_processes
+	if [ $? -eq 0 ]; then
```
grahamwhaley: check_processes is now a `die` case on failure I think - so we can probably never get to this kill loop. Can you confirm @GabyCT ? I probably vote we then leave check_processes as a die, and drop the kill loop here.

lib/common.bash
GabyCT: @grahamwhaley , yeah I got confused by this https://github.com/kata-containers/tests/blob/master/.ci/aarch64/clean_up_aarch64.sh#L27, I will modified to be neutral and in for the `clean_up_aarch64.sh` create another function to check `docker-containerd-shim`, what do you think? thanks

metrics/lib/common.bash
GabyCT: @grahamwhaley perfect so for the `kill_processes_before_start`, I will leave the `check_processes` and remove the kill loop, thanks

lib/common.bash
grahamwhaley: Ah, that is where it came from. Right. yeah, you can probably just open code the pgrep over in the aarch64 code in place of the call to check_processes - or, you could make a 'check_process' that takes a single argument (like check_processes used to), and then call that from both the aarch64 code and from check_processes - that is, factor out the pgrep/kill stuff :-). Either way works for me right now.

GabyCT:
@grahamwhaley thanks for the feedback, changes applied

GabyCT:
@grahamwhaley ping, do you agree with these changes? thanks

Issue #735, opened 2018-09-11
add ksm-throttler to the CI systems


chavafg:
We need to run the CI tests using the ksm-throttler.

jodh-intel:
Related: https://github.com/kata-containers/ci/issues/9.

chavafg:
already addressed by https://github.com/kata-containers/tests/commit/e5e98b397e3c49d6301a399a7fb04ac2556d687f

PR #736, opened 2018-09-11
CI: Add ksm-throttler

chavafg:
Install kata-ksm-throttler to the CI systems
before running the tests.

Fixes: #735

### DIFF
--- a/.ci/install_ksm_throttler.sh
+++ b/.ci/install_ksm_throttler.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+METRICS_CI="${METRICS_CI:-}"
+throttler_repo="github.com/kata-containers/ksm-throttler"
+throttler_dir="${GOPATH}/src/${throttler_repo}"
+
+go get -d "$throttler_repo"
+
+pushd "$throttler_dir"
+make
+sudo -E PATH=$PATH make install
+popd
+
+# If we are running under the metrics CI system then we do not want the throttler
+# to be dynmaically changing the KSM settings under us - we need control of them
+# ourselves
+if [ -z "${METRICS_CI}" ]; then
+	sudo systemctl daemon-reload
+	sudo systemctl enable kata-vc-throttler.service
+	sudo systemctl start kata-vc-throttler.service
+fi
--- a/.ci/install_kata.sh
+++ b/.ci/install_kata.sh
@@ -30,3 +30,6 @@ echo "Install proxy"
 
 echo "Install runtime"
 "${cidir}/install_runtime.sh"
+
+echo "Install KSM throttler"
+"${cidir}/install_ksm_throttler.sh"


.ci/install_ksm_throttler.sh
```diff
@@ -0,0 +1,28 @@
...
+throttler_dir="${GOPATH}/src/${throttler_repo}"
+
+go get -d "$throttler_repo"
+
+pushd "$throttler_dir"
+make
+sudo make install
+popd
+
+# If we are running under the metrics CI system then we do not want the proxy
```
grahamwhaley: s/proxy/throttler/ :-)

grahamwhaley:
lgtm
but, the CIs are not happy - probably simple env/path/make issue:
```
Install KSM throttler
/tmp/jenkins/workspace/kata-containers-tests-ubuntu-16-04-PR/go/src/github.com/kata-containers/ksm-throttler /tmp/jenkins/workspace/kata-containers-tests-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests
     GOBUILD  build
     GOBUILD  kata-ksm-throttler
     GOBUILD  kicker
     GOBUILD  virtcontainers
     GOBUILD  virtcontainers
/bin/sh: 1: go: not found
Makefile:54: recipe for target 'virtcontainers' failed
make: *** [virtcontainers] Error 127
```

.ci/install_ksm_throttler.sh
chavafg: fixed.

chavafg:
ups, forgot to add the `-E PATH=$PATH` to `sudo`. Now it should be fine.

jcvenegas:
@chavafg 
```
/tmp/jenkins/workspace/kata-containers-tests-vsock-fedora-27-PR/go/src/github.com/kata-containers/tests
.ci/install_ksm_throttler.sh: line 25: METRICS_CI: unbound variable
```

grahamwhaley:
oooh, bash-lint eh?

jcvenegas:
@chavafg @grahamwhaley  hehe yep, `${METRICS_CI:-}` should fix our problem

chavafg:
fixed :)

GabyCT:
lgtm

.ci/install_ksm_throttler.sh
```diff
@@ -0,0 +1,29 @@
...
+pushd "$throttler_dir"
+make
+sudo -E PATH=$PATH make install
+popd
+
+# If we are running under the metrics CI system then we do not want the throttler
+# to be dynmaically changing the KSM settings under us - we need control of them
+# ourselves
+if [ -z "${METRICS_CI}" ]; then
+	sudo systemctl enable kata-vc-throttler.service
```
jodh-intel: Since the `make install` installs systemd units, you'll need to do the following before enabling and starting the service:

```
$ sudo systemctl daemon-reload
```

chavafg:
I see two issues here: 

1. When running the docker tests with ksm-throttler, sometimes I got this error:
```
Stderr: docker: Error response from daemon: OCI runtime create failed: Failed to check if grpc server is working: context deadline exceeded: unknown.
```
Still need to investigate more on this. 

2. The `soak_parallel_rm.sh` test is taking more than 5 minutes to complete, which makes me think that when using ksm-throttler, the container boot time is increasing

I ran a `true` workload to check the times and it seems that we are very slow:

`kata-vc-throttler` enabled:
```
ubuntu@fuentess-test:~/go/src/github.com/kata-containers/tests$ time sudo docker run -ti ubuntu true

real    0m10.485s
user    0m0.092s
sys     0m0.008s
```
`kata-vc-throttler` disabled:
```
ubuntu@fuentess-test:~/go/src/github.com/kata-containers/tests$ time sudo docker run -ti ubuntu true

real    0m6.596s
user    0m0.072s
sys     0m0.004s
```

is it expected to have this impact when enabling ksm? 
also, those 6 seconds is too slow, right? 

/cc @sboeuf @jodh-intel @grahamwhaley @jcvenegas 

Also, taking a look at the jobs, the docker tests went from taking ~1000 secs to ~2300 secs in the execution, which would mean ~20 minutes more for the CI.

grahamwhaley:
Yeah, that feels too high - feels like the throttler default enable/ramp is way too aggressive would be my guess. You can probably check this by watching a 'top' - you should see I think it is `ksmd` slam up to the top. We might want to dial that back.
@sameo was the originator of the ksm agression profiles, and may have thoughts.

grahamwhaley:
Have we had any thoughts or more investigation here?
@jodh-intel - are you able to easily extract/describe/point towards the ksm-throttler aggresiveness algo - that is, can you describe how the throttler will configure KSM the instant after a container is launched, so we can consider the potential performance impact?

chavafg:
Opened https://github.com/kata-containers/runtime/issues/746 and https://github.com/kata-containers/ksm-throttler/issues/50 for the issues I'm seeing. In the meantime, adding DNM flag.

caoruidong:
@chavafg Any updates?

raravena80:
@chavafg ping (from your weekly Kata herder)

jodh-intel:
Ping @chavafg - how's this looking?

jodh-intel:
Oh - all these CI results have also disappeared ;(

grahamwhaley:
it may be they were just too old - I think we have Jenkins set to prune at 30 days unless it is explicitly told to hold a result forever (which I'm not sure I've ever seen us do).
we can '/ t e s t' it again when needed...

jodh-intel:
Ping @chavafg.

jodh-intel:
@chavafg, @GabyCT - presumably #1077 is related to this?

chavafg:
@jodh-intel maybe it is related. The issues I saw is the increase in boot time, which led to timeouts in the CI, and actually these timeouts were happening on tests executed before running the `soak_parallel_rm` test. 
And this haven't been merged because the CI would last too long (maybe more that 2 hrs).
But now that we have more CI resources, we can think of running a subset of the tests using the throttler.

grahamwhaley:
Hi @chavafg @jodh-intel - well, I think the throttler should not kill the CI (making it timeout and extending its runtime by 2x) - I'd see that as a potential throttler bug (likely a bad choice of initial KSM settings and scaling would be my guess). Something we should investigate if we do support the throttler as part of Kata and a Kata installation.

chavafg:
closing this one as it is being managed by #1122

PR #738, opened 2018-09-11
metrics: Check subdir exists while running makereport script

### LINKED ISSUE #737: Check subdir exists while running makereport script
To avoid the issue of `ls: cannot access '*/': No such file or directory` while running the `./makereport.sh`, we need to check that there is a subdirectory at least in `metrics/results`.

GabyCT:
To avoid the issue of ls: cannot access '*/': No such file or directory,
while running the ./makereport.sh, we need to check that there is a
subdirectory at least in metrics/results

Fixes #737

### DIFF
--- a/metrics/report/makereport.sh
+++ b/metrics/report/makereport.sh
@@ -23,6 +23,12 @@ GUESTINPUTDIR="/inputdir/"
 GUESTOUTPUTDIR="/outputdir/"
 
 setup() {
+	echo "Checking subdirectories"
+	check_subdir="$(cd ${HOSTINPUTDIR}; ls -dx */ > /dev/null 2>&1 | wc -l)"
+	if [ $check_subdir -eq 0 ]; then
+		die "Subdirectory not found at metrics/results to store JSON results"
+	fi
+
 	echo "Checking Dockerfile"
 	check_dockerfiles_images "$IMAGE" "$DOCKERFILE"
 


GabyCT:
@grahamwhaley changes applied

metrics/report/makereport.sh
```diff
@@ -23,6 +23,12 @@ GUESTINPUTDIR="/inputdir/"
 GUESTOUTPUTDIR="/outputdir/"
 
 setup() {
+	echo "Checking subdirectories"
+	check_subdir="$(cd ${HOSTINPUTDIR}; ls -dx */ | wc -l)"
```
chavafg: I think we can send the `stderr` to `/dev/null` here.
I see that this line may send this error:
```
ls: cannot access '*/': No such file or directory
```

metrics/report/makereport.sh
```diff
@@ -23,6 +23,12 @@ GUESTINPUTDIR="/inputdir/"
...
+	if [ $check_subdir -eq 0 ]; then
+		die "Subdirectory not found"
```
chavafg: Maybe we need to be more explicit here? 
What kind of subdirectory does this script needs? How can it be created?

GabyCT:
@chavafg changes applied

chavafg:
lgtm

jodh-intel:
>  But, tbh, if you have spaces in your path names then maybe you deserve to find out the hard way :-)

Couldn't agree more - the sort of pain that causes tends to make you learn *really* fast :)

+1

PR #743, opened 2018-09-13
wip: test commit for docker connect

bergwolf:
@sboeuf and I were getting different results w.r.t. network hotplug functionality. Still not sure what makes the difference. Let's see what we can get in CI environment. So we can have a base line for environment comparison.

Please DO NOT MERGE!


bergwolf:
I've triggered tests for this for three times, once passed and twice failed on getting the IP of the added nic. Still not sure what makes the difference. Seems like CI successfully simulated both @sboeuf's case and mine. @sboeuf @devimc @amshinde any ideas? This basically tests network hotplug using the netmon PR.

caoruidong:
@bergwolf Can this be closed?

PR #748, opened 2018-09-17
CI: Fix teardown for throttler

### LINKED ISSUE #747: CI: fix teardown for throttler
The teardown script isn't handling the `kata-vc-throttler` systemd unit and is incorrectly treating the `kata-ksm-throttler` as a tag/identifier (technically, it's a unit).

jodh-intel:
Change the teardown script to treat the `kata-ksm-throttler` as a systemd unit, not just a tag/identifier.

Also, update the teardown script to include the log for the `kata-vc-throttler` systemd unit and save the throttler logs.

Fixes #747.


chavafg:
lgtm

PR #749, opened 2018-09-17
ci: Add support  for new CI job (k8s + containerd)

jcvenegas:
Handle the case when `"$CI_JOB" == "CRI_CONTAINERD_K8S"` this should be set by the CI to indicate K8s + containerd testing should be done.

### DIFF
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -41,7 +41,7 @@ if [ -n "$CI" ]; then
 	sudo mv /etc/cni/net.d/10-containerd-net.conflist  "$cni_test_dir"
 fi
 
-cat | sudo tee /etc/systemd/system/containerd.service  << EOT
+cat <<EOT | sudo tee /etc/systemd/system/containerd.service
 [Unit]
 Description=containerd container runtime
 Documentation=https://containerd.io
@@ -62,6 +62,12 @@ ExecStart=/usr/local/bin/containerd
 [Install]
 WantedBy=containerd.target
 EOT
+
+sudo mkdir -p  /etc/systemd/system/kubelet.service.d/
+cat << EOF | sudo tee  /etc/systemd/system/kubelet.service.d/0-containerd.conf
+[Service]                                                 
+Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
+EOF
 
 sudo systemctl daemon-reload
 
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -41,4 +41,28 @@ if [ -n "$CI" ]; then
 	sudo mv /etc/cni/net.d/10-containerd-net.conflist  "$cni_test_dir"
 fi
 
+cat | sudo tee /etc/systemd/system/containerd.service  << EOT
+[Unit]
+Description=containerd container runtime
+Documentation=https://containerd.io
+After=containerd-installation.service
+[Service]
+Restart=always
+RestartSec=5
+Delegate=yes
+KillMode=process
+OOMScoreAdjust=-999
+LimitNOFILE=1048576
+# Having non-zero Limit*s causes performance problems due to accounting overhead
+# in the kernel. We recommend using cgroups to do container-local accounting.
+LimitNPROC=infinity
+LimitCORE=infinity
+ExecStartPre=/sbin/modprobe overlay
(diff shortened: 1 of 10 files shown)


jcvenegas:
@chavafg  CI keeps working with this changes, I also added (and removed) a temporal commit to tests if CI_JOB is handled when k8s+contaienrd is set . It works now we only need to add the missing job.
Let me know how can I help with.
BTW it is cool we can use `\ t e s t  ` to trigger CI.

.ci/install_cri_containerd.sh
```diff
@@ -41,4 +41,34 @@ if [ -n "$CI" ]; then
 	sudo mv /etc/cni/net.d/10-containerd-net.conflist  "$cni_test_dir"
 fi
 
+cat | sudo tee /etc/systemd/system/containerd.service  << EOT
```
jodh-intel: This looks wrong to me - shouldn't it be:

```
cat <<EOT | sudo tee /etc/systemd/system/containerd.service
```

Same comment for `0-crio.conf` below.

.ci/setup.sh
```diff
@@ -11,88 +11,121 @@ cidir=$(dirname "$0")
...
+setup_distro_env(){
+	echo "Set up environment"
+	if [ "$ID" == ubuntu ];then
+		bash -f "${cidir}/setup_env_ubuntu.sh"
+	elif [ "$ID" == fedora ];then
+		bash -f "${cidir}/setup_env_fedora.sh"
+	elif [ "$ID" == centos ];then
+		bash -f "${cidir}/setup_env_centos.sh"
+	else
+		die "ERROR: Unrecognised distribution."
```
jodh-intel: Admittedly, this isn't in the original code, but it might be worth logging `$ID` here so it's clear what distro was specified.

.ci/setup.sh
```diff
@@ -11,88 +11,121 @@ cidir=$(dirname "$0")
...
+			if [ "$CI" == true ] && grep -q "N" /sys/module/kvm_intel/parameters/nested 2> /dev/null; then
+				echo "enable Nested Virtualization"
+				sudo modprobe -r kvm_intel
+				sudo modprobe kvm_intel nested=1
+				if grep -q "N" /sys/module/kvm_intel/parameters/nested 2> /dev/null; then
+					die "Failed to find or enable Nested virtualization"
+				fi
+			fi
+			;;
+		aarch64)
```
jodh-intel: @nitkon - fyi, looks like this might need updating at some point in the future for PPC64le.

integration/kubernetes/init.sh
```diff
@@ -11,6 +11,21 @@ SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/../../.ci/lib.sh"
 source "${SCRIPT_PATH}/../../metrics/lib/common.bash"
 
+cri_runtime="${CRI_RUNTIME:-crio}"
+
+case "${cri_runtime}" in
+crio)
+    cri_runtime_socket="/var/run/crio/crio.sock"
+    ;;
+containerd)
```
jodh-intel: Nit: You could sort alphabetically so `crio` comes first.

.ci/setup.sh
```diff
@@ -11,88 +11,121 @@ cidir=$(dirname "$0")
...
+CRIO="${CRIO:-yes}"
+CRI_CONTAINERD="${CRI_CONTAINERD:-no}"
+KUBERNETES="${KUBERNETES:-yes}"
+OPENSHIFT="${OPENSHIFT:-yes}"
+
+setup_distro_env(){
+	echo "Set up environment"
+	if [ "$ID" == ubuntu ];then
+		bash -f "${cidir}/setup_env_ubuntu.sh"
+	elif [ "$ID" == fedora ];then
```
jodh-intel: Nit: Not a blocker as this isn't in the original code, but you could sort these distros alphabetically.

jcvenegas:
@jodh-intel @chavafg lest test again.

integration/kubernetes/k8s-cpu-ns.bats
```diff
@@ -20,6 +20,7 @@ setup() {
 }
 
 @test "Check CPU constraints" {
+	[ "${CRI_RUNTIME}" == "containerd" ] && skip "test not working with ${CRI_RUNTIME}"
```
chavafg: can you open an issue for this and add it here?

jcvenegas:
@chavafg  so I think did not consider that use case when OBS is not available, we keep build from source in case the package can not be installed but I think not check if the repository can be added.

PR #751, opened 2018-09-17
security: Add security tests

### LINKED ISSUE #750: Add security tests
Add common security tests just to be sure they are not reproducible in kata-containers

devimc:
Add security tests for:
* dirtycow

Add `security` Makefile target

fixes #750

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,9 @@ openshift:
 	bash -f .ci/install_bats.sh
 	bash -f integration/openshift/run_openshift_tests.sh
 
+pentest:
+	bash -f pentest/all.sh
+
 vm-factory:
 	bash -f integration/vm_factory/vm_templating_test.sh
 
@@ -104,6 +107,7 @@ check: checkcommits log-parser
 	kubernetes \
 	log-parser \
 	openshift \
+	pentest \
 	swarm \
 	test \
 	vm-factory
--- a/pentest/all.sh
+++ b/pentest/all.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+tests_dir=$(dirname $0)
+
+bash -f "${tests_dir}/dirtycow.sh" &> /dev/null
+if [ $? != 0 ]; then
+	echo "[NOT OK] dirtycow test"
+	exit 1
+fi
+echo "[OK] dirtycow test"
--- a/pentest/dirtycow.sh
+++ b/pentest/dirtycow.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+source "$(dirname $0)/lib.sh"
+
+container_name="dirtycow"
+runtime=${RUNTIME:-kata-runtime}
+test_repo="https://github.com/dirtycow/dirtycow.github.io"
+test_dir="/root/dirtycow"
+test_file="$test_dir/test"
+test_file_content="Hello"
+dirty_file_content="pwned"
+
+finish() {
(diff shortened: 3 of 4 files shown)


security/dirtycow.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+trap finish EXIT
+
+# Run a gcc container
+docker run --runtime="$runtime" --name="$container_name" -dti gcc bash
+
+# Turning off periodic writeback makes exploit stable
+# This should fail inside a container
+docker exec "$container_name" bash -c "echo 0 > /proc/sys/vm/dirty_writeback_centisecs"
+[ $? == 0 ] && exit 1
```
jodh-intel: Might be worth displaying a message for the two failure scenarios to make it clear in the logs what happened.

Makefile
```diff
@@ -101,4 +104,5 @@ check: checkcommits log-parser
 	log-parser \
 	openshift \
 	swarm \
+	security \
```
grahamwhaley: Surprised @jodh-intel did not spot this is not actually alphabetical ;-)

security/dirtycow.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+container_name="dirtycow"
+runtime=${RUNTIME:-kata-runtime}
```
grahamwhaley: Do we need this? We set the default RUNTIME in a number of places, but tbh, I'm not sure how or if we set it 'globally'. I think having less of these default RUNTIME sets is probably the way forwards though. /cc @chavafg

Makefile
jodh-intel: I blame lack of :coffee: :smile:

security/dirtycow.sh
chavafg: At least for the CI, this is not needed as it is exported on the `.ci/run.sh`. But it may be needed if you are only running this test in a local environment. So I think it is ok.

devimc: I prefer to keep it, to make life easier to users that just want to check if their custom kernel is secure

devimc:
@jodh-intel @chavafg @grahamwhaley  changes applied, thanks

security/dirtycow.sh
```diff
@@ -0,0 +1,47 @@
...
+# Run a gcc container
+docker run --runtime="$runtime" --name="$container_name" -dti gcc bash
+
+# Turning off periodic writeback makes exploit stable
+# This should fail inside a container
+docker exec "$container_name" bash -c "echo 0 > /proc/sys/vm/dirty_writeback_centisecs"
+[ $? == 0 ] && die "Turned off periodic writeback"
+
+# Clone dirtycow repo
+docker exec "$container_name" git clone "$test_repo" "$test_dir"
```
jodh-intel: It seems unlikely that this will be changed again but if it does, we might want to consider adding a particular `HEAD` commit to `versions.yaml`.

jodh-intel:
lgtm

grahamwhaley:
Looks like some CIs are timing out though @devimc  From bunty 16.04 initrd:

```
bash -f security/all.sh
Build timed out (after 5 minutes). Marking the build as aborted.
```

chavafg:
seems like the test does not send output in 5 minutes, which makes the job to fail. @devimc do you know how much time the test takes? Only option I see is to change the timeout that we have from 5 to maybe 10 minutes.

egernst:
Just to calrify @chavafg 's comments -- is this test adding 5-10 minutes to each PR run?  I don't want to limit tests, but we should be judicious until we have our CI acting in a pipeline (parallel) fashion?

devimc:
yes, this test takes for around ~7-8 minutes

devimc:
ready for merge :smile:

jodh-intel:
*"Katanauts, let's do this!"*

Issue #758, opened 2018-09-19
set gometalinter concurrency


bergwolf:
The default concurrency is 16 and it is quite likely to oom even on a system with 4GB memory. We should set it to a more reasonable value (e.g., `nproc`).

PR #759, opened 2018-09-19
static-checks: limit gometalinter concurrency

bergwolf:
The default concurrency is 16 and it is quite likely to oom even on
a system with 4GB memory. Set it to `nproc` so that it scales
automatically. And it allow `make check` to run on developer's
local environment more easily.

Fixes: #758

Signed-off-by: Peng Tao <bergwolf@gmail.com>


chavafg:
lgtm

travis stuck again 😕

grahamwhaley:
Travis travis travis. If anybody has any ideas on that, pls speak up.
force-merging then...

jodh-intel:
Good catch @bergwolf!

Issue #760, opened 2018-09-19
vm-factory tests are failing in the initrd CI


devimc:
```
+ sudo docker rm -f kata-lWQnAR
kata-lWQnAR
+ destroy_vm_template
+ echo 'destroy vm template'
destroy vm template
+ sudo -E PATH=/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/bin:/usr/local/go/bin:/usr/sbin:/sbin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games /usr/local/bin/kata-runtime factory destroy
vm factory destroyed
++ mount
++ grep template
++ wc -l
+ res=7
+ '[' 7 -eq 0 ']'
+ die 'template factory is not cleaned up'
+ msg='template factory is not cleaned up'
+ echo 'ERROR: template factory is not cleaned up'
ERROR: template factory is not cleaned up
+ exit 1
Makefile:88: recipe for target 'vm-factory' failed
make: *** [vm-factory] Error 1
Build step 'Execute shell' marked build as failure
Performing Post build task...
Match found for :.* : True
Logical operation result is TRUE
Running script  : #!/bin/bash
```

see http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-16-04-PR-initrd/593/consoleFull

cc @bergwolf

bergwolf:
Looks like there a many leftover `/run/vc/vm/template` mountpoints from previous kata runtime UTs.
```
init vm template
+ sudo -E PATH=/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/bin:/usr/local/go/bin:/usr/sbin:/sbin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games /usr/local/bin/kata-runtime factory init
vm factory initialized
+ check_vm_template_factory
+ echo 'checking vm template factory'
checking vm template factory
+ mount
+ grep /run/vc/vm/template
+ grep tmpfs
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,size=2105344k)
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,size=2105344k)
[...]

PR #761, opened 2018-09-20
ci: make sure template mountpoints are cleaned up

bergwolf:
Since we relay on the mountpoints to check for test results,
let's make sure it is cleaned up before and after the tests.

Fixes: #760

### DIFF
--- a/integration/vm_factory/vm_templating_test.sh
+++ b/integration/vm_factory/vm_templating_test.sh
@@ -54,7 +54,14 @@ check_vm_template_factory() {
 	[ $res -eq 1 ] || die "template factory is not set up, missing state file"
 }
 
+cleanup_template_mountpoints() {
+	while $(mount|grep tmpfs | grep ${template_tmpfs_path}>/dev/null); do
+		sudo umount ${template_tmpfs_path}
+	done
+}
+
 setup() {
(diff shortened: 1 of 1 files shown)


bergwolf:
ping @jodh-intel @grahamwhaley. This is blocking all the runtime repo PRs.

grahamwhaley:
I can see the principle, to clean up before we run the test (and iirc, we do that for processes with some kill_all_processes() in the metrics and maybe QA tests), but, this is also hiding the problem that somebody seems to be leaving those mount points behind in the first place....
We have the [rm soak test](https://github.com/kata-containers/tests/blob/master/integration/stability/soak_parallel_rm.sh) that does a global sanity check, and we are working on getting those checks to run after every test is run, so we can pick these things up earlier.

I'm prone to say - this patch is fine, but we should identify who left the mounts around in the first place, and have either a check or a BUG and/or Issue for that - ideally before we merge this (or we'll forget ;-) ).

bergwolf:
@grahamwhaley Fair enough. You've beaten me to it :) See https://github.com/kata-containers/runtime/pull/764

bergwolf:
Let's close this one and let it catch future leaks if possible.

PR #774, opened 2018-09-26
CI: teardown fixes

chavafg:
Update `teardown.sh` to now use the full path of `kata-collect-data.sh` when it is called. This is needed on Fedora and Centos, otherwise a `command not found` error appears. 
Also add a change to create the directory (if it doesn't exists) where the log files will be stored.


jodh-intel:
lgtm

/test

PR #775, opened 2018-09-26
metrics: Fix percentage difference row

### LINKED ISSUE #709: metrics: report: % diff row seems to calculate incorrectly
Looking at some reports, it seems the % row calculation for at least the footprint might be incorrect.

If you have a look at the metrics_report.pdf attached at https://github.com/kata-containers/agent/pull/353#issuecomment-418419932 then I think we can see the match is wrong.

The relevant code to inspect is at:
https://github.com/kata-containers/tests/blob/master/metrics/report/report_dockerfile/memory-footprint.R#L79-L92

Note - the % calculation will always have to be a '% of one entry vs the other' - presently I believe it is meant to be 'the % difference the second set of data has relative to the first set' iyswim. We'll have to keep that in mind, and how to best present it, whilst fixing the issue.

GabyCT:
Fix the percentage difference calculation for the metrics report generation.

Fixes #709


GabyCT:
@grahamwhaley , yes I saw that formula too, I got confused if we really wanted to see the + and - signs but for me that looks better, I will work on the changes and apply them in the rest of diffs that we have in the dockerfiles thanks

GabyCT:
@grahamwhaley changes applied thanks

metrics/report/report_dockerfile/lifecycle-time.R
```diff
@@ -98,7 +98,8 @@ if (length(resultdirs) == 2) {
 	# This is a touch hard wired - but we *know* we only have two
 	# datasets...
 	for( i in 1:5) {
-		val = ((as.double(rstats[1,i]) / as.double(rstats[2,i])) * 100) - 100
+		diference = as.double(rstats[2,i]) - as.double(rstats[1,i])
```
grahamwhaley: yay, this is looking good to me, and generated good output in my test. Just one tiny spelling niggle....
s/diference/difference/ throughout :-)

GabyCT:
@grahamwhaley changes applied

Issue #776, opened 2018-09-27
metrics: report: results subdir check is broken


grahamwhaley:
Over in merge https://github.com/kata-containers/tests/pull/738, when we added a subdir check to ensure we had data to process before generating the report, a last minute change got in that redirected all output to /dev/null, and thus failed the subsequent `wc` check.
It can be fixed by moving where the error output is re-directed.
At the same time, the check first does a `chdir` before the `ls`, and if the root dir does not exist (which it does not if you have never run any metrics tests), then we fail in a different way, and run the report build, and fail later on. Fix that whilst there - the `chdir` at this point is not necessary.

PR #777, opened 2018-09-27
metrics: report: fix subdir check in makereport

grahamwhaley:
An early change slide in a stderr re-direct that broke the check
in the good case as well.
Fix that by moving the stderr redirect. Also, drop the chdir as it
is not needed, and can fail in itself.
Whilst here, improve the error message for when we do fail the dir
check.

Fixes: #776


jodh-intel:
lgtm

Travis restarted (net error).

chavafg:
lgtm

Issue #779, opened 2018-09-27
cri-o: do not use devicemapper on Zuul CI 


chavafg:
On Zuul slaves, we might not be able to get a block device to test CRI-O, depending on the region where the slave is launched. 
So instead of using loopback devices (since we have experience some instabilities), lets use overlay storage driver.

PR #780, opened 2018-09-27
tests: Use overlay for CRI-O when running on zuul

chavafg:
When running on Zuul, we might not get a block device
for testing CRI-O using devicemapper driver. This depends
on the region where the zuul slave is launched. Instead,
use overlay storage driver.

Fixes: #779


chavafg:
Depends-on: https://review.openstack.org/#/c/605773/

GabyCT:
lgtm

PR #784, opened 2018-09-27
teardown: Remove uncompressed log files before archiving them

chavafg:
the `teardown.sh` collects and compresses log files, so that
jenkins can archive them, but we are also archiving the
uncompressed files, meaning that we are using more space
than the needed. With this, we remove the uncompressed log
files and save more space in the jenkins master.

Also, kernel, kata-vc-throttler and kata-ksm-throttler logs
will now be compressed.

### DIFF
--- a/.ci/teardown.sh
+++ b/.ci/teardown.sh
@@ -125,6 +125,7 @@ collect_logs()
 
 		local prefix
 
+		# Compress log files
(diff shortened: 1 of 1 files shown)


GabyCT:
lgtm

jodh-intel:
lgtm

PR #787, opened 2018-10-01
tests: Verify hypervisor, shim and proxy are not running

### LINKED ISSUE #786: Verify that hypervisor, shim and proxy is not running for docker integration tests
In order to ensure a clean environment, we need to verify that hypervisor, shim and proxy is not running before running our docker integration tests. Related with https://github.com/kata-containers/tests/issues/215.

GabyCT:
Check that hypervisor, shim and proxy processes are not running before
we run our docker integration tests.

Fixes #786

### DIFF
--- a/integration/docker/main_test.go
+++ b/integration/docker/main_test.go
@@ -62,6 +62,20 @@ func TestIntegration(t *testing.T) {
 		}
 	}
 
+	// we need to check that processes like hypervisor, shim and proxy are not running
+	hypervisorPath := KataConfig.Hypervisor[DefaultHypervisor].Path
+	proxyPath := KataConfig.Proxy[DefaultProxy].Path
+	shimPath := KataConfig.Shim[DefaultShim].Path
+	generalProcesses := []string{hypervisorPath, proxyPath, shimPath}
+
+	for _, j := range generalProcesses {
+		cmd := NewCommand("pgrep", "-f", j)
+		_, _, exitCode := cmd.Run()
+		if exitCode == 0 {
+			t.Fatal("Process found", j)
+		}
+	}
+
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Integration Suite")
 }


integration/docker/main_test.go
```diff
@@ -62,6 +62,20 @@ func TestIntegration(t *testing.T) {
...
 	}
 
+	// we need to check that processes like hypervisor, shim and proxy are not running
+	hypervisorPath := KataConfig.Hypervisor[DefaultHypervisor].Path
+	proxyPath := KataConfig.Proxy[DefaultProxy].Path
+	shimPath := KataConfig.Shim[DefaultShim].Path
+	generalProcesses := []string{hypervisorPath, proxyPath, shimPath}
+
+	for _, j := range generalProcesses {
+		cmd := NewCommand("pgrep", "-f", j)
```
grahamwhaley: Apparently, as go is sort of mostly OS neutral, and there is no standard way to do this, invoking the commands directly is probably fine :-) I know normally we'd recommend seeking out the appropriate golang library.

integration/docker/main_test.go
```diff
@@ -62,6 +62,20 @@ func TestIntegration(t *testing.T) {
 		}
 	}
 
+	// we need to check that processes like hypervisor, shim and proxy are not running
+	hypervisorPath := KataConfig.Hypervisor[DefaultHypervisor].Path
```
grahamwhaley: I have a feeling, and probably correctly, that @jodh-intel is going to ask if we can somehow factor our the list of things we check to somewhere more 'central'. I'm wondering if it is possible to do this via a yaml or something. It might not be, but the concern will be we are probably going to duplicate this list in a number of places (this is already at least the second place), and if/when we change something later (like add the net monitor or ksm throttler like we did recently, or change the proxy or shim architecture like we did a while back), then we're going to have to locate and modify multiple places (on the upside.... hopefully any change will trigger them to fail, and thus maybe they will be easy to find :-) ).

Maybe we need a little out of the box thinking and brainstorming here. I wonder if we should have a single external command that knows how to sanity check, and we call that command from everywhere (including here, and in the metrics tests). That command could be in `bash`, or `go` - I'm not too fussed.

wdyt @jodh-intel @GabyCT @chavafg 

Other than that - the notion is right - I just want to make sure we centralise and future proof as much as we can (and I'd like to also be checking mounts and /var dirs and net connections etc. like we do in the rm soak test ).

jodh-intel: @grahamwhaley - That sounds good to me!

devimc: I'd prefer to keep the current implementation and just validate (check no empty values, paths, etc) the configuration toml in `loadKataConfiguration` method

GabyCT: ^ @grahamwhaley and @jodh-intel

grahamwhaley: @devimc can you expand on that. IIUC the intent of this PR, Ideally we should be checking all system level sanity items we can, including things that are not in the config file (like the number pods and vms down in /var match our container count, and that we have the correct number of file system and network mounts/connections etc.).
The majority of that checking already exists over at [the rm soak test](https://github.com/kata-containers/tests/blob/master/integration/stability/soak_parallel_rm.sh#L55-L130). My idea is that we do that rigorous system sanity check after every test, to catch dangling issues early.

GabyCT:
Closing this issue as an improvement will be submitted.

PR #796, opened 2018-10-03
tests: Define Alpine version in the proper versions.yaml

### LINKED ISSUE #795: Move Alpine version to the proper versions.yaml
For docker integration tests and memory tests, we are using a specific Alpine image. The version of this image should be placed in the `versions.yaml` of our test repository.

GabyCT:
For docker integration tests and memory tests, we are using a specific version
for the Alpine image. This version should be placed at versions.yaml in our
repository.

Fixes #795

### DIFF
--- a/integration/docker/docker.go
+++ b/integration/docker/docker.go
@@ -7,6 +7,7 @@ package docker
 import (
 	"bytes"
 	"fmt"
+	"gopkg.in/yaml.v2"
 	"io/ioutil"
 	"log"
 	"os"
@@ -26,9 +27,6 @@ const (
 	// Image used to run containers
 	Image = "busybox"
 
-	// AlpineImage is the alpine image
-	AlpineImage = "alpine:3.7"
-
 	// PostgresImage is the postgres image
 	PostgresImage = "postgres"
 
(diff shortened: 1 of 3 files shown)


metrics/density/memory_usage_inside_container.sh
```diff
@@ -15,7 +15,9 @@ SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/../lib/common.bash"
 
 TEST_NAME="memory footprint inside container"
-IMAGE="alpine:3.7"
+VERSIONS_FILE="${SCRIPT_PATH}/../../versions.yaml"
+ALPINE_VERSION=$("${GOPATH}/bin/yq" read "$VERSIONS_FILE" "docker_images.alpine.version")
+IMAGE="alpine:$ALPINE_VERSION"
```
grahamwhaley: I don't know if we can extract this 'alpine' word from from the yaml as well, or use the `alpine.url` - but even then it would feel a bit self referential - if we did not know we were using `alpine` here, we would not know to look up the `alpine` section in the yaml. Only way out I can think of would be to make the yaml section named after the test (that is, configure all the information about which container this test uses inside the yaml). That feels one step too far for me right now - so ignore this ramble - we'll live with the wiring of `alpine` here I think.

grahamwhaley:
Travis not happy with formatting I think:
```
integration/docker/docker.go:59:6:warning: type name will be used as docker.DockerImage by other packages, and that stutters; consider calling this Image (golint)
integration/docker/docker.go:61:2:warning: struct field Url should be URL (golint)
integration/docker/docker.go:66:6:warning: type name will be used as docker.DockerVersions by other packages, and that stutters; consider calling this Versions (golint)
integration/docker/docker.go:1::warning: file is not gofmted with -s (gofmt)
```

GabyCT:
fixed the formatting

integration/docker/docker.go
```diff
@@ -46,17 +44,61 @@ const (
...
+
+// versionDockerImage is the definition in the yaml for the Alpine image
+type versionDockerImage struct {
+	Description string `yaml:"description"`
+	URL         string `yaml:"url"`
+	Version     string `yaml:"version"`
+}
+
+// versionDockerVersions is the complete information for docker images in the versions yaml
+type versionDockerVersions struct {
```
devimc: versionDockerImages

integration/docker/docker.go
```diff
@@ -46,17 +44,61 @@ const (
...
+	Description string `yaml:"description"`
+	Alpine      versionDockerImage
+}
+
+// Config will be used to parse the versions yaml
+type Config struct {
```
devimc: Versions

chavafg:
lgtm

chavafg:
seems like pullapprove is stuck, as there are 3 lgtm, merging...

PR #801, opened 2018-10-04
tests: Add sanity test after running integration and functional tests

### LINKED ISSUE #800: Add sanity test before running integration and functional tests
Verify that processes like shim, hypervisor and proxy are not left as well as that we do not have pods or runtimes running before we run our functional and integration tests.

GabyCT:
This will verify that processes like shim, hypervisor and proxy are not
left as well as that we do not have pods or runtimes running after
running our functional and integration tests.

Fixes #800

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -7,6 +7,9 @@
 # This file contains common functions that
 # are being used by our metrics and integration tests
 
+# Place where virtcontainers keeps its active pod info
+VC_POD_DIR="${VC_POD_DIR:-/var/lib/vc/sbs}"
+
 die(){
 	msg="$*"
 	echo "ERROR: $msg" >&2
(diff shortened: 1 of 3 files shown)


GabyCT:
@grahamwhaley changes applied

chavafg:
lgtm

GabyCT:
the errors present at the CI is mainly because of `open /sys/module/kvm_intel/parameters/nested: no such file or directory`

GabyCT:
@jodh-intel and @grahamwhaley changes applied

lib/common.bash
```diff
@@ -49,6 +52,27 @@ check_processes() {
...
+# Checks that pods were not left
+check_pods() {
+	if [ -d ${VC_POD_DIR} ]; then
+		# Verify that pods were not left
+		pods_number=$(ls ${VC_POD_DIR} | wc -l)
+		if [ ${pods_number} -ne 0 ]; then
+			die "${pods_number} pods left and found at ${VC_POD_DIR}"
+		fi
+	else
+		echo "Not ${VC_POD_DIR} directory found"
```
jodh-intel: - s/Not/No/
- Shouldn't this be a `die` scenario?

jodh-intel: True, but we're running `check_pods()` after the tests have run so we would expect that directory to exist at that point wouldn't we?

grahamwhaley: Today, that is probably true. OK, if the first batch of tests maybe only test container creation failure cases, or in the future we have a test that does a docker/kata full teardown and reconfig, then the premise may not be true.
We could set it to a `die` right now, and spin the CI to see how it goes. Later on it would be obvious that it needs changing if necessary, as the CIs would fail on the PR making any relevant change (one hopes ;-) ).
So, I'm fine making it a `die` right now it that passes the CIs.

PR #802, opened 2018-10-05
tests: Check netmon process for the soak parallel test

### LINKED ISSUE #727: Update soak test for netmon
`integration/stability/soak_parallel_rm.sh` needs to take account of the new network monitor process when https://github.com/kata-containers/runtime/pull/534 lands.

GabyCT:
This will enable the network monitor process at the configuration file as
well that it will verify that the number of netmons that are being launched
by this test are properly deleted.

Fixes #727

### DIFF
--- a/integration/stability/soak_parallel_rm.sh
+++ b/integration/stability/soak_parallel_rm.sh
@@ -48,6 +48,16 @@ check_vsock_active() {
 	fi
 }
 
+enable_netmon() {
+	echo "Enabling netmon at the configuration file"
+	sed -i 's|#enable_netmon = true|enable_netmon = true|' ${RUNTIME_CONFIG_PATH}
+}
+
+disable_netmon() {
+	echo "Disabling netmon at the configuration file"
+	sed -i 's|enable_netmon = true|#enable_netmon = true|' ${RUNTIME_CONFIG_PATH}
+}
+
 count_containers() {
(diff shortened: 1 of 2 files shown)


integration/stability/soak_parallel_rm.sh
```diff
@@ -48,6 +51,20 @@ check_vsock_active() {
 	fi
 }
 
+enable_netmon() {
+	echo "Enabling netmon at the configuration file"
+	sed -i 's|#enable_netmon = true|enable_netmon = true|' ${DEFAULT_KATA_CONFIG}
+	echo "Reloading unit files"
+	sudo systemctl daemon-reload
```
jodh-intel: Similar to the docker comment, you don't need to call this (here or below) as no systemd units are being touched.

grahamwhaley: Historical reference: I think looong looong ago the addition of the systemctl reload and the docker restart were added at some generic outer level to try and ensure we had a 'known stable starting point' when running the metrics (as we really like repeatable results there).
As per other comments, we probably don't need to do that here.

GabyCT:
@jodh-intel changes applied

GabyCT:
@grahamwhaley changes applied

chavafg:
lgtm

PR #811, opened 2018-10-10
test: Add ramdisk test

### LINKED ISSUE #429: create docker ramdisk test
Once https://github.com/kata-containers/runtime/pull/412 lands, we need a test that checks to ensure a container can be run in a ramdisk.

GabyCT:
This will ensure that we can run with DOCKER_RAMDISK and than --no-pivot
flag is supported.

Fixes #429

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@
 TIMEOUT := 60
 
 # union for 'make test'
-UNION := functional docker crio docker-compose docker-stability openshift kubernetes swarm vm-factory
+UNION := functional docker crio docker-compose docker-stability openshift kubernetes swarm vm-factory ramdisk
 
 # skipped test suites for docker integration tests
 SKIP :=
@@ -95,6 +95,9 @@ network:
 	cd integration/network/macvlan && \
 	bats macvlan_driver.bats
 
+ramdisk:
+	bash -f integration/ramdisk/ramdisk.sh
+
 test: ${UNION}
 
 check: checkcommits log-parser
@@ -115,5 +118,6 @@ check: checkcommits log-parser
 	pentest \
 	swarm \
 	network \
+	ramdisk \
 	test \
 	vm-factory
--- a/integration/ramdisk/ramdisk.sh
+++ b/integration/ramdisk/ramdisk.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+cidir=$(dirname "$0")
+
+source "${cidir}/../../lib/common.bash"
+
+set -e
+
+# Environment variables
+IMAGE="${IMAGE:-busybox}"
+CONTAINER_NAME="${CONTAINER_NAME:-test}"
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-sleep 30}"
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
+
+# File to save ramdisk results
+TMP_FILE=$(mktemp ramdisk.XXXXXXXXXX)
+
+remove_tmp_file() {
+	rm -rf $TMP_FILE
+}
+
+trap remove_tmp_file EXIT
+
+setup() {
+	clean_env
+
+	# Stop docker
+	sudo -E systemctl stop docker
+}
+
+test_ramdisk() {
+	extract_kata_env
+
+	# Grab the time (filter journalctl)
+	grab_time=$(date +"%H:%M:%S")
+
+	# Enable ramdisk with dockerd
+	(sudo -E DOCKER_RAMDISK=true dockerd -D --add-runtime ${RUNTIME}=${RUNTIME_PATH} 2>/dev/null) &
+
+	# Run container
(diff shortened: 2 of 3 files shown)


lib/common.bash
```diff
@@ -25,6 +25,7 @@ extract_kata_env(){
 	RUNTIME_CONFIG_PATH=$(awk '/^\[Runtime\]$/ {foundit=1} /^    Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
```
grahamwhaley: :-) I've just realised this line would be better with a `Runtime.Config` section regexp - that would help make it clear how this is picking out a different line from the one you just added for the actual runtime Path. If you have to do any edits, could you slip that one in here mayber?

integration/ramdisk/ramdisk.sh
```diff
@@ -0,0 +1,81 @@
...
+
+test_ramdisk() {
+	# Enable ramdisk with dockerd
+	(sudo -E DOCKER_RAMDISK=true dockerd -D --add-runtime ${RUNTIME}=${RUNTIME_PATH} 2>/dev/null) &
+
+	# Run container
+	docker run --rm -d --name=${CONTAINER_NAME} --runtime=${RUNTIME} ${IMAGE} sh -c ${PAYLOAD_ARGS}
+
+	# Extract the logs of kata-runtime
+	sudo journalctl -t ${RUNTIME} > ${TMP_FILE}
```
grahamwhaley: It might be an idea to limit this journalctl to a just the entries generated by this test. Maybe we can do that by using `date` to grab a start time, and then use a `--since` on the journalctl to grab only entries generated since this test started.
I know we don't have afaik any other tests that will place a no-pivot into the logs, but you never know ;-)

integration/ramdisk/ramdisk.sh
```diff
@@ -0,0 +1,81 @@
...
+	if [ $check_pivot -eq 0 ]; then
+		echo >&2 "ERROR: --no-pivot flag was not found"
+		exit 1
+	fi
+}
+
+teardown() {
+	dockerd_id=$(pgrep -f dockerd)
+	for i in "${dockerd_id[@]}"; do
+		sudo kill -9 $i
```
grahamwhaley: you might just be able to do a `pkill` here?

integration/ramdisk/ramdisk.sh
```diff
@@ -0,0 +1,81 @@
...
+	rm -rf $TMP_FILE
+}
+
+trap remove_tmp_file EXIT
+
+setup() {
+	clean_env
+
+	# Check that processes are not running
+	check_processes
```
grahamwhaley: If we dictate that all tests must do their own checks at the end of them, then you might not need a pre-check here. Not a big problem, but at this rate we will end up calling the check before and after each test, which is probably theoretically twice as often as we actually need to.

integration/ramdisk/ramdisk.sh
GabyCT: `pkill` did not work for me, I tried to do `pkill` of the `shim` for example and the process was still there, `kill -9` worked

GabyCT:
@grahamwhaley changes applied

Makefile
```diff
@@ -95,6 +95,9 @@ network:
 	cd integration/network/macvlan && \
 	bats macvlan_driver.bats
 
+ramdisk:
```
grahamwhaley: Silly question - are we actually running the test? Does it need to be added to the `UNION` at the top of the file so the `test:` rule invokes it for instance?

integration/ramdisk/ramdisk.sh
```diff
@@ -0,0 +1,81 @@
...
+setup() {
+	clean_env
+
+	# Stop docker
+	sudo -E systemctl stop docker
+}
+
+test_ramdisk() {
+	# Grab the time (filter journalctl)
+	grab_time=$(date +"%H:%M:%S")
```
grahamwhaley: nice, thanks for adding the time grab :-)

devimc:
lgtm

integration/ramdisk/ramdisk.sh
```diff
@@ -0,0 +1,81 @@
...
+
+	# Enable ramdisk with dockerd
+	(sudo -E DOCKER_RAMDISK=true dockerd -D --add-runtime ${RUNTIME}=${RUNTIME_PATH} 2>/dev/null) &
+
+	# Run container
+	docker run --rm -d --name=${CONTAINER_NAME} --runtime=${RUNTIME} ${IMAGE} sh -c ${PAYLOAD_ARGS}
```
chavafg: Saw this at the log:
```
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

It seems that `dockerd` is not ready when calling `docker run ...`, you'll need to wait until `dockerd` is ready to receive requests.

integration/ramdisk/ramdisk.sh
marcov: @GabyCT did you try with `pkill --signal 9` ?

GabyCT: @marcov yes, I tried that and at least in my environment for example the `shim` was still there, that is why I am using the `kill`

chavafg:
lgtm

Issue #816, opened 2018-10-17
ci: fix broken clean-up work for building up CI on aarch64


Pennyzct:
The former clean-up work was broken given to the latest patch  #732. The refined check_processes will break if there exist hanging-around processes, such like proxy, or qemu, etc. The cleaning-up process (`clean_env`) what lib/common.bash provides only focuses on cleaning containers running in the background. 
It's not enough for building up CI on bare-mental machine, where we can't count on hypervisor to provide clean environment. Sometimes former tests could leave stale processes, such as when container couldn't be fully started up, some intermediate processes would hang there, and couldn't be killed by `docker kill` or `docker rm`.
relevant pr will be uploaded asap.
@grahamwhaley @GabyCT

PR #817, opened 2018-10-17
ci: fix broken clean-up work for building up CI on aarch64

Pennyzct:
The former clean-up work was sort of broken by pr#732, here to fix
related codes.

Fixes: #816

Signed-off-by: Penny Zheng penny.zheng@arm.com

### DIFF
--- a/.ci/aarch64/clean_up_aarch64.sh
+++ b/.ci/aarch64/clean_up_aarch64.sh
@@ -13,22 +13,24 @@ stale_docker_mount_point_union=( "/var/lib/docker/containers" "/var/lib/docker/$
 stale_docker_dir_union=( "/var/lib/docker" )
 stale_kata_dir_union=( "/var/lib/vc" "/run/vc" )
 
-lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
+lib_script="${GOPATH}/src/${tests_repo}/lib/common.bash"
+
 source "${lib_script}"
 
-metrics_lib_script="${GOPATH}/src/${tests_repo}/metrics/lib/common.bash"
-source "${metrics_lib_script}"
+info() {
+	echo -e "INFO: $*"
+}
 
 kill_stale_process()
 {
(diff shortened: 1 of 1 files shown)


marcov:
lgtm

hi @Pennyzct, I had a quick look at this file, and at line 59 and 69 the `done <<< "${...}"` should be changed to `done`. It would be nice if you fix those too.

grahamwhaley:
Heh, you scared me a little with the title @Pennyzct , as it didn't mention aarch64 - maybe it should say aarch64 rather than baremetal machine :-)
so, yes, lgtm me too - but, please fold in that fix suggestion from @marcov. Thus, I'm going to add a DNM label to give you a chance to do that. thx!

Pennyzct:
Hi~@marcov thanks for the reminding. mixed up with `while loop` and `for loop`. already updated. ptal. ;)

Pennyzct:
Hi~ @grahamwhaley  my bad😝. refined the title and commit info. ptal.

marcov:
Thanks @Pennyzct, it looks good now :)

grahamwhaley:
metrics CI we can probably ignore for the minute - I need to go fixup some of that later.
The F27 CI fail looks odd, but I suspect unrelated, so I'll give it a nudge:
```
Ensure crio service is stopped before running the tests
Ensure docker service is stopped before running the tests
++ time bats --tap ctr.bats
1..32
ok 1 ctr not found correct error message
ok 2 ctr termination reason Completed
ok 3 ctr termination reason Error
ok 4 ctr remove
ok 5 ctr lifecycle
ok 6 ctr logging
not ok 7 ctr logging [tty=true]
# (in test file ctr.bats, line 239)
#   `grep --binary -P "^[^\n]+ stdout F here is some output\x0d$" "$logpath"' failed
# 0
```

I think that is the error - but the 22Mb log file is not helping me root cause ...

marcov:
> I think that is the error - but the 22~Mb~MB log file is not helping me root cause ...

fixed it for you @grahamwhaley  :)

chavafg:
lgtm

I'll go add an ARM job to the tests repo, so we know if we break the scripts.

Issue #822, opened 2018-10-17
CI: run kata-check with sudo 


chavafg:
This will be needed for https://github.com/kata-containers/runtime/pull/819

PR #823, opened 2018-10-17
CI: run kata-check with sudo

chavafg:
If/when https://github.com/kata-containers/runtime/pull/819
gets merged, kata-check will need to run with sudo as
it will try to load the kernel modules if they are not
already loaded.

Fixes: #822.

Depends-on: github.com/kata-containers/runtime#819


jodh-intel:
lgtm

PR #832, opened 2018-10-19
tests: Update popular docker images tests

### LINKED ISSUE #831: update popular docker hub images
Docker hub is not supporting the manifest for images like kibana and rails.
We need to remove these tests.

GabyCT:
Docker hub is not supporting the manifest for rails. Update kibana working version.

Fixes #831

### DIFF
--- a/integration/popular_docker_hub_images/popular_docker_images.bats
+++ b/integration/popular_docker_hub_images/popular_docker_images.bats
@@ -9,6 +9,10 @@
 
 source ${BATS_TEST_DIRNAME}/../../lib/common.bash
 
+versions_file="${BATS_TEST_DIRNAME}/../../versions.yaml"
+kibana_version=$("${GOPATH}/bin/yq" read "$versions_file" "docker_images.kibana.version")
+kibana_image="kibana:$kibana_version"
+
 setup() {
 	# Check that processes are not running
 	run check_processes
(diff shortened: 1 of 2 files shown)


chavafg:
I see in docker hub that rails is deprecated in favor of ruby image (which we already have in the tests), so agree to remove this one. 
For kibana, there is still a kibana image, but it does not have a latest tag. I could download it using `docker pull kibana:6.4.1`, we could use that tag, wdyt?

GabyCT:
@chavafg , I am having issues in running older versions of kibana, `docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not run process`.

chavafg:
lgtm

grahamwhaley:
Notes then:
- With that failure with older versions of kibana - do they work with `runc`? If so, we might want to confirm what the problem is, and that we are happy it is something Kata has knowingly fixed etc.
- Given this used to be the 'top 100', and we have just dropped one (rails), should we consider adding in a new item to replace it?
- @chavafg - it looks like your experimental? Jenkins pipeline job is posting a fail item as 'jenkins-ci' :-) We probably want to:
  - rename that appropriately
  - fix or disable it until it is working?

For now, we should probably ignore the 'jenkins-ci' fail then...

PR #839, opened 2018-10-23
CI: Fix xurls URL

jodh-intel:
The `xurls` tool used by the static checker seems to have changed its
URL so switch to the (hopefully more reliable!) github one.

Fixes: #835.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #838, opened 2018-10-23
CI: Make static check ignore local URLs

jodh-intel:
Update the static check script to ignore `file://` URLs which won't
exist (since they only appear in examples).

Fixes #837.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #843, opened 2018-10-23
CI: Really fix xurls URL issue

jodh-intel:
Effectively reverted #839 as the real problem appears to be that although the CI is *downloading and installing* the correct version of golang, it's not actually *using* it.

I noticed this because xurls has recently changed the minimum golang version needed to run it to 1.10.3+:

- https://github.com/mvdan/xurls/blob/master/README.md

Testing locally, the old `go get` for xurls works fine but not with versions of go < 1.10.3.

However, although this PR resolves the issue in this repo, we still need to update golang to 1.10.3+ so this PR will fail until the following one lands:

- https://github.com/kata-containers/runtime/pull/744

### DIFF
--- a/.ci/install_go.sh
+++ b/.ci/install_go.sh
@@ -101,9 +101,11 @@ case "$(arch)" in
 		;;
 esac
 
+archive="go${go_version}.linux-${goarch}.tar.gz"
+
 info "Download go version ${go_version}"
-curl -OL "https://storage.googleapis.com/golang/go${go_version}.linux-${goarch}.tar.gz"
+curl -OL "https://storage.googleapis.com/golang/${archive}"
 info "Install go"
 mkdir -p "${install_dest}"
-sudo tar -C "${install_dest}" -xzf "go${go_version}.linux-${goarch}.tar.gz"
+sudo tar -C "${install_dest}" -xzf "${archive}"
 popd
--- a/.ci/install_go.sh
+++ b/.ci/install_go.sh
@@ -109,3 +109,17 @@ info "Install go"
 mkdir -p "${install_dest}"
 sudo tar -C "${install_dest}" -xzf "${archive}"
 popd
+
+# Make sure we use the golang version we just installed
+goroot="${install_dest}/go"
+
+[ -d "${goroot}" ] || die "failed to find expected golang path ${goroot} (from ${archive})"
+
+export GOROOT="${goroot}"
+
+gorootbin="${GOROOT}/bin"
+[ -d "${gorootbin}" ] || die "failed to find expected golang binary path ${gorootbin} (from ${archive})"
+
+export PATH="${gorootbin}:$PATH"
+
+info "Using golang binary $(command -v go) version $(go version)"
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -385,7 +385,7 @@ check_docs()
 	if [ ! "$(command -v $cmd)" ]
 	then
 		info "Installing $cmd utility"
-		go get -u "github.com/mvdan/xurls/cmd/${cmd}"
+		go get -u -v "mvdan.cc/xurls/cmd/${cmd}" 2>&1
 	fi
 
 	info "Checking documentation"


jodh-intel:
Hmm - the go "fix" may actually be bogus since the Jenkins job does set `PATH` correctly:

- https://github.com/kata-containers/ci/blob/master/jenkins/jobs/kata-containers-runtime-ubuntu-16-04-PR/config.xml#L127

In which case, either that isn't propagating to the static checks script or this isn't the problem with xurls ;(

.ci/install_go.sh
```diff
@@ -109,3 +109,17 @@ info "Install go"
...
+goroot="${install_dest}/go"
+
+[ -d "${goroot}" ] || die "failed to find expected golang path ${goroot} (from ${archive})"
+
+export GOROOT="${goroot}"
+
+gorootbin="${GOROOT}/bin"
+[ -d "${gorootbin}" ] || die "failed to find expected golang binary path ${gorootbin} (from ${archive})"
+
+export PATH="${gorootbin}:$PATH"
```
grahamwhaley: iiuc, this script is called (nor sourced) from the jenkins script - so, these exports will be happening in a sub-shell, and not propagate up/out will they?
Maybe there needs to be a GOPATH type fix in either the jenkins script (caller) or the static script (user) ?

grahamwhaley: Hold on - you say we need golang >= 1.10.3 - but, our versions.yaml is not fixed for that yet (as you noted on the runtime repo PR).
So, we should expect this not to work until we get golang updated, with the pending PRs.

Let me go check that PR - last I remember upstream crio got patched so we could make the version bump, let me see where we are stuck....

jodh-intel:
Re-pushed now that https://github.com/kata-containers/runtime/pull/744 has landed.

jodh-intel:
From the Travis log:

```
INFO: Installing xurls utility
Fetching https://mvdan.cc/xurls/cmd/xurls?go-get=1
Parsing meta tags from https://mvdan.cc/xurls/cmd/xurls?go-get=1 (status code 200)
get "mvdan.cc/xurls/cmd/xurls": found meta tag get.metaImport{Prefix:"mvdan.cc/xurls", VCS:"git", RepoRoot:"https://github.com/mvdan/xurls"} at https://mvdan.cc/xurls/cmd/xurls?go-get=1
get "mvdan.cc/xurls/cmd/xurls": verifying non-authoritative meta tag
Fetching https://mvdan.cc/xurls?go-get=1
Parsing meta tags from https://mvdan.cc/xurls?go-get=1 (status code 200)
mvdan.cc/xurls (download)
Fetching https://mvdan.cc/xurls?go-get=1
Parsing meta tags from https://mvdan.cc/xurls?go-get=1 (status code 200)
get "mvdan.cc/xurls": found meta tag get.metaImport{Prefix:"mvdan.cc/xurls", VCS:"git", RepoRoot:"https://github.com/mvdan/xurls"} at https://mvdan.cc/xurls?go-get=1
mvdan.cc/xurls
mvdan.cc/xurls/cmd/xurls
INFO: Checking documentation
INFO: Checking local branch for changed documents only
INFO: No documentation to check
```

All that xurls spew is good by the way (but ugly as sin :)

grahamwhaley:
You could `chronic` it to hide the mess :-)

jodh-intel:
Ah - sorry, I should have explained - I actually wanted to see this output so I added the `-v` to `go get` - I just didn't want us to hide all the spew given all the trouble this package has caused us ;)

chavafg:
lgtm

jodh-intel:
The fact that #838 is now showing passed CI jobs suggests that this PR is indeed bogus - the real issue being that xurls changed the minimum version of golang it requires to work so it wasn't an issue with the xurls projects URL:

- https://github.com/kata-containers/runtime/pull/744

The CI installs golang version `languages.golang.meta.newest-version` (**not** `languages.golang.version`):

- https://github.com/kata-containers/tests/blob/master/.ci/install_go.sh#L76

Previous to https://github.com/kata-containers/runtime/pull/744 landing, the CI would have been installing golang 1.10 which is too old for the latest xurls which needs 1.10.3):

- https://github.com/mvdan/xurls/blob/master/README.md

I'm going to close this PR but if we get any more issues with xurls, we should consider versioning it in:

- https://github.com/kata-containers/runtime/blob/master/versions.yaml

Issue #857, opened 2018-10-25
add -f flag to install_go script 


chavafg:
we need a `-f` flag to force go installations when a the installed one is not correct.

PR #858, opened 2018-10-25
CI: add force flag to install_go.sh

chavafg:
force flag will be used in the CI scripts to remove the
current golang version and install the correct one.
This is useful on baremetal machines where golang is already
installed and versions differ from the one needed to test kata.

Fixes: #857.


jodh-intel:
lgtm

grahamwhaley:
Do we need to CI test this?

chavafg:
I think so, the metrics job should be a good candidate for this, as it is already stuck when trying to install a newer golang version.

chavafg:
ohh, forgot to add the flag when it is called on `jenkins_job_build.sh`

Issue #862, opened 2018-10-26
metrics: report: improve report gen speed using base R docker images


grahamwhaley:
Presently we build the R docker image from bare components (apt, R). The R libraries get built from sources, which can take a long time (~1h on some systems).
You only have to do this one generally, but the long runtime is laborious and may put people off running the code.

This can be improved by using the [base R docker images](https://hub.docker.com/r/rocker/tidyverse/) - if only I could figure out how to get that to add the extra R libraries and latex stuff we need....

PR #863, opened 2018-10-26
metrics: report: optimise Dockerfile image creation

grahamwhaley:
We were creating the Rmarkdown processing docker image pretty
much by hand (with apt and R installs), but this takes a lot of
time, particularly for the R parts, as they install from source.
Change the Dockerfile to use the official 'tidyverse' base, and
then install the R parts using the install2.r, which uses binaries
where possible. This speeds up the first-time image build.

Fixes: #862

### DIFF
--- a/metrics/report/report_dockerfile/Dockerfile
+++ b/metrics/report/report_dockerfile/Dockerfile
@@ -9,29 +9,29 @@
 #  - pandoc
 #  - The report generation R files and helper scripts
 
-# I would have liked to have used from base rocker/verse, but that
-# does not have all the tools we need, and it seems that apt install
-# does not simply work - texlive for instance did not install.
-#
-# So, build up our image by hand then...
-FROM ubuntu
+# Start with the base rocker tidyverse.
+# We would have used the 'verse' base, that already has some of the docs processing
+# installed, but I could not figure out how to add in the extra bits we needed to
+# the lite tex version is uses.
+FROM rocker/tidyverse
 
 # Version of the Dockerfile
-LABEL DOCKERFILE_VERSION="1.0"
+LABEL DOCKERFILE_VERSION="1.1"
 
 # Without this some of the package installs stop to try and ask questions...
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && \
-	apt-get install -y tzdata r-base pandoc texlive
+# Install the extra doc processing parts we need for our Rmarkdown PDF flow.
+RUN apt-get update -qq && \
+  apt-get install -y \
+    texlive-latex-base \
+    texlive-fonts-recommended \
+    latex-xcolor
 
-# R packages get built from source, and can take some time, so split them
-# into individual RUNs so we can make some later modifications without maybe taking
-# the full image rebuild hit.
-RUN Rscript -e "install.packages('tidyverse')"
-RUN Rscript -e "install.packages('knitr')"
-RUN Rscript -e "install.packages('gridExtra')"
-RUN Rscript -e "install.packages('ggpubr')"
+# Install the extra R packages we need.
+RUN install2.r --error --deps TRUE \
+	gridExtra \
+	ggpubr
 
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@egernst - when I saw a report image build take 1h18m for me, I thought I'd invest some effort whilst watching TV to figure out how to build on the base R docker images.
With the base image already downloaded (which is a rather hefty 2.6g or something), the image build for the report gen took more like 5m30 now.

metrics/report/report_dockerfile/Dockerfile
```diff
@@ -9,29 +9,28 @@
...
+LABEL DOCKERFILE_VERSION="1.1"
 
 # Without this some of the package installs stop to try and ask questions...
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && \
-	apt-get install -y tzdata r-base pandoc texlive
+# Install the extra doc processing parts we need for our Rmarkdown PDF flow.
+RUN apt-get update
+RUN apt-get install -y texlive-latex-base
```
jodh-intel: Running each `RUN` command separately is very clear, but I think every `RUN` gets converted to an image layer, so it's more efficient to chain all the commands you need together using oodles of `&&` and `\`'s.

grahamwhaley: It's true. The opposite side of that is that if one 'later' item changes, the whole layer has to be rebuilt.
When I'm developing the Dockerfiles, I do them as separate `RUN`s, as then generally changes take effect quicker. But, sure, this is normally, hopefully, a one-off thing, so I'll squash some of those together.

grahamwhaley: and pushed...

jodh-intel:
lgtm

metrics/report/report_dockerfile/Dockerfile
```diff
@@ -9,29 +9,29 @@
...
-# R packages get built from source, and can take some time, so split them
-# into individual RUNs so we can make some later modifications without maybe taking
-# the full image rebuild hit.
-RUN Rscript -e "install.packages('tidyverse')"
-RUN Rscript -e "install.packages('knitr')"
-RUN Rscript -e "install.packages('gridExtra')"
-RUN Rscript -e "install.packages('ggpubr')"
+# Install the extra doc processing parts we need for our Rmarkdown PDF flow.
+RUN apt-get update -qq && \
+  apt-get install -y \
```
chavafg: change spaces to tabs here and in the lines below...

grahamwhaley: :-)
 I was taking the style/inspiration from the `tidyverse` Dockerfile this is layed over - at https://github.com/rocker-org/rocker-versioned/blob/master/tidyverse/Dockerfile#L3-L21
Is there, or do we, have a standard for our Dockerfiles - tabs, 2-space-yaml-like etc.?
I'm happy to change if there is some standard we are adhering to?

chavafg: Specifically for Dockerfiles we haven't set a standard, but I think we have chosen to use tabs instead of spaces for all scripts. 
Although taking a look at other Dockerfiles in this repo, I see that they use 4-space indents... so now I am not sure.

grahamwhaley: arf! - thanks for checking @chavafg - let me see if I can go find any global guidance out on t'intinet....

chavafg: took a look at this page: https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/ but it does not says anything about indenting Dockerfiles and in their examples they use spaces, so I think this is ok ;-)

GabyCT:
yeiii @grahamwhaley 👍 as someone that uses this Dockerfile this is a great improvement

GabyCT:
lgtm

PR #865, opened 2018-10-26
ci: build/install: [RFC] enable versioned builds from ci scripts

grahamwhaley:
Enable the build and install of kata components of a specified version, or default to HEAD if none is specified.

### DIFF
--- a/.ci/install_agent.sh
+++ b/.ci/install_agent.sh
@@ -11,4 +11,4 @@ cidir=$(dirname "$0")
 
 source "${cidir}/lib.sh"
 
-build_and_install "github.com/kata-containers/agent"
+build_and_install "github.com/kata-containers/agent" "" "$KATA_VERSION"
--- a/.ci/install_proxy.sh
+++ b/.ci/install_proxy.sh
@@ -11,4 +11,4 @@ cidir=$(dirname "$0")
 
 source "${cidir}/lib.sh"
 
-build_and_install "github.com/kata-containers/proxy"
+build_and_install "github.com/kata-containers/proxy" "" "$KATA_VERSION"
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -33,7 +33,7 @@ runtime_config_path="${SYSCONFDIR}/kata-containers/configuration.toml"
 PKGDEFAULTSDIR="${SHAREDIR}/defaults/kata-containers"
 NEW_RUNTIME_CONFIG="${PKGDEFAULTSDIR}/configuration.toml"
 # Note: This will also install the config file.
-build_and_install "github.com/kata-containers/runtime"
+build_and_install "github.com/kata-containers/runtime" "" "$KATA_VERSION"
 
 # Check system supports running Kata Containers
 kata_runtime_path=$(command -v kata-runtime)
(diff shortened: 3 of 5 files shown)


chavafg:
I'll prefer that these changes go to https://github.com/kata-containers/tests/blob/master/.ci/resolve-kata-dependencies.sh, which is where we manage the versions for the kata components.

grahamwhaley:
@chavafg - I don't have a major issue trying to add the functionality to the `resolve` script, but...
- that is only ever called from the jenkins script before `setup.sh`, which is I guess OK if somebody would then need to run (either by hand or in an uber-script) both the resolve and the setup scripts to get a set version of Kata built and installed. Not a biggie, just 'one more call/script'.
- the function `build()` still by default passes `HEAD` to `build_version()` - and that does not look at all tied into the `resolve` script - any thoughts on if we'd have to do some work there as well? It feels like if we add version checkout into `resolve`, then the `build_version()` function is just going to undo that anyway. Maybe you were thinking we'd drop the `HEAD` defaulting and leave the trees at whatever version `resolve` put them to.

I might be missing something though - I admit I don't know who/when/how our depends-on stuff works for instance in the CI scripts.

chavafg:
ohh, you are right, the `resolve` script only runs on the `jenkins_job_build`.
Well, then I think we can try to do it here. We just need to make sure we don't end testing HEAD when there is a `Depends-on` tag

grahamwhaley:
OK, great :-) Let's figure out if I have more work to do here then...
I think for me the sticker is the question 'should we also update the kernel and image files to the specified version', which leads to 'can we even update the kernel and image files to the specified version?'.
@jcvenegas - what do you think - any not-overly-complex way to extend this VERSION stuff so we get a full set of kata components installed when we ask for a specific version?

raravena80:
@grahamwhaley ping (from your weekly Kata herder)

jodh-intel:
At this stage, I'd say we should be handling the image - since this PR updates `install_agent.sh`, assuming `$KATA_VERSION` is exported, won't this already dtrt if `install_kata_image.sh` gets called (since it calls `install_agent.sh`)?

I'd suggest opening a new issue to discuss the kernel issue. That's no going to be so trivial given the package version is the kernel version, not the Kata version:

```sh
$ dpkg -l|egrep -i "kata"|awk '{printf("%-25s %s\n", $2, $3)}'
kata-containers-image     1.4.0-10
kata-ksm-throttler        1.4.0.git+1212de2-12
kata-linux-container      4.14.67.17-11                # <-- odd one out :)
kata-proxy                1.4.0+git.e1856c2-11
kata-runtime              1.4.0+git.21f0059-15
kata-shim                 1.4.0+git.b02868b-9
```

caoruidong:
@grahamwhaley ping

grahamwhaley:
Actually, now we have the static binaries built and uploaded into our [releases area](https://github.com/kata-containers/runtime/releases), I'm prone to drop this PR - as I was only really using it to compare old versions using the metrics report tests.
OK, so, in theory maybe this modification could also choose by a specific 'date', which might be useful one day for bisecting - but for now I'm happy to close this.
OK with you @jodh-intel @chavafg (and everybody else..)

jodh-intel:
@grahamwhaley - idm either way - @chavafg ?

chavafg:
@grahamwhaley ok for me.

jodh-intel:
@grahamwhaley - what's the plan - forge ahead or close?

grahamwhaley:
I'll close it - I've not used it for a bit, and next time I'll try to use the static builds...

PR #870, opened 2018-10-29
CI: Specify tmpdir for running golang unit tests

### LINKED ISSUE #869: Specify output directory for "go test" to ensure tests don't fail
Running the `go-test.sh` script can fail when writing `testlog.txt` with the message:

```
testlog.txt: file already closed
```

See:

- https://github.com/kata-containers/runtime/issues/860

Fix by explicitly specifying an output directory.

jodh-intel:
It is possible for the `go-test.sh` script to fail when attemping to
write to a file as it can already be closed. This problem seems to
relate to the golang `testing` package itself so work around the
problem by specifying an explicit temporary directory for `testing` to
create files in.

Fixes #869.

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -246,8 +246,11 @@ main()
 
 	[ -z "$test_packages" ] && echo "INFO: no golang code to test" && exit 0
 
+	# directory to use for temporary files
+	golang_tmp=$(mktemp -d)
+
 	# KATA_GO_TEST_FLAGS can be set to change the flags passed to "go test".
-	go_test_flags=${KATA_GO_TEST_FLAGS:-"-v $race -timeout $timeout_value"}
+	go_test_flags=${KATA_GO_TEST_FLAGS:-"-v $race -timeout $timeout_value -outputdir \"${golang_tmp}\""}
 
 	if [ "$1" = "html-coverage" ]; then
 		test_html_coverage
@@ -256,6 +259,8 @@ main()
 	else
 		test_local
 	fi
+
+	[ -d "${golang_tmp}" ] && [ "${golang_tmp}" != "/" ] && rm -rf "${golang_tmp}"
 }
 
 main "$@"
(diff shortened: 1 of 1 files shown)


.ci/go-test.sh
```diff
@@ -10,6 +10,9 @@ set -e
 script_name=${0##*/}
 typeset -A long_options
 
+# directory to use for temporary files
+TMPDIR=${TMPDIR:-/tmp}
```
marcov: is `mktemp -d`  overkill :) ?

grahamwhaley: you mean specifically make a new tmp directory for the tests to use @marcov ? I could go for that.

jodh-intel:
@marcov - good call - branch updated!

marcov:
lgtm

.ci/go-test.sh
```diff
@@ -249,13 +246,21 @@ main()
...
+
 	if [ "$1" = "html-coverage" ]; then
 		test_html_coverage
 	elif [ "$run_coverage" = yes ]; then
 		test_coverage
 	else
 		test_local
 	fi
+
+	[ -d "${golang_tmp}" ] && [ "${golang_tmp}" != "/" ] && rm -rf "${golang_tmp}"
```
grahamwhaley: as the script has a `-e` set, would this be safer as a shell TRAP to catch the case where the tests fail?

jodh-intel: I did consider that but thought it would be better to leave the tmpfiles in the case of test failures as they might help us diagnose what went wrong.

bergwolf:
@jodh-intel What happens to the commit? I can't find the corresponding code in tests master and I'm seeing the exact same failure when running `make test` with runtime master tip.

bergwolf:
FYI, the root cause is how we flush coverage report in TestMain(). I fixed it in https://github.com/kata-containers/runtime/pull/2371

jodh-intel:
@bergwolf - The code is in `master` fwics? Good catch on 2371 though!

Issue #883, opened 2018-11-02
.ci/setup.sh and .ci/install_go.sh fails on ppc64le


sudeeshjohn:
Need to add ppc64le support for CI

PR #884, opened 2018-11-02
Adding ppc64le support for CI

sudeeshjohn:
Fixes: #883

Signed-off-by: Sudeesh John sudeeshjohn@in.ibm.com

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -64,7 +64,7 @@ enable_nested_virtualization() {
 			fi
 		fi
 		;;
-	aarch64)
+	aarch64 | ppc64le)
 		info "CI running in bare machine"
 		;;
 	*)
--- a/.ci/install_go.sh
+++ b/.ci/install_go.sh
@@ -104,6 +104,11 @@ case "$(arch)" in
 	"x86_64")
 		goarch=amd64
 		;;
+
+	"ppc64le")
+		goarch=ppc64le
+		;;
+		
 	"*")
 		die "Arch $(arch) not supported"
 		;;


nitkon:
Lgtm.

sudeeshjohn:
@grahamwhaley I guess this can group under "build" ?

grahamwhaley:
@sudeeshjohn yep, that would be fine.
I don't remember if checkcommits is going to complain that you don't have much 'body' to your commit - you may wish to add a single line explanation just above the 'Fixes' line. I know, with a small patch there is not much to say, but a one line summary, even if it reflects the same as the subject line, is useful when looking at git logs.
And.. .thanks for the contribution! :-)

sudeeshjohn:
updated commit message.

.ci/setup.sh
```diff
@@ -67,6 +67,9 @@ enable_nested_virtualization() {
 	aarch64)
 		info "CI running in bare machine"
 		;;
+	ppc64le)
```
marcov: just a hint, to avoid repeating the same `info` message you could do:
```bash
aarch64 | ppc64le)
		info "CI running in bare machine"
		;;
```

marcov:
Once we have this merge, it would be useful to have a bare metal server set up that could run tests for PPC, similarly with what we are doing for ARM (reference: kata-containers/ci#30)
lgtm

marcov:
..and thanks @sudeeshjohn for your contribution.

jodh-intel:
lgtm

nitkon:
@marcov: Yes, we are in the process of getting a bare metal Power server with a public IP for our CI. Are there any other specific requirements for a server that we may miss for CI? (going through the link you provided as well).

.ci/setup.sh
sudeeshjohn: @marcov thanks for the hint .. updated the patch accordingly

grahamwhaley:
There are no really hard/strict requirements that I'm aware of. You need the 'slave' set up enough to be a Jenkins slave, and be able to launch our [jenkins script](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh).

We should maybe clarify if this is a single slave machine that will be tied into our [kata jenkins master](http://jenkins.katacontainers.io/), of if this is maybe a single or pair of machines that will run a complete separate Jenkins instance just for ppc64.
Also, we should clarify if this will run the QA CI (tests), or the metrics CI (performance), or both. I think just the QA CI to begin with at least.

We currently run the x86 QA CI on oneshot cloud instances. That gives us a nice degree of isolation from a test run corrupting somehow a machine (it does happen). I wrote up some notes over [here](https://github.com/kata-containers/ci/issues/39) a while back. I also wrote up and coded a way to use oneshot VMs on the machines to alleviate this [here](https://github.com/kata-containers/ci/tree/master/VMs/metrics). Although I put that together for running the metrics CI, I believe that would work for QA CI as well.
The ARM folks could do neither of those (cloud instance or VMs), so have put together [some scripts](https://github.com/kata-containers/tests/blob/master/.ci/aarch64/clean_up_aarch64.sh) to try to clean up the machine back to a known state after each CI run. To note, I am about to utilise those on the x86 metrics machines, as I've come to the [conclusion](https://github.com/kata-containers/ci/issues/72) that the nested VMs create to much variance in the metrics results to be useful. They (the nested VMs) should still be useable for QA CI though (which are not performance sensitive).

I am also about to post some ansible scripts that can deploy a machine with enough bits to run either a jenkins slave directly, or setup the nested VMs. Not quite there yet - but if you could do with them, nudge me, and I could push a prelim branch you can stare at.

chavafg:
lgtm

Hi @nitkon, in addition to the good summary of @grahamwhaley, our current VMs are launched with 4 vCPUs and 16 GB of memory. I remember that CPU tests failed when the machine had less than 4 CPUs, so it would be good to at least match that number. I the case of the memory, I think 8GB has also worked in the past. Please let me know when you have the resource, I can help you set up as a slave.

PR #887, opened 2018-11-05
test: Add entropy test

### LINKED ISSUE #734: tests: Add test to verify container entropy.
After https://github.com/kata-containers/runtime/pull/676  is merged the containers entropy increase lets add a new tests to verify the entropy level.

GabyCT:
The main purpose of this test is to print the entropy level
inside the container after we installed the haveged package.
We need to verify that the entropy level is not < 1000
https://wiki.archlinux.org/index.php/Haveged.

Fixes #734

### DIFF
--- a/integration/entropy/entropy_test.bats
+++ b/integration/entropy/entropy_test.bats
@@ -1,0 +1,69 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# The main purpose of this test is to print
+# the entropy level inside the container after
+# we installed the haveged package. We need to
+# verify that the entropy level is not < 1000
+# https://wiki.archlinux.org/index.php/Haveged
+
+source /etc/os-release || source /usr/lib/os-release
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+# Environment variables
+IMAGE="busybox"
+# This the minimum entropy level produced
+# by haveged is 1000 see https://wiki.archlinux.org/index.php/Haveged
+# Less than 1000 could potentially slow down cryptographic
+# applications see https://www.suse.com/support/kb/doc/?id=7011351
+ENTROPY_LEVEL="1000"
+
+setup() {
+	clean_env
+
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+
+	# Check if haveged package is installed
+	check_package=$(which haveged | wc -l)
+
+	# Install haveged package if is not installed
+	if [ $check_package -eq 0 ]; then
+		case "$ID" in
+			ubuntu )
+				sudo -E apt install -y haveged
+				;;
+			fedora )
+				sudo -E dnf -y install haveged
+				;;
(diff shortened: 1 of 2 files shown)


integration/entropy/entropy_test.bats
```diff
@@ -0,0 +1,55 @@
...
+	check_package=$(which haveged | wc -l)
+
+	# Install haveged package if is not installed
+	if [ $check_package -eq 0 ]; then
+		if [ "$ID" == ubuntu ]; then
+			sudo -E apt install -y haveged
+		elif [ "$ID" == fedora ]; then
+			sudo -E dnf -y install haveged
+		elif [ "$ID" == centos ]; then
+			sudo -E yum install -y haveged
```
grahamwhaley: @marcov maybe you can suggest an addition here for SUSE? :-)

integration/entropy/entropy_test.bats
```diff
@@ -0,0 +1,55 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+source /etc/os-release || source /usr/lib/os-release
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+# Environment variables
+IMAGE="busybox"
+MIN_ENTROPY_LEVEL="1000"
```
grahamwhaley: We should probably have a comment stating why 1000.
We should probably also have a top level comment describing what this test is testing, and why.

One thing - with the kernels larger than, what was it, 4.14.36 maybe, won't the VM kernel wait (block) until it has enough entropy, and then start - in which case, if we are testing inside the container, what exactly are we testing for and expecting? Hence, the request for clarification above with the addition of some comments ;-)
/cc @jcvenegas

sboeuf:
@GabyCT please be more verbose in the commit message and the test itself, explaining the rationale behind the need for such level of entropy.

GabyCT:
@sboeuf and @grahamwhaley changes applied

sboeuf:
@GabyCT thanks for explaining what this test does. Also, sorry to be picky, but could you explain why we need to check for a certain amount of entropy, and why `1000` sounds like a good value?

integration/entropy/entropy_test.bats
```diff
@@ -0,0 +1,67 @@
...
+
+source /etc/os-release || source /usr/lib/os-release
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+# Environment variables
+IMAGE="busybox"
+# This the minimum entropy level produced
+# by haveged is 1000 see https://wiki.archlinux.org/index.php/Haveged
+# Less than 1000 could potentially slow down cryptographic
+# applications see https://www.suse.com/support/kb/doc/?id=7011351
```
GabyCT: @sboeuf , here I am putting an explanation about why this number and with their respective reference, thanks

sboeuf: Oh sorry I missed that...

grahamwhaley:
l g t m. If we apply those additions for SUSE @GabyCT , then I think we'll be good to go.

GabyCT:
@marcov changes applied thanks

marcov:
thanks @GabyCT 
lgtm

Issue #888, opened 2018-11-05
ci: bare metal: make the cleanup scripts more generic


grahamwhaley:
aarch64 has some [cleanup scripts](https://github.com/kata-containers/tests/blob/master/.ci/aarch64/clean_up_aarch64.sh) to ensure a bare metal CI slave is 'stable' after each run.
Now we are migrating the x86 metrics CI machines [back to using bare metal](https://github.com/kata-containers/ci/issues/72#issuecomment-434658732) to get more stable results, we can re-use most of these scripts across more than architecture.
Factor the scripts out to make them multi-arch useable.

PR #868, opened 2018-10-29
ci: detect the existence of yq before using filter scheme on aarch64

Pennyzct:
Since that filter schme on aarch64 bare-metal machine would fail if it is lack of yq, we should detect the existence of yq before going into and also install it if it is really missing.

Fixes: #867

Signed-off-by: Penny Zheng penny.zheng@arm.com

PR #889, opened 2018-11-05
CI: baremetal: make cleanup scripts more generic

grahamwhaley:
Enable the bare metal slave machine cleanup scripts to be
called from any arch (lifting out the good work done for the
ARM slave cleanups).

Fixes: #888

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -52,8 +52,15 @@ mkdir -p $(dirname "${kata_repo_dir}")
 
 # If CI running on bare-metal, a few clean-up work before walking into test repo
 if [ "${BAREMETAL}" == true ]; then
+	arch=$("${tests_repo_dir}/.ci/kata-arch.sh")
+	echo "Looking for baremetal cleanup script for arch ${arch}"
 	clean_up_script="${tests_repo_dir}/.ci/${arch}/clean_up_${arch}.sh"
-	[ -f "${clean_up_script}" ] && source "${clean_up_script}"
+	if [ -f "${clean_up_script}" ]; then
+		echo "Running baremetal cleanup script for arch ${arch}"
+		"${clean_up_script}"
+	else
+		echo "No baremetal cleanup script for arch ${arch}"
+	fi
 fi
 
 pushd "${kata_repo_dir}"
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 #
 # Copyright (c) 2017-2018 Intel Corporation
+# Copyright (c) 2018 ARM Limited
 #
 # SPDX-License-Identifier: Apache-2.0
 #
@@ -8,6 +9,10 @@
 set -e
 
 export KATA_RUNTIME=${KATA_RUNTIME:-kata-runtime}
+
+tests_repo="${tests_repo:-github.com/kata-containers/tests}"
+lib_script="${GOPATH}/src/${tests_repo}/lib/common.bash"
+source "${lib_script}"
 
 # If we fail for any reason a message will be displayed
 die() {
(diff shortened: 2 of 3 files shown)


grahamwhaley:
Pushing this mostly so we can see if I bust or not the ARM CI... @Pennyzct PTAL - I pulled a bunch of your code out into the lib code.
Let's see if it flies...
/test

grahamwhaley:
Of course, does the ARM CI only run on the proxy still??? @chavafg @Pennyzct , in which case this will be harder to test (I'll see if I can test it locally)...

sboeuf:
@grahamwhaley 
>Of course, does the ARM CI only run on the proxy still??? @chavafg @Pennyzct , in which case this will be harder to test (I'll see if I can test it locally)...

Please open an issue for this, as the ARM CI should be able to handle the configuration without proxy. And maybe assign it to @Pennyzct

chavafg:
@grahamwhaley it als runs here... as it is not stable yet, you can trigger it using:

/test-arm

chavafg:
btw, @Pennyzct is trying to fix some issues with ARM on #868

Pennyzct:
hi~ @grahamwhaley after importing `tests_repo`, you will notice `the missing arch` issue ╥﹏╥, which I found it during #868 and created an additional commit [here](https://github.com/kata-containers/tests/pull/868/commits/1f4a65f3eb305cb858d3716fb427ad6a23796eab) to fix it.
lgtm

grahamwhaley:
/test
/test-arm
even though we are pretty sure it will fail due to #868

grahamwhaley:
I'm going to mark this as DNM for the minute whilst I add and test the x86 cleanup call. I need to do a little 'dance' to do that as we don't have a bare metal x86 Jenkins instance in the main Jenkins setup right now... I'll do it locally, and then unblock when working/tested.

grahamwhaley:
Trying to test this locally, I found a couple of glitches I think...

1) @Pennyzct - in the existing jenkins script, where it tries to call out to the cleanup scripts:
```bash
# If CI running on bare-metal, a few clean-up work before walking into test repo
if [ "${BAREMETAL}" == true ]; then
	clean_up_script="${tests_repo_dir}/.ci/${arch}/clean_up_${arch}.sh"
	[ -f "${clean_up_script}" ] && source "${clean_up_script}"
fi
```

afaict, `${arch}` has not been set at this point - not in the Jenkins slave or job setups nor by this script or anything it `source`'s - so, I have a feeling the cleanup scripts may not be running at the moment. let me fix that in this PR (I'll find a way to get `arch` set nicely).

2) @chavafg - remember a long time ago we tried to make it so we could override which tests repo/branch we used from a CI job, so we could test things (like this) before merging them by setting up a special Jenkins job... I have a feeling we broke it at some point, but I've not quite figured out where. It seems that we stay on the repo, but the branch gets reset to master by something. That threw me for a bit, as my personal fork master branch is very old, so it took me ages to figure out why I was seeing messages that didn't exist in the code base :-)
Not a biggie, just something we should remember for the future.

I'll test and fix by hand.... updates coming here I expect then.

grahamwhaley:
phew, this is quite hard to test outside of a Jenkins that is tracking the HEAD....
I hand-ran the script for x86, and it looked to do the right thing.
@Pennyzct @chavafg - I updated the BAREMETAL test in the jenkins script itself to be a little more verbose, and to handle `arch` now. ptal at that bit at least.
I think the only real way we can test this is if we land it then I will tie in the bare metal packet.net metrics CI machine and spin a test run...
Given the two machines this will currently effect (ARM QA CI and x86 metrics CI) are not active anyway, I don't think this will break anything.

I'll drop the DNM and run a new set of CI on it anyway.

Pennyzct:
Hi~@grahamwhaley you may not notice that I have addressed `the missing arch issue` and fixed it in #868 in my last comment. 😄23333. 
> hi~ @grahamwhaley after importing tests_repo, you will notice the missing arch issue ╥﹏╥, which I found it during #868 and created an additional commit here to fix it😄
but never mind, since the CI is broken on arm due to the false timeline in container, I don't think that #868 could be merged in short time. let's fix this here. 
lgtm

grahamwhaley:
Ah, @Pennyzct , no, I didn't notice :-). We have a saying 'great minds think alike' :-)
In which case, sure, let's land this - just needs one more review, and we can skip the broken ARM CI as you say. thx!

grahamwhaley:
well, we had to merge this to really test it - and, it looks like there is something not quite right.
On the x86 bare metal metrics CI machines I am seeing:
```
16:21:20 Running baremetal cleanup script for arch x86_64
16:21:20 /home/jenkins/workspace/workspace/kata-metrics-agent-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/.ci/x86_64/clean_up_x86_64.sh: line 11: /home/jenkins/workspace/workspace/kata-metrics-agent-ubuntu-16-04-PR/go/src//.ci/lib.sh: No such file or directory
16:21:20 Build step 'Execute shell' marked build as failure
```

I'll go hunt that down and PR...

PR #890, opened 2018-11-06
test: Add test to exercise netmon

### LINKED ISSUE #728: write test to exercise netmon
When https://github.com/kata-containers/runtime/pull/534 lands, we'll need a test in this repo to exercise it.

GabyCT:
Enable netmon at the configuration.toml and then exercise docker network
connect and docker network disconnect.

Fixes #728

### DIFF
--- a/integration/netmon/netmon_test.bats
+++ b/integration/netmon/netmon_test.bats
@@ -1,0 +1,79 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+# Environment variables
+IMAGE="busybox"
+NETWORK_NAME="test"
+CONTAINER_NAME="containerA"
+PAYLOAD_ARGS="tail -f /dev/null"
+
+setup() {
+	clean_env
+
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+
+	extract_kata_env
+
+	# Enabling netmon at the configuration file
+	sudo sed -i -e 's/^#\(enable_netmon =\).*$/\1 true/g' ${RUNTIME_CONFIG_PATH}
+}
+
+@test "test netmon" {
+	# Create network
+	docker network create $NETWORK_NAME
+
+	# Run container
+	docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+
+	# Check the number of interfaces before network connect
+	before_interfaces=$(docker exec $CONTAINER_NAME ip addr show | awk '/inet.*brd/{print $NF}' | wc -l)
+
+	docker network connect $NETWORK_NAME $CONTAINER_NAME
+
+	# Check the number of interfaces after network connect
+	after_interfaces=$(docker exec $CONTAINER_NAME ip addr show | awk '/inet.*brd/{print $NF}' | wc -l)
+
+	# Compare interface numbers before and after network connect
+	[ "$after_interfaces" -gt "$before_interfaces" ]
+
+	# Check the ip address of the network
+	network_address=$(docker network inspect test | grep IPv4Address | cut -d '/' -f1 | cut -d '"' -f4)
+
+	# Check the ip address of the container of the new interface
+	container_address=$(docker inspect --format "{{.NetworkSettings.Networks.${NETWORK_NAME}.IPAddress}}" $CONTAINER_NAME)
+
+	# Check ip address of the network matches ip address of the container with the new interface
+	[ "$network_address" == "$container_address" ]
+
+	docker network disconnect $NETWORK_NAME $CONTAINER_NAME
+
+	# Check that we got back to the original number of interfaces
+	final_interfaces=$(docker exec $CONTAINER_NAME ip addr show | awk '/inet.*brd/{print $NF}' | wc -l)
+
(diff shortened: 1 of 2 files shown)


integration/netmon/netmon_test.bats
```diff
@@ -0,0 +1,55 @@
...
+	clean_env
+
+	extract_kata_env
+
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+
+	docker network disconnect $NETWORK_NAME $CONTAINER_NAME
```
grahamwhaley: You probably want to do this before calling `clean_env` above, as I suspect that deleted your container...

GabyCT: @grahamwhaley done :)

integration/netmon/netmon_test.bats
```diff
@@ -0,0 +1,54 @@
...
+	[ "$status" -eq 0 ]
+
+	extract_kata_env
+
+	# Enabling netmon at the configuration file
+	sed -i 's|#enable_netmon = true|enable_netmon = true|' ${RUNTIME_CONFIG_PATH}
+}
+
+@test "test netmon" {
+	docker network create $NETWORK_NAME
```
sboeuf: Nit: either you choose to put `docker network create` in `setup()` to match with `docker network rm` in `teardown()`, or you can have both `docker network create` and `docker network rm` in the main test function.

integration/netmon/netmon_test.bats
```diff
@@ -0,0 +1,54 @@
...
+	docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE sh -c $PAYLOAD_ARGS
+	docker network connect $NETWORK_NAME $CONTAINER_NAME
+	# Get ip address
+	run docker inspect --format "{{.NetworkSettings.IPAddress}}" $CONTAINER_NAME
```
sboeuf: Ok so this does not retrieve the right information, and it does check the IP address either.

I would suggest that we test the addition of the extra network worked by:
- `docker exec $CONTAINER_NAME ifconfig` or `docker exec $CONTAINER_NAME ip a` before we add the new network, this way, we know about what was there as a reference.
- `docker exec $CONTAINER_NAME ifconfig` or `docker exec $CONTAINER_NAME ip a` after the new network has been added. By comparing to the reference, we should find that a new interface has been added.
- Check the IP address of the new interface, and compare it to the IP of the network created
- Now remove the network by using `docker disconnect`
- Check `docker exec $CONTAINER_NAME ifconfig` or `docker exec $CONTAINER_NAME ip a` again, to verify you got back to the initial reference.

This would check that the network being added is properly added and shows up inside the container. This would also make sure the removal works properly.
And last thing we could add, would be to check the connectivity, by maybe pinging the bridge on the host.

GabyCT: @sboeuf thanks for the feedback, let me work on this

GabyCT: @sboeuf changes applied :)

integration/netmon/netmon_test.bats
```diff
@@ -0,0 +1,74 @@
...
+	[ "$after_interfaces" -gt "$before_interfaces" ]
+
+	# Check the ip address of the network
+	network_address=$(docker network inspect $NETWORK_NAME | grep "IPv4Address" | cut -d ':' -f2 | cut -d '"' -f2 | cut -d '/' -f1)
+
+	# Check the ip address of the container of the new interface
+	container_address=$(docker inspect $CONTAINER_NAME | grep IPAddress | tail -1 | cut -d ':' -f2 | cut -d '"' -f2)
+
+	# Check ip address of the network matches ip address of the container with the new interface
+	[ "$network_address" == "$container_address" ]
```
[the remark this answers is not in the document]
GabyCT: @sboeuf changes applied, I am only missing this last comment, I am kind of lost here, why we want to check the ip address of `docker exec $CONTAINER_NAME ip addr` with `docker inspect --format "{{.NetworkSettings.Networks.${NETWORK_NAME}.IPAddress}}" $CONTAINER_NAME`? Maybe, I did not understand well your comment :S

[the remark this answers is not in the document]
sboeuf: No nevermind, the changes you did are fine!

GabyCT:
The CI is not happy which is kind of strange as I can run the script using ubuntu and I have not failures. It is complaining at `[ "$after_interfaces" -gt "$before_interfaces" ]' failed` which it looks that there was not add a new interface

sboeuf:
@GabyCT add some `sleep` in the script to make sure you leave enough time for the interface to be created. The device is added right away, but the network interface creation can take some time, and that might be the reason why you don't see it.

integration/netmon/netmon_test.bats
```diff
@@ -0,0 +1,81 @@
...
+
+	# Run container
+	docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+
+	# Check the number of interfaces before network connect
+	before_interfaces=$(docker exec $CONTAINER_NAME ip addr show | awk '/inet.*brd/{print $NF}' | wc -l)
+
+	docker network connect $NETWORK_NAME $CONTAINER_NAME
+
+	sleep 15
```
chavafg: do we still need this sleep?

grahamwhaley: Really we should make it a sleepy poll loop with a long timeout - so, poll every 5s, and if after 60s or something the IP has not appeared, then fail. Relying on a hard-wired sleep is going to fail one day when we change some CI infra somewhere...

chavafg:
lgtm

Issue #891, opened 2018-11-07
test: ci: CI wont launch containers with k8s and containerd job.


jcvenegas:
The configuration file is only created for containerd integration test
but to when k8s runs.

PR #892, opened 2018-11-07
test: setup: configure containerd at install

jcvenegas:
containerd is configured at integration test, configure it at installation
to ensure it works for any other test, like k8s.

Fixes: #891


.ci/install_cri_containerd.sh
```diff
@@ -69,6 +70,15 @@ cat << EOF | sudo tee  /etc/systemd/system/kubelet.service.d/0-containerd.conf
 Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
```
egernst: I agree with your commit and note this is not part of your change, but while I'm here...  @jcvenegas @chavafg I question configuring kubernetes in a containerd installation script?  Seems that should be done elsewhere?

jcvenegas: Hey @egernst , It is done at k8s setup, depending on a job flag, will decide to pass the socket for containerd or cri-o.

jcvenegas: https://github.com/kata-containers/tests/blob/master/integration/kubernetes/init.sh#L52

Issue #897, opened 2018-11-08
ci: cleanup: need to pass `tests_repo` to sub-shell cleaner scripts


grahamwhaley:
#889 had a bug, where the sub-script in the arch dirs uses `${tests_repo)`, which is not exported from the caller. This is exposed as I changed the `source` of the script to a call, as sourcing the script felt a bit unclean.

Symptoms in the CI look like:
```
16:21:20 Running baremetal cleanup script for arch x86_64
16:21:20 /home/jenkins/workspace/workspace/kata-metrics-agent-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/.ci/x86_64/clean_up_x86_64.sh: line 11: /home/jenkins/workspace/workspace/kata-metrics-agent-ubuntu-16-04-PR/go/src//.ci/lib.sh: No such file or directory
16:21:20 Build step 'Execute shell' marked build as failure
```

Fix is probably to do something like:

```bash
                tests_repo="$tests_repo" "${clean_up_script}"
```

PR #898, opened 2018-11-08
ci: cleanup: pass tests_repo to cleanup scripts

grahamwhaley:
The baremetal cleanup scripts use `tests_repo` to find the funcs
to call, but that is not currently propagated from the caller.
Rather than export or source, let's pass it on the commandline
when calling the sub-scripts.

Fixes: #897

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -57,7 +57,7 @@ if [ "${BAREMETAL}" == true ]; then
 	clean_up_script="${tests_repo_dir}/.ci/${arch}/clean_up_${arch}.sh"
 	if [ -f "${clean_up_script}" ]; then
 		echo "Running baremetal cleanup script for arch ${arch}"
-		"${clean_up_script}"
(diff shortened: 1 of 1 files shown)


grahamwhaley:
A key item to watch here is how the metrics CI tests work...
/test

chavafg:
weird... still no metrics job, let me kick

/test

chavafg:
now it is there... not sure what happened.

grahamwhaley:
thx @chavafg I have a feeling (as noted on slack) maybe I have the jenkins metrics jobs not configured with the github api key secret, so maybe they are running in github api bandwidth reduced mode and effectively polling. If you confirm then I/we can fix :-)

grahamwhaley:
Yay @chavafg - metrics CI passed - I'm going to hand-retrigger that so we can check it will run twice in a row (as we are effectively testing the baremetal cleanup scripts....)

grahamwhaley:
metrics CI passed twice in a row - feeling pretty stable with this fix.

PR #914, opened 2018-11-19
ci: bare cleanup: never fail the cleanup

grahamwhaley:
We should probably never fail the cleanup script call, as we may be trying to clean
an unstable or part-cleaned system. Stop requiring `-e`, and add an `|| true` to the top
level call.

### DIFF
--- a/.ci/aarch64/clean_up_aarch64.sh
+++ b/.ci/aarch64/clean_up_aarch64.sh
@@ -4,10 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-set -e
-
 tests_repo="${tests_repo:-github.com/kata-containers/tests}"
 lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
 source "${lib_script}"
 
-gen_clean_arch
+gen_clean_arch || info "Arch cleanup scripts failed"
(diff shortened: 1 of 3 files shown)


grahamwhaley:
I want to see if this flies on the new metrics slave, so...
/test

.ci/aarch64/clean_up_aarch64.sh
```diff
@@ -4,10 +4,8 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
-set -e
-
 tests_repo="${tests_repo:-github.com/kata-containers/tests}"
 lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
 source "${lib_script}"
 
-gen_clean_arch
+gen_clean_arch || true
```
marcov: - Having both "`set -e` removed" AND "`|| true` added" is probably redundant?
- What about replacing the `||true` with something like:  `|| echo "WARNING: Arch clean up failed"`

grahamwhaley: Yeah, I like that - I'll do the rework.
The patches got the metrics slave running (it now fails as I have to set the check bounds for that machine....)

grahamwhaley: Updated. I used the available `info` func call to dump the info.

marcov:
lgtm

chavafg:
lgtm

PR #922, opened 2018-11-21
tests: Add K8s tests for persistent volumes

### LINKED ISSUE #921: Add K8S for persistent volumes
This test will create a persistent volume that is backed by physical storage and then we will create a persistent volume claim which will automatically bound to the persistent volume and finally we will create a pod that uses a persistent volume claim as a storage.

GabyCT:
This test will create a persistent volume that is backed
by physical storage and then we will create a persistent
volume claim which will automatically bound to the persistent
volume and finally we will create a pod that uses a persistent
volume claim as a storage.

Fixes #921

### DIFF
--- a/integration/kubernetes/k8s-volume.bats
+++ b/integration/kubernetes/k8s-volume.bats
@@ -1,0 +1,50 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+
+setup() {
+	export KUBECONFIG=/etc/kubernetes/admin.conf
+	pod_config_dir="${BATS_TEST_DIRNAME}/untrusted_workloads"
+	tmp_file=$(mktemp -d /tmp/data.XXXX)
+	msg="Hello from Kubernetes"
+	echo $msg > $tmp_file/index.html
+	pod_name="pv-pod"
+	# Define temporary file at yaml
+	sed -i "s|tmp_data|${tmp_file}|g" ${pod_config_dir}/pv-volume.yaml
+}
+
+@test "Create Persistent Volume" {
+	volume_name="pv-volume"
+	volume_claim="pv-claim"
+
+	# Create the persistent volume
+	sudo -E kubectl create -f "${pod_config_dir}/pv-volume.yaml"
+
+	# Check the persistent volume
+	sudo -E kubectl get pv $volume_name | grep Available
+
+	# Create the persistent volume claim
+	sudo -E kubectl create -f "${pod_config_dir}/volume-claim.yaml"
+
+	# Check the persistent volume claim
+	sudo -E kubectl get pvc $volume_claim | grep Bound
+
+	# Create pod
+	sudo -E kubectl create -f "${pod_config_dir}/pv-pod.yaml"
+
+	# Check pod creation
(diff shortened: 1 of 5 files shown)


GabyCT:
@chavafg changes applied

integration/kubernetes/k8s-volume.bats
```diff
@@ -0,0 +1,49 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+
+setup() {
+	export KUBECONFIG=/etc/kubernetes/admin.conf
+	pod_config_dir="${BATS_TEST_DIRNAME}/untrusted_workloads"
+	path="/mnt/data"
```
grahamwhaley: Might be better (safer from a whole system point of view) if we made this path a mktmpdir rather than hard wire it to a subdir of /mnt

integration/kubernetes/k8s-volume.bats
```diff
@@ -0,0 +1,49 @@
...
+	# Check pod creation
+	sudo -E kubectl wait --for=condition=Ready pod "$pod_name"
+
+	cmd="apt-get -y update && apt-get install -y curl && curl localhost"
+	sudo -E kubectl exec $pod_name -- sh -c "$cmd" | grep "$msg"
+}
+
+teardown() {
+	sudo -E kubectl delete pod "$pod_name"
+	sudo rm -rf $path
```
grahamwhaley: I'd say we should do the delete in a TRAP, so we cleanup on fail - but, maybe bats does that for the teardown func anyway?, and/or maybe bat doesn't support TRAP.

integration/kubernetes/untrusted_workloads/pod-pv-volume.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    type: local
+spec:
+  storageClassName: manual
+  capacity:
+    storage: 10Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: "/mnt/data"
```
grahamwhaley: and sure, if we used a tmpdir we'd have to dynamically adjust this file I guess.

integration/kubernetes/k8s-volume.bats
chavafg: `teardown` will be executed even if the test fails, so we are fine here.

integration/kubernetes/untrusted_workloads/pod-pv-volume.yaml
```diff
@@ -0,0 +1,17 @@
+kind: PersistentVolume
+apiVersion: v1
+metadata:
+  name: pv-volume
+  annotations:
```
chavafg: these annotations are not needed here as this is a persistent volume, not a pod. 
same comment for the `pod-volume-claim.yaml`

GabyCT:
@chavafg , I tried it however, I am getting the following error
```
```
Here it is the log
```

GabyCT:
@chavafg changes applied

jodh-intel:
lgtm - seems that the ARM CI is "stuck" and I don't have the super-powers to unstick it.

/cc @Pennyzct, @chavafg.

chavafg:
I re triggered, but the status didn't reflect here.
http://jenkins.katacontainers.io/job/kata-containers-tests-ARM-18.04-PR/91/ 
The build passed, merging this change

PR #938, opened 2018-11-26
test: Docker integration test for disable networking

### LINKED ISSUE #932: We need a test for disable_new_netns=true
Currently, it appears to be broken: https://github.com/kata-containers/runtime/issues/935.

GabyCT:
We are disabling networking at the configuration.toml as well as running
a container with the flag of --net=none.

Fixes #932

### DIFF
--- a/integration/network/disable_net/net_none.bats
+++ b/integration/network/disable_net/net_none.bats
@@ -1,0 +1,63 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../../lib/common.bash"
+
+setup () {
+	clean_env
+
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+}
+
+@test "Disable net" {
+	IMAGE="busybox"
+	PAYLOAD="tail -f /dev/null"
+	NAME="test"
+
+	extract_kata_env
+
+	# Get the name of the network name at the configuration.toml
+	NETWORK_NAME=$(grep -E "internetworking_model=" ${RUNTIME_CONFIG_PATH} | head -1 | cut -d '"' -f2)
+
+	# Disable the network
+	sudo sed -i 's/#disable_new_netns = true/disable_new_netns = true/g' ${RUNTIME_CONFIG_PATH}
+	sudo sed -i 's/internetworking_model=".*"/internetworking_model="none"/g' ${RUNTIME_CONFIG_PATH}
+
+	# Run a container without network
+	docker run -d --runtime=${RUNTIME} --name=${NAME} --net=none ${IMAGE} ${PAYLOAD}
+
+	# Check namespaces of host init daemon with no network
+	no_network_ns=$(sudo stat -L -c "%i" /proc/1/ns/net)
+
+	# Check namespaces of the processes with no network
+	general_processes=( ${PROXY_PATH} ${HYPERVISOR_PATH} ${SHIM_PATH} )
+	for i in "${general_processes[@]}"; do
+		process_pid=$(pgrep -f "$i")
+		process_ns=$(sudo stat -L -c "%i" /proc/$process_pid/ns/net)
+		# Compare namespace of host init daemon is equal to namespace of the process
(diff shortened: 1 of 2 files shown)


integration/network/disable_net/net_none.bats
```diff
@@ -0,0 +1,52 @@
...
+
+	# Get the name of the network name at the configuration.toml
+	NETWORK_NAME=$(grep -E "internetworking_model=" /usr/share/defaults/kata-containers/configuration.toml | head -1 | cut -d '"' -f2)
+
+	# Disable the network
+	sudo sed -i 's/#disable_new_netns = true/disable_new_netns = true/g' ${RUNTIME_CONFIG_PATH}
+	sudo sed -i 's/internetworking_model=".*"/internetworking_model="none"/g' ${RUNTIME_CONFIG_PATH}
+
+	# Run a container without network
+	run docker run -d --rm --runtime=${RUNTIME} ${IMAGE} ${PAYLOAD}
```
grahamwhaley: From the commit - did you intend to run this command with '--net=none' as well? Maybe it is implicit somewhere/somehow?

jodh-intel:
Hi @GabyCT - this test proves we can still create containers with `disable_new_netns = true`. However it doesn't check that the components have the correct (host) namespaces.

You could test this as follows for example:

- Start a long-running container:
  ```
  $ sudo docker run -it --runtime kata-runtime --net=none busybox tail -f /dev/null
  ```
- Get the namespaces for the host init daemon. For example, the net ns:
  ```
  $ sudo ls -l /proc/1/ns|awk '{print $NF}'|grep net:
  ```
- Do the same for `kata-proxy`, `kata-shim` and `qemu` and check that their net ns is the same as the host init's net ns:
  ```
  $ sudo ls -l /proc/$pid/ns|awk '{print $NF}'|grep net:
  ```

Bonus points for running the namespace tests twice - once with `disable_new_netns = true` to ensure the namespaces are the same as the hosts, and once more with `disable_new_netns = false` to assert that they are *not* the same as the namespaces in the host namespace context.

integration/network/disable_net/net_none.bats
```diff
@@ -0,0 +1,63 @@
...
+
+@test "Disable net" {
+	IMAGE="busybox"
+	PAYLOAD="tail -f /dev/null"
+	NAME="test"
+
+	extract_kata_env
+
+	# Get the name of the network name at the configuration.toml
+	NETWORK_NAME=$(grep -E "internetworking_model=" /usr/share/defaults/kata-containers/configuration.toml | head -1 | cut -d '"' -f2)
```
devimc: might be here you should use `$RUNTIME_CONFIG_PATH`

integration/network/disable_net/net_none.bats
```diff
@@ -0,0 +1,63 @@
...
+
+	# Disable the network
+	sudo sed -i 's/#disable_new_netns = true/disable_new_netns = true/g' ${RUNTIME_CONFIG_PATH}
+	sudo sed -i 's/internetworking_model=".*"/internetworking_model="none"/g' ${RUNTIME_CONFIG_PATH}
+
+	# Run a container without network
+	docker run -d --runtime=${RUNTIME} --name=${NAME} --net=none ${IMAGE} ${PAYLOAD}
+
+	# Check namespaces of host init daemon with no network
+	no_network_ns=$(sudo ls -l /proc/1/ns|awk '{print $NF}'|grep net:|cut -d '[' -f2|cut -d ']' -f1)
```
devimc: `sudo stat -L -c "%i" /proc/1/ns/net`

it's the same :wink:

integration/network/disable_net/net_none.bats
```diff
@@ -0,0 +1,63 @@
...
+
+	# Check namespaces of the processes with no network
+	general_processes=( ${PROXY_PATH} ${HYPERVISOR_PATH} ${SHIM_PATH} )
+	for i in "${general_processes[@]}"; do
+		process_pid=$(pgrep -f "$i")
+		process_ns=$(sudo ls -l /proc/$process_pid/ns|awk '{print $NF}'|grep net:|cut -d '[' -f2|cut -d ']' -f1)
```
devimc: `sudo stat -L -c "%i" /proc/$process_pid/ns/net`

GabyCT:
@devimc changes applied

jodh-intel:
Nice!

lgtm

PR #946, opened 2018-11-27
integration/cri-o: skip crio test 'ctr list filtering'

### LINKED ISSUE #945: Skip crio test 'ctr list filtering'
Skip crio test 'ctr list filtering' to allow https://github.com/kata-containers/runtime/pull/847 lands

see https://github.com/kata-containers/runtime/issues/944

devimc:
Skip crio test 'ctr list filtering' to allow
kata-containers/runtime#847 lands

fixes #945

### DIFF
--- a/integration/cri-o/crio_skip_tests.sh
+++ b/integration/cri-o/crio_skip_tests.sh
@@ -10,4 +10,5 @@
 declare -a skipCRIOTests=(
 'test "ctr oom"'
 'test "ulimits"'
+'test "ctr list filtering"'
 );


grahamwhaley:
For ref (as it is many hops to find) https://github.com/kata-containers/runtime/issues/944
@devimc - is there any way we can skip this *only* for devicemapper based CIs or similar? Also, is there an upstream k8s/crio issue open for this anywhere?
Do we have a plan to ever turn this back on? I don't like the idea of disabling a test just to land a single PR etc.

sboeuf:
@devimc 
>Skip crio test 'ctr list filtering' to allow kata-containers/runtime#847 lands

Maybe I'm missing something here, but why would we skip a test if it was working fine before and without the PR kata-containers/runtime#847?

devimc:
@grahamwhaley 
> is there any way we can skip this only for devicemapper based CIs or similar? 

No, but I can implemented

> Also, is there an upstream k8s/crio issue open for this anywhere?

No, I spent a week debugging the issue, and seems like the problem is with the mount namespace and its propagation, might be the problem is the test

> Do we have a plan to ever turn this back on? I don't like the idea of disabling a test just to land a single 
PR etc.

Yes, https://github.com/kata-containers/runtime/issues/944, and I'll be working on it

@sboeuf 

> why would we skip a test if it was working fine before and without the PR kata-containers/runtime#847?

I understand your concerns, it's just one test that fails with devicemapper (with overlay it works correctly), I'd like to see 847 landing and then fix this specific test

CI in https://github.com/kata-containers/runtime/pull/847 is green, but first we need this change

sboeuf:
@devimc 
>I understand your concerns, it's just one test that fails with devicemapper (with overlay it works correctly), I'd like to see 847 landing and then fix this specific test

Well if it's only one test to fix, then let's fix it first right? There's no rush having kata-containers/runtime#847 merged since it needs to be discussed with the Kata community during the AC meeting. This gives you some time to fix the test in the meantime.

jodh-intel:
Ping @devimc - any update?

devimc:
Hi @jodh-intel nop :smile:

ydjainopensource:
Ping @devimc - any update?

PR #948, opened 2018-11-27
ci: install kata-agent as init

### LINKED ISSUE #947: Install kata-agent as init
Install kata-agent as init to rid of systemd

devimc:
install kata-agent as init to rid of systemd

fixes #947

### DIFF
--- a/.ci/install_kata_image.sh
+++ b/.ci/install_kata_image.sh
@@ -84,7 +84,8 @@ update_agent() {
 	"${IMG_MOUNT_DIR}/usr/bin/kata-agent" --version
 
 	echo "Install new agent"
-	sudo -E PATH="$PATH" bash -c "make install DESTDIR=$IMG_MOUNT_DIR"
+	sudo -E PATH="$PATH" bash -c "make install INIT=yes DESTDIR=$IMG_MOUNT_DIR"
+	sudo -E PATH="$PATH" bash -c "ln -sf kata-agent $IMG_MOUNT_DIR/sbin/init"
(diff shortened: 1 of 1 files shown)


.ci/install_kata_image.sh
```diff
@@ -84,7 +84,8 @@ update_agent() {
 	"${IMG_MOUNT_DIR}/usr/bin/kata-agent" --version
 
 	echo "Install new agent"
-	sudo -E PATH="$PATH" bash -c "make install DESTDIR=$IMG_MOUNT_DIR"
+	sudo -E PATH="$PATH" bash -c "make install INIT=yes DESTDIR=$IMG_MOUNT_DIR"
+	sudo -E PATH="$PATH" bash -c "ln -sf kata-agent $IMG_MOUNT_DIR/sbin/init"
```
jodh-intel: As mentioned on https://github.com/kata-containers/osbuilder/pull/210, doesn't this need a `chroot` call?

jodh-intel:
+1. Also, I think it might be helpful to [raise a top-level issue }(https://github.com/kata-containers/kata-containers/issues/new) with links to the work being done in the other repos for this change.

Note that if we did switch entirely to using the agent as the init daemon, that would break the [agent tracing work](https://github.com/kata-containers/agent/pull/415) as currently that only works if the agent is *not* the init daemon.

devimc:
@jodh-intel 

> Note that if we did switch entirely to using the agent as the init daemon, that would break the agent tracing work as currently that only works if the agent is not the init daemon.

not necessary, because users can still generate images with systemd, 

@grahamwhaley  I haven't documented anything yet, first I'd like to discuss this in the next arch forum

Issue #903, opened 2018-11-12
Search which PR introduce memory regression in K8S


GabyCT:
Research which PR introduce a memory constraint regression in K8s (https://github.com/kata-containers/tests/pull/805)

GabyCT:
This already closed

PR #805, opened 2018-10-08
test: Update k8s memory constraints tests

GabyCT:
Update the value for allocated size memory for the k8s memory constraint test.  The amount of memory that can be allocated has decreased from 150M to 100M as it looks that the kernel is taking more memory.

Fixes #680

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

Issue #680, opened 2018-08-30
Investigate k8s test issue: not ok 2 Running within memory constraints


chavafg:
This k8s test have been failing sometimes on the CI this week. We need to investigate...
```
not ok 2 Running within memory constraints
# (from function `waitForProcess' in file ../../.ci/lib.sh, line 164,
#  in test file k8s-memory.bats, line 47)
#   `waitForProcess "$wait_time" "$sleep_time" "$pod_status_cmd"' failed
# pod "memory-test" created
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
# Flag --show-all has been deprecated, will be removed in an upcoming release
Makefile:67: recipe for target 'kubernetes' failed
```

devimc:
is this a random issue?

chavafg:
yes, it is random and started to happen this week I think.

WeiZhang555:
But it happens a lot, I see a lot failures in different PR
[...]

PR #954, opened 2018-11-28
test: Update allocated memory size value for K8s test

### LINKED ISSUE #953: Update allocated memory size value for K8s test
Update allocated memory size value for memory constraints K8s test.

GabyCT:
Revert the memory value to the original value (https://github.com/kata-containers/tests/pull/534).
The main purpose of this is to fixed the bug (https://github.com/kata-containers/tests/issues/903)
where we could not allocated a memory size more than 100M.

Fixes #953

### DIFF
--- a/integration/kubernetes/k8s-memory.bats
+++ b/integration/kubernetes/k8s-memory.bats
@@ -36,7 +36,7 @@ setup() {
 
 @test "Running within memory constraints" {
 	memory_limit_size="200Mi"
-	allocated_size="100M"
+	allocated_size="150M"
 	# Create test .yaml
         sed \
             -e "s/\${memory_size}/${memory_limit_size}/" \
--- a/integration/kubernetes/run_kubernetes_tests.sh
+++ b/integration/kubernetes/run_kubernetes_tests.sh
@@ -33,12 +33,12 @@ bats k8s-pid-ns.bats
 bats k8s-cpu-ns.bats
 bats k8s-parallel.bats
 bats k8s-security-context.bats
-bats k8s-memory.bats
 bats k8s-liveness-probes.bats
 bats k8s-attach-handlers.bats
 bats k8s-qos-pods.bats
 bats k8s-pod-quota.bats
 bats k8s-volume.bats
 bats k8s-projected-volume.bats
+bats k8s-memory.bats
 ./cleanup_env.sh
 popd


grahamwhaley:
Hi @GabyCT I know you were investigating if/where we had a regression on this - looks like you found something? can you add details into probably both the Issue and the Commit so we can contemplate.
thx

GabyCT:
@grahamwhaley , yes I have been looking this, I tested around 12 commits (starting from Agust 30, which was the day this test start failing) for the runtime (using kubernetes version 1.10.5), however I could not find the regression. Then, I moved to the agent, where I tested also around 12 commits (build the agent using osbuilder), however, there I updated the kubernetes version to 1.12) and again I could not find the regression. Then, I checked with the latest agent, runtime and kubernetes version and I noticed that the test was working. I do not know if it was a combination of kubernetes version with maybe something else.

jodh-intel:
Hi @GabyCT - I'm not quite clear - Please could you summarise what is causing us to need to allocate 50M more in this test?

GabyCT:
@jodh-intel , I could not find the commit which introduce the regression (handle 100M), I just notice that with the latest changes we have 150M of allocated memory (size that we were using before)

ydjainopensource:
@GabyCT I don't understand. do you wish to say that we introduced an allocation somewhere which increased our memory footprint by 50MB and that s the reason why we need to change it here or are you saying that the newer version of k8ns needs this increase?

GabyCT:
@ydjainopensource , I am updating the value for the allocation of the memory before like we had before we had the regression. I noticed that with the latest agent, runtime and kubernetes version we can come back to the original amount of memory that we had before.

jodh-intel:
Hi @GabyCT - I'm still not totally clear what's going on here. It would be **really** helpful if you put details into the issue and the commit itself explaining this change as making seemingly arbitrary changes like this with no explanation is confusing to those (like me) who don't know the history of this file.

From what I can find out quickly, when you originally wrote this test, the original values was 150M:

- https://github.com/kata-containers/tests/pull/534

But then something happened and the value was modified (please provide references to what happened including issue(s)/PR(s)).

This PR is therefore reverting the value back to the original one I think?

GabyCT:
@jodh-intel , yes this PR is reverting the original value that we have from this test. The failures that we start seing was addressed in this issue https://github.com/kata-containers/tests/issues/680. As I mention above, we still do not know which PR introduced the regression as the runtime and agent were verified and we could not find anything.

jodh-intel:
I'm not very comfortable with the idea of making a change if we don't understand exactly why we're doing it.

At the very least though, please can you update the commit message to provide further details - it should explain:

- That the memory value is being reverted back to its original value of `150M` as "somehow", the bug(?) that was stopping 150M from being allocated has now been fixed / gone away.
- Provide a reference to the issue/PR where it was changed to `100M` (#805).

GabyCT:
@jodh-intel changes applied

jodh-intel:
CI failing due to the OSX brew bash issue here too (/cc @raravena80).

Thanks @GabyCT. Let's keep an eagle eye on this if we ever need to change it again.

lgtm

jodh-intel:
We need another ack over here @kata-containers/tests.

PR #955, opened 2018-11-29
metrics: Add sleep before gathering memory inside the container

### LINKED ISSUE #931: metrics: mem-in-container: may need to nap before measure
As found over on https://github.com/kata-containers/runtime/issues/933, now we hotplug memory into the container, that takes a short time to appear to the system.
Therefore, in the 'memory in container' test, we may want to add a short nap/delay before we take the measure. Cmd is [here](https://github.com/kata-containers/tests/blob/master/metrics/density/memory_usage_inside_container.sh#L21)

GabyCT:
We need a sleep before taking the measurement of the memory inside
of a container. This change is necessary as the kernel takes time
to hotplug the memory.

Fixes #931

### DIFF
--- a/metrics/density/memory_usage_inside_container.sh
+++ b/metrics/density/memory_usage_inside_container.sh
@@ -18,7 +18,7 @@ TEST_NAME="memory footprint inside container"
 VERSIONS_FILE="${SCRIPT_PATH}/../../versions.yaml"
 ALPINE_VERSION=$("${GOPATH}/bin/yq" read "$VERSIONS_FILE" "docker_images.alpine.version")
 IMAGE="alpine:$ALPINE_VERSION"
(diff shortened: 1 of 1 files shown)


GabyCT:
@chavafg something got wrong in the CI, I already rebuild this PR and it passed in `jenkins-ci-fedora-27-vsocks`, however, it is not updated the latest status and it is showing as failed

chavafg:
right, I see that http://jenkins.katacontainers.io/job/kata-containers-tests-vsock-fedora-27-PR/522/ is the latest status and it passed.

chavafg:
ohh, this happened because we no longer have fedora 27 job, now it is 28. you can see it in the list. So this is ready to be merged.

PR #957, opened 2018-11-29
functional: Add test checking for container state

### LINKED ISSUE #956: Check kata-runtime status after the container terminated
We need to make sure the `kata-runtime state` command will return the value `stopped` after a simple workload like `sleep 1` returns after 1 second, instead of getting `running` followed by `stopped` if we reiterate the command `kata-runtime state`.

This is related to the fix https://github.com/kata-containers/runtime/pull/950

sboeuf:
Following an issue that we found with the kata-runtime regarding
a wrong state being reported after the container workload terminated,
this patch introduces a test that runs a simple container workload
and validate that the value obtained after a few seconds is "stopped".

Previously, this test would have failed as we would have retrieved
"running", which is a wrong value in this case.

Fixes #956

### DIFF
--- a/bundle.go
+++ b/bundle.go
@@ -60,6 +60,9 @@ func NewBundle(workload []string) (*Bundle, error) {
 		return nil, err
 	}
 
+	// By default, let's not create a terminal
+	config.Process.Terminal = false
+
 	config.Process.Args = workload
 
 	bundle := &Bundle{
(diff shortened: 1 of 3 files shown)


sboeuf:
@jodh-intel @GabyCT PTAL

chavafg:
the arm CI is currently disabled and the fedora28 job failed on `apt-get update` test, maybe a network issue, sent rebuild.

chavafg:
lgtm

Issue #959, opened 2018-12-04
kernel: add dev-kernel build process for CI environment on aarch64


Pennyzct:
when $DEV has been set as true, it means that the whole kata-ecosystem on aarch64 (mainly including kernel, qemu, etc) is built for development use, not stable. Here is about adding the dev-kernel(4.20-rc3+) build process on aarch64.

PR #269, opened 2018-04-30
ci: use fedora as base OS for osbuilder

devimc:
Use a more reliable distro as base OS for osbuilder

fixes #

Signed-off-by: Julio Montes <julio.montes@intel.com>

PR #960, opened 2018-12-04
kernel: add dev-kernel build process for CI environment on aarch64

Pennyzct:
when $DEV has been set as true, it means that the whole kata-ecosystem on aarch64 (mainly including kernel, qemu, etc) is built for development use, not stable. Here is about adding the dev-kernel(4.20-rc3+)
build process on aarch64.

Fixes: #959

Signed-off-by: Wei Chen <Wei.Chen@arm.com>

### DIFF
--- a/.ci/install_kata_kernel.sh
+++ b/.ci/install_kata_kernel.sh
@@ -40,7 +40,11 @@ download_repo() {
 }
 
 get_current_kernel_version() {
-	kernel_version=$(get_version "assets.kernel.version")
+	if [ "$DEV" == false ]; then
+		kernel_version=$(get_version "assets.kernel.version")
+	else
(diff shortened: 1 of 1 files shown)


Pennyzct:
This is related to [#973](https://github.com/kata-containers/runtime/pull/973) in runtime repo and [#269](https://github.com/kata-containers/packaging/pull/269) packaging repo.

jodh-intel:
I wonder if we should make it `$KATA_DEV` for consistency with:

- https://github.com/kata-containers/tests#developer-mode

Pennyzct:
Since we all agreed to drop the `dev/` solution and switched to the stable 4.19.x. This pr is useless and need to be closed. :)

Issue #963, opened 2018-12-04
cri-o: tests on centos started to fail.


chavafg:
cri-o tests on centos are failing on:
```
helpers.bash: line 128: /tmp/jenkins/workspace/kata-containers-agent-centos-7-4-PR/go/src/github.com/kubernetes-sigs/cri-o/test/copyimg/copyimg: No such file or directory
Error pulling quay.io/crio/redis
make: *** [crio] Error 1
```
Seems that this started happening yesterday afternoon.

chavafg:
taking a look at the log, I see a failrue when cri-o is being built and installed
```
Installing CRI-O
rm -f "/tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/.gopathok"
rm -rf _output
rm -f docs/*.5 docs/*.8
rm -fr test/testdata/redis-image
find . -name \*~ -delete
find . -name \#\* -delete
rm -f bin/crio
rm -f bin/crio.cross.*
make -C conmon clean
make[1]: Entering directory `/tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/src/github.com/kubernetes-sigs/cri-o/conmon'
rm -f conmon.o cmsg.o ../bin/conmon
rm -f config.h
make[1]: Leaving directory `/tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/src/github.com/kubernetes-sigs/cri-o/conmon'
make -C pause clean
make[1]: Entering directory `/tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/src/github.com/kubernetes-sigs/cri-o/pause'
rm -f pause.o ../bin/pause
make[1]: Leaving directory `/tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/src/github.com/kubernetes-sigs/cri-o/pause'
rm -f test/bin2img/bin2img
rm -f test/copyimg/copyimg
rm -f test/checkseccomp/checkseccomp
touch "/tmp/jenkins/workspace/kata-containers-tests-centos-7-4-PR/go/.gopathok"
go build -i -ldflags '-s -w -X main.gitCommit="fa540c8e806d28c2cbcd157bdf8acf2b20990ab6-dirty" -X main.buildInfo=1543868920' -tags "exclude_graphdriver_devicemapper libdm_no_deferred_remove containers_image_ostree_stub" -o bin/crio github.com/kubernetes-sigs/cri-o/cmd/crio
[...]

PR #965, opened 2018-12-04
centos: fix cri-o issues

chavafg:
the centos image that is used in the vexxhost cloud was updated to 7.6, which is good, but the go-md2man package is no longer available, this is why the cri-o tests started to fail. 
Instead of getting the `go-md2man` from the centos repositories, build and install it from sources.
Also, add yum configuration to fail when a package is not available. 

Fixes: #963.


GabyCT:
lgtm

PR #977, opened 2018-12-06
test: K8s test to distribute credentials using secrets

### LINKED ISSUE #976: Add K8s that distribute credentials using secrets
Add a test that will distribute credentials securely using secrets

GabyCT:
This test will distribute credentials securely using secrets.

Fixes #976

### DIFF
--- a/integration/kubernetes/k8s-credentials-secrets.bats
+++ b/integration/kubernetes/k8s-credentials-secrets.bats
@@ -1,0 +1,56 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+
+setup() {
+	export KUBECONFIG=/etc/kubernetes/admin.conf
+	if sudo -E kubectl get runtimeclass | grep kata; then
+		pod_config_dir="${BATS_TEST_DIRNAME}/runtimeclass_workloads"
+	else
+		pod_config_dir="${BATS_TEST_DIRNAME}/untrusted_workloads"
+	fi
+}
+
+@test "Credentials using secrets" {
+	secret_name="test-secret"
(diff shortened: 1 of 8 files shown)


integration/kubernetes/untrusted_workloads/inyect_secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: Secret
+metadata:
+  name: test-secret
+data:
+  username: bXktYXBw
```
jodh-intel: - Is this value a username or a password? I'm confused. Maybe it's just a secret ;)
- Should this file be called `inject_secret.yaml`?

integration/kubernetes/k8s-credentials-secrets.bats
```diff
@@ -0,0 +1,50 @@
...
+
+	# Create a pod that has acces to the secret through a volume
+	sudo -E kubectl create -f "${pod_config_dir}/pod-secret.yaml"
+
+	# Check pod creation
+	sudo -E kubectl wait --for=condition=Ready pod "$pod_name"
+
+	# List the files
+	cmd="ls /tmp/secret-volume"
+	sudo -E kubectl exec $pod_name -- sh -c "$cmd" | grep -E "password|username"
```
jodh-intel: I don't really understand why alternation is being used here and below in the `grep` call? Shouldn't we be checking for both username *and* password (rather than accepting either)?

jodh-intel: Yes, but that `grep` will succeed in **all** the following scenarios:

1. if it **only** finds `password`.
1. if it **only** finds `username`.
1. if it find `username` **and** `password`.

But that's not what we want aiui - we need to guarantee point (3) above happens right?

GabyCT:
@jodh-intel and @chavafg changes applied

Issue #982, opened 2018-12-07
add the option to use runtimeclass feature from k8s 


chavafg:
In addition to the untrusted/trusted tags, with k8s 1.12 we can use `runtimeClass` feature

PR #983, opened 2018-12-07
k8s: Add runtimeclass configuration 

chavafg:
We will now be able to use the runtimeClass feature.
In addition, the old trusted/untrusted annotations 
will also work.

To configure cri-o and k8s to use the new runtimeClass feature,
simply set `use_runtime_class=true` as an environment variable.

For the tests, use runtimeclass when the cluster is configured with 
this feature. If not, continue using old trusted/untrusted annotations.

Fixes: #982.


GabyCT:
lgtm

integration/kubernetes/runtimeclass_workloads/lifecycle-events.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: handlers
+spec:
+  runtimeClassName: kata
+  containers:
+  - name: handlers-container
+    image: nginx
+    lifecycle:
+      postStart:
+        exec:
+          command: ["/bin/sh", "-c", "echo Hello from the postStart handler > /usr/share/message"]
```
jodh-intel: ooi, why `/usr/share/` ?

integration/kubernetes/runtimeclass_workloads/busybox-pod.yaml
```diff
@@ -0,0 +1,25 @@
...
+  containers:
+  - name: first-test-container
+    image: busybox
+    command:
+        - sleep
+        - "30"
+  - name: second-test-container
+    image: busybox
+    command:
+        - sleep
```
jodh-intel: Might be worth replacing the sleeps with `tail -f /dev/null` to make it clear these containers should "run forever" ?

integration/kubernetes/runtimeclass_workloads/pod-liveness.yaml
```diff
@@ -0,0 +1,27 @@
...
+  name: liveness-exec
+spec:
+  runtimeClassName: kata
+  containers:
+  - name: liveness
+    image: busybox
+    args:
+    - /bin/sh
+    - -c
+    - touch /tmp/healthy; echo "Check status"; sleep 6; rm -rf /tmp/healthy; echo "Check dead"; sleep 12
```
jodh-intel: These `sleep` values seem pretty specific - is there any way to avoid sleeping to make this more robust?

integration/kubernetes/runtimeclass_workloads/runtimeclass_crd.yaml
```diff
@@ -0,0 +1,26 @@
...
+    kind:     RuntimeClass
+  scope: Cluster
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            runtimeHandler:
+              type: string
+              pattern: '^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)?$'
```
jodh-intel: ? :) Could you put a comment line above this one explaining what this is doing? Also, could it be simplified to use the `\w+` form for example?

.ci/configure_crio_for_kata.sh
```diff
@@ -11,10 +11,22 @@ set -o pipefail
...
-
+# `use_runtime_class` should be set to:
+# - true if we will test using k8s RuntimeClass feature (default) or
+# - false if we will test using the old trusted/untrusted annotations.
+use_runtime_class=${use_runtime_class:-true}
 
+if [ "${use_runtime_class}"  == true ]; then
+	echo "Configure runtimes map for RuntimeClass feature"
+	echo "- Set runc as default runtime"
+	sudo sed -i 's/runtime_path =.*/runtime_path = "\/usr\/local\/bin\/crio-runc"/' "$crio_config_file"
```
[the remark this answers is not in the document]
chavafg: agree, changed  the delimiter.

chavafg:
@jodh-intel @devimc this PR only adds the yaml files for the new runtimeClass annotations (taken from the yaml files that already exist on the `untrusted_workloads` directory). If you think the commands of some of the yaml files should be changed, I can handle it in another PR.

integration/kubernetes/runtimeclass_workloads/lifecycle-events.yaml
jodh-intel: Oh right. In that case, maybe keep it as close to those original files as possible then. In fact, I'd mention this fact in the commit (ideally also mentioning the commit version of the files you modified).

jodh-intel:
As mentioned above, I hadn't realised that so we prolly want to modify them as "lightly" as possible, but document where these files came from in the commit(s).

devimc:
@chavafg  ok :man_teacher:

PR #996, opened 2018-12-12
CI: bare metal: do not fail on bare metal cleanup

grahamwhaley:
We should probably never fail on bare metal cleanup, as it is trying to potentially
clean up a corrupt system, and may hit things it does not expect.
Instead, try our hardest, and use some defaults if we cannot find the actual
runtime defaults.


grahamwhaley:
This PR looks to have cleared the hang up/error on the metrics slave we had overnight:
http://jenkins.katacontainers.io/job/kata-metrics-tests-ubuntu-16-04-PR/212/

devimc:
restarting jenkins-ci-ubuntu-16-04

Issue #1000, opened 2018-12-13
CI: reduce log dump size when errors are found


grahamwhaley:
Currently when an error is found in the logs in the [teardown script](https://github.com/kata-containers/tests/blob/master/.ci/teardown.sh#L232-L261), it does a fancy sed trick to pull out the error line and the following likely trail of messages - but, quite often that results in us having huge log files (Mb's - for instance in a local test from a real CI run, for two errors, I got 28715 lines in a 5.5Mb file...).
I suspect that means at the moment:
 - as we normally see the tail of the log in Jenkins, we only ever see a tail of log lines that mean very little out of context
 - few of us dare click the 'download the 48Mb full console log link`
 - and even if we do, we scroll back up over all the log stuff back to the real CI error

Let's reduce the log output to just the critical lines that the sed/searc/grep located. The full logs are attached as artifacts to the build run anyway, so if somebody does need to go look at the logs for real, then they can.

grahamwhaley:
ooh, Issue **1000** :-) Do I get a prize?

jodh-intel:
@grahamwhaley - here ya go.... :champagne: :wine_glass: :chocolate_bar: :cake: :ice_cream: :1st_place_medal: ! :smile:

PR #1001, opened 2018-12-13
CI: logs: reduce the log output upon error

grahamwhaley:
Rather than dump the error line found and all its 'freinds',
which can result in massive log dumps, let's just dump the log
line we found.

If somebody does need to look at the complete log, the CI keeps
them as artifact attachments to the build run anyway.

Fixes: #1000

### DIFF
--- a/.ci/teardown.sh
+++ b/.ci/teardown.sh
@@ -236,22 +236,11 @@ check_log_files()
 
 		# Display *all* errors caused by runtime exceptions and fatal
 		# signals.
-		for pattern in "fatal error" "fatal signal"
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@jodh-intel @chavafg ptal.
I guess we only really get to see if this works if the CIs fail....
/test

jodh-intel:
> freinds

Fiends? Or friends? :smile:

lgtm

I'm actually wondering if we can remove that block entirely as we're already running [the `collect` script](https://github.com/kata-containers/runtime/blob/master/data/kata-collect-data.sh.in) in the teardown and that looks for errors pretty aggressively (but could do with an explicit check on "segfault" maybe).

Issue #997, opened 2018-12-12
soak parallel test is taking more than 5 mins to complete making the CI to timeout


chavafg:
the `soak_parallel_rm` test is taking more than 5 minutes to complete, which makes the CI timeout. When we added this test, it was taking around 4 minutes. 

```
cd integration/stability && \
export ITERATIONS=2 && export MAX_CONTAINERS=20 && chronic ./soak_parallel_rm.sh
Build timed out (after 5 minutes). Marking the build as aborted.
```

chavafg:
@grahamwhaley do you know if something in the performance side has increased that could make this test to timeout?

grahamwhaley:
we launch quite a lot of containers, repeatedly, in that test - so, yes, I suspect we have slowly drifted up in time and thus exceeded our timeout.
If I could get back to looking at the [pending chronic PR](https://github.com/kata-containers/tests/pull/810), then that would help by giving out the 'heartbeat' ticker...
In the meantime, we could reduce the MAX_CONTAINERS... or drop the chronic temporarily - or, bump the timeout up ;-)

PR #1003, opened 2018-12-13
test: remove chronic from the soak_parallel_rm test

chavafg:
this test started timing out after 5 minutes of inactivity.
Seems like this test sometimes taks more minutes than before
as we started to see this issue this week.
removing chronic here would let the test avoid the timeout.

Fixes: #997.


devimc:
in the commit message  `s/taks/takes/`

PR #1007, opened 2018-12-13
k8s: tests: wait for volume to be bounded and delete pv and pvc at teardown.

chavafg:
Sometimes the volume is not bounded immediately. Instead of
checking the status of the volume claim right away after it is created,
check for the status every 3 sec to a maximum of 10 sec.
If not bounded on the 10th sec, then we fail.

Also, delete the persistent volume and persistent volume claim
on teardown() to have the system clean after running the
test.


integration/kubernetes/k8s-volume.bats
```diff
@@ -25,6 +25,8 @@ setup() {
 }
 
 @test "Create Persistent Volume" {
+	wait_time=10
+	sleep_time=3
```
devimc: I'd prefer a small wait time and more tries, how about we use 5 and 6 ?

chavafg: ok, changed to try every 2s.

integration/kubernetes/k8s-volume.bats
```diff
@@ -25,6 +25,8 @@ setup() {
 }
 
 @test "Create Persistent Volume" {
+	wait_time=10
+	sleep_time=2
```
devimc: that make sense, `sleep_time` must be a multiple of `wait_time`

GabyCT:
lgtm

jcvenegas:
```
#cri-containerd job
ok 1 Exceeding memory constraints
not ok 2 Running within memory constraints
# (in test file k8s-memory.bats, line 50)
#   `sudo -E kubectl wait --for=condition=Ready pod "$pod_name"' failed
# error: the server doesn't have a resource type "runtimeclass"
# pod/memory-test created
# error: timed out waiting for the condition
```

jcvenegas:
Fedora timeout out restaring ....
```
Adding repo from: https://download.docker.com/linux/fedora/docker-ce.repo
Last metadata expiration check: 0:00:00 ago on Thu 13 Dec 2018 11:05:30 PM UTC.
Metadata cache created.
Build timed out (after 5 minutes). Marking the build as aborted.
Build was aborted
Performing Post build task...
Match found for :.* : True
Logical operation result is TRUE
Running script  : #!/bin/bash
```

PR #1014, opened 2018-12-18
Fix package manager tests

devimc:


GabyCT:
lgtm

devimc:
uhmm the problem are not the tests, the problem is the network in the CI

```
Stderr: http://mirror.calgah.com/centos/7.6.1810/extras/x86_64/repodata/repomd.xml: [Errno 12] Timeout on http://mirror.calgah.com/centos/7.6.1810/extras/x86_64/repodata/repomd.xml: (28, 'Operation too slow. Less than 1000 bytes/sec transferred the last 30 seconds')
Trying other mirror.
```

Issue #1016, opened 2018-12-25
Test for bind-mounted volumes


running99:
runtime/issues/965

PR #1017, opened 2018-12-25
integration/stability: test for bind-mounted volumes

running99:
Test if a container stops properly:
start a container with a bind mount and set bind propagation to a sub-directory of the mount point.

Fixes: #1016

Depends-on:github.com/kata-containers/runtime#963

Signed-off-by: <crossrunning@outlook.com>

### DIFF
--- a/integration/stability/bind_mount_linux.sh
+++ b/integration/stability/bind_mount_linux.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Ning Lu
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test will start a container with a bind mount
+# and set bind propagation, the purpose of this
+# test is to check if the container stops properly
+
+set -e
+
+cidir=$(dirname "$0")
+testname="${0##*/}"
+sysname=$(uname)
+
+if [ "${sysname}" != "Linux" ]; then
+    echo "Skip ${testname} on ${sysname}"
+    exit 0
+fi
+
+source "${cidir}/../../lib/common.bash"
+
+# Environment variables
+IMAGE="${IMAGE:-busybox}"
+CONTAINER_NAME="${CONTAINER_NAME:-test}"
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+TMP_DIR=$(mktemp -d --tmpdir=/tmp ${testname}.XXX)
+MOUNT_DIR="${TMP_DIR}/mount"
+BIND_DST="${MOUNT_DIR}/dst"
+BIND_SRC="${TMP_DIR}/src"
+DOCKER_ARGS="-v ${MOUNT_DIR}:${MOUNT_DIR}:rslave"
+CONTAINER_ID=
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
+
+function setup {
+	clean_env
+	docker run --runtime=${RUNTIME} -d ${DOCKER_ARGS} --name ${CONTAINER_NAME} ${IMAGE} ${PAYLOAD_ARGS}
+	CONTAINER_ID=$(docker ps -q -f "name=${CONTAINER_NAME}")
+}
+
+function cmd_bind_mount {
+	mkdir -p ${BIND_SRC}
+	mkdir -p ${BIND_DST}
+	mount --bind ${BIND_SRC} ${BIND_DST}
+	docker rm -f ${CONTAINER_NAME}
+
+	KATA_PROC=$(ps aux | grep ${CONTAINER_ID} | grep -v grep | tee)
+}
+
+function clean_kata_proc {
+	kata_pids=$(echo -n "${KATA_PROC}" | awk '{print $2}')
+	[ -n "${kata_pids}" ] && echo "${kata_pids}" | xargs kill
+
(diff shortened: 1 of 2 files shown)


integration/stability/bind_mount_linux.sh
```diff
@@ -0,0 +1,62 @@
...
+
+set -e
+
+cidir=$(dirname "$0")
+testname="${0##*/}"
+sysname=$(uname)
+
+[ "$sysname"x == "Linux"x ] || exit 0
+
+source "${cidir}/../../metrics/lib/common.bash"
```
chavafg: If this is just for the `clean_env` function, it would be better to only source `"${cidir}/../../lib/common.bash"`

integration/stability/bind_mount_linux.sh
```diff
@@ -0,0 +1,62 @@
...
+BIND_DIR="${MOUNT_DIR}/sub"
+DOCKER_ARGS="-v ${MOUNT_DIR}:${MOUNT_DIR}:rslave"
+CONTAINER_ID=
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
+
+function setup {
+	clean_env
+	docker run --runtime=$RUNTIME -d $DOCKER_ARGS --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
```
chavafg: can you please quote all variables here and below?

integration/stability/bind_mount_linux.sh
running99: udpate

integration/stability/bind_mount_linux.sh
running99: update

integration/stability/bind_mount_linux.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+#
+# SPDX-License-Identifier: Apache-2.0
```
jodh-intel: Missing copyright by the author here?

running99: update

jodh-intel: Thanks. I guess you may have written this in 2018, but should this be 2019 now? :smile:

integration/stability/bind_mount_linux.sh
```diff
@@ -0,0 +1,64 @@
...
+# and set bind propagation, the purpose of this
+# test is to check if the container stops properly
+
+set -e
+
+cidir=$(dirname "$0")
+testname="${0##*/}"
+sysname=$(uname)
+
+[ "$sysname"x == "Linux"x ] || exit 0
```
jodh-intel: I know it's traditional, but I don't think we need to use this form of syntax any more as bash is sane :) I'd remove the `x`'s and add a warning message that the test is being skipped.

integration/stability/bind_mount_linux.sh
```diff
@@ -0,0 +1,64 @@
...
+function cmd_bind_mount {
+	mkdir -p ${BIND_DIR}
+	mount --bind /tmp ${BIND_DIR}
+	docker rm -f ${CONTAINER_NAME}
+}
+
+function check {
+	kata_pid_count=$(ps aux | grep ${CONTAINER_ID} | grep -v grep | wc -l)
+	if (( kata_pid_count != 0 ));then
+		die "Left $kata_pid_count kata processes, quitting"
```
jodh-intel: How about displaying full details of these "rogue" processes before dying? That might help with debugging any issues. Also, should those processes continue to run as this is a CI test? Maybe they should be killed and the bind mount unmounted?

/cc @chavafg, @grahamwhaley.

integration/stability/bind_mount_linux.sh
running99: oh, yes😂

integration/stability/bind_mount_linux.sh
running99: update

integration/stability/bind_mount_linux.sh
running99: ok, update to clean processes and mount point

jodh-intel:
Thanks @running99!

lgtm

jodh-intel:
Ping @kata-containers/tests.

grahamwhaley:
Running this locally, as non-root, I get a 'cannot create /mnt/tmp/sub' dir perm error - as `/mnt` may need root perms (depending on the distro, user setup etc.).
Do we have to use `/mnt`? Better if we make this all use `mktemp` (and place in the system `$TMP` etc.) if we can ?

Let's try not to hard code host side paths, and use `sudo` if/where we need to.

jodh-intel:
Restarting the CIs as the 3 failure no longer have valid URLs to trigger individual restarts...

/retest

chavafg:
```
10:11:23 Running stability test: bind_mount_linux.sh
10:11:25 test
10:11:25 Ending stability test: bind_mount_linux.sh
10:11:25 rm: cannot remove '/tmp/bind_mount_linux.sh.76n/mount/dst': Device or resource busy
10:11:25 make: *** [Makefile:66: docker-stability] Error 1
10:11:25 Build step 'Execute shell' marked build as failure
```

Test still failing, we need to umount resource before deleting it, right?

running99:
Yes, in the new version, need umount.

chavafg:
@GabyCT seems like firecracker job started right away when pushing the change, can you check the trigger setup?

chavafg:
merging this one, all CI passed and have several acks.

PR #1028, opened 2019-01-07
metrics: Remove exit after using die function

### LINKED ISSUE #1027: Remove exit after using die function
Die function performs already an exit 1 so with we need to remove the duplicated exits.

GabyCT:
Die function performs an exit 1 already so with this PR we will remove
the duplicated exits.

Fixes #1027


jodh-intel:
lgtm

Issue #1036, opened 2019-01-09
metrics: boot time: kernel 'last line' has changed upstream


grahamwhaley:
Looks like the kernel has changed the exact lines it prints out when finally freeing up un-used memory after initial boot - which we use in the [boot time metrics test](https://github.com/kata-containers/tests/blob/master/metrics/time/launch_times.sh#L82-L93) to grab a timestamp...

Should be easy enough to make that check handle both the old and the new style with some extra grepping or some regexp etc.

I've not tracked down specifically which kernel version the change got merged in, but the patch is [here](https://lkml.org/lkml/2018/8/5/172) I think.

This cropped up over at https://github.com/kata-containers/runtime/issues/1100
where you can see the output of a newer kernel:

```
14:19:07 >: [[    0.886457] Freeing unused kernel image memory: 836K
14:19:07 ]
14:19:07 >: [[    0.887657] Write protecting the kernel read-only data: 16384k
14:19:07 ]
14:19:07 >: [[    0.888820] Freeing unused kernel image memory: 2016K
14:19:07 ]
14:19:07 >: [[    0.889388] Freeing unused kernel image memory: 1208K
```

PR #1044, opened 2019-01-10
metrics: launch times: update end check for new kernel dmesg

grahamwhaley:
The kernel dmesg output has changed around the 'Freeing kernel memory',
which we search for to try and identify the 'end of kernel' boot time.
Update the search expression we use to find the last instance of these
lines - which was as easy as trimming the 'memory' word from the end
of the expression, which now works for both older and newer kernel
demsg outputs.

As an example of how the outputs changed, here is a grep of two logs
(taken from bare metal hosts btw, not Kata Container boots):

$ grep "Freeing unused kernel" *log
newlog:[    2.243048] Freeing unused kernel image memory: 2180K
newlog:[    2.248191] Freeing unused kernel image memory: 2012K
newlog:[    2.248337] Freeing unused kernel image memory: 120K
oldlog:[    2.462519] Freeing unused kernel memory: 1528K
oldlog:[    2.483732] Freeing unused kernel memory: 1664K
oldlog:[    2.484299] Freeing unused kernel memory: 92K

Fixes: #1036

### DIFF
--- a/metrics/time/launch_times.sh
+++ b/metrics/time/launch_times.sh
@@ -82,12 +82,12 @@ run_workload() {
 		# Grab the last kernel dmesg time
 		# In our case, we need to find the last real kernel line before
 		# the systemd lines begin. The last:
-		# 'Freeing unused kernel memory' line is a reasonable
+		# 'Freeing unused kernel' line is a reasonable
 		# 'last in kernel line' to look for.
 		# We make a presumption here that as we are in a cold-boot VM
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@chavafg @sboeuf @egernst - this should enable the boot time test to work with both old and newer kernel dmesg outputs.

jodh-intel:
lgtm!

/test

sboeuf:
we have enough LGTM, and travis is only failing on macOS, let's merge!

PR #1064, opened 2019-01-17
ci: Skip running kata-runtime kata-env for firecracker

### LINKED ISSUE #1063: ci: Skip running kata-runtime kata-env for Firecracker
kata-runtime kata-env does not work with Firecracker so we need to skip it.

GabyCT:
kata-runtime kata-env does not work with firecracker so we need to skip
running it.

Fixes #1063


GabyCT:
@grahamwhaley https://github.com/kata-containers/runtime/issues/1144

PR #1073, opened 2019-01-21
test: Fix check kata-ksm-throttler process only in Debian

### LINKED ISSUE #1072: Fix check kata-ksm-throttler process only in Debian
We need to check kata-ksm-throttler process only in Debian

GabyCT:
We need to check kata-ksm-throttler process only in Debian.

Fixes #1072

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -208,10 +208,12 @@ show_system_state() {
 	done
 
 	# Verify kata-ksm-throttler on Debian
-	debian_process="kata-ksm-throttler"
-	debian_process_path=$(whereis ${debian_process} | tr -d '[:space:]'| cut -d ':' -f2)
-	echo " --pgrep ${debian_process}--"
-	pgrep -f ${debian_process_path} > /dev/null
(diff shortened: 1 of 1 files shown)


metrics/lib/common.bash
```diff
@@ -208,10 +208,12 @@ show_system_state() {
 	done
 
 	# Verify kata-ksm-throttler on Debian
-	debian_process="kata-ksm-throttler"
-	debian_process_path=$(whereis ${debian_process} | tr -d '[:space:]'| cut -d ':' -f2)
-	echo " --pgrep ${debian_process}--"
-	pgrep -f ${debian_process_path} > /dev/null
+	if [ "$ID" == "debian" ]; then
```
grahamwhaley: Just to check then - we only run and test the `kata-ksm-throttler` on the Debian CI run? /cc @jodh-intel

jodh-intel: @GabyCT - I don't understand why we're doing this? Debian is no more special than any other distro and we should be testing all distro versions in the same way imho.

Please could you either enable this test for all other distros or explain why Debian needs to be different?

PR #1074, opened 2019-01-22
test: Modify soak test to support Firecracker

### LINKED ISSUE #1053: Refactor soak test for firecracker
Currently the soak test is looking for processes which are not available in firecracker so we need to refactor this test to support firecracker.

GabyCT:
This will enable and modify the soak test to run with Firecracker.
We need it to skip checking the processes of proxy and netmon as they
are not present when running with Firecracker.

Fixes #1053


jodh-intel:
I'd probably say "tolerate firecracker" rather than "support firecracker" :smile:

lgtm

Ping @kata-containers/tests.

PR #1076, opened 2019-01-22
ci: Modify Firecracker installation

### LINKED ISSUE #1075: Modify firecracker installation
As we are installing from the tar, we are getting the static binaries for runtime, shim, proxy, etc. With this change, we will avoid the duplication of installing twice the kata elements.

GabyCT:
As we are installing from the tar, we are getting the static binaries for
runtime, shim, proxy, etc. With this change, we will avoid the duplication
of installing twice the kata elements.

Fixes #1075

### DIFF
--- a/.ci/install_firecracker.sh
+++ b/.ci/install_firecracker.sh
@@ -85,3 +85,8 @@ check_vsock=$(sudo modprobe vhost_vsock)
 if [ $? != 0 ]; then
 	die "vsock is not supported on your host system"
 fi
+
+# FIXME - we need to create a symbolic link for kata-runtime
+# in order that kata-runtime kata-env works
+# https://github.com/kata-containers/runtime/issues/1144
+ln -s /opt/kata/bin/kata-runtime /usr/local/bin/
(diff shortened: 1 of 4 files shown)


.ci/install_firecracker.sh
```diff
@@ -67,7 +67,7 @@ else
 	cat <<-EOF | sudo tee "$docker_configuration_file"
 	{
 	 "runtimes": {
-	  "kata": {
+	  "kata-runtime": {
```
jodh-intel: Why has this changed?

GabyCT: @jodh-intel , if we leave it as `kata` all the tests are going to fail as they are looking for runtime `kata-runtime` that is why I am changing it.

jodh-intel: Oh - so this never worked? I understand now.

But this change should really be on a separate commit as it's not strictly related to the rest of the changes on this commit.

GabyCT:
@jodh-intel changes applied, I will add another PR for the change in the name of the runtime

jodh-intel:
lgtm

PR #1080, opened 2019-01-23
kata-webhook: Add a simple admission controller for Kata

mcastelino:
Given that crio no longer has support for default workload trust and
containerd never had the notion of default workload trust, an user will
need to explicitly tag workloads with a runtime class for them to be
deployed with Kata. This is something most users will not do (or for
that matter should be able to choose).

A cluster operator normally defines policy based on RBAC and namespaces.

To show how Kata can be setup seamlessly by a cluster operator it makes
sense to publish a simple policy implementation using an admission
controller.

This admission controller is also needed in order to run kubernetes
conformance tests with Kata. This will allow the conformance tests
to run on all pods which can be supported with Kata. The admission
controller will exclude workloads which cannot be run in kata, while
running the bulk of the tests in Kata containers. This will give us
significantly higher kubernetes integration testing, specially as
the conformance test suite evolves over time.

Fixes: #1102
Fixes: https://github.com/kata-containers/runtime/issues/1102

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

PR #1082, opened 2019-01-23
tests: Add kubernetes e2e conformance test script

chavafg:
Add script that runs the heptio/sonobuoy e2e conformance
tests.

Fixes: #1081
Related: https://github.com/kata-containers/runtime/issues/1108

### DIFF
--- a/integration/kubernetes/e2e_conformance/run.sh
+++ b/integration/kubernetes/e2e_conformance/run.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script runs the Sonobuoy e2e Conformance tests.
+# Run this script once your K8s cluster is running.
+# WARNING: it is prefered to use containerd as the 
+# runtime interface instead of cri-o as we have seen
+# errors with cri-o that still need to be debugged.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+export KUBECONFIG=$HOME/.kube/config
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../../../lib/common.bash"
+
+RUNTIME="${RUNTIME:-kata-runtime}"
+
+# Check if Sonobuoy is still running every 5 minutes.
+WAIT_TIME=300
+
+# Add a global timeout of 2 hours to stop the execution
+# in case Sonobuoy gets hanged.
+GLOBAL_TIMEOUT=$((WAIT_TIME*24))
+
+create_kata_webhook() {
+	pushd "${SCRIPT_PATH}/../../../kata-webhook" >> /dev/null
+	# Create certificates for the kata webhook
+	./create-certs.sh
+
+	# Apply kata-webhook deployment
+	kubectl apply -f deploy/
+	popd
+}
+
+run_sonobuoy() {
+	sonobuoy_repo="github.com/heptio/sonobuoy"
+	go get -u "$sonobuoy_repo"
+
+	# Run Sonobuoy e2e tests
+	info "Starting sonobuoy execution."
+	info "When using kata as k8s runtime, the tests take around 2 hours to finish."
+	sonobuoy run
+
+	start_time=$(date +%s)
+	estimated_end_time=$((start_time + GLOBAL_TIMEOUT))
+
+	# Wait for the sonobuoy pod to be running.
+	kubectl wait --for condition=Ready pod sonobuoy -n heptio-sonobuoy
+
+	while sonobuoy status | grep -Eq "running|pending" && [ "$(date +%s)" -le "$estimated_end_time" ]; do
+		info "sonobuoy still running, sleeping $WAIT_TIME seconds"
+		sleep "$WAIT_TIME"
+	done
+
+	# Retrieve results
+	e2e_result_dir="$(mktemp -d /tmp/kata_e2e_results.XXXXX)"
+	sonobuoy retrieve "$e2e_result_dir" || \
+		die "Couldn't retrieve sonobuoy results, please check status using: sonobuoy status"
+	pushd "$e2e_result_dir" >> /dev/null
+
+	# Uncompress results
+	ls | grep tar.gz | xargs tar -xvf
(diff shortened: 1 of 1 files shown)


chavafg:
Depends on: https://github.com/kata-containers/tests/pull/1080

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,59 @@
...
+# Get Sonobuoy repo
+sonobuoy_repo="github.com/heptio/sonobuoy"
+go get -u "$sonobuoy_repo"
+
+# Run Sonobuoy e2e tests
+sudo -E "$GOPATH/bin/sonobuoy" run
+info "When using kata as k8s runtime, the tests take around 2 hrs to finish."
+
+wait_time=300
+while sudo -E "$GOPATH/bin/sonobuoy" status | grep running > /dev/null; do
```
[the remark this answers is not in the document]
chavafg: yes, agree, I'll add global timeout.

ydjainopensource:
@chavafg #1080 has been merged now.

chavafg:
@kata-containers/tests ptal

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,86 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script runs the Sonobuoy e2e Conformance tests.
+# Run this script once your K8s cluster is running.
+# WARNING: it is prefered to use containerd as the 
+# runtime interface instead of cri-o as we have seen
+# errors with cri-o that still need to be debugged.
+
+set -e
```
jcvenegas: We can be more strict here 
```
set -o errexit
set -o nounset
set -o pipefail
```

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,86 @@
...
+	# Create certificates for the kata webhook
+	./create-certs.sh
+
+	# Apply kata-webhook deployment
+	kubectl apply -f deploy/
+	popd
+}
+
+run_sonobuoy() {
+	# Get Sonobuoy repo
```
jcvenegas: Comment not needed

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,86 @@
...
+	sonobuoy_repo="github.com/heptio/sonobuoy"
+	go get -u "$sonobuoy_repo"
```
jcvenegas: Not needed to pin the test?

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,86 @@
...
+	# Wait for the sonobuoy pod to be running.
+	kubectl wait --for condition=Ready pod sonobuoy -n heptio-sonobuoy
+
+	while sonobuoy status | grep -Eq "running|pending" && [ "$(date +%s)" -le "$estimated_end_time" ]; do
+		info "sonobuoy still running, sleeping $WAIT_TIME seconds"
+		sleep "$WAIT_TIME"
+	done
+
+	# Retrieve results
+	e2e_result_dir="$(mktemp -d kata_e2e_results)"
```
jcvenegas: This is not missing `X` ?: mktemp -d kata_e2e_results.XXXXX

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,86 @@
...
+	sonobuoy retrieve "$e2e_result_dir" || \
+		die "Couldn't retrieve sonobuoy results, please run: sonobuoy status"
```
jcvenegas: Missing cleanup function if something goes wrong.

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,86 @@
...
+	pushd "$e2e_result_dir"
```
jcvenegas: I like to add >> /dev/null , to make cleaner the output

GabyCT:
@chavafg one question so this e2e will run in a different jenkins job? or how this script will be triggered?

chavafg:
@GabyCT yes, it will run on a new daily job, I am setting up this one: http://jenkins.katacontainers.io/job/kata-containers-containerd-k8s-conformance-e2e

integration/kubernetes/e2e_conformance/run.sh
GabyCT: should we remove this after the end of this test?

integration/kubernetes/e2e_conformance/run.sh
chavafg: I would not want to add a cleanup function here as the user may want to debug what went wrong with the tests. That is why it is asked to run `sonobuoy status`. 
I added a cleanup function at the end, which deletes the sonobuoy pods that are in charge of running the tests and it sends a message that the results directory will not be deleted, so that a user can consult it after the tests run.

integration/kubernetes/e2e_conformance/run.sh
chavafg: official documentation does not says that it is required to move to another commit/branch

integration/kubernetes/e2e_conformance/run.sh
chavafg: it would be under /tmp/ and I prefer to not remove this as the user might want to check the results directory after the tests are executed.

GabyCT: @chavafg thanks for the explanation :)

chavafg:
@jcvenegas addressed your comments, please take a look

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -0,0 +1,94 @@
...
+	kubectl apply -f deploy/
+	popd
+}
+
+run_sonobuoy() {
+	sonobuoy_repo="github.com/heptio/sonobuoy"
+	go get -u "$sonobuoy_repo"
+
+	# Run Sonobuoy e2e tests
+	sonobuoy run
```
jodh-intel: Presumably this command returns immediately?

Nit: It still might be clearer to move the `info` line below about this line.

chavafg: agree.. moved the `info` line :)

PR #1083, opened 2019-01-24
kata-manager: Remove directory if git clone fails.

yyyeerbo:
Remove directory if git clone fails, otherwise the following installation
fails.


jodh-intel:
lgtm

PR #1101, opened 2019-01-30
ci: Exclude *.service.in files from license static check

### LINKED ISSUE #1100: Exclude *.service.in files from license static check
In order to prevent CI failures when systemd service template files are modified, the license static check script should exclude `*.service.in` files.

sboeuf:
In order to prevent CI failures when systemd service template
files are modified, this commit excludes *.service.in files from
the license static check script.

Fixes #1100

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -360,6 +360,7 @@ check_license_headers()
 		--exclude="*.png" \
 		--exclude="*.pub" \
 		--exclude="*.service" \
+		--exclude="*.service.in" \
 		--exclude="*.svg" \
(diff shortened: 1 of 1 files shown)


jodh-intel:
I guess we could do this, but since systemd units support standard hash comments, couldn't (in fact shouldn't) we also just add the copyright header to the following?:

- https://github.com/kata-containers/agent/blob/master/kata-agent.service.in
- https://github.com/kata-containers/agent/blob/master/kata-containers.target

sboeuf:
@jodh-intel Fair enough, let's drop this PR and add the license.

Issue #1129, opened 2019-02-04
Add s390x support for CI scripts


jschintag:
Currently s390x is not supported by the CI.

PR #1130, opened 2019-02-04
Add s390x support for CI scripts

jschintag:
Fixes: #1129

Further Todos:
- build and install qemu will be implemented as soon as the configure hypervisor script supports s390x
- currently the Hot Plug CPU Feature is not working on s390x

### DIFF
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -93,25 +93,31 @@ if [ -f "${cidir}/${QEMU_ARCH}/lib_install_qemu_${QEMU_ARCH}.sh" ]; then
 fi
 
 main() {
-	if [ "$QEMU_ARCH" == "x86_64" ]; then
-		packaged_qemu_commit=$(get_packaged_qemu_commit)
-		short_current_qemu_commit=${CURRENT_QEMU_COMMIT:0:10}
-		if [ "$packaged_qemu_commit" == "$short_current_qemu_commit" ]; then
-			# If installing packaged qemu from OBS fails,
-			# then build and install it from sources.
-			install_packaged_qemu || build_and_install_qemu
-		else
-			build_and_install_qemu
-		fi
-	elif [ "$QEMU_ARCH" == "aarch64" ] || [ "$QEMU_ARCH" == "ppc64le" ]; then
-		packaged_qemu_version=$(get_packaged_qemu_version)
-		short_current_qemu_version=${CURRENT_QEMU_VERSION#*-}
-		if [ "$packaged_qemu_version" == "$short_current_qemu_version" ] && [ -z "${CURRENT_QEMU_COMMIT}" ]; then
-			install_packaged_qemu || build_and_install_qemu
-		else
-			build_and_install_qemu
-		fi
-	fi
+	case "$QEMU_ARCH" in
+		"x86_64")
+			packaged_qemu_commit=$(get_packaged_qemu_commit)
+			short_current_qemu_commit=${CURRENT_QEMU_COMMIT:0:10}
+			if [ "$packaged_qemu_commit" == "$short_current_qemu_commit" ]; then
+				# If installing packaged qemu from OBS fails,
+				# then build and install it from sources.
+				install_packaged_qemu || build_and_install_qemu
+			else
+				build_and_install_qemu
+			fi
+			;;
+		"aarch64"|"ppc64le"|"s390x")
+			packaged_qemu_version=$(get_packaged_qemu_version)
+			short_current_qemu_version=${CURRENT_QEMU_VERSION#*-}
+			if [ "$packaged_qemu_version" == "$short_current_qemu_version" ] && [ -z "${CURRENT_QEMU_COMMIT}" ] || [ "${QEMU_ARCH}" == "s390x" ]; then
+				install_packaged_qemu || build_and_install_qemu
+			else
+				build_and_install_qemu
+			fi
+			;;
+		*)
+			die "Architecture $QEMU_ARCH not supported"
+			;;
+	esac
 }
 
 main
--- a/.ci/s390x/clean_up_s390x.sh
+++ b/.ci/s390x/clean_up_s390x.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
+source "${lib_script}"
+
+gen_clean_arch || info "Arch cleanup scripts failed"
--- a/.ci/s390x/configuration_s390x.yaml
+++ b/.ci/s390x/configuration_s390x.yaml
@@ -1,0 +1,22 @@
+#
+# Copyright (c) 2019 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+test:
+  - functional
+  - docker
+
+# Same model as is for aarch64
+docker:
+  Describe:
+    - CPUs and CPU set
+    - Update number of CPUs
+    - Hot plug CPUs
+    - Update CPU constraints
+    - Hotplug memory
+    - update memory constraints
+  Context:
+    - remove bind-mount source before container exits
+    - run container exceeding memory constraints
+  It:
--- a/.ci/s390x/filter_docker_s390x.sh
+++ b/.ci/s390x/filter_docker_s390x.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+GOPATH_LOCAL="${GOPATH%%:*}"
+kata_dir="${GOPATH_LOCAL}/src/github.com/kata-containers"
+test_dir="${kata_dir}/tests"
+ci_dir="${test_dir}/.ci"
+test_config_file="${ci_dir}/s390x/configuration_s390x.yaml"
+
+describe_skip_flag="docker.Describe"
+context_skip_flag="docker.Context"
+it_skip_flag="docker.It"
+
+# value for '-skip' in ginkgo
+_skip_options=()
+
+filter_and_build()
+{
+	local dependency="$1"
+	local array_docker=$("${GOPATH_LOCAL}/bin/yq" read "${test_config_file}" "${dependency}")
+	[ "${array_docker}" = "null" ] && return
+	mapfile -t _array_docker <<< "${array_docker}"
+	for entry in "${_array_docker[@]}"
+	do
+		_skip_options+=("${entry#- }|")
+	done
+}
+
+main()
+{
+	# build skip option based on Describe block
+	filter_and_build "${describe_skip_flag}"
+
+	# build skip option based on context block
+	filter_and_build "${context_skip_flag}"
+
+	# build skip option based on it block
+	filter_and_build "${it_skip_flag}"
+
+	skip_options=$(IFS= ; echo "${_skip_options[*]}")
+
+	echo "${skip_options%|}"
+}
+
+main
--- a/.ci/s390x/lib_install_qemu_s390x.sh
+++ b/.ci/s390x/lib_install_qemu_s390x.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
+PACKAGED_QEMU="qemu"
+
+[ "$ID" == "ubuntu" ] || die "Unsupported distro: $ID"
+
+get_packaged_qemu_version() {
+	if [ "$ID" == "ubuntu" ]; then
+		sudo apt-get update > /dev/null
+		qemu_version=$(apt-cache madison $PACKAGED_QEMU \
+		| awk '{print $3}' | cut -d':' -f2 | cut -d'+' -f1 | head -n 1 )
+	fi
+
+	if [ -z "$qemu_version" ]; then
+		die "unknown qemu version"
+	else
+		echo "${qemu_version}"
+	fi
+}
+
+install_packaged_qemu() {
+	sudo apt install -y "$PACKAGED_QEMU"
+}
+
+build_and_install_qemu() {
+	QEMU_REPO=$(get_version "assets.hypervisor.qemu.url")
+	# Remove 'https://' from the repo url to be able to clone the repo using 'go get'
+	QEMU_REPO_PATH=${QEMU_REPO/https:\/\//}
+
+	PACKAGING_REPO="github.com/kata-containers/packaging"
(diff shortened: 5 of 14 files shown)


.ci/s390x/clean_up_s390x.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 IBM
```
jodh-intel: 2019 (here and below)?

.ci/s390x/configuration_s390x.yaml
```diff
@@ -0,0 +1,22 @@
+#
+# Copyright (c) 2018 IBM
+#
+# SPDX-License-Identifier: Apache-2.0#
```
jodh-intel: Nit: trailing `#`.

.ci/s390x/lib_install_qemu_s390x.sh
```diff
@@ -0,0 +1,39 @@
...
+#
+
+set -e
+
+CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
+PACKAGED_QEMU="qemu"
+
+get_packaged_qemu_version() {
+        if [ "$ID" == "ubuntu" ]; then
+		#output redirected to /dev/null
```
jodh-intel: I'd drop this comment tbh.

.ci/s390x/lib_install_qemu_s390x.sh
```diff
@@ -0,0 +1,39 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
+PACKAGED_QEMU="qemu"
+
+get_packaged_qemu_version() {
+        if [ "$ID" == "ubuntu" ]; then
```
jodh-intel: Idea: I'd be tempted to deal with the error scenario first, so in each function:

```sh
[ "$ID" == "ubuntu" ] || die "Unsupported distro: $ID"

 :
```

That will also simplify the remainder of the code as you can remove one layer of `if` testing. Might be worth putting that `die` test at the top global level to avoid having to duplicate either the code in each function, or duplication a call to a function that runs that test?

.ci/setup.sh
```diff
@@ -70,6 +70,16 @@ enable_nested_virtualization() {
 	aarch64 | ppc64le)
 		info "CI running in bare machine"
 		;;
+	s390x)
+		if [ $(systemd-detect-virt) == "kvm" ]; then
+			info "Running in kvm guest. Nested KVM is needed"
+			sudo modprobe -r kvm
+			sudo modprobe kvm nested=1
+			if [ grep -q "N" /sys/module/kvm/parameters/nested 2>/dev/null ]; then
+				die "Failed to find or enable Nested virtualization"
```
jodh-intel: This block looks almost identical to `enable_nested_virtualization()` in `.ci/setup.sh`. Could you just update that to handle os390x?

.ci/s390x/configuration_s390x.yaml
```diff
@@ -0,0 +1,22 @@
...
+#
+test:
+  - functional
+  - docker
+
+# Same model as is for aarch64
+docker:
```
jodh-intel: Comparing these files with the `aarch64` ones, they are *very* similar. I can see that there are small differences but I'm wondering if we can create something like:

```
.ci/generic/*
```

... and have `.ci/aarch64/*` and `.ci/s390x/*` include the generic files if possible. That should work for the shell scripts. I guess it won't work as-is for the YAML as that language doesn't support including natively. But, we could either consider breaking the YAML into chunks and reconstructing or having the scripts use tooling to convert the generic YAML files into arch-specific ones?

Bottom line: it's a maintenance burden on the architecture maintainers to ensure multiple very similar files remain updated. But maybe these files won't change much?

alicefr: hi @jodh-intel now the config looks pretty similar to arm, but on the other hand this file lists (or better black list) what's not working/have been tested on s390x. It could be that in the future the 2 architectures files diverge. On s390x for example, qemu hasn't supported memory hotplug  yet. On s390x, the CPU hotplug has probably a bug with Kata but it should be supported. The file could be a bit redundant but it clearly list what has been tested on s390x. Otherwise, we can have a discussion with @Pennyzct and @nitkon how to integrate the tested features in a more reasonable way for all the arch :)

jodh-intel: Hi @alicefr - Yes, let's leave these files as separate arch-specific ones for now. We can always re-assess the situation again in the future.

.ci/s390x/clean_up_s390x.sh
jschintag: Thank you, i just copied them and forgot to change the year

.ci/setup.sh
jschintag: sure, i can do that

.ci/s390x/lib_install_qemu_s390x.sh
jschintag: Wouldn't this just mean that i check the content of $ID at two places? I would instead use a switch case to replace this if elif structures and have the error scenario as a default.

jodh-intel: Well, as I say, you could make this a global check in `.ci/s390x/lib_install_qemu_s390x.sh`. Something like:

```sh
[ "$ID" == "ubuntu" ] || die "Unsupported distro: $ID"

get_packaged_qemu_version() {
   # ...
}
```

I think that should work since the caller already sources `/etc/os-release` and `lib.sh` which gives access to `die()`.

jschintag: Yes, it should work at that position. I changed it in the code.

.ci/s390x/filter_docker_s390x.sh
```diff
@@ -0,0 +1,51 @@
...
+	local array_docker=$("${GOPATH_LOCAL}/bin/yq" read "${test_config_file}" "${dependency}")
+	[ "${array_docker}" = "null" ] && return
+	mapfile -t _array_docker <<< "${array_docker}"
+	for entry in "${_array_docker[@]}"
+	do
+		_skip_options+=("${entry#- }|")
+	done
+}
+
+main()
```
grahamwhaley: Similar to some other items, much of this filter code looks arch generic - at some point we should probably pull a bunch of this out into library functions.
Does not have to block this PR, but will /cc @nitkon @GabyCT on it.

.ci/setup.sh
```diff
@@ -57,11 +57,16 @@ install_docker() {
 
 enable_nested_virtualization() {
 	case "$arch" in
-	x86_64)
+	x86_64 | s390x)
 		if [ "$CI" == true ] && grep -q "N" /sys/module/kvm_intel/parameters/nested 2>/dev/null; then
```
grahamwhaley: given you are doing a non-intel modprobe below, is this `kvm_intel` enabled check also not valid for s390x ?

integration/docker/docker.go
```diff
@@ -34,10 +34,7 @@ const (
 	DebianImage = "debian"
 
 	// FedoraImage is the fedora image
-	FedoraImage = "fedora"
-
-	// CentosImage is the centos image
-	CentosImage = "centos"
+        FedoraImage = "fedora"
```
grahamwhaley: Is this an intentional change - unclear to me (might have been clearer if it was in its own commit ;-) )

integration/docker/docker_arch_base.go
```diff
@@ -0,0 +1,6 @@
+//+build !s390x
```
grahamwhaley: I don't think we generally use in-file golang build constraints in kata - the constraint via the filename _extention is our normal way. wdty @jodh-intel ?

grahamwhaley:
/test
coz I want to see how it flies....

chavafg:
I see `.ci/.install_qemu.sh.swp`, which shouldn't be part of this change

.ci/s390x/lib_install_qemu_s390x.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
```
chavafg: seems like this is not being used, so it can be removed until we use it for building qemu.

.ci/install_qemu.sh
```diff
@@ -91,25 +91,34 @@ if [ -f "${cidir}/${QEMU_ARCH}/lib_install_qemu_${QEMU_ARCH}.sh" ]; then
...
+			packaged_qemu_version=$(get_packaged_qemu_version)
+			short_current_qemu_version=${CURRENT_QEMU_VERSION#*-}
+			if [ "$packaged_qemu_version" == "$short_current_qemu_version" ] && [ -z "${CURRENT_QEMU_COMMIT}" ]; then
+				install_packaged_qemu || build_and_install_qemu
+			else
+				build_and_install_qemu
+			fi
+			;;
+		"s390x")
+			install_packaged_qemu
```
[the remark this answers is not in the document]
jschintag: No, it will not build the source. As i've written in the PR-Message, building Qemu from source is currently not implemented. It will be added later, as soon as the configure_hypervisor.sh script from packaging supports s390x. The code will then look similar to x86_64.

.ci/setup.sh
jschintag: Yes it isn't. I overlooked that one.

jschintag:
I added the feedback from @grahamwhaley and @chavafg

.ci/install_qemu.sh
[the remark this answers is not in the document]
alicefr: @grahamwhaley if this is ok for you we can add the qemu building later. Otherwise, this PR is blocked by the missing support in [configure-hypervisor.sh](https://github.com/kata-containers/packaging/blob/master/scripts/configure-hypervisor.sh). As we obtained the CI integrated, we thought to open the pr to be able at least to install the dependencies and build Kata.

.ci/setup.sh
```diff
@@ -57,12 +57,15 @@ install_docker() {
 
 enable_nested_virtualization() {
 	case "$arch" in
-	x86_64)
-		if [ "$CI" == true ] && grep -q "N" /sys/module/kvm_intel/parameters/nested 2>/dev/null; then
+	x86_64 | s390x)
+		kvm_arch="kvm"
+		[ ${arch} == "x86_64" ] && kvm_arch="kvm_intel"
+		echo "${kvm_arch}"
```
alicefr: Do we need this print?

jschintag: no we don't

.ci/install_qemu.sh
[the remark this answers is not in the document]
grahamwhaley: Sure, its fine - was just checking...

grahamwhaley:
Hi @jschintag - thanks for the updates. We are getting there :-)
The CI check is moaning that you have no text body to your commit though - can you add a brief description (or, as full as you like ;-) ) pls.

```
Found 1 commit between commit 5ea2931de74c090c3246ab52db64bcc59e32177f and branch master
ERROR: Commit 5ea2931de74c090c3246ab52db64bcc59e32177f: single-line "Signed-off-by" body not permitted
ERROR: checkcommits failed. See the document below for help on formatting
commits for the project.
https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format
```

integration/docker/docker_arch_base.go
jodh-intel: Yes, but you can't negate using the filename afaik, so this one is fine.

integration/docker/docker_s390x.go
```diff
@@ -0,0 +1,8 @@
+//+build s390x
```
jodh-intel: Nit: not required as you've already encoded the platform in the name of the file itself.

jschintag:
@grahamwhaley  Ok, i added a body to the commit message. Hopefully the CI check likes it now.
@jodh-intel I also removed the build constraint header in integration/docker/docker_s390x.go

grahamwhaley:
Hi @jschintag - CI happy with the commit now, but it has noticed two files don't have the SPDX copyright headers...

```
All commit checks passed.
INFO: Checking for SPDX license headers
ERROR: Required license identifier ('SPDX-License-Identifier: Apache-2.0') missing from following files:
integration/docker/docker_arch_base.go
integration/docker/docker_s390x.go
The command "bash .ci/static-checks.sh github.com/kata-containers/tests" exited with 1.
```

I know, battling the CIs... but, I'm hoping we are nearly there!

jschintag:
@grahamwhaley No problem, i'll fix all the Problems the CI has.

grahamwhaley:
side note - @ttx, look, Zuul SoB/WIP checks active and working on this tests repo PR :-)

grahamwhaley:
@jschintag thanks - Travis now happy, along with the new Zuul checks (that replace some of the broken pullapprove CI). I'll fire the tests again so we can check we have not broken any other arch's...

grahamwhaley:
From the 16.04 CI, looks like we have some code issue:

```
14:39:05 bash sanity/check_sanity.sh
14:39:05 ./ginkgo -v -focus "" -skip "" ./integration/docker/ -- -runtime=kata-runtime -timeout=60
14:39:05 Failed to compile docker:
14:39:05 
14:39:05 can't load package: package github.com/kata-containers/tests/integration/docker: 
14:39:05 integration/docker/docker_arch_base.go:8:1: expected 'package', found 'const'
14:39:05 integration/docker/docker_arch_base.go:10:9: expected ';', found CentosImage
```

jschintag:
I fixed the issue

jodh-intel:
Thanks @jschintag.

/retest

jschintag:
@grahamwhaley @jodh-intel Since https://github.com/kata-containers/packaging/pull/333 was merged i added the support for building qemu on s390x.

integration/docker/docker_arch_base.go
```diff
@@ -0,0 +1,12 @@
+// +build !s390x
+
+// Copyright (c) 2019 IBM
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+package docker
```
grahamwhaley: odd, I think the CI is moaning about this, but did not moan about the one in docker_s390x.go?
```
integration/docker/docker_arch_base.go:3:1:warning: package comment should be of the form "Package docker ..." (golint)
The command "bash .ci/static-checks.sh github.com/kata-containers/tests" exited with 1.
```

@sameo - is this maybe at all related to the static check stuff you were looking at yesterday? (that is, not that you broke it, but maybe you know if there is something related to what you were debugging/fixing etc.).

jschintag: It seems the CI wants at least one empty line between the package and the License

jodh-intel:
@chavafg - the firecracker CI is failing but that doesn't appear to be related to this PR. Can we just merge this as all other CI's are passing?

chavafg:
sure, lets merge, but I think this needs 1 more ack

grahamwhaley:
Only fc failing - let's merge... and, let's figure out what the deal with fc is...

Issue #1154, opened 2019-02-06
do not use sudo when calling kubectl 


chavafg:
kubectl should not need to be called using sudo.

PR #1155, opened 2019-02-06
k8s: do not use sudo when making calls to kubectl 

chavafg:
Instead of using `/etc/kubernetes/admin.conf`, copy it to the
`$HOME/.kube/` of the non-root user to be able to make calls
to the kubernetes cluster without being root.
Also remove sudo from `kubectl` calls.

Fixes: #1154.


GabyCT:
lgtm

PR #1159, opened 2019-02-07
test: Check KATA_HYPERVISOR variable at the Makefile

### LINKED ISSUE #1158: Check KATA_HYPERVISOR at the Makefile
Check the KATA_HYPERVISOR variable at the Makefile to avoid .ci/hypervisors//filter_docker_.sh: No such file or directory errors.

GabyCT:
Check the KATA_HYPERVISOR variable to avoid the message of
`ci/hypervisors//filter_docker.sh: No such file`.

Fixes #1158

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,10 @@ TIMEOUT := 60
 UNION := functional docker crio docker-compose network netmon docker-stability oci openshift kubernetes swarm vm-factory entropy ramdisk shimv2
 
 # skipped test suites for docker integration tests
-SKIP := $(shell bash -f .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh)
+FILTER_FILE = .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh
+ifneq ($(wildcard $(FILTER_FILE)),)
+	SKIP := $(shell bash -f $(FILTER_FILE))
+endif
 
 # get arch
 ARCH := $(shell bash -c '.ci/kata-arch.sh -d')


Makefile
```diff
@@ -11,7 +11,11 @@ TIMEOUT := 60
 UNION := functional docker crio docker-compose network netmon docker-stability oci openshift kubernetes swarm vm-factory entropy ramdisk shimv2
 
 # skipped test suites for docker integration tests
-SKIP := $(shell bash -f .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh)
+ifeq ($(KATA_HYPERVISOR),)
+	SKIP :=
+else
+	SKIP := $(shell bash -f .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh)
```
grahamwhaley: Oh, OK. I was imagining we might do an 'if-exists' on the actual destination file to check it existed before we tried to run it. That way, if KATA_HYPERVISOR is set, but the file does not exist, we would still silently ignore it. With the above, I think if KATA_HYPERVISOR is set, but the relevant file does not exist in the repo, we will still get the error won't we?

GabyCT:
@grahamwhaley so what we need is `ifeq ($(KATA_HYPERVISOR),Firecracker)` ?

grahamwhaley:
I would code it so it can cope with any future hypervisors being added to kata, without having to recode it.
Thus, if we can check if the file exists, that should cover all future cases. In bash, we would:

```bash
if [ -x .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh ]; then
  SKIP := $(shell bash -f .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh)
else
  SKIP :=
fi
```

iyswim

GabyCT:
@grahamwhaley changes applied, thanks

Makefile
```diff
@@ -11,7 +11,10 @@ TIMEOUT := 60
 UNION := functional docker crio docker-compose network netmon docker-stability oci openshift kubernetes swarm vm-factory entropy ramdisk shimv2
 
 # skipped test suites for docker integration tests
-SKIP := $(shell bash -f .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh)
+FILE = .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh
```
devimc: `FILE` is a very generic name, can you change it to `FILTER_FILE` or something like that?

GabyCT: @devimc changes applied

devimc: Thanks @GabyCT

GabyCT:
weird the WIP-check failed even if I do not have a label or a tittle
/cc @grahamwhaley

grahamwhaley:
Hi @GabyCT - the WIP check is now showing passed - did you knowingly change something to make that pass? Just checking in case there are wrinkles still to iron out...

Issue #1147, opened 2019-02-06
metrics: docker_mem_usage: sometimes shows syntax error


grahamwhaley:
Occasionally, when running the report generator with `RUNTIME=runc` at least, I see a shell math syntax error come out from what I think is the PSS measurements:

```bash
density/docker_memory_usage.sh: line 72: ((: 632
37738 > 0 : syntax error in expression (error token is "37738 > 0 ")
```

and, I don't know why. It definitely does not happen all the time. Needs some debug (`bash -x ...`). Probably a relatively simple thing to do with the output of `smem` at my best guess.

jodh-intel:
It looks like this line:

- https://github.com/kata-containers/tests/blob/master/metrics/density/docker_memory_usage.sh#L67

... is returning multiple pids so you end up with two numbers (`632` and `37738`) rather than the expected one?

grahamwhaley:
hmm, which would imply multiple shims active or similar... thx @jodh-intel - that should make it easier to track (and, probably add a sanity check maybe to ensure we are getting the number of children we expect)

grahamwhaley:
This is now bugging me a bit. I can't reproduce it every time. I have a feeling it involves me running the `runc` tests after I've run either the `kata-fc` or `kata-qemu` tests - maybe something is dying and leaving an extra shim around is my current best theory. I will continue to try and reproduce, as it is hindering me doing some measure/compares :-)

grahamwhaley:
ah. ah ah... how about....

```bash
 child_mem=$(sudo "$SMEM_BIN" -c "pid pss" | \
                                grep "$child_pid" | awk '{print $2}')
```

*if*, one PID is a subset of another PID (like you have valid PIDs 123 and 1234), then I reckon just now and again the grep | awk will give you a double answer....

PR #1163, opened 2019-02-08
metrics: docker memory: runc: Handle multiple sub-set PIDs

grahamwhaley:
It looks like our use of 'smem | grep' would fail if one of the PIDs
in the list was a subset of another PID - as the grep would extract
both PIDs, which we did not want (the script would fail on the later
shell math).

Fix this by using the awk itself to do the PID line extraction from
the smem with an exact PID match, which also has the nice benefit of
dropping the grep.

Fixes: #1147


chavafg:
seems like several jobs didn't report back, lets run the tests again

/test

grahamwhaley:
Only failing fc CI, which I believe is a current known issue....

PR #1165, opened 2019-02-08
metrics: fix tests to work with static kata (including fc)

grahamwhaley:
Some of our metrics tests code presumed that kata was called `kata-runtime`, and that `kata-runtime` was always installed. That is not now the case with the advent of the static builds and `kata-fc`.
Update the scripts to handle the new runtimes and build types.

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -22,6 +22,24 @@ info() {
         echo -e "INFO: $*"
 }
 
+# Check if the $1 argument is the name of a 'known'
+# Kata runtime. Of course, the end user can choose any name they
+# want in reality, but this function knows the names of the default
+# and recommended Kata docker runtime install names.
+is_a_kata_runtime(){
+	case "$1" in
+	"kata-runtime") ;&	# fallthrough
+	"kata-qemu") ;&		# fallthrough
+	"kata-fc")
+		echo "1"
+		return
+		;;
+	esac
+
+	echo "0"
+}
+
+
 # Gets versions and paths of all the components
 # list in kata-env
 extract_kata_env(){
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -42,7 +42,7 @@ is_a_kata_runtime(){
 
 # Try to find the real runtime path for the docker runtime passed in $1
 get_docker_kata_path(){
-	local jpaths=$(docker info --format "{{json .Runtimes}}")
+	local jpaths=$(docker info --format "{{json .Runtimes}}" || true)
 	local rpath=$(jq .\"$1\".path <<< "$jpaths")
 	# Now we have to de-quote it..
 	rpath="${rpath%\"}"
@@ -99,7 +99,13 @@ extract_kata_env(){
 	RUNTIME_CONFIG_PATH="/usr/share/defaults/kata-containers/configuration.toml"
 	RUNTIME_VERSION="0.0.0"
 	RUNTIME_COMMIT="unknown"
-	RUNTIME_PATH="$rpath"
+	# If docker is broken, disabled or not installed then we may not get a runtime
+	# path from it...
+	if [ -z "$RUNTIME_PATH" ]; then
+		RUNTIME_PATH="/usr/bin/kata-runtime"
+	else
+		RUNTIME_PATH="$rpath"
+	fi
 	SHIM_PATH="/usr/libexec/kata-containers/kata-shim"
 	SHIM_VERSION="0.0.0"
 	PROXY_PATH="/usr/libexec/kata-containers/kata-proxy"
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -56,53 +56,64 @@ extract_kata_env(){
 	local toml
 	local rpath=$(get_docker_kata_path "$RUNTIME")
 
-	# If we cannot find the runtime, or it fails to run for some reason, do not die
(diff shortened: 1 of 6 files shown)


grahamwhaley:
If you want to run this with `runc` btw, you might also want #1163 layered in to prevent `runc` result oddities.

chavafg:
ramdisk test failing:
```
13:11:42 Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
13:11:42 INFO: Runtime environment not found - setting defaults
13:11:42 docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
13:11:42 See 'docker run --help'.
13:11:42 make: *** [ramdisk] Error 125
13:11:42 Build step 'Execute shell' marked build as failure
```

grahamwhaley:
Heh heh. Well, I know why it happens.... (came as a surprise to me!)
in the ramdisk test, as it reconfigures docker, it stops docker, and then tries to extract the kata env to work out which runtime to re-instantiate - the problem is now the `extract_kata_env` code is now trying to use `docker` to ask for the actual paths to the runtimes...

```bash
setup() {
	clean_env

	# Stop docker
	sudo -E systemctl stop docker
}

test_ramdisk() {
    extract_kata_env
```

Let me see what the best solution is - whether I make `extract_kata_env` do something sensible in the absence of `docker` working, or if I just need to fix this test so it grabs the env before it stops docker..

grahamwhaley:
OK, went with both belt and braces....
 - make the lib code more robust if we cannot get a runtime path from docker
 - re-arrange the ramdisk test code to extract the kata env before it stops docker running.

grahamwhaley:
16.04 CI failed in crio pv tests, but it looks like those tests do something different/unique on 16.04, so I will re-spin to see if that is persistent. for ref, the check is:

```
# run CRI-O tests using devicemapper on ubuntu 16.04
MAJOR=$(echo "$VERSION_ID"|cut -d\. -f1)
if [ "$ID" == "ubuntu" ] && [ "$MAJOR" -eq 16 ]; then
        # Block device attached to the VM where we run the CI
        # If the block device has a partition, cri-o will not be able to use it.
        export LVM_DEVICE
        if sudo fdisk -l "$LVM_DEVICE" | grep "${LVM_DEVICE}[1-9]"; then
                die "detected partitions on block device: ${LVM_DEVICE}. Will not continue"
        fi
        export STORAGE_OPTIONS="$DM_STORAGE_OPTIONS"
fi
```

and the error is:

```
13:12:12 ok 1 Pod quota
13:12:12 1..1
13:12:13 not ok 1 Create Persistent Volume
13:12:13 # (in test file k8s-volume.bats, line 42)
13:12:13 #   `kubectl get pv $volume_name | grep Available' failed
13:12:13 # kata   8m
13:12:13 # persistentvolume/pv-volume created
13:12:13 # Error from server (NotFound): pods "pv-pod" not found
13:12:13 # Error from server (NotFound): persistentvolumeclaims "pv-claim" not found
13:12:13 # persistentvolume "pv-volume" deleted
13:12:13 Makefile:69: recipe for target 'kubernetes' failed
```

I'm going to see if I can figure out why the WIP check is marked as failed...

lib/common.bash
```diff
@@ -22,65 +22,116 @@ info() {
...
+		RUNTIME_PATH="$rpath"
+	fi
+	SHIM_PATH="/usr/libexec/kata-containers/kata-shim"
+	SHIM_VERSION="0.0.0"
+	PROXY_PATH="/usr/libexec/kata-containers/kata-proxy"
+	PROXY_VERSION="0.0.0"
+	if [ "$KATA_HYPERVISOR" == firecracker ]; then
+		HYPERVISOR_PATH="/usr/bin/firecracker"
+	 else
+		HYPERVISOR_PATH="/usr/bin/qemu-system-x86_64"
```
chavafg: what about not hardcoding the `x86_64` arch and use the `kata-arch.sh` here, something like:
```
HYPERVISOR_ARCH=$(${cidir}/kata-arch.sh -d)
HYPERVISOR_PATH="/usr/bin/qemu-system-${HYPERVISOR_ARCH}"
```

metrics/density/docker_memory_usage.sh
```diff
@@ -236,7 +236,7 @@ get_docker_memory_usage(){
 EOF
 )"
 
-	elif [ "$RUNTIME" == "cor" ] || [ "$RUNTIME" == "cc-runtime" ] || [ "$RUNTIME" == "kata-runtime" ]; then
```
chavafg: cool, removing cc bits 👍

lib/common.bash
grahamwhaley: good idea, let me update.

metrics/density/docker_memory_usage.sh
grahamwhaley: ;-) I meant to note about that...

grahamwhaley:
going to wiggle a label to re-try the WIP check...

jcvenegas:
I think we still need to ignore fc job. LGTM @chavafg  please ack and merge if you are ok with it.

chavafg:
I think @grahamwhaley still needs to update the PR

grahamwhaley:
updated, re-pushed.

PR #1175, opened 2019-02-12
test: Update popular docker hub images test

GabyCT:
We need to remove the images after running the containers of the popular docker hub images test as we are running out of space and also we need to remove the kibana image as while doing a docker pull is showing the error of `docker: Error response from daemon: manifest for kibana:latest not found.`

### DIFF
--- a/integration/popular_docker_hub_images/popular_docker_images.bats
+++ b/integration/popular_docker_hub_images/popular_docker_images.bats
@@ -22,11 +22,13 @@ setup() {
 }
 
 @test "[insert data] insert data in an aerospike container" {
-	docker run --runtime=$RUNTIME -d --name aerospike aerospike/aerospike-server
+	image="aerospike/aerospike-server"
+	docker run --runtime=$RUNTIME -d --name aerospike $image
 	status=1
 	set +e
 	for i in $(seq 1 5); do
-		docker run --rm --runtime=$RUNTIME -i aerospike/aerospike-tools aql -h $(docker inspect -f '{{.NetworkSettings.IPAddress}}' aerospike) -c "insert into test.foo (PK, foo) values ('123','any'); select * from test.foo"
+		image2="aerospike/aerospike-tools"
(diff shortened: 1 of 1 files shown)


integration/popular_docker_hub_images/popular_docker_images.bats
```diff
@@ -35,347 +37,428 @@ setup() {
...
+	image="julia"
+	docker run --rm --runtime=$RUNTIME -i $image bash -c "julia -e 'println(\"this is a test\")'"
 }
 
 @test "[display configuration] run kapacitor container" {
-	docker run --rm --runtime=$RUNTIME -i kapacitor bash -c "kapacitord config > kapacitor.conf | ls -l kapacitor.conf"
-}
-
-@test "[display version] display information kibana container" {
-	docker run --rm --runtime=$RUNTIME -i $kibana_image kibana --version
```
chavafg: I see on the [kibana - docker hub page](https://hub.docker.com/_/kibana) that the latest tag is not supported and I remember that this was the reason we added a version of this image to https://github.com/kata-containers/tests/blob/master/versions.yaml#L28-L31

I would prefer to update the version that we track and continue testing this image.

integration/popular_docker_hub_images/popular_docker_images.bats
```diff
@@ -35,343 +37,433 @@ setup() {
...
 
 @test "[run application] start zookeeper container" {
-	docker run --rm --runtime=$RUNTIME -i zookeeper zkServer.sh start
+	image="zookeeper"
+	docker run --rm --runtime=$RUNTIME -i $image zkServer.sh start
 }
 
 teardown() {
 	clean_env
+	docker rmi $image
```
grahamwhaley: I think the idea of dropping/removing/cleaning up the images after each test is probably a sound one. It might end up with us being a bit less efficient on the download side as we might drop some layers that are then used in a later test, but if we need to save the space to run the tests, then so be it.

integration/popular_docker_hub_images/popular_docker_images.bats
```diff
@@ -251,6 +251,10 @@ setup() {
 	docker run --rm --runtime=$RUNTIME -i $kibana_image kibana --version
 }
 
+@test "[display version] display information kibana container" {
```
grahamwhaley: heh, if you resubmit, pls squash :-)

PR #1176, opened 2019-02-12
test: Update popular docker hub images test

### LINKED ISSUE #1174: Remove image after running container
We are running out of space while running the popular docker hub images test so with this change we will remove the images after running the containers to avoid this issue.

GabyCT:
We need to remove the images after running the containers
of the popular docker hub images test as we are running out of space.

Fixes #1174

### DIFF
--- a/integration/popular_docker_hub_images/popular_docker_images.bats
+++ b/integration/popular_docker_hub_images/popular_docker_images.bats
@@ -22,11 +22,13 @@ setup() {
 }
 
 @test "[insert data] insert data in an aerospike container" {
-	docker run --runtime=$RUNTIME -d --name aerospike aerospike/aerospike-server
+	image="aerospike/aerospike-server"
+	docker run --runtime=$RUNTIME -d --name aerospike $image
 	status=1
 	set +e
 	for i in $(seq 1 5); do
-		docker run --rm --runtime=$RUNTIME -i aerospike/aerospike-tools aql -h $(docker inspect -f '{{.NetworkSettings.IPAddress}}' aerospike) -c "insert into test.foo (PK, foo) values ('123','any'); select * from test.foo"
+		image2="aerospike/aerospike-tools"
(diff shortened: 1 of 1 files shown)


GabyCT:
@chavafg changes applied

integration/popular_docker_hub_images/popular_docker_images.bats
```diff
@@ -35,347 +37,433 @@ setup() {
 	done
 	set -e
 	docker rm -f aerospike
+	docker rmi $image2
```
grahamwhaley: maybe we can add an:

```bash
[ -n "$image2" ] docker rmi "$image2"
```

to the end where we do the other docker rm. In fact, maybe we should predicate the `$image` rm with a `-n` as well, just in case ;-)

grahamwhaley:
Interesting centos fail. Something failed, but then we seem to have leaked file descriptors (which is probably a bug, as that smells like we didn't clean up after the failure maybe).

```
01:36:38 ok 1 check entropy level
01:36:38 bash -f integration/ramdisk/ramdisk.sh
01:36:38 Running setup
01:36:39 Running ramdisk test
01:36:40 docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
01:36:40 See 'docker run --help'.
01:36:40 make: *** [ramdisk] Error 125
01:36:50 Process leaked file descriptors. See https://jenkins.io/redirect/troubleshooting/process-leaked-file-descriptors for more information
01:36:50 Build step 'Execute shell' marked build as failure
```

I think it is unrelated to this PR, so will re-trigger it.

GabyCT:
@grahamwhaley , mm it looks like the WIP check  and SoB check was triggered but the status was never reported :S

grahamwhaley:
@GabyCT - indeed, I don't see them in the zuul status log either (but the log does not have a long history)...
I'll try to nudge them here. also /cc @ttx, as we were looking at where delays etc. were in the pipelines recently.
/zuul-recheck

PR #1181, opened 2019-02-13
test: Add K8s test for tcp liveness probe

### LINKED ISSUE #1180: Add K8s test for tcp liveness probe
This is related with the request #1127. We need to add a test for tcp.

GabyCT:
This will add a K8s test for tcp liveness probe.

Fixes #1180

### DIFF
--- a/integration/kubernetes/k8s-liveness-tcp-probes.bats
+++ b/integration/kubernetes/k8s-liveness-tcp-probes.bats
@@ -1,0 +1,40 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+
+setup() {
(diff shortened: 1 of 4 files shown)


integration/kubernetes/k8s-liveness-tcp-probes.bats
```diff
@@ -0,0 +1,40 @@
...
+	pod_name="tcptest"
+
+	if kubectl get runtimeclass | grep kata; then
+		pod_config_dir="${BATS_TEST_DIRNAME}/runtimeclass_workloads"
+	else
+		pod_config_dir="${BATS_TEST_DIRNAME}/untrusted_workloads"
+	fi
+}
+
+@test "Liveness tcp probe" {
```
chavafg: @GabyCT do you think we can have this test included in this file? https://github.com/kata-containers/tests/blob/master/integration/kubernetes/k8s-liveness-probes.bats 

I see that the code is pretty similar and we can reuse code. Same for the http liveness probe test on #1179.

Issue #1047, opened 2019-01-10
shimv2 test flaky on centos CI


chavafg:
```
./hack/test-utils.sh: line 63:  4148 Terminated              keepalive "sudo PATH=${PATH} ${ROOT}/_output/containerd ${CONTAINERD_FLAGS}" ${RESTART_WAIT_PERIOD} &>${report_dir}/containerd.log
make[1]: Leaving directory `/tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/containerd/cri'
INFO: containerd(cri): Running test-integration
make[1]: Entering directory `/tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/containerd/cri'
1 attempt "sudo /tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/bin/crictl --runtime-endpoint=unix:///run/containerd/containerd.sock info"! Trying again in 1 seconds...
2 attempt "sudo /tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/bin/crictl --runtime-endpoint=unix:///run/containerd/containerd.sock info"! Trying again in 2 seconds...
3 attempt "sudo /tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/bin/crictl --runtime-endpoint=unix:///run/containerd/containerd.sock info"! Trying again in 3 seconds...
4 attempt "sudo /tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/bin/crictl --runtime-endpoint=unix:///run/containerd/containerd.sock info"! Trying again in 4 seconds...
E0110 19:10:02.451501    5127 remote_runtime.go:266] ListContainers with filter &ContainerFilter{Id:,State:nil,PodSandboxId:,LabelSelector:map[string]string{},} from runtime service failed: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService
time="2019-01-10T19:10:02Z" level=fatal msg="Failed to connect daemons" error="failed to list containers: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService" 
./hack/test-utils.sh: line 63:  5029 Terminated              keepalive "sudo PATH=${PATH} ${ROOT}/_output/containerd ${CONTAINERD_FLAGS}" ${RESTART_WAIT_PERIOD} &>${report_dir}/containerd.log
[...]

PR #1184, opened 2019-02-14
test: fix the shimv2 case random failed

lifupan:
Test if there is some legacy containerd process running,
kill them before try to run another test case, otherwise
it will be failed to relaunch another containerd daemon
which used by following testcase.

Fixes:#1047

Signed-off-by: Fupan Li <lifupan@gmail.com>

### DIFF
--- a/integration/containerd/shimv2/shimv2-tests.sh
+++ b/integration/containerd/shimv2/shimv2-tests.sh
@@ -25,9 +25,4 @@ if [[ "$ID" =~ ^opensuse.*$ ]] || [ "$ID" == sles ] || [ "$ID" == rhel ]; then
 	exit
 fi
 
-if [ "$ID" != "centos" ]; then
-	${SCRIPT_PATH}/../cri/integration-tests.sh
-else
-	issue="https://github.com/kata-containers/tests/issues/1047"
-	echo "Skip shimv2 on $ID, see: $issue"
-fi
+${SCRIPT_PATH}/../cri/integration-tests.sh


integration/containerd/cri/integration-tests.sh
```diff
@@ -182,6 +182,11 @@ main() {
 
 	for t in "${passing_test[@]}"
 	do
+		# In case there is some lagacy containerd process left, 
+		# try to kill them.
+		if [ -f /run/containerd/containerd.sock ]; then
```
grahamwhaley: It's good to check if there are things around we don't expect :-), but, are they a sign that something has gone wrong with an earlier test? If so, I'd rather we caught that and error'd out (like we do for stuck processes over in https://github.com/kata-containers/tests/blob/master/lib/common.bash#L80 for instance).
Would this be better added to `check_processes` so we can capture a broken test as early as possible (and not have to try and debug it after all other tests have run etc.)?

lifupan:
> Marking as request changes until we have discussed the right time and place to do this sanity check and warn/die.

Hi @grahamwhaley, I'm debugging the root cause of why some legacy Containerd process left, and once this issue is fixed, I think there will be no need for this sanity check.

grahamwhaley:
:-) @lifupan excellent you are tracking it down - but - normally we would leave the sanity check in place to *make sure the problem never comes back* :-) So, pls do add the sanity check :-)

jodh-intel:
Ping @lifupan.

jodh-intel:
Hi @lifupan - any update on this?

jodh-intel:
@lifupan - is this PR still needed?

integration/containerd/shimv2/shimv2-tests.sh
```diff
@@ -25,9 +25,4 @@ if [[ "$ID" =~ ^opensuse.*$ ]]; then
 	exit
 fi
 
-if [ "$ID" != "centos" ]; then
```
grahamwhaley: this looks like quite a big change/replace that is not called out in the commit - can you explain? Am I missing something?

lifupan: Hi @grahamwhaley Since the shimv2 testcases would fail randomly on "centos", and it was skipped on "centos", this fix tried to fix the random failure, thus we can drop this skip now.

jodh-intel:
@lifupan - is this PR ready for re-review?

lifupan:
> @lifupan - is this PR ready for re-review?

No, I'll update it  ASAP.

amshinde:
@lifupan Any updates on this PR?

PR #1186, opened 2019-02-14
metrics: Fix tittle for percentiles

### LINKED ISSUE #1089: metrics: report: fio: prints '98' where it should say '99' for percentiles
In the metrics report generator, for both fio read and write pages, it says `95&98 percentiles, boxplot over jobs` on one of the graphs, when it is actually using the 95 and 99'th percentiles. The problem line is at:
https://github.com/kata-containers/tests/blob/master/metrics/report/report_dockerfile/fio-reads.R#L180

Needs fixing in both fio read and write files.

GabyCT:
The report is printing 98 for percentiles but it should be 99 for the fio
write and fio reads operations.

Fixes #1089


grahamwhaley:
I don't think we even need to run the CIs on this (they don't process the touched files anyway)
@jodh-intel @jcvenegas if you concur, let's merge it.

Issue #1196, opened 2019-02-18
k8s tests: mute message when runtimeclass is not configured


chavafg:
when `runtimeclass` is not configured, we sometimes see this message when a test fail:
```
# error: the server doesn't have a resource type "runtimeclass"
```
but this actually is not the cause of a test failure as when `runtimeclass` is not configured, we use the old `trusted` and `untrusted` labels. 

We should mute this message in order to avoid confusion when a test fail for other reasons.

PR #1197, opened 2019-02-18
k8s: mute message when runtimeclass is not configured

chavafg:
k8s: mute message when runtimeclass is not configured

When runtimeclass is not configured on a test environment,
it is normal that we get this message:
```
error: the server doesn't have a resource type "runtimeclass"
```

This message is not the cause of a test failure, since we
use old trusted and untrusted tags when runtimeclass is not
configured.
To avoid confusion, lets send the message to /dev/null.

Also,  add information so that we know if we are running the
tests using runtimeclass or trusted/untrusted annotations.

Fixes: #1196.

### DIFF
--- a/integration/kubernetes/k8s-attach-handlers.bats
+++ b/integration/kubernetes/k8s-attach-handlers.bats
@@ -6,16 +6,13 @@
 #
 
 load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
 
 setup() {
 	export KUBECONFIG="$HOME/.kube/config"
 	pod_name="handlers"
 
-	if kubectl get runtimeclass | grep kata; then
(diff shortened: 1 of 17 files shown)


integration/kubernetes/k8s-attach-handlers.bats
```diff
@@ -11,10 +11,12 @@ setup() {
 	export KUBECONFIG="$HOME/.kube/config"
 	pod_name="handlers"
 
-	if kubectl get runtimeclass | grep kata; then
+	if kubectl get runtimeclass 2> /dev/null | grep kata; then
```
jodh-intel: Nit: I appreciate this is a minimal change, but it might be worth making the logic consistent since sometimes the tests are doing `grep kata` and other times `grep -q kata` (probably better).

But looking at the diff, I wonder if this logic could be centralised into a shell function as the `if` test and the variable being set looks seems to be the same each time.

Just a thought.

chavafg: @jodh-intel addressed both comments, ptal

PR #1208, opened 2019-02-20
metrics: Check entropy for boot time tests

### LINKED ISSUE #1167: metrics: check for entropy before running tests
If the system is short on entropy, the kernel inside the VM may stall/wait for enough entropy before carrying on. This can add a large artificial overhead/delay into tests such as the launch times test.
We could probably do something a little smart and have the tests wait on the host side until they know enough entropy is present to launch a container - and, have a sensible timeout so they die if entropy does not become available.

GabyCT:
This will check the entropy level at the host before running the
boot time tests in order to ensure stability.

Fixes #1167

### DIFF
--- a/metrics/time/launch_times.sh
+++ b/metrics/time/launch_times.sh
@@ -38,6 +38,24 @@ REQUIRED_CMDS=("bc" "awk")
 # Note the no-0-padding - 0 padding the results breaks bc in some cases
 DATECMD="date -u +%-s:%-N"
 
+# This the minimum entropy level produced
+# by haveged is 1000 see https://wiki.archlinux.org/index.php/Haveged
+# Less than 1000 could potentially slow down cryptographic
+# applications see https://www.suse.com/support/kb/doc/?id=7011351
+entropy_level="1000"
+
+check_entropy_level() {
+	retries="10"
+	for i in $(seq 1 "$retries"); do
+		if [ $(cat /proc/sys/kernel/random/entropy_avail) -ge ${entropy_level} ]; then
+			break;
+		fi
+		sleep 1
+	done
+	if [ $(cat /proc/sys/kernel/random/entropy_avail) -le ${entropy_level} ]; then
+		die "Not enough entropy level to run this test"
+	fi
+}
 
 # convert a 'seconds:nanoseconds' string into nanoseconds
 sn_to_ns() {
@@ -53,6 +71,9 @@ ns_to_s() {
 
 run_workload() {
 	start_time=$($DATECMD)
+
+	# Check entropy level of the host
+	check_entropy_level
 
 	# Run the image and command and capture the results into an array...
 	declare workload_result


metrics/time/launch_times.sh
```diff
@@ -38,6 +38,31 @@ REQUIRED_CMDS=("bc" "awk")
...
+	# Check that service is running
+	sudo service ${entropy_package} status | grep "running"
+}
+
+measure_entropy() {
+	retries="10"
+	for i in $(seq 1 "$retries"); do
+		if [ $(cat /proc/sys/kernel/random/entropy_avail) -ge ${entropy_level} ]; then
+			break;
+		fi
```
devimc: missing sleep ?

metrics/time/launch_times.sh
```diff
@@ -38,6 +38,31 @@ REQUIRED_CMDS=("bc" "awk")
...
+	# Check that entropy package is installed
+	check_package=$(which ${entropy_package} | wc -l)
+	if [ ${check_package} -eq 0 ]; then
+		die "No entropy package installed"
+	fi
+	# Check that service is running
+	sudo service ${entropy_package} status | grep "running"
+}
+
+measure_entropy() {
```
devimc: I think that `check_entropy_level` is a better name for this function, what do you think ?

metrics/time/launch_times.sh
```diff
@@ -38,6 +38,31 @@ REQUIRED_CMDS=("bc" "awk")
...
+	retries="10"
+	for i in $(seq 1 "$retries"); do
+		if [ $(cat /proc/sys/kernel/random/entropy_avail) -ge ${entropy_level} ]; then
+			break;
+		fi
+	done
+}
```
devimc: should the process die if entropy level is not `>= ${entropy_level}` ?

GabyCT:
@devimc changes applied

metrics/time/launch_times.sh
```diff
@@ -38,6 +38,34 @@ REQUIRED_CMDS=("bc" "awk")
 # Note the no-0-padding - 0 padding the results breaks bc in some cases
 DATECMD="date -u +%-s:%-N"
 
+# This the minimum entropy level produced
+# by haveged is 1000 see https://wiki.archlinux.org/index.php/Haveged
+# Less than 1000 could potentially slow down cryptographic
+# applications see https://www.suse.com/support/kb/doc/?id=7011351
```
grahamwhaley: iirc, the real issue we hit for kata boot is the kernel waiting for entropy to be 'enough' before continuing to boot. Let me see if I can dig up some of the details.
The original patch that introduced the 'slow boot' was to fix a CVE - https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=43838a23a05f
This caused boot slowdown hang issues for a number of situations. A follow up patch added RDRAND support to help get around this:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a26fb01c2879ed7026e6cbd78bb701912d249eef
but I think that patch landed in v4.18, and we are not quite on that yet... so we still see the stall/hangs, and hence recommend adding an entropy daemon on the host.
There is an overview here https://www.phoronix.com/scan.php?page=news_item&px=Linux-Protect-User-Entropy

The bit that I think is hard to work out is, how much entropy needs to be available for the boot to actually happen. It's not obvious to me from the code (welcome to the world of crypto/random code... maybe you are not meant to be able to understand it ;-) ).

The '1000' might be enough - @GabyCT , did you test this? Did you try running the test on a machine without the entropy daemon or this patch, and get a 'spiky' result, and then re-run it with just this patch to see if it fixes the issue (just this patch, as if you add in the entropy daemon I think you will always have >1000 and thus always just boot anyway).

metrics/time/launch_times.sh
```diff
@@ -38,6 +38,34 @@ REQUIRED_CMDS=("bc" "awk")
...
+entropy_level="1000"
+
+check_entropy_package() {
```
grahamwhaley: I'm not sure we should do a hard check for a specific entropy daemon. People might be running this test via the metrics report generator on a distro or setup that uses a different entropy source. This is not just for our CI systems.
We probably don't need to check there is entropy at all. What we can do is maybe have a timeout on waiting for entropy - if we don't get enough entropy in say 10s, then make the test fail saying 'not enough entropy to perform a reliable test' for instance.

metrics/time/launch_times.sh
```diff
@@ -38,6 +38,34 @@ REQUIRED_CMDS=("bc" "awk")
...
+	sudo service ${entropy_package} status | grep "running"
+}
+
+check_entropy_level() {
+	retries="10"
+	for i in $(seq 1 "$retries"); do
+		if [ $(cat /proc/sys/kernel/random/entropy_avail) -ge ${entropy_level} ]; then
+			break;
+		else
+			die "Low entropy level"
```
grahamwhaley: won't this just die on the first time around the loop - should the die be after the `for` loop?
But, yay, for the sleepy retry :-)

metrics/time/launch_times.sh
```diff
@@ -232,6 +263,7 @@ main() {
 	# it, so we may not fail if it is not set on the command line...
 	[ -z "$RUNTIME" ] && help && die "Mandatory runtime argument not supplied"
 
+	check_entropy_package
```
grahamwhaley: as above, I think we can drop checking for a specific entropy source daemon.

metrics/time/launch_times.sh
GabyCT: @grahamwhaley thanks for the explanation, I ran without and with entropy running 1000 containers, I will examine the results running the report generation (for both cases I ran twice). I am also trying to see if 1000 is enough or we need more.

jodh-intel: Any update on this @GabyCT?

GabyCT:
@grahamwhaley , I ran the tests for `1000` containers with entropy and without entropy twice and from the report I can see that randomly in both cases we have a spike
comparison

The results that we are getting with the `haveged` package is looking for an entropy level of minimum `1000`. I will try to run this test but now applying the patch to see the difference.

grahamwhaley:
thx @GabyCT - wow, the graphs coped with 1000 iterations ;-)
Those look pretty identical to me (if you ignore the fact the y-scales are different due to that bigger single spike on the haveged case). Looks to me like adding haveged is not making much difference in that case. Yeah, let's see what your next set of results look like.

GabyCT:
@grahamwhaley so it looks like the patch (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a26fb01c2879ed7026e6cbd78bb701912d249eef)  of is already applied because when I try to add it is saying
```
INFO: Applying patch /home/popular/go/src/github.com/kata-containers/packaging/kernel/patches/0004-rdrand.patch
patching file drivers/char/random.c
Reversed (or previously applied) patch detected!  Assume -R? [n]
```
Yes, it is already applied, I checked so it looks like entropy has no major impact on the stability of the tests results.

jodh-intel:
Any update @GabyCT, @grahamwhaley?

grahamwhaley:
Hi @jodh-intel @GabyCT - sorry, this feel off my end of week radar....
@GabyCT - I'm a bit confused, why are you trying to apply a kernel patch? afaik, that patch is in the upstream kernels, so is already there?
The goal here was to see if having an entropy source `haveged` enabled or not made any difference to the results and stability of the boot time tests.
From your metrics report run, can you also post the [boxplot](https://en.wikipedia.org/wiki/Box_plot) picture, so we can try to assess the 'spread' or noise in the results. Yeah, the report doesn't put a number in the results table for that (maybe that is something we can add). From the graphs you posted the spread looks about the same between the two, but the blue lines may be obscuring the red ones, and the boxplot separates them out, so should be easier to assess.

GabyCT:
@grahamwhaley , here are the graphs

speed

grahamwhaley:
Thanks @GabyCT So, from those graphs - and the 'nohaveged-vs-haveged' is the one that is interesting - I think we can see that there is a little more 'noise' in the no-haveged case, but the difference looks so small as to not have any real effect on the results.
Thus, right now I'm happy to move on - and just leave a:

> **HEADS UP** for everybody, that next time we see a time graph with a lot of spikes or times way beyond what we expect, keep 'lack of entropy' in the front of your mind as a potential cause. I think I've seen that maybe 3 times in the last year in user and developer reports for instance...

Here is a thought then - @egernst @sboeuf et. al. - maybe we should have the runtime do an entropy check before it launches a container (on the host) and issue a warning in the logs at least if it deems entropy is too low and may hinder the container boot.

PR #1217, opened 2019-02-21
k8s: fix tests to be able to run them several times

chavafg:
Do not overwrite on pod template. Instead create a
temporary file from the template with the needed changes.
This will let us run the tests several times without messing
up with the template files.

Also, remove `declare -a` from `name` array declaration on `k8s-parallel.bats`

For some reason, when using `declare -a` on the `name` array,
the values of the array cannot be retrieved on the `teardown()`
function. This was preventing the `teardown()` to correctly delete
some auto generated pod config files.
So lets remove the `declare -a` and move the declaration
of it to the `setup()`.

### DIFF
--- a/integration/kubernetes/k8s-volume.bats
+++ b/integration/kubernetes/k8s-volume.bats
@@ -17,11 +17,12 @@ setup() {
 	get_pod_config_dir
 
 	tmp_file=$(mktemp -d /tmp/data.XXXX)
+	pod_yaml=$(mktemp --tmpdir pod_config.XXXXXX.yaml)
 	msg="Hello from Kubernetes"
 	echo $msg > $tmp_file/index.html
 	pod_name="pv-pod"
 	# Define temporary file at yaml
-	sed -i "s|tmp_data|${tmp_file}|g" ${pod_config_dir}/pv-volume.yaml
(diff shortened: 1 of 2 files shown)


integration/kubernetes/k8s-volume.bats
```diff
@@ -17,11 +17,12 @@ setup() {
 	get_pod_config_dir
 
 	tmp_file=$(mktemp -d /tmp/data.XXXX)
+	pod_yaml=$(mktemp --tmpdir pod_config.XXXXXX.yaml)
 	msg="Hello from Kubernetes"
 	echo $msg > $tmp_file/index.html
 	pod_name="pv-pod"
 	# Define temporary file at yaml
-	sed -i "s|tmp_data|${tmp_file}|g" ${pod_config_dir}/pv-volume.yaml
+	sed -e "s|tmp_data|${tmp_file}|g" ${pod_config_dir}/pv-volume.yaml > "$pod_yaml"
```
grahamwhaley: do we remove that tmpfile later on? I don't see it in the patch. maybe we need a bash trap exit hook to do that?

chavafg: good catch, fixed. The `teardown` will ensure to delete the file even if something fails in the test.

PR #1221, opened 2019-02-21
test: Add K8s test for LimitRange in a namespace

### LINKED ISSUE #1220: K8s test for LimitRange
Add a test for LimitRange in a namespace

GabyCT:
This is a test for LimitRange in a namespace.

Fixes #1220

### DIFF
--- a/integration/kubernetes/k8s-limit-range.bats
+++ b/integration/kubernetes/k8s-limit-range.bats
@@ -1,0 +1,39 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+setup() {
+	export KUBECONFIG="$HOME/.kube/config"
+	get_pod_config_dir
+	namespace_name="default-cpu-example"
+	pod_name="default-cpu-test"
+}
+
+@test "Limit range for storage" {
+	# Create namespace
+	kubectl create namespace "$namespace_name"
+
+	# Create the LimitRange in the namespace
+	kubectl create -f "${pod_config_dir}/limit-range.yaml" --namespace=${namespace_name}
+
+	# Create the pod
+	kubectl create -f "${pod_config_dir}/pod-cpu-defaults.yaml" --namespace=${namespace_name}
+
+	# Get pod specification
+	kubectl wait --for=condition=Ready pod "$pod_name" --namespace="$namespace_name"
+
+	# Check limits
+	# Find the 500 millicpus specified at the yaml
+	kubectl describe pod "$pod_name" --namespace="$namespace_name" | grep "500m"
+}
+
+teardown() {
+	kubectl delete pod "$pod_name"
+	kubectl delete namespaces "$namespace_name"
+}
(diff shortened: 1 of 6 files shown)


integration/kubernetes/k8s-limit-range.bats
```diff
@@ -0,0 +1,39 @@
...
+	kubectl create namespace "$namespace_name"
+
+	# Create the LimitRange in the namespace
+	kubectl create -f "${pod_config_dir}/limit-range.yaml" --namespace=${namespace_name}
+
+	# Create the pod
+	kubectl create -f "${pod_config_dir}/pod-cpu-defaults.yaml" --namespace=${namespace_name}
+
+	# Get pod specification
+	cmd="kubectl get pod "$pod_name" -o jsonpath='{.status.phase}' --namespace="$namespace_name" | grep Running"
```
chavafg: Is this only checking that the pod is running? 
If so, we could use `kubectl wait`. 
Also, if this test is about Limits, I think that we should also check those limits were properly set,  right?

GabyCT: @chavafg , I tried it with `kubectl wait` but it does not retrieve the information so maybe it does not work with a namespace

chavafg: did you tried with `-n $namespace_name`? 
Something like: 
```
kubectl wait --for condition=Ready pod "$pod_name" -n "$namespace_name"
```

GabyCT: @chavafg , I fixed it with `wait` , I needed to add a command at the yaml so the container stays running

integration/kubernetes/k8s-limit-range.bats
```diff
@@ -0,0 +1,38 @@
...
+	kubectl create -f "${pod_config_dir}/limit-range.yaml" --namespace=${namespace_name}
+
+	# Create the pod
+	kubectl create -f "${pod_config_dir}/pod-cpu-defaults.yaml" --namespace=${namespace_name}
+
+	# Get pod specification
+	kubectl wait --for=condition=Ready pod "$pod_name" --namespace="$namespace_name"
+
+	# Check limits
+	kubectl describe pod "$pod_name" --namespace="$namespace_name" | grep "500m"
```
jodh-intel: Is this 500Mb or minutes or microseconds or miles or metres or mangos? :) I'm assuming this magic number is somehow related to the static yaml.

I do wonder if we should be generating that YAML from `.yaml.in` files for all these k8s tests so that the YAML is guaranteed to match the test script but I guess we can save that for another day.

GabyCT: @jodh-intel that is 500 millicpus and yes it is related to the static.yaml

grahamwhaley: I think @jodh-intel is really asking you to make it clear in the code - so, an explanatory comment for the 'magic numbers' :-)

integration/kubernetes/k8s-limit-range.bats
```diff
@@ -0,0 +1,39 @@
...
+	kubectl create -f "${pod_config_dir}/limit-range.yaml" --namespace=${namespace_name}
+
+	# Create the pod
+	kubectl create -f "${pod_config_dir}/pod-cpu-defaults.yaml" --namespace=${namespace_name}
+
+	# Get pod specification
+	kubectl wait --for=condition=Ready pod "$pod_name" --namespace="$namespace_name"
+
+	# Check limits
+	# Find the 500 millicpus specified at the yaml
```
GabyCT: @jodh-intel and @grahamwhaley changes applied

integration/kubernetes/k8s-limit-range.bats
GabyCT: @grahamwhaley and @jodh-intel changes applied

jodh-intel: Thanks @GabyCT - that's much clearer. Of course, it would be great if `kubectl` was a little saner, but that's another story... :-)

PR #1227, opened 2019-02-22
debug: Move sleep to gather the logs for oci test

### LINKED ISSUE #1065: oci_call_test randomly fails on Debian 9
We need to see why the oci_call_test is randomly failing on Debian

GabyCT:
Move the place where we are doing the sleep in order to gather the
logs after running a container to check the oci calls.

Fixes #1065

### DIFF
--- a/integration/oci_calls/oci_call_test.sh
+++ b/integration/oci_calls/oci_call_test.sh
@@ -12,7 +12,9 @@ set -e
 
 dir_path=$(dirname "$0")
 source "${dir_path}/../../lib/common.bash"
+source "${dir_path}/../../.ci/lib.sh"
 source /etc/os-release || source /usr/lib/os-release
+
 
 # Save logs
 TMP_FILE=$(mktemp runtimelogs.XXXXX)
@@ -35,6 +37,13 @@ function get_time() {
 # Get log for a specific time
 function get_debug_logs() {
 	sudo journalctl -q --since "$start_time" -o cat -a -t ${RUNTIME} > ${TMP_FILE}
+}
+
+function check_docker_status() {
+	wait_time=10
+	sleep_time=2
+	cmd="$1"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
 }
 
 # Find the arguments or oci calls for a specific command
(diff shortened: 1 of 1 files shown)


integration/oci_calls/oci_call_test.sh
```diff
@@ -76,14 +76,14 @@ function setup() {
...
-	# This sleep is necessary to gather the correct logs
-	sleep 10
-
 	get_time
 
 	# Start a container
 	docker run -d --runtime=${RUNTIME} --name=${NAME} ${IMAGE} ${PAYLOAD}
 
+	# This sleep is necessary to gather the correct logs
+	sleep 10
```
jodh-intel: This is an awful lot of cpu cycles. Why does it take so long to gather these logs?

chavafg: just a thought: what about removing the `-d` from the docker run, so that we gather the log after the workload has already finish, that way I guess we wont have to `sleep`?

GabyCT: @chavafg , we can not do that as part of the test is to run the container and then stop the container to gather the oci calls

grahamwhaley: I had a read of the test - it's a little convoluted and not that easy to follow, but what I think happens is this part of the test launches the container and checks the correct launch calls were made. then a later part of the test kills the same container and checks the correct kill calls were made....

I agree the 10s feels like a long time. Is there any way we can poll, with a sleep/timeout loop, to check when the container is up and running - using a docker status command etc.?
Also, given the payload is just a busybox with a tail, it should not take more than 3-5s to get it running (ideally <2).
Still, if we know it might take upto 5s on some machines to launch, maybe 10s is not a terrible thing here...

grahamwhaley:
@GabyCT - can you give us an update on this one before I ask others to re-review it. It felt like maybe you were still debugging the Debian CI case?

grahamwhaley:
Hi @GabyCT - so, I guess you are using this PR to test out the Debian CI?
Is this the right PR to do that with - as the title doesn't say anything about testing the CI. It just confuses me a bit every time I see the 'test-debian' only on this PR.
Maybe we need a separate PR just for testing Debian to make it clear?

Issue #1106, opened 2019-01-30
make CI install OBS master respository.


jcvenegas:
The "release" repository is depreacated.

All CI jobs should point to https://build.opensuse.org/project/show/home:katacontainers:releases:x86_64:master

PR #1229, opened 2019-02-22
ci: Update obs repos to stable-1.5

chavafg:
Release is not longer maintained.

Fixes: #1106


sboeuf:
I guess having `jenkins-ci-centos-7-4-firecracker` is expected here since we didn't have firecracker support in 1.4 and this is not something that has been backported.

Issue #1233, opened 2019-02-25
ci: Test install kata-runtime in in GOPATH not set mode


teawater:
Current kata-runtime makefile support install kata-runtime in in GOPATH not set mode but there is not test for it.
That is why https://github.com/kata-containers/runtime/issues/1224 happen.

PR #1234, opened 2019-02-25
ci: Test install kata-runtime in "No GO command or GOPATH not set" mode

teawater:
Add a test to install kata-runtime in "No GO command or GOPATH not set" mode.

Fixes: #1233

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -34,7 +34,7 @@ runtime_config_path="${SYSCONFDIR}/kata-containers/configuration.toml"
 PKGDEFAULTSDIR="${SHAREDIR}/defaults/kata-containers"
 NEW_RUNTIME_CONFIG="${PKGDEFAULTSDIR}/configuration.toml"
 # Note: This will also install the config file.
-build_and_install "github.com/kata-containers/runtime"
+build_and_install "github.com/kata-containers/runtime" "" "true"
 
 # Check system supports running Kata Containers
 kata_runtime_path=$(command -v kata-runtime)
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -85,8 +85,15 @@ function build() {
 function build_and_install() {
 	github_project="$1"
 	make_target="$2"
+	test_not_gopath_set="$3"
+
 	build "${github_project}" "${make_target}"
 	pushd "${GOPATH}/src/${github_project}"
+	if [ "$test_not_gopath_set" = "true" ]; then
+		info "Installing ${github_project} in No GO command or GOPATH not set mode"
+		sudo -E KATA_RUNTIME="${KATA_RUNTIME}" make install
+		[ $? -ne 0 ] && die "Fail to install ${github_project} in No GO command or GOPATH not set mode"
+	fi
 	info "Installing ${github_project}"
 	sudo -E PATH="$PATH" KATA_RUNTIME="${KATA_RUNTIME}" make install
 	popd


.ci/install_runtime.sh
```diff
@@ -34,7 +34,7 @@ runtime_config_path="${SYSCONFDIR}/kata-containers/configuration.toml"
 PKGDEFAULTSDIR="${SHAREDIR}/defaults/kata-containers"
 NEW_RUNTIME_CONFIG="${PKGDEFAULTSDIR}/configuration.toml"
 # Note: This will also install the config file.
-build_and_install "github.com/kata-containers/runtime"
+build_and_install "github.com/kata-containers/runtime" "" "1"
```
devimc: I'd prefer to use "true" and "false"

teawater: Done.

.ci/lib.sh
```diff
@@ -85,8 +85,15 @@ function build() {
 function build_and_install() {
 	github_project="$1"
 	make_target="$2"
+	test_not_gopath_set="$3"
+
 	build "${github_project}" "${make_target}"
 	pushd "${GOPATH}/src/${github_project}"
+	if [ "$test_not_gopath_set" = "true" ]; then
+		info "Installing ${github_project} in GOPATH not set mode"
+		sudo -E KATA_RUNTIME="${KATA_RUNTIME}" make install
```
jodh-intel: `GOPATH` is set on line 89 and you're using `sudo -E` so isn't `GOPATH` set here?

teawater: I think it found the issue of "GOPATH not set mode" with this test.
I need to fix it in runtime first.
https://github.com/kata-containers/runtime/pull/1286

teawater:
jenkins-metrics-ubuntu-16-04 failed:
```
Report Summary:
+-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
| P/F |         NAME         |  FLR  | MEAN  |  CEIL  |  GAP  |  MIN  |  MAX   |  RNG  | COV  | ITS |
+-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
| P   | boot-times           | 95.0% | 99.7% | 105.0% | 10.0% | 92.5% | 104.8% | 13.4% | 3.6% |  20 |
| *F* | memory-footprint     | 95.0% | 92.9% | 105.0% | 10.0% | 92.9% | 92.9%  | 0.0%  | 0.0% |   1 |
| *F* | memory-footprint-ksm | 95.0% | 69.5% | 105.0% | 10.0% | 69.5% | 69.5%  | 0.0%  | 0.0% |   1 |
+-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
Fails: 2, Passes 1
Failed
checkmetrics FAILED (1)
restoring KSM settings
```
This pr doesn't change the default behavior of kata-runtime.
So I think this fail does not relate to this fail.

grahamwhaley:
Hi @teawater - I 'adjusted' the metrics CI yesterday, so hopefully it is more stable. I have nudged a rebuild of that job, but indeed, this PR should not affect metrics.

grahamwhaley:
@jodh-intel @devimc - pls re-review

.ci/lib.sh
jodh-intel: I still don't understand this - `GOPATH` *is* set here so how can it be *"No GO command or GOPATH not set mode"* ?

teawater: @jodh-intel This part use:
```
sudo -E KATA_RUNTIME="${KATA_RUNTIME}" make install
```
without 
```
PATH="$PATH"
```
Then no go command can work.

jodh-intel: Doh! I always read `-E` as "export the entire env" rather than "export the vars you give me" :smile:

Issue #1225, opened 2019-02-22
CI: cleanup scripts should remove kata (and other?) PPAs etc. from distro repo paths


grahamwhaley:
Our CI scripts inject remote package paths into the test systems.
For one shot cloud instances this is not an issue, but for bare metal servers we should clean these paths up/out at the end of the run (or the start), as we can end up with stale URLs that then break builds (just witnessed).

Update the bare metal cleanup scripts to do this.

/cc @nitkon @alicefr @chavafg

grahamwhaley:
Related, I also found a large amount (>700) of commented out 'docker' lines in my /etc/apt/sources.list file, which I also suspect end up being injected from our CI scripts. We probably want to see if we can clean that up at the same time (ideally by not adding them in the first place ;-) ).

PR #1235, opened 2019-02-25
ci: baremetal cleanup: remove repo links and keys

grahamwhaley:
When cleaning up our servers, we should also try to remove the kata
repo links and keys. In the past we have seen issues when keys/repos
change and the old ones already exist, and it has prevented the new
ones being injected/updated. It's also just good practice to try and
remove all the things we installed...

Fixes: #1225

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -246,6 +246,41 @@ delete_stale_kata_resource()
 	done
 }
 
+delete_kata_repo_registrations()
+{
+	case "$ID" in
+		ubuntu)
+			local apt_file="/etc/apt/sources.list.d/kata-containers.list"
+			if [ -f "$apt_file" ]; then
+				info "Removing Kata apt file [$apt_file]"
+				sudo rm -f "$apt_file"
+			fi
+
+			sudo apt-key list | grep 'home:katacontainers' > /dev/null
+			if [ $? -eq 0 ]; then
+				# apt-key output format changed at ubuntu 16.10
+				if [ "$VERSION_ID" \< "16.10" ]; then
+					kata_uuid="$(sudo apt-key list | awk '$2=="home:katacontainers" {print prev} {prev=$2}')"
+					kata_uuid="${kata_uuid##*/}"
+				else
+					kata_uuid="$(sudo apt-key list | awk '$4=="home:katacontainers" {print prev} {prev=$0}')"
+				fi
+
+				if [ -n "$kata_uuid" ]; then
+					info "Removing Kata apt key [$kata_uuid]"
+					sudo apt-key del "$kata_uuid"
+				else
+					die "Failed to parse apt-key output for [$ID][$VERSION_ID]"
+				fi
+			else
+				info "No katacontainers key found - not removing"
+			fi
+			;;
+
+		*) info "Do not know how to clean repos from distro [$ID]";;
+	esac
+}
+
 gen_clean_arch() {
 	# Set up some vars
 	stale_process_union=( "docker-containerd-shim" )
@@ -268,5 +303,7 @@ gen_clean_arch() {
 		sudo rm -rf /etc/systemd/system/kubelet.service.d
 		sudo apt-get purge kubeadm kubelet kubectl -y
 	fi
+	info "Remove Kata package repo registrations"
+	delete_kata_repo_registrations
 }
 
--- a/.ci/aarch64/clean_up_aarch64.sh
+++ b/.ci/aarch64/clean_up_aarch64.sh
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+source "/etc/os-release" || source "/usr/lib/os-release"
+
 tests_repo="${tests_repo:-github.com/kata-containers/tests}"
 lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
 source "${lib_script}"
--- a/.ci/ppc64le/clean_up_ppc64le.sh
+++ b/.ci/ppc64le/clean_up_ppc64le.sh
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+source "/etc/os-release" || source "/usr/lib/os-release"
+
 tests_repo="${tests_repo:-github.com/kata-containers/tests}"
 lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
 source "${lib_script}"
--- a/.ci/s390x/clean_up_s390x.sh
+++ b/.ci/s390x/clean_up_s390x.sh
@@ -5,6 +5,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+source "/etc/os-release" || source "/usr/lib/os-release"
+
 lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
 source "${lib_script}"
 
--- a/.ci/x86_64/clean_up_x86_64.sh
+++ b/.ci/x86_64/clean_up_x86_64.sh
@@ -5,6 +5,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+source "/etc/os-release" || source "/usr/lib/os-release"
+
 lib_script="${GOPATH}/src/${tests_repo}/.ci/lib.sh"
 source "${lib_script}"
 


grahamwhaley:
Hmm, I wonder if we need a 'do-not-test' label ;-)
@chavafg @nitkon @Pennyzct @tuan-hoang1 - here is my first stab at bare metal cleanup of the apt repo links and keys - before we fire the CI on this can you give it a good eyeball.... I'm not concerned that it might damage a cloud instance, but there is always the chance it will take down our bare metal servers if I've messed up..

**PLEASE REVIEW BEFORE TESTING** ;-)

.ci/lib.sh
```diff
@@ -246,6 +246,25 @@ delete_stale_kata_resource()
...
+			if [ -f "$apt_file" ]; then
+				info "Removing Kata apt file [$apt_file]"
+				sudo rm -f "$apt_file"
+			fi
+			kata_uuid=$(sudo apt-key list | awk '$2=="home:katacontainers" {print prev} {prev=$2}')
+			if [ -n "$kata_uuid" ]; then
+				info "Removing Kata apt key [$kata_uuid]"
+				sudo apt-key del "$kata_uuid"
+			fi
+
```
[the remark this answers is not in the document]
grahamwhaley: heh - I had that in mind when I started writing the block, and then had forgotten by the time I got to the end of it!
thx, updated..

grahamwhaley:
@chavafg - if you think the code/theory is fine, then shall we drop the dnm and let it rip?

chavafg:
@grahamwhaley seems like it didn't removed the repo:
http://jenkins.katacontainers.io/job/kata-metrics-tests-ubuntu-16-04-PR/519/console

Also idk why the job didn't reported back here.

grahamwhaley:
@chavafg that is most odd. Still, I see this at least, so I can go fix that:
```
17:39:16 INFO: Remove Kata package repo registrations
17:39:16 INFO: Do not know how to clean repos from distro []
```

I thought it was already set/imported looking at the rest of the code - let me check...

grahamwhaley:
OK @chavafg - looks like as the cleanup scripts were called from higher level scripts, they never inherited the `os-release` vars they sometimes use!! Let's add that `source` to the top level cleanup scripts then.... and see how it flies now.

chavafg:
we can ignore ubuntu 16-04 job in favor to the 18.04 job which already passed.

Pennyzct:
Hi~ @grahamwhaley @chavafg Since it's designed for extra clean-up procedure on bare metal, I would like to see it tested on ARM CI. ;) any way to trigger ARM CI here?

grahamwhaley:
Hi @Pennyzct - no way to trigger on ARM CI here right now, as we have not enabled ARM (or some other arch) CIs on this repo yet...
Best I think we can do right now is:
- you visually check it is OK
- you hand-run it on an ARM machine

I coded it to try and be 'safe' if it does not find the files or keys it is looking or - but, I did note that the package install in the first place does an x86_64 specific install iirc. I suspect this patch is benign on non-x86_64 arch's right now, as they are not installing from OBS packages.

grahamwhaley:
looking at the metrics CI (which is the primary x86_64 bare metal slave this patch was aimed at), here is the relevant output (the last few lines being the new ones):

```
18:05:51 Looking for baremetal cleanup script for arch x86_64
18:05:51 Running baremetal cleanup script for arch x86_64
18:05:51 INFO: kill stale process
18:05:52 INFO: Runtime environment not found - setting defaults
18:05:52 INFO: delete stale docker resource under /var/lib/docker
18:05:54 INFO: delete stale kata resource under /var/lib/vc /run/vc
18:05:54 INFO: Remove installed kata packages
18:05:54 ERROR: packages not installed
18:05:54 INFO: Remove installed kubernetes packages and configuration
18:05:54 Reading package lists...
18:05:54 Building dependency tree...
18:05:54 Reading state information...
18:05:55 Package 'kubeadm' is not installed, so not removed
18:05:55 Package 'kubectl' is not installed, so not removed
18:05:55 Package 'kubelet' is not installed, so not removed
18:05:55 The following packages were automatically installed and are no longer required:
18:05:55   conntrack cri-tools kubernetes-cni
18:05:55 Use 'sudo apt autoremove' to remove them.
18:05:55 0 upgraded, 0 newly installed, 0 to remove and 117 not upgraded.
18:05:55 INFO: Remove Kata package repo registrations
18:05:55 INFO: Removing Kata apt file [/etc/apt/sources.list.d/kata-containers.list]
18:05:55 INFO: Removing Kata apt key [2048R/6063F3ED]
18:05:55 OK
```

and, the CI ran to completion (and failed due to a bounds check, but that is a different matter).
Thus, I think this patch is working as expected for x86_64 bare metal at least.

.ci/lib.sh
```diff
@@ -246,6 +246,26 @@ delete_stale_kata_resource()
...
+delete_kata_repo_registrations()
+{
+	case "$ID" in
+		ubuntu | debian)
+			local apt_file="/etc/apt/sources.list.d/kata-containers.list"
+			if [ -f "$apt_file" ]; then
+				info "Removing Kata apt file [$apt_file]"
+				sudo rm -f "$apt_file"
+			fi
+			kata_uuid=$(sudo apt-key list | awk '$2=="home:katacontainers" {print prev} {prev=$2}')
```
jodh-intel: This doesn't work on my system. Should it be instead:

```sh
kata_uuid=$(sudo apt-key list | awk '$4=="home:katacontainers" {print prev} {prev=$0}')
```

There is also a warning but I'm not sure what we're supposed to do about it:

```sh
Warning: apt-key output should not be parsed (stdout is not a terminal)
```

grahamwhaley: I suspect the warning is benign - like we have a lot of warnings about using apt in scripts as 'the interface is not guaranteed' throughout the logs iirc.
wrt $2 - better show me the output of your `sudo apt-key list`. The goal of the script was to:
- match the `home:katacontainers` to the second field of the `uid` line in the output
- and then pick up the stored second field of the *previous* line, which holds the key.

Let me post an example of what I see in a moment from my other machine..

grahamwhaley: Here is a snippet from my 16.04 machine. Note, this snippet does not have kata in it, as my test run already nuked it from this box - but, it gives you the idea of the format we are trying to parse:

```
pub   2048R/BA07F4FB 2018-04-01 [expires: 2021-03-31]
uid                  Google Cloud Packages Automatic Signing Key <gc-team@google.com>

pub   2048R/A7317B0F 2015-04-03 [expired: 2018-04-02]
uid                  Google Cloud Packages Automatic Signing Key <gc-team@google.com>
```

jodh-intel: I'm running on a bionic system and that code results in `kata_uuid` being empty.

grahamwhaley: .... so, show me the output ....

grahamwhaley: of `sudo apt-key list` pls

jodh-intel: Attached using the ol' fake file extension trick ;)

apt-key list gz

chavafg:
We actually have a job for arm in this repo, although not sure why it didn't report back. 
It is being triggered with the same `/test` comment. 
Anyway, here is the log
http://jenkins.katacontainers.io/job/kata-containers-tests-ARM-18.04-PR/426/ 

And got the same warning as @jodh-intel 
```
INFO: Remove Kata package repo registrations
Warning: apt-key output should not be parsed (stdout is not a terminal)
~/workspace/kata-containers-tests-ARM-18.04-PR/go/src/github.com/kata-containers/tests ~/workspace/kata-containers-tests-ARM-18.04-PR/go/src/github.com/kata-containers/tests
/tmp/install-go-tmp.OmrJn54Obj ~/workspace/kata-containers-tests-ARM-18.04-PR/go/src/github.com/kata-containers/tests
/usr/local/go/bin/go
```

.ci/lib.sh
grahamwhaley: right, so the format of the output on 18.04 is different from 16.04:

(from your file...)
```
pub   rsa2048 2018-03-13 [SC] [expires: 2020-08-20]
      9FDC 0CB6 3708 CF80 3696  E2DC D0B3 7B82 6063 F3ED
uid           [ unknown] home:katacontainers OBS Project <home:katacontainers@build.opensuse.org>
```

soo, I'm going to have to figure out how to parse that, and what to then give to `apt-key` to remove the key - it is reasonably different....
I'm not sure I have an 18.04 machine at my fingertips either... let me see what I can work out.

grahamwhaley: :-( looks like it changed at 16.10. sigh, so, if apt tells you to not parse its output - but - I can't see they offer any parseable scriptable alternative either. If anybody knows (@jodh-intel - you do have history here I say ;-) ), pls let me know!

grahamwhaley: @jodh-intel @chavafg - updated to handle both <16.10 and >16.10 apt-key output format. Can you check that for me visually and on your systems pls? I did test the awk in an 18.10 container, but it's harder for me to try the actual script. That bash string compare against "16.10" makes me twitch. Well, the whole dance makes me twitch tbh... thx!

chavafg:
@grahamwhaley tested on 18.04 and works ok, but now I think that we should have `debian` on a different case statement? or maybe only remove it from this change? I think we currently do not have debian on baremetal.

BTW, debian 9 uses the same format as 18.04.

.ci/lib.sh
```diff
@@ -246,6 +246,31 @@ delete_stale_kata_resource()
...
+{
+	case "$ID" in
+		ubuntu | debian)
+			local apt_file="/etc/apt/sources.list.d/kata-containers.list"
+			if [ -f "$apt_file" ]; then
+				info "Removing Kata apt file [$apt_file]"
+				sudo rm -f "$apt_file"
+			fi
+			# apt-key output format changed at ubuntu 16.10
+			if [ "$VERSION_ID" < "16.10" ]; then
```
jodh-intel: Alas this won't work as test only handles integers. The warning we get from `apt-key list` is clearly suggesting we don't do what we're trying to here, presumably because the format may change (as it has). But if that's the case, what *is* the correct way of parsing such info?

grahamwhaley:
Good call on debian @chavafg - yeah, I'll drop it from this PR, and we can add it in if/when we need it, along with all the other distros.

@jodh-intel , I used `<` in the [] test, as according to the bash manual, `-le` and its friends are arithmetic test, but:

```
        string1 < string2
              True if string1 sorts before string2 lexicographically.
```

(and, sadly there seemed to be no >= or I would have used that).
I was relying on the fact that stringwise lexographically "16.04" is 'smaller than "16.10" etc. wdyt?

As for what *is* the correct way to parse `apt-key` etc., afaict there isn't one, which is rather sadly disappointing. Even more frustrating that `apt-key list` cannot take a single argument for any sort of search, and `apt-key export` doesn't afaict export what we need in a nice parseable fashion. I didn't find a better solution :-(

.ci/lib.sh
jodh-intel: Ah - I see the problem - this fails currently as the shell treats `<` as a redirection (try it out ;)

The fix - quote the operator:

```sh
[ "$VERSION_ID" \< "16.10" ]
```

grahamwhaley: huh! wah wah, oh - let me look into that - bizzaro! thx!

grahamwhaley: heh, yep, need the escape with `[` but not `[[`... apparently, maybe.
escape added, debian dropped.
29th attempt gotta be the charm - let's try again...

grahamwhaley:
/test
and dropping the DNM

.ci/lib.sh
```diff
@@ -246,6 +246,31 @@ delete_stale_kata_resource()
...
+				info "Removing Kata apt file [$apt_file]"
+				sudo rm -f "$apt_file"
+			fi
+			# apt-key output format changed at ubuntu 16.10
+			if [ "$VERSION_ID" \< "16.10" ]; then
+				kata_uuid="$(sudo apt-key list | awk '$2=="home:katacontainers" {print prev} {prev=$2}')"
+			else
+				kata_uuid="$(sudo apt-key list | awk '$4=="home:katacontainers" {print prev} {prev=$0}')"
+			fi
+			if [ -n "$kata_uuid" ]; then
```
jodh-intel: I wonder if we want a `[ -z "$kata_uuid" ]` test in here so we get an alert when format changes yet again? :)

grahamwhaley: it would be nice, but the only problem I see is if we run on a machine that has no kata key installed (like the first time we use an instance, or if we ever get to bare metal one-off deployments which I hope to achieve on day), then it will fail as the kata key won't be present.

jodh-intel: In which case how about we error if the key is installed (`apt-key list | grep home:katacontainers`) but `$kata_uuid` is blank?

grahamwhaley: let me add that then..

grahamwhaley: and... one more time....
 - predicated the key parse/remove on a kata key being present
 - fixed a bug - needed to remove the 2048R/ key frontage before handing back to apt-key del

jodh-intel:
Thanks @grahamwhaley.

grahamwhaley:
Hey @jodh-intel , may have gotten a real log-parser error!:

```
16:54:12 ERROR: kata-log-parser: missing pid: {Count:0 TimeDelta:0 Filename:/tmp/jenkins/workspace/kata-containers-tests-fedora-vsocks-PR/go/src/github.com/kata-containers/tests/kata-runtime.log Line:10 Time:2019-02-27 15:42:50.154042921 +0000 UTC Pid:0 Level:debug Msg:New filesystem store backend for /var/lib/vc/sbs/UZDOk4Y2wJVAi2rfcXKR Source:virtcontainers/store Name: Container: Sandbox: Data:map[path:/var/lib/vc/sbs/UZDOk4Y2wJVAi2rfcXKR subsystem:store backend:filesystem]}
```

grahamwhaley:
I think the vsocks CI fail might be that 'don't check the proxy' item @chavafg opened a little while ago...

chavafg:
@grahamwhaley seems like different error, but maybe related. Anyway not related with this PR.
I see WIP-check still failing... should we merge?

grahamwhaley:
@chavafg zuul-recheck will only re-trigger the SoB check I think - need to edit the PR or wiggle a label change to re-trigger the WIP check. Let me wiggle the label, as I think zuul missed it earlier when I dropped the title change...

chavafg:
Thanks @grahamwhaley, now that passed. merging.

PR #1260, opened 2019-03-04
kata-webhook: Support fetching request context

mcastelino:
Not all information about the admission request is available in
the object. Some properties like Namespace is only available in
the request context.

Modify the webhook to allow accessing the full context. This will
allow for richer processing and policies.

This also eliminates removes the need for explict namespace filters
in the webhook configuration. This means user yaml files can now
remain unmodified.

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/kata-webhook/README.md
+++ b/kata-webhook/README.md
@@ -30,27 +30,10 @@ and sample admission controller we created by running
 ```bash
 $ kubectl apply -f deploy/
 ```
-The admission webhook (deploy/webhook-registration.yaml)
-is setup to exclude certian namespaces from being run with Kata using filters on namespace labels.
 
-```yaml
-    namespaceSelector:
-      matchExpressions:
-        -  {key: "kata", operator: NotIn, values: ["false"]}
-```
-The rook operators for example are marked as such.
+The webhook mutates pods to use the kata runtime class for all pods except
+those with 
 
-```yaml
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: rook-ceph-system
-  labels:
-    kata: "false"
-```
-
-Pods not explicitly excluded by the namespace filter are dynamically tagged to run with Kata with some exceptions.
-
-* `hostNetwork: true`
-* `rook-ceph` and `rook-ceph-system` namespaces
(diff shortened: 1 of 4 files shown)


chavafg:
LGTM, please add the `Fixes:` to the commit, so travis can pass.

krsna1729:
@mcastelino maybe this PR or follow-on can we add @ganeshmaharaj's suggestion? have a default list of namespaces, that can be overriden by the user?

krsna1729:
@mcastelino On merge the docker image will be automatically built? We should confirm that.

mcastelino:
> Looks good. One question though, we are hard-coding the black-list values in the program. Any chance it can be a user input (configmap, or other sorts) that can be merged along with what we have?

Makes sense. Config map would be ideal.

kata-webhook/README.md
```diff
@@ -30,26 +30,9 @@ and sample admission controller we created by running
...
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: rook-ceph-system
-  labels:
-    kata: "false"
-```
-
-Pods not explicitly excluded by the namespace filter are dynamically tagged to run with Kata with some exceptions.
+Pods are dynamically tagged to run with Kata based policies setup with the webhook.
```
grahamwhaley: Can I check the context of this paragraph is still correct? In my head it used to read for instance 'we exclude rook', and now I read it as 'we include rook' ?

kata-webhook/README.md
```diff
@@ -30,26 +30,9 @@ and sample admission controller we created by running
...
+These current include any pods with the following properties
```
jodh-intel: Nit: Add colon at end of line.

kata-webhook/README.md
mcastelino: Now that I read it I do not quite understand what I was trying to say.

In short with this the webhook can be setup to ignore pods based on the namespace, net==host or some other property that we choose to filter on. All other pods will run with Kata. Let me reword this to make it clearer.

PR #1263, opened 2019-03-05
ci: Add installation script for SLES

### LINKED ISSUE #1262: Add SLES setup script
Add a installation script for SLES

GabyCT:
This will perform the installation for SLES.

Fixes #1262

### DIFF
--- a/.ci/setup_env_sles.sh
+++ b/.ci/setup_env_sles.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+cidir=$(dirname "$0")
+source "/etc/os-release" || source "/usr/lib/os-release"
+source "${cidir}/lib.sh"
+
+echo "Add repo for perl-IPC-Run"
+perl_repo="https://download.opensuse.org/repositories/devel:languages:perl/SLE_${VERSION//-/_}/devel:languages:perl.repo"
+sudo -E zypper addrepo --no-gpgcheck ${perl_repo}
+sudo -E zypper refresh
+
+echo "Install perl-IPC-Run"
+sudo -E zypper -n install perl-IPC-Run
+
+echo "Add repo for moreutils"
+moreutils_repo="https://download.opensuse.org/repositories/utilities/SLE_${VERSION//-/_}_Backports/utilities.repo"
+sudo -E zypper addrepo --no-gpgcheck ${moreutils_repo}
+sudo -E zypper refresh
+
+echo "Install chronic"
+sudo -E zypper -n install moreutils
+
+echo "Install curl"
+chronic sudo -E zypper -n install curl
+
+echo "Install git"
+chronic sudo -E zypper -n install git
+
+echo "Install kata containers dependencies"
+chronic sudo -E zypper -n install libtool automake autoconf bc libpixman-1-0-devel coreutils
+
+echo "Install qemu dependencies"
+chronic sudo -E zypper -n install libcap-devel libattr1 libcap-ng-devel librbd-devel
+
+echo "Install kernel dependencies"
+chronic sudo -E zypper -n install libelf-devel flex
+
+echo "Install bison binary"
+chronic sudo -E zypper -n install bison
+
+echo "Install libudev-dev"
+chronic sudo -E zypper -n install libudev-devel
+
+echo "Install Build Tools"
+chronic sudo -E zypper -n install -t pattern "Basis-Devel" && sudo -E zypper -n install python zlib-devel
+
+echo "Install tools for metrics tests"
+chronic sudo -E zypper -n install  jq
+
+if [ "$(arch)" == "x86_64" ]; then
+	echo "Install Kata Containers OBS repository"
+	obs_url="${KATA_OBS_REPO_BASE}/SLE_${VERSION//-/_}/"
+	chronic sudo -E zypper addrepo --no-gpgcheck "${obs_url}/home:katacontainers:releases:$(arch):master.repo"
+fi
+
+echo -e "Install cri-containerd dependencies"
+chronic sudo -E zypper -n install libseccomp-devel libapparmor-devel make pkg-config
+
+echo "Install patch"
+chronic sudo -E zypper -n install patch
+
+echo "Add crudini repo"
+VERSIONID="12_SP1"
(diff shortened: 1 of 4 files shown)


.ci/setup_env_sles.sh
```diff
@@ -0,0 +1,70 @@
...
+
+echo "Install Build Tools"
+chronic sudo -E zypper -n install python zlib-devel pattern Basis-Devel
+
+echo "Install tools for metrics tests"
+chronic sudo -E zypper -n install  jq
+
+if [ "$(arch)" == "x86_64" ]; then
+	echo "Install Kata Containers OBS repository"
+	obs_url="${KATA_OBS_REPO_BASE}/SLE_12_SP3/"
```
GabyCT: @marcov is it possible to change the URL for the obs packages to get match with the info of `/etc/os-release`
```
NAME="SLES"
VERSION="12-SP3"
VERSION_ID="12.3"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP3"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp3"
```

marcov: Hi @GabyCT, that's something possible, but then we need to make sure we don't break something else like [installation instructions](https://github.com/kata-containers/documentation/blob/master/install/sles-installation-guide.md). 

For the moment, could you build the URL like this? :
`obs_url="${KATA_OBS_REPO_BASE}/SLE_${VERSION//-/_}"`

.ci/setup_env_sles.sh
```diff
@@ -0,0 +1,70 @@
...
+#
+
+set -e
+
+cidir=$(dirname "$0")
+source "/etc/os-release" || source "/usr/lib/os-release"
+source "${cidir}/lib.sh"
+
+echo "Add repo for perl-IPC-Run"
+perl_repo="https://download.opensuse.org/repositories/devel:languages:perl/SLE_12_SP3/devel:languages:perl.repo"
```
jodh-intel: Ouch - service pack versions encoded in the URL? If there really is no alternative to this, please create variables for `12` and `SP3` and use those throughout this file as this will clearly break at some point so we need to make it quick and easy to update to v99 / sp99, etc.

.ci/setup_env_sles.sh
```diff
@@ -0,0 +1,70 @@
...
+
+echo "Install perl-IPC-Run"
+sudo -E zypper -n install perl-IPC-Run
+
+echo "Add repo for moreutils"
+moreutils_repo="https://download.opensuse.org/repositories/utilities/SLE_12_SP3_Backports/utilities.repo"
+sudo -E zypper addrepo ${moreutils_repo}
+sudo -E zypper refresh
+
+echo "Install chronic"
```
grahamwhaley: just a thought - if we can, we could move chronic to be the first thing to install, and then we can use it for all the other item installs (like the perl stuff above) - presuming chronic doesn't need perl that is?

.ci/setup_env_sles.sh
marcov: Actually the `SPn` suffix for SLE does not identify strictly a service pack. 
It is better to consider it as a new OS release, since it has its own repositories and packages set.

.ci/setup_env_sles.sh
marcov:
```
$ file $(which chronic)                                                                                                                            
/usr/bin/chronic: Perl script text executable
```

.ci/setup_env_sles.sh
GabyCT: @marcov changes applied

marcov: thank you @GabyCT

jodh-intel:
@GabyCT - it's unclear as https://github.com/kata-containers/ci/issues/5 has been closed but are we about to add SLES slaves to Jenkins?

If so, I need to remove the SLES bits from https://github.com/kata-containers/documentation/pull/397.

jodh-intel:
We're good to merge but might be good to get an extra ack from @marcov given the nature of this PR :)

GabyCT:
@jodh-intel for the moment, I will not enable the jenkins jobs until this is solved https://github.com/kata-containers/tests/issues/1264

jodh-intel:
Hi @GabyCT - sorry. What I really mean is, can we actually obtain SLES licenses and make SLES work as part of the Jenkins CI?

GabyCT:
@jodh-intel, I do not have idea about the licenses maybe is something that we can ask

jodh-intel:
Hi @chavafg, @GabyCT - I'm going to merge but I'm still not clear if we can create SLES Jenkins nodes/slaves...?

chavafg:
@jodh-intel, I checked in the azure portal and there are 2 options to create an SLES VM.
1. Bring your own subscription. Which means we have a license and we add it to our running VMs. 

2. Pay as you go images. From azure portal: Pay-as-you-go-image with Standard Support, which includes updates, patches, and support through 12x5 web, email, chat and phone from Microsoft and SUSE.
Virtual machines created from this image incur per-hour support fees, in addition to Azure platform fees. An Azure support plan is required (developer or above). Support incidents are initiated through Azure Support.

I think we can use second option in case we do not get a license, although not sure if the license will be covered by the azure sponsorship. 

@GabyCT, when you tested this script by your own, which one of these did you use? 

/cc @egernst

jodh-intel:
Thanks @chavafg. Presumably there is a similar offering for RHEL?

Related: https://github.com/kata-containers/documentation/pull/397.

chavafg:
Yes, there is a Pay-as-you-go image for RHEL also.

jodh-intel:
@egernst, @gnawux, @ttx - could you clarify any of this? Specifically:

- Do we have (or can we get) RHEL+SLES licenses?
- If not, do we have a way to fund the PAYG option (2) @chavafg mentioned above?

ttx:
We do not currently have RHEL or SLES licenses. If RHEL/SLES testing is needed, and our current arrangement with those cloud providers do not include SLES/RHEL PAYG access, we should probably engage with Suse and Red Hat see if they would be willing to donate licenses so that we can do proper validation testing. We can engage at a higher level, but it's probably best to engage with Kata developers from those companies first, see if they can pull the required strings...

FWIW for OpenStack we do run tests on CentOS and OpenSUSE Leap to cover for RHEL and SLES. Last we engaged with Suse to get SLES licenses they told us using OpenSUSE Leap would be virtually identical to testing on SLES... Is there any specific reason why we'd need "real" RHEL/SLES for Kata testing?

Issue #1255, opened 2019-02-28
Add tracing test


jodh-intel:
As shown on https://github.com/kata-containers/runtime/issues/1277, we broke runtime tracing but didn't notice.

Clearly, we need a test to ensure tracing works that does atleast the following:

- Install the [Jaeger docker image](https://www.jaegertracing.io/docs/1.10/getting-started/#all-in-one).
- Enable tracing for all components.
- Run a docker workload.
- Check that Jaeger allows us to query a "reasonable number" of trace spans / component:
  - The minimum number of expected spans is going to differ per component.
  - We'd always expect `>0`.
  - The actual figures are a little difficult to tie down. We are going to require some magic numbers alas (shudder) since as the code changes, so do the number of trace spans generated.

jodh-intel:
Related:

- https://github.com/kata-containers/agent/issues/322
- https://github.com/kata-containers/agent/pull/415
- https://github.com/kata-containers/osbuilder/pull/200

chavafg:
Hi @jodh-intel, seems like your approach should work. I'll check this.

chavafg:
Hi @jodh-intel, 
I tried to collect some traces but I got the following issue:
BTW, I am using:  https://github.com/kata-containers/runtime/pull/1317
```
[fuentess@centos-prll kata-containers]$ sudo docker run -ti busybox true
docker: Error response from daemon: OCI runtime create failed: Shim tracing requires disable_new_netns for Jaeger agent communication: unknown.
```
I went and modify the `disable_new_netns = true`, then I got: 
```
[fuentess@centos-prll kata-containers]$ sudo docker run -ti busybox true
docker: Error response from daemon: OCI runtime create failed: config disable_new_netns only works with 'none' internetworking_model: unknown.
```
Modified the internetworking model to none and got the next error:
```
[fuentess@centos-prll kata-containers]$ sudo docker run -ti busybox true
docker: Error response from daemon: OCI runtime create failed: exit status 1: stdout: , stderr: time="2019-03-04T20:47:29Z" level=fatal msg="failed to add interface veth6d874f2 to sandbox: error renaming interface \"veth6d874f2\" to \"eth0\": file exists": unknown.
[fuentess@centos-prll kata-containers]$ sudo docker run -ti busybox true                                                                                                                                           
[...]

PR #1267, opened 2019-03-07
tracing: Add test to verify tracing works correctly

chavafg:
This tracing test runs as follows:
1. Enables tracing on runtime and shim components.
2. Runs a Jaeger container that will receive the trace spans.
3. A kata-container is executed.
4. Verifies that Jaeger has collected a number of spans.
5. Disables tracing.

Fixes: #1255.

### DIFF
--- a/.ci/configure_tracing_for_kata.sh
+++ b/.ci/configure_tracing_for_kata.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+[ "$#" -eq 1 ] || die "Specify enable or disable"
+
+kata_cfg_file=$(kata-runtime kata-env --json |jq '.Runtime | .Config | .Path' |cut -d\" -f2)
+
+enable_tracing() {
+	info "Enabling kata tracing on $kata_cfg_file"
+	sudo crudini --set "$kata_cfg_file" shim.kata enable_tracing true
+	sudo crudini --set "$kata_cfg_file" runtime enable_tracing true
+	sudo crudini --set "$kata_cfg_file" runtime internetworking_model \"none\"
+	sudo crudini --set "$kata_cfg_file" runtime disable_new_netns true
+	sudo crudini --set "$kata_cfg_file" netmon enable_netmon false
+}
+
+disable_tracing() {
+	info "Disabling kata tracing on $kata_cfg_file"
+	sudo crudini --set "$kata_cfg_file" shim.kata enable_tracing false
+	sudo crudini --set "$kata_cfg_file" runtime enable_tracing false
+	sudo crudini --set "$kata_cfg_file" runtime internetworking_model \"macvtap\"
+	sudo crudini --set "$kata_cfg_file" runtime disable_new_netns false
+	sudo crudini --set "$kata_cfg_file" netmon enable_netmon false
+}
+
+main() {
+	cmd="$1"
+	case "$cmd" in
+		enable ) enable_tracing ;;
+		disable ) disable_tracing ;;
+		*) die "invalid command: '$cmd'" ;;
+	esac
+}
+
+main "$@"
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,9 @@
 TIMEOUT := 60
 
 # union for 'make test'
-UNION := functional docker crio docker-compose network netmon docker-stability oci openshift kubernetes swarm vm-factory entropy ramdisk shimv2
+UNION := functional docker crio docker-compose network netmon \
+	docker-stability oci openshift kubernetes swarm vm-factory \
+	entropy ramdisk shimv2 tracing
 
 # skipped test suites for docker integration tests
 FILTER_FILE = .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh
@@ -143,6 +145,9 @@ netmon:
 	bash -f .ci/install_bats.sh
 	bats integration/netmon/netmon_test.bats
 
+tracing:
+	bash tracing/tracing-test.sh
+
 test: ${UNION}
 
 check: checkcommits log-parser
@@ -168,4 +173,5 @@ check: checkcommits log-parser
 	network \
 	ramdisk \
 	test \
+	tracing \
 	vm-factory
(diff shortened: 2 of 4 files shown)


.ci/configure_tracing_for_kata.sh
```diff
@@ -0,0 +1,42 @@
...
+
+[ "$#" -eq 1 ] || die "Specify enable or disable"
+
+default_kata_cfg_file="/usr/share/defaults/kata-containers/configuration.toml"
+kata_cfg_file=${kata_cfg_file:-$default_kata_cfg_file}
+
+enable_tracing() {
+	sudo crudini --set "$kata_cfg_file" shim.kata enable_tracing true
+	sudo crudini --set "$kata_cfg_file" runtime enable_tracing true
+	sudo crudini --set "$kata_cfg_file" runtime internetworking_model none
```
jodh-intel: Nice! :smile:

.ci/configure_tracing_for_kata.sh
```diff
@@ -0,0 +1,42 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+[ "$#" -eq 1 ] || die "Specify enable or disable"
+
+default_kata_cfg_file="/usr/share/defaults/kata-containers/configuration.toml"
```
grahamwhaley: Or you could use the `$RUNTIME_CONFIG` type elements, but probably have to call an init func from the lib.sh to get those setup, or you could use `kata-runtime kata-env --json | jq...` to extract the actual path (which is what the lib.sh does, but hopefully now uses the actual runtime configured in the docker setup these days, so in theory should pick up the correct runtime and thus the correct config file).

Makefile
```diff
@@ -143,6 +145,9 @@ netmon:
 	bash -f .ci/install_bats.sh
 	bats integration/netmon/netmon_test.bats
 
+tracing:
+	bash tracing/tracing-test.sh
```
grahamwhaley: my brain wants to slap a `chronic` in here ;-)

tracing/tracing-test.sh
```diff
@@ -0,0 +1,247 @@
...
+	local status=""
+
+	while [ $attempt -lt 10 ]
+	do
+		status=$(curl -s "http://${jaeger_server}:${jaeger_ui_port}/api/traces?service=${service}" 2>/dev/null)
+		local ret=$?
+
+		[ "$ret" -eq 0 ] && [ -n "$status" ] && break
+
+		attempt=$((attempt+1))
```
grahamwhaley: I think you can just `((attempt++)` here :-)

tracing/tracing-test.sh
```diff
@@ -0,0 +1,247 @@
...
+		sleep 0.1
```
grahamwhaley: This is only giving 10x0.1 == 1s for jaeger to come up - seems a bit short maybe?

tracing/tracing-test.sh
```diff
@@ -0,0 +1,247 @@
...
+	done
+
+	echo "$status"
+}
+
+# Look for any "dangling" spans that have not been reported to the Jaeger
+# agent.
+check_missing_spans()
```
grahamwhaley: both these jaeger checking funcs look great!

tracing/tracing-test.sh
```diff
@@ -0,0 +1,247 @@
...
+	runtime_service="kata-runtime"
+	shim_service="kata-shim"
+
+	[ -n "$1" ] && min_spans="$1"
+	[ "$min_spans" -le 0 ] && die "minimum spans must be positive"
+
+	.ci/configure_tracing_for_kata.sh enable
+	run_test "$runtime_min_spans" "$runtime_service"
+	run_test "$shim_min_spans" "$shim_service"
+	.ci/configure_tracing_for_kata.sh disable
```
grahamwhaley: I was contemplating if we needed the unconfigure of the config in a TRAP function - but, given that if this test dies, the CI will stop running anyway, I suspect we can just leave the file configured for jeager at that point, as the whole test setup will be thrown away and re-created for the next test run...

tracing/tracing-test.sh
jodh-intel: Yep - my bad - we should prolly sleep `0.5s` or `1s` per iteration.

.ci/configure_tracing_for_kata.sh
chavafg: agree, changed to use `kata-env`

Makefile
chavafg: this will only print around 10 lines, so I think we are fine :)

tracing/tracing-test.sh
chavafg: changed to 1s

tracing/tracing-test.sh
chavafg: added trap function

jodh-intel:
Hi @chavafg - tests are failing but it's difficult to see where/why:

```
18:13:58 INFO: Enabling kata tracing on /usr/share/defaults/kata-containers/configuration.toml
18:13:59 INFO: Checking runtime spans
18:13:59 the input device is not a TTY
18:14:00 jaeger
18:14:00 jaeger
18:14:00 INFO: Disabling kata tracing on /usr/share/defaults/kata-containers/configuration.toml
18:14:00 Makefile:149: recipe for target 'tracing' failed
```

I'd guess this is due to `set -o errexit` so might be worth removing that and/or running with `set -x` to get some details logged.

chavafg:
`docker run -ti` was the one complaining. I fixed it.

/test

jodh-intel:
Hmm, github shows all CIs are running but they were all aborted fwics. Restarting...

/retest

grahamwhaley:
if that happens again @jodh-intel , post a link to one and @chavafg and I can go chase it down to see if we can figure out why. aborts always worry me a little in the CI...

jodh-intel:
Oh - I'd assumed that was user-initiated. Any ideas @chavafg? Does Jenkins keep a log of such things?

grahamwhaley:
It may have been user initiated, but I think that is pretty rare. Sometimes I think it happens due to timeouts or other random events....
Jenkins does have some logs, but I'm not sure if they will help us. Sometimes it takes some staring at the tail end of the console logs to spot something, and sometimes there is nothing there to see :-(

jodh-intel:
I didn't see any timeouts but given the amount of log data, I *may* have missed them.

Is it too early to put [Jenkins pipelines](https://github.com/kata-containers/ci/issues/15) on my Christmas list to Santa? :smile:

chavafg:
@jodh-intel no idea what could have happened. I see 2 last jobs of this PR that passed. maybe the you were looking at a cached page?  
Anyway I see all green now.

jodh-intel:
@chavafg - thanks. Let's keep an eye on it...

PR #1272, opened 2019-03-07
test: Fix instability of the oci call test

### LINKED ISSUE #1271: Fix instability of the oci call test
Check the container status so we can guarantee that we are looking the oci calls with its correspondent status, also it removes unnecessary elements.

GabyCT:
This will check the container status so we can guarantee that we are
looking the oci calls with its correspondent status, also it removes
unnecessary elements.

Fixes #1271

### DIFF
--- a/integration/oci_calls/oci_call_test.sh
+++ b/integration/oci_calls/oci_call_test.sh
@@ -12,6 +12,7 @@ set -e
 
 dir_path=$(dirname "$0")
 source "${dir_path}/../../lib/common.bash"
+source "${dir_path}/../../.ci/lib.sh"
 source /etc/os-release || source /usr/lib/os-release
 
 # Save logs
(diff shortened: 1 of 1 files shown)


integration/oci_calls/oci_call_test.sh
```diff
@@ -74,16 +83,15 @@ function setup() {
...
-	# This sleep is necessary to gather the correct logs
-	sleep 10
-
 	get_time
+	local -a oci_call=( "create" "start" "state" )
 
 	# Start a container
 	docker run -d --runtime=${RUNTIME} --name=${NAME} ${IMAGE} ${PAYLOAD}
 
+	cmd="docker ps -a --filter status=running"
```
chavafg: `sudo docker...`? 
also, should we specify the name of the container? In case there are other containers running?

integration/oci_calls/oci_call_test.sh
```diff
@@ -92,16 +100,15 @@ function run_oci_call() {
...
-	# This sleep is necessary to gather the correct logs
-	sleep 10
-
 	get_time
+	local -a oci_call=( "kill" "delete" "state" )
 
 	# Stop a container
 	docker stop ${NAME}
 
+	cmd="docker ps -a --filter status=exited"
```
chavafg: same as above...

integration/oci_calls/oci_call_test.sh
GabyCT: @chavafg changes applied

Issue #1289, opened 2019-03-12
scripts: helper scripts: extract recent git history across the repos


grahamwhaley:
Sometimes it is very useful to be able to see the recent git commits across all our repos, particuarly when trying to track down recently introduced instabilities or changes in performance or functionality.

Add a small script that extracts and sorts `git log --oneline --no-merges` info and sorts it as a helper script.

PR #1290, opened 2019-03-12
metrics: cmd: history: add a repo history script

grahamwhaley:
When comparing versions with metrics, it is very handy
to be able to see what changed between versions. Let's
add a script to help us get that data.

Fixes: #1289

### DIFF
--- a/cmd/history/README.md
+++ b/cmd/history/README.md
@@ -1,0 +1,24 @@
+# history.sh
+
+A simple script to help extract git log history from the main Kata Containers
+github repos and sort them by date.
+
+This can be very helpful for instance when:
+
+- Searching or bisecting recent commits that may have caused a regression.
+- Examining recent commits that may be suitable for stable backports
+
+The script uses `git fetch` to try and have minimal impact on any existing
+repositories and state.
+
+The script uses the `$GOPATH/src/github.com/kata-containers/` repo paths.
+
+## Arguments
+
+The script can take a number of command line arguments or environment variables
+to configure a number of parameters, including:
+
+- The date range to extract commit information from.
+- The git remote and branch to extract commit information from.
+
+See the script source or `history.sh -h` for more information.
--- a/cmd/history/history.sh
+++ b/cmd/history/history.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Extract git history from the main Kata repos and generate two report files:
+#  report.txt - commits merged, sorted by repo section, and date.
+#  sorted_report.txt - all commits sorted by date
+#
+# This information can be useful when tracking down 'what changed when'.
+
+set -e
+
+REPORTFILE="report.txt"
+SORTEDREPORTFILE="sorted_${REPORTFILE}"
+
+# Set a date to start from, or use the default of '1 week'
+# Format is in any acceptible by 'Date -I -d'.
+defstart="${defstart:-last week}"
+defend="${defend:-now}"
+START=${START:-$(date -I -d "${defstart}")}
+END=${END:-$(date -I -d "$defend")}
+
+# Where are we looking? Default to origin/master
+remote="${remote:-origin}"
+branch="${branch:-master}"
+
+repo_base="github.com/kata-containers"
+repos="${repos:-
+	agent \
+	osbuilder \
+	packaging \
+	proxy \
+	runtime \
+	shim \
+	tests \
+	}"
+
+msg() {
+	local msg="$*"
+	echo "${msg}" | tee -a "${REPORTFILE}"
+}
+
+run() {
+	# Blank the file
+	echo "" > "${REPORTFILE}"
+	
+	msg "------------------------------------"
+	msg "Commits from $START to $END"
+	msg "------------------------------------"
+	for repo in $repos; do
+		repopath="${GOPATH}/src/${repo_base}/${repo}"
+		git -C "${repopath}" fetch ${remote} ${branch}
+		msg "---------- $repo ---------------"
+		TZ=UTC git -C "${repopath}" log --since "$START" --until "$END" --pretty="%cd: %h: ${repo}: %s" --date=format:"%Y-%m-%dT%H:%M:%S" --no-merges ${branch} | tee -a "${REPORTFILE}"
+		msg ""
+	done
+	
+	# And make a date sorted version of the file
+	sort -r < "${REPORTFILE}" > "${SORTEDREPORTFILE}"
+	# Remove all the 'fluff' from that sorted file, as it is now just noise.
+	sed -i '/^---.*$/d' "${SORTEDREPORTFILE}"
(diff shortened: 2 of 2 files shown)


grahamwhaley:
I've had this script hanging around in my local repo for quite some time. I use it when I'm trying to track down what changed the metrics ...
I thought I'd better upload it, as I keep forgetting where I left it ;-)

I have some suggestions then for discussion:
- It currently works on the local `GOPATH` repos - maybe I can use some 'git foo' to get the information directly from the upstream github repos, and not have a local copy at all (and not potentially change/corrupt the users local git branch states). Anybody know if I can do that? Got an example cmd? :-)
- maybe this script is OK here, or maybe we should create a `contrib` repo in kata where we can collect such little 3rd party helper scripts. I have some other scripts I use to grab the average metrics from recent jenkins builds for instance that I'd like to upload somewhere, and then improve. I know @jcvenegas has some scripts as well I think he'd like to host somewhere?

cmd/history/history.sh
```diff
@@ -0,0 +1,52 @@
...
+#  sorted_report.txt - all commits sorted by date
+#
+# This information can be useful when tracking down 'what changed when'.
+
+set -e
+
+REPORTFILE="report.txt"
+
+# Set a date to start from, or use the default of '1 week'
+#START="2019-03-08"
```
jodh-intel: Might be worth making these CLI args/options so you could do:

```sh
$ history.sh
$ history.sh [-h|--help]
$ history.sh [start]
$ history.sh [start] [end]

```

cmd/history/history.sh
```diff
@@ -0,0 +1,52 @@
...
+msg "===================================="
+msg "Commits from $START to $END"
+msg "===================================="
+for repo in $repos; do
+	repopath="${GOPATH}/src/$repo"
+	(go get -d -u ${repo} || true)
+	git -C "${repopath}" checkout master
+	git -C "${repopath}" pull >/dev/null
+	msg "---------- $repo ---------------"
+	TZ=UTC git -C "${repopath}" log --since "$START" --until "$END" --pretty="%h: %cd: ${repo##*/}: %s" --date=format:"%Y-%m-%dT%H:%M:%S" --no-merges | tee -a "${REPORTFILE}"
```
jodh-intel: - It might be better to put the date field first.
- How about also including tags somewhere?

cmd/history/history.sh
```diff
@@ -0,0 +1,52 @@
...
+	msg ""
+done
+
+sort -r -k 2 < "${REPORTFILE}" > "sorted_${REPORTFILE}"
```
jodh-intel: Could you remove `grep -v` the blank lines and `---- $repo ----` lines?

cmd/history/history.sh
jodh-intel: You could avoid the `##*/` by building the github url inside the loop using an array. Something like:

```sh
repos=()
repos+=("runtime")
repos+=("shim")
  :

for repo in "${repos[@]}"
do
    github_repo="${repo_base}/${repo}"
    repopath="${GOPATH}/src/${github_repo}"
        :
  
done
```

cmd/history/history.sh
grahamwhaley: I guess... I could make this a full blow tool with help and allll. I guess.... more typing....

cmd/history/history.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Blank the file
+echo "" > "${REPORTFILE}"
+
+msg "===================================="
+msg "Commits from $START to $END"
+msg "===================================="
+for repo in $repos; do
+	repopath="${GOPATH}/src/$repo"
+	(go get -d -u ${repo} || true)
```
grahamwhaley: So, I'm thinking some `git fetch` and `git log remotes/master` or similar is going to be nicer here, so we don't destroy or disrupt (or fail due to) any user state in their go repos. The script mostly works for me as-is just due to my current workflow, but I'm aware it probably won't work for others.
Let me see if I can update that soon then.

cmd/history/history.sh
jodh-intel: Whilst you're at it, could you also add i18n support... oh and colour? And sound. Also, how about a gui?

/me runs...

jodh-intel: ... and a `--display-ascii-unicorns` option just for fun :unicorn:

raravena80:
@grahamwhaley ping, any updates?

jodh-intel:
Ping @grahamwhaley :)

chavafg:
Any update on this @grahamwhaley ?

cmd/history/README.md
```diff
@@ -0,0 +1,12 @@
+# Kata repo git history
+
+This `history.sh` script checks out and switches to master branch
```
jodh-intel: s/to master/to the `master`/

grahamwhaley:
/test
@jodh-intel - updated - I think you'll like it better now ;-)
@jcvenegas @ganeshmaharaj - you might find it useful when looking at stable branches etc. as well?

cmd/history/history.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Extract git history from the main kata repos and generate two report files:
```
jodh-intel: Nit: "Kata".

cmd/history/history.sh
```diff
@@ -0,0 +1,123 @@
...
+# This information can be useful when tracking down 'what changed when'.
+
+set -e
+
+REPORTFILE="report.txt"
+SORTEDREPORTFILE="sorted_${REPORTFILE}"
+
+# Set a date to start from, or use the default of '1 week'
+# Format is in any acceptible by 'Date -I -d'.
+defstart="last week"
```
jodh-intel: Worth making these `defstart=${defstart:-last week}` for consistency with the others maybe?

cmd/history/history.sh
```diff
@@ -0,0 +1,123 @@
...
+END=${END:-$(date -I -d "$defend")}
+
+# Where are we looking? Default to origin/master
+remote="${remote:-origin}"
+branch="${branch:-master}"
+
+repo_base="github.com/kata-containers"
+repos="${repos:-runtime \
+	proxy \
+	shim \
```
jodh-intel: Micro-nit: unsorted list :smile:

cmd/history/history.sh
grahamwhaley: sure. I didn't as I was thinking folks would override the `START` etc., but no harm, so adding.

cmd/history/history.sh
grahamwhaley: I've not added tags yet - I was going to look, but forget ;-)
maybe we can land without to start with?

grahamwhaley:
thanks for the feedback @jodh-intel updated and pushed
/test

cmd/history/history.sh
jodh-intel: wfm - already approved ;)

PR #1301, opened 2019-03-14
metrics: Skip KSM to settle down with runc or firecracker

### LINKED ISSUE #1164: metrics: ksm: can skip settle detection in certain circumstances
the metrics tests ksm settle detection code (https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L281) could probably figure out that there are no MERGEABLE pages mapped in, and hence KSM is idle, and will never settle. In this case, it could skip the settle wait, saving us upto 5 minutes when running the report generator for instance.
I think you can work it out using the /proc KSM files that we already read some of in the idle loop.
This is particularly relevant when running runtimes that do not have MERGEABLE pages, like `runc` and `kata-fc`.

marking as something easy for somebody to pick up.

One day I'll probably get bored waiting for the tests, and fix it whilst I'm waiting, if nobody beats me to it ;-)

GabyCT:
When we have enable KSM on density metrics we can skip the part where
we are waiting to KSM to settle down for runc or firecracker.

Fixes #1164

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -284,6 +284,7 @@ check_for_ksm(){
 #
 # arg1 - timeout in seconds
 wait_ksm_settle(){
+	[[ "$RUNTIME" == "runc" ]] || [[ "$RUNTIME" == "kata-fc" ]] && return
 	local t pcnt
 	local oldscan=-1 newscan
 	local oldpages=-1 newpages
@@ -295,12 +296,15 @@ wait_ksm_settle(){
 	for ((t=0; t<$1; t++)); do
 
 		newscan=$(cat /sys/kernel/mm/ksm/full_scans)
+		newpages=$(cat /sys/kernel/mm/ksm/pages_shared)
+		[[ "$newpages" -eq 0 ]] && echo "No need to wait for KSM to settle" && return
+
 		if (( newscan != oldscan )); then
 			echo -e "\nnew full_scan ($oldscan to $newscan)"
 
-			newpages=$(cat /sys/kernel/mm/ksm/pages_shared)
 			# Do we have a previous scan to compare with
 			echo "check pages $oldpages to $newpages"
+
 			if (( oldpages != -1 )); then
 				# avoid divide by zero problems
 				if (( $oldpages > 0 )); then


metrics/lib/common.bash
```diff
@@ -284,48 +284,50 @@ check_for_ksm(){
...
-					pcnt=$(( 100 - ((newpages * 100) / oldpages) ))
-					# abs()
-					pcnt=$(( $pcnt * -1 ))
-
-					echo "$oldpages to $newpages is ${pcnt}%"
-
-					if (( $pcnt <= 5 )); then
-						echo "KSM stabilised at ${t}s"
-						return
+	if [ "$RUNTIME" = "kata-runtime" ] || [ "$RUNTIME" == "kata-qemu" ]; then
```
grahamwhaley: Just a thought - we may have been able to avoid adding more indentation here by doing a 'negative check and quick return', like:

```bash
 if [ "$RUNTIME" != "kata-runtime"] && return
```

grahamwhaley: I was really hoping there was a way to ask KSM now many pages or zones had been registered as MERGEABLE, and thus to quickly ask it if there was any point waiting for it to settle - but, afaict there is no one easy node to get that information from :-( So, maybe this hard-ish wired runtime check is the easiest way. But, two ideas:
 - we could check for the runtimes we know *not* to check, which means we default to the 'safe but slow' route for any future runtimes or runtime renames that are added'
 - we might be able to wait for the first KSM 'full run' to complete, and then check if any pages have been found at all, and if they have not, then quit at that point. I'll see if I can annotate what I mean below...

metrics/lib/common.bash
```diff
@@ -284,48 +284,50 @@ check_for_ksm(){
...
+
+			newscan=$(cat /sys/kernel/mm/ksm/full_scans)
+			if (( newscan != oldscan )); then
+				echo -e "\nnew full_scan ($oldscan to $newscan)"
+
+				newpages=$(cat /sys/kernel/mm/ksm/pages_shared)
+				# Do we have a previous scan to compare with
+				echo "check pages $oldpages to $newpages"
+				if (( oldpages != -1 )); then
+					# avoid divide by zero problems
```
grahamwhaley: Somewhere around here, I think we maybe can code up that 'we know we have done a complete ksm scan', and that if there have been no pages found to merge yet, then we could probably fairly safely assume the test is not utilising KSM, and thus can quit early.
wdyt @GabyCT - I think it will take a little bit of concentration to ensure we get the right spot here to do that check...

GabyCT:
@grahamwhaley changes applied

GabyCT:
I tested with runc and I noticed that `newpages` stays always 0 and that is an indicator that KSM will not settle and it will timeout without seeing changes.

chavafg:
ping @grahamwhaley, ok with the changes?

metrics/lib/common.bash
```diff
@@ -295,12 +296,15 @@ wait_ksm_settle(){
 	for ((t=0; t<$1; t++)); do
 
 		newscan=$(cat /sys/kernel/mm/ksm/full_scans)
+		newpages=$(cat /sys/kernel/mm/ksm/pages_shared)
+		[[ "$newpages" -eq 0 ]] && echo "No need to wait for KSM to settle" && return
+
 		if (( newscan != oldscan )); then
 			echo -e "\nnew full_scan ($oldscan to $newscan)"
```
GabyCT: @grahamwhaley if I move the newpages check to 0 here at the 'newscan' loop the comparison even if I got a 0 value for newscan the loop will not end and it will timeout waiting for KSM to settle, that is the reason why I put it before

metrics/lib/common.bash
```diff
@@ -295,12 +296,15 @@ wait_ksm_settle(){
 	for ((t=0; t<$1; t++)); do
 
 		newscan=$(cat /sys/kernel/mm/ksm/full_scans)
+		newpages=$(cat /sys/kernel/mm/ksm/pages_shared)
+		[[ "$newpages" -eq 0 ]] && echo "No need to wait for KSM to settle" && return
```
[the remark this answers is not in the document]
GabyCT: @grahamwhaley if I move the newpages check to 0 inside the 'newscan' loop the comparison even if I got a 0 value for newscan the loop will not end and it will timeout waiting for KSM to settle, that is the reason why I put it before

metrics/lib/common.bash
grahamwhaley: Ah, OK - understood.
In that case, let's land as is, and see how it goes...

PR #1307, opened 2019-03-14
makefile: honor FOCUS environment variable

### LINKED ISSUE #1306: integration/docker: FOCUS envar is not honored
`FOCUS` envar is not honored because all serial test are started after all the tests that match with `FOCUS` finish, for example,  `RUNTIME=kata-runtime FOCUS="CPU" make -e docker` runs fist the cpu tests and then all the serial tests

devimc:
Run sequentially only the tests that match with `FOCUS`.
The `FOCUS` environment variable is used to debug and put attention on
certain tests.

fixes #1306


devimc:
CI broken, this patch is required to fix one part of the CI, instead of having different PRs, let's move all the fixes in a single PR, see https://github.com/kata-containers/tests/pull/1309

PR #1313, opened 2019-03-18
integration: cri-o: Skip 'ctr stats'

### LINKED ISSUE #1311: integration: cri-o: Skip 'ctr stats'
With the recent fix of the bats from CRI-O https://github.com/kubernetes-sigs/cri-o/pull/2054, the test `ctr stats` does not work with Kata Containers. This is actually an expected behavior since we know stats don't work with Kata v1. Once we get to Kata v2, this will be solved, but for now, we have to skip this test.

sboeuf:
Based on recent fix of the CRI-O bats 'ctr stats' from this following
PR https://github.com/kubernetes-sigs/cri-o/pull/2054, this test is
failing for Kata since Kata v1 does not support proper stats.

This is something that needs to be fixed once the CI for Kata v2 will
be in place.

For now, let's simply skip this test.

Fixes #1311

### DIFF
--- a/integration/cri-o/crio_skip_tests.sh
+++ b/integration/cri-o/crio_skip_tests.sh
@@ -9,4 +9,5 @@
 
 declare -a skipCRIOTests=(
 'test "ctr oom"'
+'test "ctr stats"'
 );


sboeuf:
FYI, metrics CI can be considered as passing since we hit 85% of the expected value. This is expected since we recently improved Kata footprint.

sboeuf:
@chavafg @GabyCT let's try to merge this quickly to unblock the CRI-O/Kata CI.

PR #1319, opened 2019-03-19
cri-o: fix name of skipped test

chavafg:
The name of the test is 'ctr stats output'.
Add the complete name of the test in the list instead
of only part of it.

Fixes: #1311.


sboeuf:
@chavafg thanks for fixing this, and sorry I didn't realize...

Issue #1345, opened 2019-03-23
agent and runtime cannot pass travis-ci after switch to golangci-lint


teawater:
After merge https://github.com/kata-containers/tests/pull/1315
travis-ci failed in:
https://github.com/kata-containers/runtime/pull/1409
https://github.com/kata-containers/runtime/pull/1410
https://github.com/kata-containers/agent/pull/493
```
=== CONT  TestSourcesFromTestdataWithIssuesDir/depguard.go
--- FAIL: TestSourcesFromTestdataWithIssuesDir (0.30s)
	linters_test.go:38: testdata/*.go
    --- FAIL: TestSourcesFromTestdataWithIssuesDir/ineffassign.go (0.26s)
    	linters_test.go:137: [run --disable-all --print-issued-lines=false --print-linter-name=false --out-format=line-number -Eineffassign --no-config testdata/ineffassign.go]
    	require.go:794: 
    			Error Trace:	linters_test.go:34
    			            				linters_test.go:138
    			            				linters_test.go:54
    			Error:      	Received unexpected error:
    			            	
    			            	unmatched errors
    			            	2019/03/23 00:41:15 unknown architecture wasm
    			Test:       	TestSourcesFromTestdataWithIssuesDir/ineffassign.go
    --- FAIL: TestSourcesFromTestdataWithIssuesDir/deadcode.go (0.26s)
...
...
```

teawater:
Just reproduced the issue in my part.
Upstream golangci-lint cannot pass test with go1.10.8.  But OK with go1.12.1.

teawater:
```
  golangci-lint:
    description: "utility to run various golang linters"
    url: "github.com/golangci/golangci-lint"
    # Using HEAD for now as v1.15.0 does not compile with golang 1.12+. Will be
    # pegged to the next release once it comes out.
    # https://github.com/kata-containers/tests/issues/1329
    version: "HEAD"
```
golangci-lint v1.15.0 cannot work with golang 1.12+
golangci-lint HEAD cannot work with go1.10.8

Issue #1344, opened 2019-03-22
yq_linux_s390x/ppc64le release is missing 


alicefr:
The project https://github.com/mikefarah/yq has released a new version 2.3 but the yq_linux_s390x and also yq_linux_ppc64le are missing. The ci script installs the latest version and is failing on s390x because the file is missing. I already opened an issue in the project. See https://github.com/mikefarah/yq/issues/225

/cc @nitkon

alicefr:
@nitkon opened a similar issue for power. See https://github.com/mikefarah/yq/issues/224

alicefr:
The latest releases for power and z has been updated

PR #1346, opened 2019-03-23
lint: set golangci-lint.version to d278457390d8c314192c08a7024e4648c4…

teawater:
…0883af

Using d278457390d8c314192c08a7024e4648c40883af for now as
v1.15.0 does not compile with golang 1.12+ and HEAD cannot
pass test with golang 1.10.8.
Will be pegged to other version once there is a verion
can work OK with golang 1.12+ and golang 1.10.8.

Fixes: #1345

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -52,7 +52,11 @@ externals:
   golangci-lint:
     description: "utility to run various golang linters"
     url: "github.com/golangci/golangci-lint"
-    # Using HEAD for now as v1.15.0 does not compile with golang 1.12+. Will be
-    # pegged to the next release once it comes out.
+    # Using d278457390d8c314192c08a7024e4648c40883af for now as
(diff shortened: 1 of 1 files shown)


caoruidong:
lgtm. But got 'yq not found' error

gnawux:
@caoruidong that's another weird issue, only on ppc64le platform

gnawux:
@caoruidong that's another weird issue, only on ppc64le platform

gnawux:
travis on ppc64le failed because of #1344 

others good, merge this PR

ganeshmaharaj:
Should we move our base golang version to 1.11? Are there still use cases where 1.10 is needed?

gnawux:
@ganeshmaharaj good point.

I think we should use a newer stable golang version by default, and keep more old versions being supported if we could.

Issue #1347, opened 2019-03-24
cannot pass vm_templating_test.sh if kata-runtime is not pathname in docker config


teawater:
```
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# cat /etc/docker/daemon.json
{
  "registry-mirrors": ["https://p2ynl829.mirror.aliyuncs.com"],
  "default-runtime": "runc",
  "runtimes": {
    "kata": {
      "path": "kata-runtime"
    },
    "kata-runtime": {
      "path": "kata-runtime"
  }
}
}

root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# bash -f integration/vm_factory/vm_templating_test.sh
++ dirname integration/vm_factory/vm_templating_test.sh
+ cidir=integration/vm_factory
+ source integration/vm_factory/../../metrics/lib/common.bash
+++ readlink -f integration/vm_factory/../../metrics/lib/common.bash
++ THIS_FILE=/mnt/hgfs/gopath/src/github.com/kata-containers/tests/metrics/lib/common.bash
++ LIB_DIR=/mnt/hgfs/gopath/src/github.com/kata-containers/tests/metrics/lib
++ RESULT_DIR=/mnt/hgfs/gopath/src/github.com/kata-containers/tests/metrics/lib/../results
++ source /mnt/hgfs/gopath/src/github.com/kata-containers/tests/metrics/lib/../../lib/common.bash
+++ VC_POD_DIR=/var/lib/vc/sbs
+++ KATA_HYPERVISOR=qemu
++ source /mnt/hgfs/gopath/src/github.com/kata-containers/tests/metrics/lib/json.bash
+++ declare -a json_result_array
+++ declare -a json_array_array
++ source /etc/os-release
+++ NAME=Ubuntu
+++ VERSION='16.04.6 LTS (Xenial Xerus)'
+++ ID=ubuntu
+++ ID_LIKE=debian
+++ PRETTY_NAME='Ubuntu 16.04.6 LTS'
+++ VERSION_ID=16.04
+++ HOME_URL=http://www.ubuntu.com/
+++ SUPPORT_URL=http://help.ubuntu.com/
+++ BUG_REPORT_URL=http://bugs.launchpad.net/ubuntu/
+++ VERSION_CODENAME=xenial
+++ UBUNTU_CODENAME=xenial
++ KATA_KSM_THROTTLER=no
++ DOCKER_EXE=docker
[...]

PR #1348, opened 2019-03-24
lib/common: extract_kata_env: Convert $rpath to pathname

teawater:
Cannot pass vm_templating_test.sh if kata-runtime is not pathname
in docker config because extract_kata_env determines $rpath
cannot execute.

Convert $rpath to pathname to handle the issue.

Fixes: #1347

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -61,6 +61,7 @@ get_docker_kata_path(){
 extract_kata_env(){
 	local toml
 	local rpath=$(get_docker_kata_path "$RUNTIME")
+	rpath=$(command -v "$rpath")
 
 	# If we can execute the path handed back to us
 	if [ -x "$rpath" ]; then


chavafg:
should we then delete the `get_docker_kata_path()` function, which no longer make sense to have it?

grahamwhaley:
heh, posted on the Issue not the PR - here is a copy - let's continue the chat here:
-----

@chavafg - I'm not sure we should drop that function - it asks `docker` to give us the actual name/path of the runtime it is using. It is the most accurate method we have to determine exactly what binary is being used - iirc, I added it when we added static builds of kata - as they are not in the path either.
@teawater - so, you have a binary name registered with `docker` with no path associated with it? I guess it works - but, what are the `docker` search rules to find that binary in that case? Do they match the same rules as `command -v`.

I think the change is good btw - it will try to find the correct binary in the case that:
- it does not have a fully qualified path in the docker config files
- it is not on the `PATH`

I don't think I've seen a case before where the `docker` runtime path was not a fully qualified path though ;-)

chavafg:
ok, then disregard my comment above...

PR #1351, opened 2019-03-25
static-checks: Exclude systemd unit files from SPDX license check

### LINKED ISSUE #1350: Exclude systemd units from SPDX license checks

amshinde:
This is required for systemd target and service files we have in
agent repo. While at it, adding all systemd units types here.

Fixes #1350

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -310,6 +310,12 @@ check_license_headers()
 		--exclude="*.yaml" \
 		--exclude="*.pb.go" \
 		--exclude="*.gpl.c" \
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Can we not add the headers? They are copyrightable files I presume?? And, systemd does support comment fields afaict:
https://www.freedesktop.org/software/systemd/man/systemd.syntax.html

> Empty lines and lines starting with "#" or ";" are ignored, which may be used for commenting.

amshinde:
@grahamwhaley Fair enough. I'll close this one

Issue #1356, opened 2019-03-26
lib: extract_kata_env fails when docker service is not up 


chavafg:
when running `extract_kata_env` we shouldn't rely on docker output as it may not be running.

PR #1357, opened 2019-03-26
lib: extract_kata_env: don't rely on docker output

chavafg:
We are getting the runtime path (rpath) from docker info, but
it may be the case where we try to run `extract_kata_env` and
not running docker, so we shouldn't fail to get the kata
runtime path if docker service is down or not healthy.

Fixes: #1356.

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -61,7 +61,9 @@ get_docker_kata_path(){
 extract_kata_env(){
 	local toml
 	local rpath=$(get_docker_kata_path "$RUNTIME")
-	rpath=$(command -v "$rpath")
+	if [ -n "$rpath" ]; then
+		rpath=$(command -v "$rpath")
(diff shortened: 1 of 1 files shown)


lib/common.bash
```diff
@@ -61,7 +61,9 @@ get_docker_kata_path(){
 extract_kata_env(){
 	local toml
 	local rpath=$(get_docker_kata_path "$RUNTIME")
-	rpath=$(command -v "$rpath")
+	if [ -n "$rpath" ]; then
+		rpath=$(command -v "$rpath")
+	fi
 
 	# If we can execute the path handed back to us
 	if [ -x "$rpath" ]; then
```
grahamwhaley: ah, I thought I'd already done something similar ;-) - but, with the new addition, I guess not good/early enough now...

PR #1360, opened 2019-03-26
test: Add K8s tests for sysctl

### LINKED ISSUE #1339: Add sysctls with Kubernetes test

GabyCT:
This sets sysctls for a kubernetes pod.

Fixes #1339

### DIFF
--- a/integration/kubernetes/k8s-sysctls.bats
+++ b/integration/kubernetes/k8s-sysctls.bats
@@ -1,0 +1,32 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+setup() {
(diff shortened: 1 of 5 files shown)


integration/kubernetes/k8s-sysctls.bats
```diff
@@ -0,0 +1,31 @@
...
+@test "Setting sysctl" {
+	# Create pod
+	kubectl apply -f "${pod_config_dir}/pod-sysctl.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready pod "$pod_name"
+
+	# Check sysctl configuration
+	cmd="cat /proc/sys/kernel/shm_rmid_forced"
+	kubectl exec $pod_name -- sh -c "$cmd" | grep 0
```
jodh-intel: This looks correct, but why not anchor to guarantee we're getting what we expect? For example, what if -- somehow -- this file contained the value `10`?

Maybe something like:

```sh
result=$(kubectl exec $pod_name -- sh -c "$cmd")
[ "${result}" = 0 ]
```

GabyCT: @jodh-intel changes applied

Issue #1358, opened 2019-03-26
k8s: add test to verify kata support for block volumes


chavafg:
related:
https://github.com/kata-containers/runtime/issues/1354

PR #1362, opened 2019-03-26
k8s: add test to verify block volume support

chavafg:
This test creates a block loop device and passes it
to a pod as a persistent volume. Then we create a
file system on the device inside the container and
write to it.

Fixes: #1358.

### DIFF
--- a/integration/kubernetes/k8s-block-volume.bats
+++ b/integration/kubernetes/k8s-block-volume.bats
@@ -1,0 +1,77 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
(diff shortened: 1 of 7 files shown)


integration/kubernetes/k8s-block-volume.bats
```diff
@@ -0,0 +1,77 @@
...
+	export KUBECONFIG="$HOME/.kube/config"
+	get_pod_config_dir
+
+	pod_name="pod-block-pv"
+	ctr_dev_path="/dev/xda"
+	vol_capacity="500M"
+
+	# Create Loop Device
+	tmp_disk_image=$(mktemp --tmpdir disk.XXXXXX.img)
+	sudo truncate "$tmp_disk_image" --size 500M
```
jodh-intel: Maybe:

```sh
sudo truncate "$tmp_disk_image" --size "${vol_capacity}"
```

Issue #1365, opened 2019-03-28
gocache: clean GOCACHE on AArch64


Pennyzct:
Go command caches build outputs under `$GOCACHE`, The default location for cache data is a subdirectory named `go-build/` in the standard user cache directory, like `$XDG_CACHE_HOME` in linux.
Since some CI, including ARM CI, is running on bare metal. We need to clean` GOCACHE` before each run.

PR #1366, opened 2019-03-28
gocache: clean GOCACHE on AArch64

Pennyzct:
Go command caches build outputs under `$GOCACHE`, The default location for cache data is a subdirectory named `go-build/` in the standard user cache directory, like `$XDG_CACHE_HOME` in linux.
Since some CI, including ARM CI, is running on bare metal. We need to clean` GOCACHE` before each run.

Fixes: #1365

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -317,5 +317,14 @@ gen_clean_arch() {
 	fi
 	info "Remove Kata package repo registrations"
 	delete_kata_repo_registrations
+
+	info "Clean GOCACHE"
+	if command -v go > /dev/null; then
+		GOCACHE=${GOCACHE:-$(go env GOCACHE)}
+	else
+		# if go isn't installed, try default dir
+		GOCACHE=${GOCACHE:-$HOME/.cache/go-build}
+	fi
+	[ -d "$GOCACHE" ] && sudo rm -rf ${GOCACHE}/*
 }
 


.ci/lib.sh
```diff
@@ -317,5 +317,14 @@ gen_clean_arch() {
 	fi
 	info "Remove Kata package repo registrations"
 	delete_kata_repo_registrations
+
+	info "Clean GOCACHE"
+	if command -v go > /dev/null; then
+		GOCACHE=${GOCACHE:-$(go env GOCACHE)}
+	else
+		# if go isn't installled, try default dir
```
jodh-intel: Nit: typo: "installed".

Pennyzct: nice catch! thanks. ;)

Pennyzct:
Hi~ @chavafg @grahamwhaley @jodh-intel For jenkins-ci-opensuse-15 failure `ERROR: Error cloning remote repo 'origin'`, i get a similar one in my another pr [runtime/#1433](https://github.com/kata-containers/runtime/pull/1433), looks like a pattern, you may want to have a look. ;)

.ci/lib.sh
Pennyzct: updated, ptal. ;)

grahamwhaley:
Hi @Pennyzct - I think the opensuse clone issue is a known problem right now - I think we can ignore that on this PR.

grahamwhaley:
We have a few CI fails, but, this change should only affect bare metal machines, which means pretty much ARM, IBM and the metrics. The only one of those CIs to fail is the metrics, and that was happy with this code change (it failed as we need to adjust the metrics bounds check due to recent other code changes).

```
02:41:51 OK
02:41:51 INFO: Clean GOCACHE
02:41:51 ~/workspace/workspace/kata-metrics-tests-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests ~/workspace/workspace/kata-metrics-tests-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests
```

So, I'm happy this code change is good - if somebody else agrees, let's merge it.

chavafg:
Hi @Pennyzct, yes, we are having problems with the opensuse job, we can ignore it. Restarted the ubuntu and fedora job which failed on the shimv2 tests.

Issue #1373, opened 2019-03-29
clean-up: clean stale kata resource on ARM CI


Pennyzct:
Directories `/usr/share/kata-containers` and `/usr/share/defaults/kata-containers` must be cleaned up before each run.
PR is on the way. ;)

PR #1374, opened 2019-03-29
clean-up: clean stale kata resource on ARM CI

Pennyzct:
Logging into ARM CI, Found out quite a big surprise there.
```
root@testing-1:/usr/share# du -sh -- *
142G    kata-containers
```
I suppose, it won't cost them too much time to eat up my whole `/`
Directories `/usr/share/kata-containers` and `/usr/share/defaults/kata-containers` must be cleaned up before each run.

Fixes: #1373


grahamwhaley:
Again, metrics CI bare metal is happy with this code change (it failed on the bounds check).
I'm happy if we merge this with the ARM CI passing and the known metrics fail. (the change only applies to bare metal servers).

Issue #1381, opened 2019-03-30
Speed up travis-ci


teawater:
Travis-ci become slow after lint was changed to golangci-lint because golangci-lint will do static check with itself when it is built.
And there is not a target that just built but doesn't static check itself in its makefile.
https://github.com/golangci/golangci-lint/blob/master/Makefile

So I submitted a PR https://github.com/golangci/golangci-lint/pull/459 to add a build target to its makefile.
Travis-ci can use this target build golangci-lint to speed up itself.

teawater:
https://github.com/golangci/golangci-lint/pull/459 was merged.

PR #1427, opened 2019-04-08
lint: set golangci-lint.version to v1.16.0

teawater:
I would like to handle issue #1381 in 3 steps:
1 step is change golangci-lint.version to introduce v1.16.0 to introduce 8319caf63f0f5fd38b2439c7379823d656fb404e(Makefile: Add target build).
2 step is wait 1 or 2 weeks to make sure golangci-lint v1.16.0 is work OK for each project in multi-arch.
3 step add a patch to change golangci-lint build command from "make" to "make build".

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -52,11 +52,4 @@ externals:
   golangci-lint:
     description: "utility to run various golang linters"
     url: "github.com/golangci/golangci-lint"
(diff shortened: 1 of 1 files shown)


devimc:
cc @ganeshmaharaj

Issue #1438, opened 2019-04-09
ci: install parallel for cri-o tests


chavafg:
cri-o is parallelizing its tests and needs parallel installed.

PR #1439, opened 2019-04-09
ci: Install gnu/parallel

chavafg:
cri-o is parallelizing its tests and needs gnu/parallel.

Fixes: #1438.


chavafg:
on the ones we build is because the package is not available on their repos.

Issue #1095, opened 2019-01-28
metrics: report: elasticsearch density: scripts fail to find results


grahamwhaley:
the [metrics report](https://github.com/kata-containers/tests/tree/master/metrics/report) is failing to pick up the [elastic search density results](https://github.com/kata-containers/tests/blob/master/metrics/report/grabdata.sh#L120-L126), because when we [pinned the elasticsearch version](https://github.com/kata-containers/tests/pull/871), that changed the name format of the results file (that now includes the pinned version of the container under test):

```bash
$ ls results/*elastic*
results/footprint-elasticsearch:6.4.0-ksm.json
```

but the script [only looks for a vanilla `footprint-elasticsearch[-ksm].json`](https://github.com/kata-containers/tests/blob/master/metrics/report/report_dockerfile/footprint-density.R#L18).

We have a few options to fix this:
- modify the R to do a wildcard search
- modify the [json saving code](https://github.com/kata-containers/tests/blob/master/metrics/lib/json.bash#L40) to strip out container/image versions
- modify the tests to [not include any version information in their teat names](https://github.com/kata-containers/tests/blob/master/metrics/density/footprint_data.sh#L88)

Although I'd quite like to not have versions in the results file names, having them there does make it clear we pinned the image. I'd probably opt to fix the R search routine to cater for wildcard matching. Up for debate though if somebody wants.

PR #1447, opened 2019-04-10
metrics: report: handle pinned container names

grahamwhaley:
The footprint density report script did not know how to handle the
container version turning up in the results (which happens if we
pin a container rather than use 'latest', as we do with elastic).

Add wildcard file searching and the appropriate de-nesting of data
to the script to find and use all 'XXX.*\.json' results files it
finds for each container name.

Fixes: #1095


grahamwhaley:
we may not need to 'test' this with CI, as the CIs do not run this code - the basic static checks should be fine I think.

chavafg:
I think we still need to run CI as we are now not allowed to merged without having the `Required` items passed.
/test

Issue #1426, opened 2019-04-08
Update vm_templating_test.sh to check store related vm template leak


teawater:
https://github.com/kata-containers/runtime/issues/1452

PR #1452, opened 2019-04-11
vm_factory: Add check for storage related vm template leak

teawater:
According to the issue in
https://github.com/kata-containers/runtime/issues/1452,
There is storage related vm template leak in /var/lib/vc/sbs and
/run/vc/sbs that is fixed in
https://github.com/kata-containers/runtime/pull/1491.

Add check for this issue.

Fixes: #1426

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -10,6 +10,9 @@
 # Place where virtcontainers keeps its active pod info
 VC_POD_DIR="${VC_POD_DIR:-/var/lib/vc/sbs}"
 
+# Sandbox runtime directory
+RUN_SBS_DIR="${RUN_SBS_DIR:-/run/vc/sbs}"
+
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 
 die() {
@@ -155,17 +158,24 @@ check_processes() {
 	done
 }
 
-# Checks that pods were not left
-check_pods() {
-	if [ -d ${VC_POD_DIR} ]; then
+# Checks that pods were not left in a directory
+check_pods_in_dir() {
+    local DIR=$1
+    if [ -d ${DIR} ]; then
 		# Verify that pods were not left
-		pods_number=$(ls ${VC_POD_DIR} | wc -l)
+		pods_number=$(ls ${DIR} | wc -l)
 		if [ ${pods_number} -ne 0 ]; then
-			die "${pods_number} pods left and found at ${VC_POD_DIR}"
+            ls ${DIR}
+			die "${pods_number} pods left and found at ${DIR}"
 		fi
 	else
-		echo "Not ${VC_POD_DIR} directory found"
+		echo "Not ${DIR} directory found"
 	fi
+}
+
+# Checks that pods were not left
+check_pods() {
+	check_pods_in_dir ${VC_POD_DIR}
 }
 
 # Check that runtimes are not running, they should be transient
--- a/integration/vm_factory/vm_templating_test.sh
+++ b/integration/vm_factory/vm_templating_test.sh
@@ -54,6 +54,22 @@ check_vm_template_factory() {
 	[ $res -eq 1 ] || die "template factory is not set up, missing state file"
 }
 
+clean_storage_in_dir() {
+	for file in $(ls $1); do
+		sudo rm -rf $1/$file
+	done
+}
+
+clean_storage() {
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -40,6 +46,10 @@ ginkgo:
 	GOPATH=$(PWD)/vendor go build ./vendor/github.com/onsi/ginkgo/ginkgo
 	unlink vendor/src
 
+clean:
+	rm $(VC_POD_DIR)/*
+	rm $(RUN_SBS_DIR)/*
```
devimc: missing `-rf` ?

teawater: > missing `-rf` ?

@devimc Thanks!  Updated.

chavafg:
taking in consideration that this will be running on a non-production testing environment, this should be fine. 
wdyt of this change @grahamwhaley @jodh-intel

chavafg:
I see this error on the intrd job:
```
09:26:11 + ls /var/lib/vc/sbs
09:26:11 a15d85881cae5d70b1ea537d41bbd64b30d527336f5c7b3f54dfd622d7d5daa8
09:26:11 f8c3ace381b3ded5642cef37944cd0b2d80344f2b6516e0b5c89bfec3ad65dff
09:26:11 + die '2 pods left and found at /var/lib/vc/sbs'
09:26:11 + msg='2 pods left and found at /var/lib/vc/sbs'
09:26:11 + echo 'ERROR: 2 pods left and found at /var/lib/vc/sbs'
09:26:11 ERROR: 2 pods left and found at /var/lib/vc/sbs
09:26:11 + exit 1
09:26:11 Makefile:133: recipe for target 'vm-factory' failed
```

grahamwhaley:
heh, err, did its job? :-) /cc @teawater :-)

teawater:
> heh, err, did its job? :-) /cc @teawater :-)

@grahamwhaley @chavafg the log is:
```
check storage leak before vm templating test
+ check_storage_leak
+ check_pods_in_dir /var/lib/vc/sbs
+ local DIR=/var/lib/vc/sbs
+ '[' -d /var/lib/vc/sbs ']'
++ ls /var/lib/vc/sbs
++ wc -l
+ pods_number=2
+ '[' 2 -ne 0 ']'
+ ls /var/lib/vc/sbs
a15d85881cae5d70b1ea537d41bbd64b30d527336f5c7b3f54dfd622d7d5daa8
f8c3ace381b3ded5642cef37944cd0b2d80344f2b6516e0b5c89bfec3ad65dff
```
This is the check before vm template test.  So it is not related with vm template.
And I think add check in current PR to check storage leak issue before vm template is not good idea.
I will push a new commit to remove it.

grahamwhaley:
np @teawater . But, that pre-check failing, is that an indication that one of the previous tests left some 'cruft' in the sbs directory? So, maybe it found a bug, just not a bug in the vmtemplating code?
@chavafg wdyt? I suspect we should add the check after every test, if something is leaking **anything** in our dirs, I think that is a bug...

lib/common.bash
```diff
@@ -155,19 +158,26 @@ check_processes() {
...
 		fi
 	else
-		echo "Not ${VC_POD_DIR} directory found"
+		echo "Not ${DIR} directory found"
 	fi
 }
 
+# Checks that pods were not left
+check_pods() {
+	check_pods_in_dir ${VC_POD_DIR}
```
grahamwhaley: So, I reckon we should add the SBS check here as well... then it will check sbs after every test, and not just vm template tests.... as nothing should ever leave a dirty sbs dir, should it?

teawater: > So, I reckon we should add the SBS check here as well... then it will check sbs after every test, and not just vm template tests.... as nothing should ever leave a dirty sbs dir, should it?

@grahamwhaley If we want to do that.  what I think update all the check function to check_sanity.sh and call in after each test in Makefile.

grahamwhaley: Sure - and I'm happy if we do that on a different PR - as I suspect it will break under the CI right now and may take some time to find/fix all the issues it shows - and I don't think we should have to hold up this PR .

chavafg:
@grahamwhaley, @teawater opened https://github.com/kata-containers/tests/issues/1467 to check which test is leaving pods behind. In the meantime I think we can merge this one. CI is happy.

Issue #1464, opened 2019-04-12
ci: always install/update golang


grahamwhaley:
We have an `install_go.sh` script in the .ci subdir, but presently it is only ever called from the jenkins CI kickoff script.
Over on https://github.com/kata-containers/proxy/pull/154, Zuul presently does not invoke the jenkins script, but installs golang on the node/distro and then invokes the `.ci/setup.sh` directly.
As some distros (e.g. Ubuntu et. al.), by default install old versions of go (v1.6 for instance iirc), the CI run then fails to build Kata, as the golang is too old.

Unpicking the CI scripts, we do have a sequence of items that must be installed for things to proceed, and they are likely something like:
- must have git, to clone some initial repos to get the scripts
- must have curl so we can install `yq` to then query the version.yaml
- must have `go` so we can get and build the other repos

Let's try to always install/update golang at the start of the setup then. There may be an inter-dependence here between git/go/curl/yq - we'll find out when we PR the update.

PR #1465, opened 2019-04-12
ci: setup: always install/update golang on setup

grahamwhaley:
Always try to install/update the version of golang.sh we have when
running setup.sh. For Jenkins, this is done already by the jenkins
specific script, but for other CIs and for hand-running the setup
it would be nice to always try to do it before we do anything else.

Fixes: #1464

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -128,6 +128,7 @@ install_extra_tools() {
 }
 
 main() {
+	bash -f "${cidir}/install_go.sh" -p -f
 	setup_distro_env
 	install_docker
 	enable_nested_virtualization


grahamwhaley:
/test
As I want to see if this breaks any of the existing CIs.

grahamwhaley:
OOh, it didn't immediately explode!
I think the SLES failure is unrelated:
```
12:18:12 Summarizing 1 Failure:
12:18:12 
12:18:12 [Fail] [Serial Test] Update number of CPUs Update CPU constraint [It] should have 2 CPUs 
12:18:12 /tmp/jenkins/workspace/kata-containers-tests-sles-12-PR/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:299
12:18:12 
12:18:12 Ran 19 of 247 Specs in 107.619 seconds
12:18:12 FAIL! -- 18 Passed | 1 Failed | 0 Pending | 228 Skipped --- FAIL: TestIntegration (107.62s)
12:18:12 FAIL
```

Ah, but I think I need to add some args to the install call:

```
11:36:54 /tmp/install-go-tmp.EsABHBpOdq ~/workspace/workspace/kata-metrics-tests-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests
11:36:54 Missing go version or -p option
11:36:54 Usage:
11:36:54 
11:36:54 install_go.sh [options] <args>
11:36:54 
11:36:54 Args:
11:36:54 <go-version> : Install a specific go version.
11:36:54 
11:36:54 Example:
11:36:54 install_go.sh 1.10
11:36:54 
11:36:54 Options
11:36:54 -d <path> : destination path, path where go will be installed.
11:36:54 -f        : Force remove old go version and install the specified one.
11:36:54 -h        : Show this help
11:36:54 -p        : Install go defined in Kata Containers versions file.
```

grahamwhaley:
/test
updated. try again...

grahamwhaley:
/test
heh, so, here is a funny thing - I had put:
```bash
       bash -f "${cidir}/install_go.sh -p -f"
```

and the opts inside the double quotes then make those opts part of the command, and the file is not found... duh.
Reworked to put the opts after the closing "...

Here is the question - what does `-f` on the bash commandline do? It obviously looks like it is saying 'execute this file', but staring at the bash manual, I'm not sure it is. I suspect it is setting 'disable pathname expansion', like `set -f` would..... can anybody tell me that I'm being stupid, or that we do/don't want that @chavafg @jodh-intel  ?
/me going slightly mad...

chavafg:
@grahamwhaley yes, I think it is the same as `set -f`.
I think in this case, it would be the same to use only `bash "${cidir}/install_go.sh" -p -f`

grahamwhaley:
right. I think we have `bash -f xxx.sh` all over our scripts - and maybe we don't need any of them. But, that is for another PR/day...

PR #1475, opened 2019-04-15
k8s: fix e2e script and yaml file

chavafg:
Use `die` to fail the e2e `run.sh` script when a `grep`
finds a failure in the e2e log.

In addition, remove dash from cri-o in skipped_tests_e2e.yaml
We use `CRI_RUNTIME` to specify which runtime interface we are
using to launch k8s and to know which tests we need to skip for e2e
testing.
Specifically for cri-o, we use `CRI_RUNTIME=crio`, so we
need to remove the dash (-) from cri-o in the skipped_tests_e2e.yaml
file, so that the tests listed there are properly skipped.

### DIFF
--- a/integration/kubernetes/e2e_conformance/run.sh
+++ b/integration/kubernetes/e2e_conformance/run.sh
@@ -72,7 +72,7 @@ run_sonobuoy() {
 	[ "$CI" == true ] && cat "$e2e_result_log"
 
 	# Check for Success message on the logs.
-	grep -aq " 0 Failed" "$e2e_result_log"
+	grep -aq " 0 Failed" "$e2e_result_log" || die "A failure has been found"
 	grep -aq "SUCCESS" "$e2e_result_log" && \
 		info " k8s e2e conformance using Kata runtime finished successfully"
 	popd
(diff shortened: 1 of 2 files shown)


integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -72,7 +72,7 @@ run_sonobuoy() {
 	[ "$CI" == true ] && cat "$e2e_result_log"
 
 	# Check for Success message on the logs.
-	grep -aq " 0 Failed" "$e2e_result_log"
+	grep -aq " 0 Failed" "$e2e_result_log" || die "A failure has been found"
```
jodh-intel: Nit: Is it worth referencing the log file in the error? Either specifying the name of the log or actually displaying it?

```sh
grep -aq " 0 Failed" "$e2e_result_log" || die "A failure has been found: $(cat $e2e_result_log)"
```

grahamwhaley: Do we save that log in the CI results (Jenkins artifacts?). In which case we could just log the results of the `grep` output, rather than the whole log itself.

chavafg: @jodh-intel We have an [info](https://github.com/kata-containers/tests/blob/d2ed1c0140a281dd286143987bea408e776f02e2/integration/kubernetes/e2e_conformance/run.sh#L69) where we specify the log location and we also print the whole tests output to the console log . 

@grahamwhaley We still do not save this on the artifacts when running on Jenkins, but I will add that on the jenkins job.

chavafg:
removing dnm flag... 
@grahamwhaley ptal

PR #1481, opened 2019-04-17
vm_factory: Add test for vm template init check

teawater:
According to the issue in
https://github.com/kata-containers/runtime/issues/1514,
runtime add vm template init check in
https://github.com/kata-containers/runtime/pull/1547.

Add test for this check.

Fixes: #1480

### DIFF
--- a/integration/vm_factory/vm_templating_test.sh
+++ b/integration/vm_factory/vm_templating_test.sh
@@ -35,6 +35,9 @@ disable_vm_template_config() {
 init_vm_template() {
 	echo "init vm template"
 	sudo -E PATH=$PATH "$kata_runtime_bin" factory init
+
+	{ sudo -E PATH=$PATH "$kata_runtime_bin" factory init ; res=$?; } || true
+	[ $res -ne 0 ] || die "factory init already called so expected 2nd call to fail"
 }
 
 destroy_vm_template() {


integration/vm_factory/vm_templating_test.sh
```diff
@@ -35,6 +35,9 @@ disable_vm_template_config() {
 init_vm_template() {
 	echo "init vm template"
 	sudo -E PATH=$PATH "$kata_runtime_bin" factory init
+
+	{ sudo -E PATH=$PATH "$kata_runtime_bin" factory init ; res=$?; } || true
+	[ $res -ne 0 ] || die "vm template init check not work"
```
jodh-intel: Isn't this the wrong way round? Shouldn't it be?:

```
[ $res -eq 0 ] || die "vm template init check not work"
```

teawater: > Isn't this the wrong way round? Shouldn't it be?:
> 
> ```
> [ $res -eq 0 ] || die "vm template init check not work"
> ```

@jodh-intel the $res should not be 0 because this test call factory init second time like https://github.com/kata-containers/runtime/issues/1514.
https://github.com/kata-containers/runtime/pull/1547 try to make factory init second time fail.

jodh-intel: Ah - I hadn't noticed the previous call. Could you make the die message a little clearer then. Something like:

```sh
die "factory init already called so expected 2nd call to fail"
```

teawater: > Ah - I hadn't noticed the previous call. Could you make the die message a little clearer then. Something like:
> 
> ```shell
> die "factory init already called so expected 2nd call to fail"
> ```

Updated.

PR #1482, opened 2019-04-17
k8s: volume test: wait for physical volume to be available

egernst:
backport of stability fixes for kubernets PV tests.


grahamwhaley:
@jodh-intel - this is a cherry pick backport I think - in which case we should probably leave it the same 'shape' as the original?

jodh-intel:
Ah - I'd missed that, and clearly we all missed the split observation on the original ;(

Issue #1483, opened 2019-04-18
CI: URL static check on GIT url failed


gnawux:
For `https://git.openstack.org/openstack`, it return `404` for https request but could be `git clone`

```bash
➜  curl -v https://git.openstack.org/openstack/zun-ui.git
*   Trying 104.130.246.85...
* TCP_NODELAY set
* Connected to git.openstack.org (104.130.246.85) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: OU=Domain Control Validated; OU=PositiveSSL; CN=git.openstack.org
*  start date: Jun 22 00:00:00 2018 GMT
*  expire date: Jun 22 23:59:59 2019 GMT
*  subjectAltName: host "git.openstack.org" matched cert's "git.openstack.org"
*  issuer: C=GB; ST=Greater Manchester; L=Salford; O=COMODO CA Limited; CN=COMODO RSA Domain Validation Secure Server CA
*  SSL certificate verify ok.
> GET /openstack/zun-ui.git HTTP/1.1
> Host: git.openstack.org
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Date: Thu, 18 Apr 2019 04:04:50 GMT
< Server: Apache/2.4.6 (CentOS)
[...]

PR #1484, opened 2019-04-18
CI: static check for openstack git URL

gnawux:
Fixes: #1483

Use git ls-remote instead of curl for git URL check

Signed-off-by: Xu Wang <xu@hyper.sh>


WeiZhang555:
LGTM

PR #1479, opened 2019-04-16
backport: 1.6.2 patches

ganeshmaharaj:
Backport includes below changes.

https://github.com/kata-containers/tests/pull/1456
https://github.com/kata-containers/tests/pull/1462
https://github.com/kata-containers/tests/pull/1471
https://github.com/kata-containers/tests/pull/1444

PR #1498, opened 2019-04-24
ci: lib.sh: set PATH for sudo even in no-go case

grahamwhaley:
If the scripts failed to find go or a GOPATH they would follow a different
path and invoke `sudo -E`, but not with the PATH=$PATH argument. This
normally works on oru CIs, but on Zuul, where the old distro go version
will be in the sudo path first, it fails with a 'go too old' error.
Pass the CI env/user PATH into the `sudo` to ensure we pick up
the correct golang.

Fixes: #1479


grahamwhaley:
/test
I was tempted to squash that function a bit, as it feels like the only difference now between the two `sudo`s is the error message, and we could handle that with a pre-check. But, no harm... let's see how it flies.

PR #1499, opened 2019-04-24
test: Reworked OCI test

### LINKED ISSUE #1065: oci_call_test randomly fails on Debian 9

GabyCT:
Debian VMs looks seems slower than other distros which makes
that we have instabilities in our OCI test. With this we will have
some retries to ensure that all the information is there while
running the test.

Fixes #1065

### DIFF
--- a/integration/oci_calls/oci_call_test.sh
+++ b/integration/oci_calls/oci_call_test.sh
@@ -20,6 +20,7 @@ TMP_FILE=$(mktemp runtimelogs.XXXXX)
 IMAGE="busybox"
 PAYLOAD="tail -f /dev/null"
 NAME="testoci"
+number_of_retries="15"
 
 function remove_tmp_file {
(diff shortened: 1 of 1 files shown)


jodh-intel:
@GabyCT - is this PR still needed (raised for debugging only maybe)?

GabyCT:
@jodh-intel no it is not debugging, this is the PR that we need

jodh-intel:
@GabyCT - in which case we need to understand why this is only failing on Debian as we really want to avoid distro-specific tests if at all possible [1].

---
[1] - yes, even though my https://github.com/kata-containers/runtime/pull/1587 will allow a distro-specific skip if really, really, *really* required :smile:

GabyCT:
@jodh-intel ok, I will open an issue on runtime

PR #1503, opened 2019-04-26
kata-manager: Check for dependencies earlier

### LINKED ISSUE #1502: kata-manager needs to check for dependencies earlier
See: https://github.com/kata-containers/documentation/issues/443.

jodh-intel:
The `kata-manager` script needs to download `git` repos. But if `git` isn't available, it can fall back to using `curl` or `wget`. However, if none of these tools are available, it should exit before any changes are
made to the system.

Note: The script _could_ install one of the tools of course. However, all known Linux distro "default installs" provide at least one of these tools, so this scenario should only be hit on very minimal systems. In that case, install one of the tools the error mentions and re-run `kata-manager`.

Fixes #1502.

### DIFF
--- a/cmd/kata-manager/kata-manager.sh
+++ b/cmd/kata-manager/kata-manager.sh
@@ -269,8 +269,6 @@ get_git_repo()
 	local -r repo_url="https://${repo_path}"
 
 	if ! command -v git >/dev/null; then
-		info "getting repo $1 using http downloader"
-		detect_downloader
 		$downloader "${repo_url}/${tarball_suffix}" | tar xz -C "$local_dest" --strip-components=1
 		return
 	fi
@@ -436,6 +434,11 @@ setup()
 	kata_repos_base=$(go env GOPATH 2>/dev/null || true)
 	if [ -z "$kata_repos_base" ]; then
 		kata_repos_base="$HOME/go"
+	fi
+
+	if ! command -v git >/dev/null; then
+		info "git not installed - trying to use a downloader instead"
+		detect_downloader
 	fi
 }
 


olberger:
I'm afraid the problem is not only in cases where git is not installed, but also lies in other uses of curl for other matters than replacing git: see https://github.com/kata-containers/documentation/issues/443#issuecomment-486211490

Hope this helps,

jodh-intel:
@olberger - I'm afraid I'm confused. Firstly, `curl` should really be installed since that's what we're telling users to run here:

- https://github.com/kata-containers/documentation/blob/master/install/installing-with-kata-manager.md#full-installation

I've now changed the script that that page tells you to run (`kata-manager`). The error in your comment is coming from when `kata-manager` actually tries to execute the installation doc. But that's *after* you run the `kata-manager` script.

I may be misunderstanding what you are saying. If so, please try to provide further details to help us understand the issue you are seeing.

olberger:
Alright. If curl is needed, the script should check that curl is there... asking users to do curl http://.../script | bash shouldn't be an excuse for not checking that curl is there.

Btw, any user doing curl | sh should be blamed, and be instructed to do curl or wget, inspect and then execute, IMHO ;-)

jodh-intel:
@olberger - I still think we might be talking "at crossed purposes" here. If you followed the instructions to do an automatic install here:

- https://github.com/kata-containers/documentation/blob/master/install/installing-with-kata-manager.md#full-installation

Then, clearly, you will need `curl` installed to even attempt to run `kata-manager`. That doc is designed to be read by users and the `bash -c "$(curl ...)"` will fail immediately if `curl` is not installed on your system.

> Btw, any user doing curl | sh should be blamed, and be instructed to do curl or wget, inspect and then execute, IMHO ;-)

I understand. And that is *exactly* why we offer a variety of different installation options here:

- https://github.com/kata-containers/documentation/tree/master/install#installing-on-a-linux-system

If you have concerns, I would recommend either the "manual" or "scripted" installation option.

PR #1514, opened 2019-04-29
test: Fix pod quota test for kubernetes

### LINKED ISSUE #1513: fix pod quota test for kubernetes
Fix pod quota test to avoid leaving pod information at /var/lib/vc/sbs and verify after running the tests that not pod information is left

GabyCT:
Fix pod quota test to avoid leaving pod information at /var/lib/vc/sbs (issue: https://github.com/kata-containers/runtime/issues/1611) and verify after running the tests that not pod information is left.

Fixes #1513

### DIFF
--- a/integration/kubernetes/cleanup_env.sh
+++ b/integration/kubernetes/cleanup_env.sh
@@ -35,3 +35,6 @@ sudo ip link del flannel.1
 
 # Check no kata processes are left behind after reseting kubernetes
 check_processes
+
+# Checks that pods were not left
+check_pods
--- a/integration/kubernetes/runtimeclass_workloads/pod-quota-deployment.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-quota-deployment.yaml
@@ -11,7 +11,7 @@ spec:
   selector:
     matchLabels:
       purpose: quota-demo
-  replicas: 3
+  replicas: 2
   template:
     metadata:
       labels:
--- a/integration/kubernetes/k8s-pod-quota.bats
+++ b/integration/kubernetes/k8s-pod-quota.bats
@@ -15,8 +15,6 @@ setup() {
 @test "Pod quota" {
 	resource_name="pod-quota"
 	deployment_name="deploymenttest"
-	wait_time=10
-	sleep_time=2
 
 	# Create the resourcequota
 	kubectl create -f "${pod_config_dir}/resource-quota.yaml"
@@ -27,9 +25,8 @@ setup() {
 	# Create deployment
 	kubectl create -f "${pod_config_dir}/pod-quota-deployment.yaml"
 
-	# View information about the deployment
-	cmd="kubectl get deployment \"$deployment_name\" --output=yaml | grep -q 'forbidden: exceeded quota'"
-	waitForProcess "$wait_time" "$sleep_time" "$cmd"
+	# View deployment
+	kubectl wait --for=condition=Available deployment/${deployment_name}
 }
 
 teardown() {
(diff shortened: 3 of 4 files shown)


integration/kubernetes/runtimeclass_workloads/pod-quota-deployment.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   selector:
     matchLabels:
       purpose: quota-demo
-  replicas: 3
+  replicas: 2
```
grahamwhaley: why this change? Is it related to not leaving pods around?

integration/kubernetes/cleanup_env.sh
```diff
@@ -35,3 +35,6 @@ sudo ip link del flannel.1
 
 # Check no kata processes are left behind after reseting kubernetes
 check_processes
+
+# Checks that pods were not left
+check_pods
```
grahamwhaley: nice - I presume this will not check on all k8s tests we run?

integration/kubernetes/runtimeclass_workloads/pod-quota-deployment.yaml
GabyCT: @grahamwhaley yes, with this we will not leave pods

integration/kubernetes/cleanup_env.sh
GabyCT: @grahamwhaley yes it will

grahamwhaley: Oops, I meant 'will check' - good !

integration/kubernetes/runtimeclass_workloads/pod-quota-deployment.yaml
grahamwhaley: but... why? Can you explain (maybe in the commit message) why 3->2 makes it not leak?

chavafg: @GabyCT can you open an issue in the runtime repo and add it as a reference maybe in the commit message?

GabyCT: @chavafg and @grahamwhaley changes applied

jodh-intel: Like @grahamwhaley, I'm still not clear on this change. It might "work" with `2`, but why did we choose `3` initially then? What did we break?

Maybe @egernst or @mcastelino have thoughts?

mcastelino: @GabyCT there does not seem to be a co-relation between the bug and this fix. Can you expand? Were we failing deployment initially?

GabyCT: @jodh-intel and @mcastelino before this PR, we never check if kubernetes tests were leaving pod information behind. Until recently, we found out that this test was leaving pod information that is why as part of this PR, I am including to check that after running all the kubernetes tests. If we leave it like it is with 3 replicas then we will have pod info at /var/lib/vc/sbs. The deployment of the replicas is done successfully and even it says that when we remove the deployment is donde successfully, however, after cleaning everything the pod info is at /var/lib/vc/sbs.

GabyCT: @jodh-intel , we choose 3 before because the test was trying to run 3 containers but the resource quota was only 2 so the test verified that it was not possible to run 3 containers and that only 2 were running

PR #1518, opened 2019-04-30
checkcommits: recognize jenkins CI environment

chavafg:
Add jenkins github pull request builder env variables
to set commitID, source and dest branches.

Fixes: #1516

### DIFF
--- a/cmd/checkcommits/checkcommits.go
+++ b/cmd/checkcommits/checkcommits.go
@@ -467,6 +467,13 @@ func detectCIEnvironment() (commit, dstBranch, srcBranch string) {
 			// contains that branch: master doesn't exist.
 			dstBranch = "origin"
 		}
+	} else if os.Getenv("ghprbPullId") != "" {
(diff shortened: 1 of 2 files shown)


cmd/checkcommits/checkcommits.go
```diff
@@ -467,6 +467,13 @@ func detectCIEnvironment() (commit, dstBranch, srcBranch string) {
 			// contains that branch: master doesn't exist.
 			dstBranch = "origin"
 		}
+	} else if os.Getenv("CI") != "" {
```
grahamwhaley: I'm not sure `CI` is the most accurate thing we can check for Jenkins. There are some more Jenkins specific env vars I think: https://wiki.jenkins.io/display/JENKINS/Building+a+software+project
Including:
- `WORKSPACE`
- `JENKINS_URL`
- or we could even check we have some/one/all of the `ghprb` vars set?

PR #1520, opened 2019-04-30
test: Add K8s test for containers with shared volume

### LINKED ISSUE #1519: k8s test for containers with shared volume
Add a kubernetes test where we have a pod that runs two containers with a shared volume

GabyCT:
This adds a kubernetes test where we have a pod that runs
two containers with a shared volume.

Fixes #1519

### DIFF
--- a/integration/kubernetes/runtimeclass_workloads/pod-shared-volume.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-shared-volume.yaml
@@ -1,0 +1,28 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-shared-volume
+spec:
+  runtimeClassName: kata
+  restartPolicy: Never
+  volumes:
+  - name: shared-data
+    emptyDir: {}
+  containers:
+  - name: nginx-container
+    image: nginx
+    volumeMounts:
+    - name: shared-data
+      mountPath: /usr/share/nginx/html
+  - name: debian-container
+    image: debian
+    volumeMounts:
+    - name: shared-data
+      mountPath: /pod-data
+    command: ["/bin/sh"]
+    args: ["-c", "echo Hello from the debian-container > /pod-data/index.html && tail -f /dev/null"]
--- a/integration/kubernetes/untrusted_workloads/pod-shared-volume.yaml
+++ b/integration/kubernetes/untrusted_workloads/pod-shared-volume.yaml
@@ -1,0 +1,30 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-shared-volume
+  annotations:
+    io.kubernetes.cri-o.TrustedSandbox: "false"
+    io.kubernetes.cri.untrusted-workload: "true"
+spec:
+  restartPolicy: Never
+  volumes:
+  - name: shared-data
+    emptyDir: {}
+  containers:
+  - name: nginx-container
+    image: nginx
+    volumeMounts:
+    - name: shared-data
+      mountPath: /usr/share/nginx/html
+  - name: debian-container
+    image: debian
+    volumeMounts:
+    - name: shared-data
+      mountPath: /pod-data
+    command: ["/bin/sh"]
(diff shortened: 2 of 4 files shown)


integration/kubernetes/untrusted_workloads/pod-shared-volume.yaml
```diff
@@ -0,0 +1,30 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-shared-volume
+  annotations:
```
grahamwhaley: @GabyCT @egernst Do we still use the annotations for trust/untrust? Feels maybe we can drop this, now we have moved wholly to runtimeClass ?

integration/kubernetes/runtimeclass_workloads/pod-shared-volume.yaml
```diff
@@ -0,0 +1,28 @@
...
+  name: test-shared-volume
+spec:
+  runtimeClassName: kata
+  restartPolicy: Never
+  volumes:
+  - name: shared-data
+    emptyDir: {}
+  containers:
+  - name: nginx-container
+    image: nginx:1.14
```
jodh-intel: Why this version? Can't we just use the latest (same comment for config below)?

integration/kubernetes/untrusted_workloads/pod-shared-volume.yaml
GabyCT: @grahamwhaley, yes as far I can see we still use the annotations

integration/kubernetes/runtimeclass_workloads/pod-shared-volume.yaml
GabyCT: @jodh-intel fixed

integration/kubernetes/untrusted_workloads/pod-shared-volume.yaml
grahamwhaley: Oh, OK. I thought they'd been deprecated .... can you show/tell us where they are used @GabyCT ?
@ganeshmaharaj @amshinde @krsna1729  - any insights or clarification here?

GabyCT: @grahamwhaley 
https://github.com/kata-containers/tests/blob/4ccdac9df33918350bf3ca95a83fbf3cd91bbaa3/integration/kubernetes/untrusted_workloads/lifecycle-events.yaml
https://github.com/kata-containers/tests/blob/48ec7a396cdd660bcb4b312a599da72990cb7822/integration/kubernetes/untrusted_workloads/busybox-pod.yaml
https://github.com/kata-containers/tests/blob/48ec7a396cdd660bcb4b312a599da72990cb7822/integration/kubernetes/untrusted_workloads/busybox-template.yaml
......
and all the rests you can check it at untrusted_workloads. I talked with @chavafg and he said that after we move to 1.14 we will remove this.

grahamwhaley: ah - they are all in test yaml files.... I was more thinking does k8s actually use/support this any more.
Maybe with the move to runtimeClass with k8s v1.14, it will finally be officially deprecated. Maybe those tests will start to fail when we move to v1.14 anyway ;-)
@chavafg - do we have a plan/timeline of when to move the CI to v1.14 ? I don't know if we have a rule, like 'after first minor update' or similar?

krsna1729: I would think k8s itself does not care about this. containerd/crio would.

krsna1729: 1.14.1 is already here https://github.com/kubernetes/kubernetes/releases/tag/v1.14.1

chavafg: No timeline defined but the easiest would be to open a PR in runtime and change our current version with 1.14 and see if CI passes. If it is ok, we can merge the change and then get back here and make the necessary changes to remove the old annotations. 
I'll try to make some time tomorrow for this, but if someone else want to try please go ahead.

grahamwhaley: If you do (which would be nice), then I presume you are aware the runtimeClass moved from an alpha to a beta feature, which means it no longer needs a featuretweak, and the yaml that sets it up for kata hooks has to be updated s/alpha/beta/ - I'm presuming our CI does that, so it would need a tweak ...

Issue #1522, opened 2019-04-30
openshift tests blocked


chavafg:
All Fedora jobs will be failing due to an openshift issue we have.
Kubernetes is removing all images from `docker.io` [1] and for some reason the `kubernetes/pause` image in the docker registry is being used by openshift. 
I am not sure if there is a way to configure openshift and get the image from another registry, but in the meantime you may see failures. 

btw, I have tried to look for this image at `gcr.io` and couldn't find it.

[1] https://github.com/kubernetes/k8s.io/issues/237

spiffxp:
Should be using `k8s.gcr.io/pause` instead ref: https://github.com/kubernetes/kubernetes/pull/69631

chavafg:
Thanks @spiffxp, 
I opened #1526 that changes our `cri-o` configuration to use `k8s.gcr.io/pause` image.

PR #1526, opened 2019-05-02
tests:openshift: use k8s.gcr.io/pause image

chavafg:
Openshift v3.10 uses cri-o v1.10, which still makes reference
to the kubernetes/pause image and this image at docker.io
registry is being removed, so we need to use the pause image
from gcr.

Fixes: #1522.


chavafg:
hmm, seems like it doesn't solve the issue...
Still see that `docker.io/kubernetes/pause` image is being downloaded.

Issue #1540, opened 2019-05-07
arm ci failed on installing qemu


bergwolf:
ref:http://jenkins.katacontainers.io/job/kata-containers-runtime-ARM-18.04-PR/578/console

>00:47:20 INFO: Clone config and patches
00:47:20 lrwxrwxrwx 1 root root 18 May  7 00:46 /usr/share/kata-containers/vmlinux.container -> vmlinux-4.19.28-36
00:47:20 lrwxrwxrwx 1 root root 18 May  7 00:46 /usr/share/kata-containers/vmlinuz.container -> vmlinuz-4.19.28-36
00:47:20 Install Qemu
00:52:13 package github.com/qemu/qemu: no Go files in /home/jenkins/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/qemu/qemu
00:52:13 can't load package: package github.com/kata-containers/packaging: no Go files in /home/jenkins/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/kata-containers/packaging
00:52:13 ~/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/qemu/qemu ~/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/kata-containers/tests
00:52:17 Already on 'master'
00:52:17 Your branch is up to date with 'origin/master'.
00:52:18 error: The following untracked working tree files would be overwritten by checkout:
00:52:18 	slirp/COPYRIGHT
00:52:18 Please move or remove them before you switch branches.
00:52:18 Aborting
00:52:18 Build step 'Execute shell' marked build as failure
00:52:19 Performing Post build task...
00:52:19 Match found for :.* : True
00:52:19 Logical operation result is TRUE
00:52:19 Running script  : #!/bin/bash

Pennyzct:
Hi~ @bergwolf  thanks for the raising issue ;). I've been offline a few days for attending Open Infrastructure summit. First thing After coming back was to solve this issue. Please see the related [PR#1541](https://github.com/kata-containers/tests/pull/1541).

PR #1541, opened 2019-05-08
qemu-aarch64: replace go get with git clone when retrieving qemu

Pennyzct:
`go get` will cause error of untracking files from sub-modules being overwritten when checking out, so we use `git clone` to avoid.

Fixes: #1540

### DIFF
--- a/.ci/aarch64/lib_install_qemu_aarch64.sh
+++ b/.ci/aarch64/lib_install_qemu_aarch64.sh
@@ -8,7 +8,7 @@ set -e
 
 CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
 PACKAGED_QEMU="qemu"
-CURRENT_QEMU_PATCHES_BRANCH=$(get_version "assets.hypervisor.qemu.architecture.aarch64.branch")
+CURRENT_QEMU_BRANCH=$(get_version "assets.hypervisor.qemu.architecture.aarch64.branch")
 CURRENT_QEMU_COMMIT=$(get_version "assets.hypervisor.qemu.architecture.aarch64.commit")
 
 get_packaged_qemu_version() {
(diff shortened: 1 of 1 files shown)


.ci/aarch64/lib_install_qemu_aarch64.sh
```diff
@@ -45,34 +45,28 @@ install_packaged_qemu() {
...
+        QEMU_REPO_URL=$(get_version "assets.hypervisor.qemu.url")
         # Remove 'https://' from the repo url to be able to clone the repo using 'go get'
-        QEMU_REPO=${QEMU_REPO/https:\/\//}
+        QEMU_REPO=${QEMU_REPO_URL/https:\/\//}
         PACKAGING_REPO="github.com/kata-containers/packaging"
         QEMU_CONFIG_SCRIPT="${GOPATH}/src/${PACKAGING_REPO}/scripts/configure-hypervisor.sh"
 
-        go get -d "${QEMU_REPO}" || true
-        go get -d "$PACKAGING_REPO" || true
+        git clone --branch "$CURRENT_QEMU_BRANCH" --single-branch "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
```
grahamwhaley: if we do go with `git clone`, and afaik we are just building from the HEAD of the code, maybe we can do this with a `--depth 1` to make a shallow clone, and save us time and bandwidth? :-)
iirc, there may be submodules in the qemu tree as well, so we may also want/need `--shallow-submodules` as well?

Pennyzct: Hi~ @grahamwhaley @devimc I have addressed your comments in another PR #1555, ptal. ;)

Pennyzct:
Hi~ @chavafg I have addressed @grahamwhaley's comments in another PR #1555, ptal. ;)

PR #1542, opened 2019-05-08
CI: Add markdown checker

### LINKED ISSUE #1485: Add a markdown checker to the CI
Add a tool to more rigorously check our markdown docs for errors and inconsistencies.

jodh-intel:
Add a simple tool that checks markdown files to ensure all links are valid.

Updated the static checker script to call this tool to ensure that documents with broken links are not merged.

Note that when the tool detects a broken link, it is often able to suggest the correct link name so looking at its output is beneficial.

Fixes #1485.


cmd/check-markdown/hack.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"strings"
+
+	bf "gopkg.in/russross/blackfriday.v2"
```
grahamwhaley: I still suspect using the same markup tools that github itself uses to do the check might have made more sense:
https://github.com/github/markup
https://github.com/gjtorikian/commonmarker
https://github.com/github/cmark-gfm

The last one looks to be the core library, which is in C, so not too avant-garde or arduous maybe. c'est la vie.

jodh-intel: Point already taken but:

- This tool was based on a test tool I had already half-written a while back.

  Although I think there is a lot more this tool could be made to do, I just want the checking integrated into the CI asap as we continue to see doc breakages and this tool will stop them (frankly, I'm surprised GitHub doesn't offer this sort of functionality already).

  Once we have "something" in the CI to perform the checks, we can always replace it with "something else" later.

- markup is ruby and I'm very rusty on that.

- the other 2 packages are written in C which -- although I love it -- is not a language we should probably be writing new code in.
- writing it in go means the rest of the team will understand it (hopefully :-) but we can also leverage all the go goodness we rely on so much (`go test`, `gofmt`, etc).

Of course, if you or anyone else in the community would like to *take up the gauntlet* and rewrite this in ruby/rust/blah, be my guest - I won't be offended! :smile:

jodh-intel:
The sooner we have "borked link protection" in the CI, the better for everyone imho. Hence, working on the assumption that "something is better than nothing", let's merge this and fix up any doc issues on a follow-up PR...

klynnrif:
@jodh-intel apologies, I missed this ping. Would you still like a review on this?

jodh-intel:
No problem - you can review if you wish and I'll raise a new PR. But I'm about to raise a follow-up PR which will have a few doc changes so you could wait for that if you want?

klynnrif:
@jodh-intel I'll wait for the follow-up PR. Thanks!

Issue #1544, opened 2019-05-08
qemu-ppc64le: Replace go get with git clone when retrieving qemu


nitkon:
Go get will cause error of untracking files from sub-modules being overwritten when checking out, so we use git clone to avoid. This issue is similar to that of ARM CI: #1540

PR #1545, opened 2019-05-08
qemu-ppc64le: Use "git clone" instead of "go get" when retrieving qemu

nitkon:
'go get' will cause error of untracking files from submodules
being overwritten when checking out, so we use 'git clone' to avoid.

Fixes:  #1544

Signed-off-by: Nitesh Konkar niteshkonkar@in.ibm.com


jodh-intel:
OOI, do you need to add `--shallow-submodules` to the `git clone` also then?

Issue #1543, opened 2019-05-08
entropy/randomness: tests need improving


grahamwhaley:
Enabling the kernel `CONFIG_RANDOM_TRUST_CPU` option over in
https://github.com/kata-containers/packaging/pull/482
has highlighted that when we added the [entropy test](https://github.com/kata-containers/tests/pull/887) to the CI, maybe it is not quite testing what we thought, and maybe we really now need to test the availability of random data, not just the amount of entropy available?

Maybe we can do something along the lines of:

```bash
$ dd if=/dev/random of=/dev/null bs=4b count=1000000
```

and time it - and ensure it finishes in 'a reasonable time' ?
I have also wondered if we can run something like the openssl benchmarks, maybe they consume randomness?

GabyCT:
@grahamwhaley, in order to get this `reasonable time` should we need to perform this operation several times and get the `maximum` and `minimum` in order to set a time limit? or do we have an idea of this `reasonable time`?

GabyCT:
@grahamwhaley also what do you think if I add this test as part of PnP metrics?

grahamwhaley:
@GabyCT I don't have specific numbers in mind - we are going to have to do a bit of experimentation and thought to figure out what feels 'good' and where that becomes 'bad'.
The only number I have right now is that 1000000 /dev/random reads inside a kata on my local NUC took ~35s iirc. I want to look around to see if maybe we can find an off the shelf test that eats randomness as well to see if that will be useful.
I think, yes, maybe we will have to do re-runs and average if we find there is variation in the results. But, I think we should maybe move away from measuring the entropy value, and shift towards the availability of randomness - as I think that is ultimately what a real container workload would care about.

grahamwhaley:
oh, and wrt adding to the pnp metrics - let's get some numbers first and see how it looks - it is possible, but we need to see how good and stable the tests we come up with are first.

PR #1547, opened 2019-05-08
test: Add retries to get entropy level

### LINKED ISSUE #1546: improve entropy test
We need to wait a couple of seconds in order to get the
entropy level https://github.com/kata-containers/packaging/pull/482.

GabyCT:
We need to wait a couple of seconds in order to get the
entropy level https://github.com/kata-containers/packaging/pull/482.

Fixes #1546

### DIFF
--- a/integration/entropy/entropy_test.bats
+++ b/integration/entropy/entropy_test.bats
@@ -20,7 +20,11 @@ IMAGE="busybox"
 # by haveged is 1000 see https://wiki.archlinux.org/index.php/Haveged
 # Less than 1000 could potentially slow down cryptographic
 # applications see https://www.suse.com/support/kb/doc/?id=7011351
-ENTROPY_LEVEL="1000"
+entropy_level="1000"
+
+# This is needed to wait for a couple of seconds
+# until we try to reach the minimum entropy level
+retries="10"
 
 setup() {
 	clean_env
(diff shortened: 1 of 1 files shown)


GabyCT:
@ganeshmaharaj @jcvenegas @grahamwhaley @mcastelino this can be a temporary solution of the problem seen at kata-containers/packaging#482. This can help until we have this set of test https://github.com/kata-containers/tests/issues/1543

grahamwhaley:
/test
Indeed, I think this is a bit of a flaky sticky plaster - this *might* work, today, with our current kernel, config, and CI machines - but, as noted on the other items/issues, I think checking the entropy level is probably not the right test - we probably need to check how much 'randomness' is available, and at what rate. Let's see how it flies in the CI for now though...

jcvenegas:
@grahamwhaley  I think it is still value on check the entropy level at container startup at at least to ensure that we provide it in a low time after the container has boot.

Correct me if I am wrong but just dumping some megabytes from /dev/random does not tell us the quality of the entropy if this is low (less that 1000) our level is not good as we want for users.

jcvenegas:
@grahamwhaley  the other option is to force the agent finish the boot until finds the entropy level is good but will result in the same issue @mcastelino is trying to solve (reduce boot time).

I agree that we should add an additional test to see how fast we provide random numbers `/dev/random` should be the one we pick as this is a blocking device (not use pool as urandom).

grahamwhaley:
afaict, reading from `/dev/random` does block until entropy is good enough. Reading from `/dev/urandom` does not block, and will continue to provide *potentially* less random numbers if entropy runs out...
Best document I found to read was at https://www.2uo.de/myths-about-urandom

What concerns me most I guess about us measuring `entropy >= 1000` is, do we know how long it will, or should, take to achieve that level of entropy inside a kata container? How is it getting filled up - is there a rate limit to it filling, is it via `virtio-rng`, or ??? I think we have shown it is not instant already, by failing the test just by modifying the kernel config and thus the PRNG init sequence and boot time... on my local system for instance I think I was seeing an extra 30s was not enough, but an extra 60s was - but, this probably comes down to 'lack of activity' happening inside the container post-boot etc.

Reading from `/dev/random` I think will maybe force the Kata kernel to go access the `virtio-rng` device to get more entropy, and thus really test that we can provide entropy, post boot, at a good enough rate.

jcvenegas:
@GabyCT  after talk with @grahamwhaley and you  in slack, lets not run this test for now until we can find if we can have a reliable that we agree that gives value to our test suite.

GabyCT:
@jcvenegas ok I will close this PR thanks

Issue #1552, opened 2019-05-09
tracing tests getting hanged


chavafg:
tracing test may get hanged due to bad increment to the `$attempt` variable

```
xal:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xal:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xal:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xal:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xal:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xbi:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xbi:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xbi:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xdw:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
xdw:WARNING: Found invalid parent span errors (attempt 0):             "invalid parent span IDs=1ca0cd7c50369650; skipping clock skew adjustment",
```

chavafg:
reopening as issue still occurs as we are exiting the loop iteration without increasing the `$attempt` value.

PR #1553, opened 2019-05-09
tracing: Fix tracing test script.

chavafg:
We use an `$attempt` variable to have a maximum
of 3 attempts to validate tracing spans, but
the variable increment was not correct, leading
to having the script getting hanged.

Fixes: #1552.

### DIFF
--- a/tracing/tracing-test.sh
+++ b/tracing/tracing-test.sh
@@ -76,7 +76,7 @@ get_jaeger_status()
 
 		[ "$ret" -eq 0 ] && [ -n "$status" ] && break
 
-		attempt=$((attempt++))
+		attempt=$((attempt+1))
 		sleep 1
 	done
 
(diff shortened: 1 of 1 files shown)


tracing/tracing-test.sh
```diff
@@ -76,7 +76,7 @@ get_jaeger_status()
 
 		[ "$ret" -eq 0 ] && [ -n "$status" ] && break
 
-		attempt=$((attempt++))
+		attempt=$((attempt+1))
```
devimc: `((attempt++))` is be enough, isn't it?

chavafg: no, it doesn't do the work

grahamwhaley: Hmm, I use that idom elsewhere:
https://github.com/kata-containers/tests/blob/master/metrics/density/docker_memory_usage.sh#L74
```
(( count++ ))
```

Maybe it needs the spaces to invoke the commands `((` and `))` rather than maybe using a shell builting. No big matter for the merge, the intent is still clear :-)

grahamwhaley: @chavafg - might be good to search the rest of the repo/scripts to see if we have the same issue lurking anywhere else though ;-)

Issue #1554, opened 2019-05-10
qemu-aarch64: use shadow clone to save time and bandwidth


Pennyzct:
For now, we often use one specific qemu-system-aarch64 version referred by `assets.hypervisor.qemu.architecture.aarch64.tag` in `version.yaml`.
we could use shadow clone to save time and bandwidth.

PR #1555, opened 2019-05-10
qemu-aarch64: use shadow clone to save time and bandwidth

Pennyzct:
For now, we often use one specific qemu-system-aarch64 version referred by `assets.hypervisor.qemu.architecture.aarch64.tag` in `version.yaml.`
we could use shadow clone to save time and bandwidth.

Fixes: #1554

### DIFF
--- a/.ci/aarch64/lib_install_qemu_aarch64.sh
+++ b/.ci/aarch64/lib_install_qemu_aarch64.sh
@@ -8,8 +8,10 @@ set -e
 
 CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
 PACKAGED_QEMU="qemu"
-CURRENT_QEMU_BRANCH=$(get_version "assets.hypervisor.qemu.architecture.aarch64.branch")
-CURRENT_QEMU_COMMIT=$(get_version "assets.hypervisor.qemu.architecture.aarch64.commit")
+CURRENT_QEMU_BRANCH=$(get_version "assets.hypervisor.qemu.architecture.aarch64.tag")
+QEMU_REPO_URL=$(get_version "assets.hypervisor.qemu.url")
+# Remove 'https://' from the repo url to be able to git clone the repo
+QEMU_REPO=${QEMU_REPO_URL/https:\/\//}
 
 get_packaged_qemu_version() {
         if [ "$ID" == "ubuntu" ]; then
@@ -45,20 +47,17 @@ install_packaged_qemu() {
 }
 
 build_and_install_qemu() {
-        QEMU_REPO_URL=$(get_version "assets.hypervisor.qemu.url")
-        # Remove 'https://' from the repo url to be able to clone the repo using 'go get'
-        QEMU_REPO=${QEMU_REPO_URL/https:\/\//}
         PACKAGING_REPO="github.com/kata-containers/packaging"
         QEMU_CONFIG_SCRIPT="${GOPATH}/src/${PACKAGING_REPO}/scripts/configure-hypervisor.sh"
 
-        git clone --branch "$CURRENT_QEMU_BRANCH" --single-branch "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
+	clone_qemu_repo
+
 	go get -d "$PACKAGING_REPO" || true
 
         pushd "${GOPATH}/src/${QEMU_REPO}"
         git fetch
-        git checkout "$CURRENT_QEMU_COMMIT"
-        [ -d "capstone" ] || git clone https://github.com/qemu/capstone.git capstone
-        [ -d "ui/keycodemapdb" ] || git clone  https://github.com/qemu/keycodemapdb.git ui/keycodemapdb
+        [ -d "capstone" ] || git clone https://github.com/qemu/capstone.git --depth 1 capstone
+        [ -d "ui/keycodemapdb" ] || git clone  https://github.com/qemu/keycodemapdb.git --depth 1 ui/keycodemapdb
 
         # Apply required patches
         QEMU_PATCHES_PATH="${GOPATH}/src/${PACKAGING_REPO}/obs-packaging/qemu-aarch64/patches"
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -13,6 +13,9 @@ source /etc/os-release || source /usr/lib/os-release
 
 CURRENT_QEMU_BRANCH=$(get_version "assets.hypervisor.qemu-lite.branch")
 CURRENT_QEMU_COMMIT=$(get_version "assets.hypervisor.qemu-lite.commit")
+QEMU_REPO_URL=$(get_version "assets.hypervisor.qemu-lite.url")
+# Remove 'https://' from the repo url to be able to git clone the repo
+QEMU_REPO=${QEMU_REPO_URL/https:\/\//}
 PACKAGED_QEMU="qemu-lite"
 QEMU_ARCH=$(${cidir}/kata-arch.sh -d)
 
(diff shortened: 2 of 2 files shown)


.ci/aarch64/lib_install_qemu_aarch64.sh
```diff
@@ -51,14 +50,13 @@ build_and_install_qemu() {
         PACKAGING_REPO="github.com/kata-containers/packaging"
         QEMU_CONFIG_SCRIPT="${GOPATH}/src/${PACKAGING_REPO}/scripts/configure-hypervisor.sh"
 
-        git clone --branch "$CURRENT_QEMU_BRANCH" --single-branch "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
+        git clone --branch "$CURRENT_QEMU_BRANCH" --single-branch --depth 1 --shallow-submodules "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
```
jodh-intel: I was suggesting @nitkon used this on https://github.com/kata-containers/tests/pull/1545#issuecomment-490443025, but that wasn't applied.

I do think this is similar to https://github.com/kata-containers/runtime/pull/1660 - this code is not architecture-specific so we really need a `clone_qemu_repo()` function that accepts as `$branch` param in `.ci/install_qemu.sh` to avoid having to fix the same bug for every architecture.

Pennyzct: Hi~ @grahamwhaley @jodh-intel I'm fine with spreading it to more archs and use `clone_qemu_repo()` as common function. 
And I'm thinking that if we are using `--depth 1`, we are definitely to be sure that we are building from the HEAD, `git checkout "$CURRENT_QEMU_COMMIT"` is unnecessary and if we failed to update to keep it the same as HEAD, it could cause problem.

Pennyzct:
Hi~ @grahamwhaley @jodh-intel I've already done the update. ;) but I only wrote code for AArch64 and Amd64, since i could only touch these two machines locally.

jodh-intel:
Thanks @Pennyzct.

Maybe @nitkon and @alicefr can update the other arches to use your new generic `clone_qemu_repo` call.

All the CI's are failing on a `docker run debian sh` test:

```
08:12:33 Stderr: E: Unable to determine file size for fd 13 - fstat (2: No such file or directory)
08:12:33 W: The repository 'http://security.debian.org/debian-security stretch/updates InRelease' provides only weak security information.
08:12:33 W: Conflicting distribution: http://security-cdn.debian.org/debian-security stretch/updates InRelease (expected stretch/updates but got )
08:12:33 E: Unable to determine file size for fd 15 - fstat (2: No such file or directory)
08:12:33 W: The repository 'http://deb.debian.org/debian stretch-updates InRelease' provides only weak security information.
08:12:33 W: Conflicting distribution: http://cdn-fastly.deb.debian.org/debian stretch-updates InRelease (expected stretch-updates but got )
08:12:33 E: Unable to determine file size for fd 4 - fstat (2: No such file or directory)
08:12:33 E: The package lists or status file could not be parsed or opened.
```

Looks like it might be a transitory error, so let's restart them all...

/retest

alicefr:
I'm just wondering if we can reduce the 3 arch scripts to one as they're pretty similar. Actually, all the scripts in `.ci/arch/` are almost equal, or do we want to keep those scripts separated?

Pennyzct:
Hi~@alicefr FWIT, I'm quite hoping that different arch maintainers maintain different arch scripts. ;). For example, just for `qemu`, on AArch64, I'm not using` qemu-lite`, and I'm also not using stable version 2.11, you could see from [version.yaml](https://github.com/kata-containers/runtime/blob/master/versions.yaml#L116), I have specific aarch64 entry.

jodh-intel:
I guess it's ultimately up to each arch maintainer, but the ideal scenario would be to make code common where possible as we've had a few PRs recently to make the same changes for different arches and that seems a little inefficient to me.

alicefr:
@Pennyzct I agree but as @jodh-intel said we have a lot of common code duplicated. We could have a clean up later and try to unify the scripts. I'll update the s390x scripts up to now :)

Pennyzct:
@alicefr @jodh-intel Ok. I have skipped three different `arch/` dir, and I think that we are sharing the same filter scheme to skip integration tests, which we could replace the filter scheme into `.ci/` dir and just leave the specific configuration toml file in different arch dir.

alicefr:
@Pennyzct  thanks :)

Pennyzct:
Hi~ @jodh-intel @grahamwhaley @chavafg 
I think all failed CIs are failing the same error.
```
10:12:07 [Serial Test] package manager update test
10:12:07 /tmp/jenkins/workspace/kata-containers-tests-fedora-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:32
10:12:07   check apt-get update and upgrade
10:12:07   /tmp/jenkins/workspace/kata-containers-tests-fedora-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:54
10:12:07     should not fail [It]
10:12:07     /tmp/jenkins/workspace/kata-containers-tests-fedora-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:55
10:12:07 
10:12:07     Expected
10:12:07         <int>: 100
10:12:07     to be zero-valued
10:12:07 
10:12:07     /tmp/jenkins/workspace/kata-containers-tests-fedora-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:61
10:12:07 ------------------------------
10:12:07 SSSSSS
10:12:07 
10:12:07 Summarizing 1 Failure:
10:12:07 
10:12:07 [Fail] [Serial Test] package manager update test check apt-get update and upgrade [It] should not fail 
10:12:07 /tmp/jenkins/workspace/kata-containers-tests-fedora-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:61
10:12:07 
10:12:07 Ran 19 of 249 Specs in 139.722 seconds
10:12:07 FAIL! -- 18 Passed | 1 Failed | 0 Pending | 230 Skipped --- FAIL: TestIntegration (139.72s)
```
It should be related to the missing 9p kernel patches, so what's status of this issue([packaging#505](https://github.com/kata-containers/packaging/issues/505))?

grahamwhaley:
@Pennyzct - indeed, I was hoping this got fixed. I've triggered some other PR CIs, so we will find out in the next few hours. /cc @chavafg @jcvenegas for update on if the fix landed and is working.

chavafg:
Hi, the failure on the `package manager update`  test should now be fixed. I've triggered failed jobs from yesterday, lets see.

nitkon:
lgtm
good to merge.

egernst:
optimization - not a stable candidate

grahamwhaley:
Just for future reference, in case somebody else hits this and does a search, if your `git` is very old, you will stumble into:

```
Install Qemu
can't load package: package github.com/kata-containers/packaging: no Go files in /home/gwhaley/go/src/github.com/kata-containers/packaging
error: unknown option `shallow-submodules'
usage: git clone [<options>] [--] <repo> [<dir>]
```

My git was v2.7.4 - from March 2016 !! (outdated Ubuntu 16.04 machine).
afaict, the `--shallow-submodules` was introduced in `git` v2.9.0, released in June 2016. So, not really an issue for us, but a historical note/KnowledgeBase entry.

grahamwhaley:
Oh, a further note, it appears that git v2.7.4 **is** the latest on Ubuntu 16.04 ?? So, let's watch out for that. I guess we are not running any 16.04 CI's any more @chavafg , or we might have hit this?
https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=git

chavafg:
hmmm, at least on this PR, the metrics job run on an ubuntu 16.04, but qemu is not being built, instead it is being downloaded from the obs packages. so we didn't hit this error, but we might hit it in the future, when we need to build qemu because the packaged one is not the latest.
We might also hit this on the cri-o job, which runs ubuntu 16.04

nitkon:
@grahamwhaley @chavafg: FYI Power CI runs on Ubuntu 1604.

Pennyzct:
Hi~ @grahamwhaley @chavafg So, do we need to add extra check for git version, in case it's too old to fail???

grahamwhaley:
@Pennyzct - a version check is probably going to be the easiest workaround I think, yes... sad, but true.

Pennyzct:
@grahamwhaley ok, I will raise a new PR to include the version check. ;)

grahamwhaley:
Here is a thought @Pennyzct (that maybe you already had..) - we can check on the `git` version rather than the distro type/version ? :-)

Pennyzct:
yes, share the same thought, that (`git version`) will be the most accurate check. ;)

Issue #1556, opened 2019-05-10
docker: docker images latest version available


alicefr:
Hi,

it happens sometime for s390x that the `latest` version of a docker image is not correctly update for a couple of days. We're monitoring that but it causes the `docker` testsuite failing because it's trying to pull an image that has not been correctly update. The error I get is:
```
no matching manifest for unknown in the manifest list entries
```
this issue breaks the kata CI. If we plan to run the kata CI for every project and ideally fo each PR, we should solve the issue.

A possible work-around is to refer to the architecture image. In this case, we avoid the broken reference of the multi-arch image. WDYT?

@Pennyzct @nitkon I'm wondering if you face a similar issue for ARM and Power.

jodh-intel:
Hi @alicefr - ooi, is this one particular image (busybox maybe?) or a variety?

alicefr:
@jodh-intel  no a variety. This morning was postgres, but it already happened with httpd, alpine and busybox too.

jodh-intel:
That almost sounds like a bug with docker hub, not giving you the latest "known working" and tested version of an image for s390x in this case presumably?

alicefr:
I guess more than a bug is a race condition. The s390x building server is not very fast and docker pushes the images when the amd64 image is ready. If you want an example, you can have a look at https://github.com/docker-library/httpd/issues/129 . The detailed explanation at https://github.com/docker-library/official-images/issues/3835

jodh-intel:
Errr, so they update `:latest` *for all architectures* only when a single architectures build passes...? Ouch!

alicefr:
yes, until Tianon finds a better solution

alicefr:
I've a patch let me pull it and then you can have a look if it sounds reasonable for you

jodh-intel:
Wow. Well, surely until they can make `:latest` change independently for each architecture, why can't they just make `:latest` change when the slowest build system completes? :)

alicefr:
@jodh-intel my proposal https://github.com/kata-containers/tests/pull/1557

jodh-intel:
Seems pretty harsh to potentially break all other arches though :(

tianon:
I commented on https://github.com/docker-library/official-images/issues/3835 (https://github.com/docker-library/official-images/issues/3835#issuecomment-491108329, https://github.com/docker-library/official-images/pull/5897), but we are hoping to have this one solved upstream by the end of this week. :smile:

[...]

PR #1557, opened 2019-05-10
docker: set the arch for the docker image

alicefr:
It happends that the latest version of the image is not available or
is broken for an architecture. To avoid this issue, we can refer
explicitly to the architecture with the tag arch/image

Fixes: #1556

Signed-off-by: Alice Frosi <afrosi@de.ibm.com>

### DIFF
--- a/integration/docker/docker.go
+++ b/integration/docker/docker.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
@@ -27,16 +28,16 @@ const (
 	Docker = "docker"
 
 	// Image used to run containers
-	Image = "busybox"
+	ImageName = "busybox"
 
 	// PostgresImage is the postgres image
-	PostgresImage = "postgres"
+	PostgresImageName = "postgres"
 
 	// DebianImage is the debian image
-	DebianImage = "debian"
+	DebianImageName = "debian"
 
 	// FedoraImage is the fedora image
-	FedoraImage = "fedora"
+	FedoraImageName = "fedora"
 
 	// StressImage is the vish/stress image
 	StressImage = "vish/stress"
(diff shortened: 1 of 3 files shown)


alicefr:
@grahamwhaley what we discussed in slack

grahamwhaley:
/test
as I'd like to see it fly...

alicefr:
@grahamwhaley you're right, this PR fixes the issue only for the docker test-suite

alicefr:
Another solution (that for the docker suite doesn't work) is to pre-pull the images with the arch and tag them locally with multi-arch name. Like
```ssh
docker pull s390x/alpine
docker tag s390x/alpine alpine
```
but this trick only works if the pull function doesn't try alwasy to pull from dockerhub. In this case, the locally tagged image will be substituted with image from Docker Hub

jodh-intel:
Just seen this problem on arm64:

```
10:09:59 Status: Downloaded newer image for alpine:3.7
10:09:59 
10:09:59 Stderr: 
10:09:59 Running command '/usr/bin/docker [docker pull postgres]'
10:09:59 command failed error 'exit status 1'
10:09:59 [docker pull postgres]
10:09:59 Timeout: 600 seconds
10:09:59 Exit Code: 1
10:09:59 Stdout: Using default tag: latest
10:09:59 latest: Pulling from library/postgres
10:09:59 
10:09:59 Stderr: no matching manifest for unknown in the manifest list entries
```

(from arm Ubuntu 18.04 CI on https://github.com/kata-containers/runtime/pull/1513).

alicefr:
If they solved the issue as Tianon said, we can leave this PR :)

jodh-intel:
Hi @alicefr - all the docker-related issues on #1556 are now closed, so presumably we can close this PR too (we can always re-open if the problem reoccurs of course )?

alicefr:
yes, I'm going to close it

PR #1571, opened 2019-05-14
filter_scheme: re-place filter scheme under .ci/filter and clean up redundant files

Pennyzct:
Hi~ @jodh-intel @alicefr
As you all suggested in #1555, we need to clean up duplicated codes under `.ci/arch`.
I think that the most major part is about we all share the same filter scheme scripts for docker integration test suites and make test suites.
here, I clean up all redundant files under `.ci/arch/aarch64`, `.ci/arch/s390x`, .`ci/arch/ppc64le`, and even `.ci/hypervisors/firecracker/`.


jcvenegas:
hi @Pennyzct seems that the refactor enabled some cpu hotplug in firecracker job that are not supported there.
@GabyCT  thanks for the heads up. I was already opening issues

Pennyzct:
Hi~@jcvenegas, my bad/(ㄒoㄒ)/~~, i will update asap ;)

Pennyzct:
Hi~ @jcvenegas @chavafg @chavafg 
For my carelessness, we should separate the `SKIP` option for firecracker and qemu, otherwise, in the future, what if AArch64 supports firecracker, it will firstly be occupied by here:
```
FILTER_FILE = .ci/hypervisors/$(KATA_HYPERVISOR)/filter_docker_$(KATA_HYPERVISOR).sh
ifneq ($(wildcard $(FILTER_FILE)),)
	SKIP := $(shell bash -f $(FILTER_FILE))
endif
```
and will be overwritten here with values defined for qemu:
```
 # Load architecture-dependent settings
ifneq ($(wildcard $(ARCH_FILE)),)
include $(ARCH_FILE)
endif
```
I have already updated the code, ptal ;)

Pennyzct:
Hi~ @jcvenegas @chavafg @jodh-intel  I get all green here~ 🎉🎊

jodh-intel:
Thanks @Pennyzct!

PR #1581, opened 2019-05-15
test: Add /dev/random entropy measurement

### LINKED ISSUE #1543: entropy/randomness: tests need improving

GabyCT:
We need to measure the availability of random data.

Fixes #1543

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,8 +148,7 @@ ramdisk:
 
 entropy:
 	bash -f .ci/install_bats.sh
-	cd integration/entropy && \
-	bats entropy_test.bats
+	bats integration/entropy/entropy_time.bats
 
 netmon:
 	systemctl is-active --quiet docker || sudo systemctl start docker
--- a/integration/entropy/entropy_time.bats
+++ b/integration/entropy/entropy_time.bats
@@ -1,0 +1,39 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+source /etc/os-release || source /usr/lib/os-release
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+setup() {
+	clean_env
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+}
+
+@test "measured time for /dev/random" {
+	output_file=$(mktemp)
+	block_size="4b"
+	expected_time="40"
+	copy_number="100000"
+	image="busybox"
+	cmd="dd if=/dev/random of=/dev/null bs=$block_size count=$copy_number"
+
+	docker run -i --runtime $RUNTIME $image sh -c "$cmd" 2> "$output_file"
+	measured_time=$(cat $output_file | grep seconds | cut -d',' -f2 | cut -d'.' -f1)
+	[ "$measured_time" -le "$expected_time" ]
+}
+
+teardown() {
+	clean_env
+	rm "$output_file"
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+}


integration/entropy/entropy_time.bats
```diff
@@ -0,0 +1,37 @@
...
+@test "entropy for /dev/random" {
+	output_file=$(mktemp)
+	block_size="4b"
+	entropy_time="20"
+	copy_number="100000"
+
+	{ TIMEFORMAT="%R"; time dd if=/dev/random of=/dev/null bs="$block_size" count="$copy_number" 2>/dev/null ; } 2> "$output_file"
+	measured_entropy=$(cat $output_file | cut -d '.' -f1)
+	[ "$measured_entropy" -le "$entropy_time" ]
+}
```
jcvenegas: Missing indentation

integration/entropy/entropy_time.bats
```diff
@@ -0,0 +1,37 @@
...
+
+setup() {
+	clean_env
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+}
+
+@test "entropy for /dev/random" {
```
chavafg: this is testing that time for this operation is reasonable, right? maybe change the name to reflect that?

integration/entropy/entropy_time.bats
```diff
@@ -0,0 +1,37 @@
...
+	output_file=$(mktemp)
+	block_size="4b"
+	entropy_time="10"
```
chavafg: should we add a comment of why this is a good value?
also, can the variable name be changed to something like `expected_time` ?

integration/entropy/entropy_time.bats
```diff
@@ -0,0 +1,37 @@
...
+	copy_number="100000"
+
+	{ TIMEFORMAT="%R"; time dd if=/dev/random of=/dev/null bs="$block_size" count="$copy_number" 2>/dev/null ; } 2> "$output_file"
+	measured_entropy=$(cat $output_file | cut -d '.' -f1)
```
chavafg: `measured_time=...`?

integration/entropy/entropy_time.bats
```diff
@@ -0,0 +1,37 @@
...
+	[ "$measured_entropy" -le "$entropy_time" ]
+}
+
+teardown() {
+	clean_env
+	sudo rm "$output_file"
```
chavafg: sudo shouldn't be needed here

GabyCT:
@chavafg changes applied

Makefile
```diff
@@ -150,6 +150,7 @@ entropy:
 	bash -f .ci/install_bats.sh
 	cd integration/entropy && \
 	bats entropy_test.bats
+	bats entropy_time.bats
```
grahamwhaley: A niggle... is this right? I'm not sure the `cd` carries over to the 2nd line does it?
We seem to have a mix through the `Makefile` where some `bats` tests use the `cd && bats` idiom, and others use the `bats fullpath` idiom. I prefer the second (the fullpath) one?
Maybe make both these entropy bats lines `bats <fullpath to batsfile>` style?

GabyCT:
@chavafg changes applied

GabyCT:
@grahamwhaley changes applied

grahamwhaley:
Thanks @GabyCT - need @chavafg to ack his change request I think.

integration/entropy/entropy_time.bats
```diff
@@ -0,0 +1,39 @@
...
+	[ "$status" -eq 0 ]
+}
+
+@test "measured time for /dev/random" {
+	output_file=$(mktemp)
+	block_size="4b"
+	expected_time="20"
+	copy_number="100000"
+	image="busybox"
+	cmd="time dd if=/dev/random of=/dev/null bs=$block_size count=$copy_number"
```
chavafg: do we really need `time` ? I see that later you are doing `.. | grep seconds`, which comes from the output of `dd`

grahamwhaley: ah, if `dd` gives the time, then sure we can use that. I my home run local example I was thinking of the seconds from the time. heh.

GabyCT:
@chavafg thanks for the feedback , changes applied

chavafg:
thanks @GabyCT

PR #1583, opened 2019-05-15
integration/kubernetes: Add test to check initContainers and shared v…

### LINKED ISSUE #1582: kubernetes: add test to check initContainer
Add test to check if the initContainer runs before the container and at the same time check if a volume can be shared between them

devimc:
…olumes

Add a simple test to check if the initContainer is started before the
container. Also, this test check if volumes can be shared between the
initContainer and the container.

fixes #1582

### DIFF
--- a/integration/kubernetes/k8s-shared-volume.bats
+++ b/integration/kubernetes/k8s-shared-volume.bats
@@ -30,6 +30,20 @@ setup() {
 	kubectl exec "$pod_name" -c "$first_container_name" -- sh -c "$cmd" | grep "$msg"
 }
 
+@test "initContainer with shared volume" {
+	pod_name="initcontainer-shared-volume"
+	last_container="last"
+
+	# Create pod
+	kubectl create -f "${pod_config_dir}/initContainer-shared-volume.yaml"
+
+	# Check pods
+	kubectl wait --for=condition=Ready pod "$pod_name"
+
+	cmd='test $(cat /volume/initContainer) -lt $(cat /volume/container)'
+	kubectl exec "$pod_name" -c "$last_container" -- sh -c "$cmd"
+}
+
 teardown() {
 	kubectl delete pod "$pod_name"
 }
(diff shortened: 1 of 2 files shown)


grahamwhaley:
nudged a rebuild on metrics CI - it only just failed the boot time test, and the metrics CI as a whole looks stable, so let's respin..

grahamwhaley:
@devimc - did you just merge your own PR? :-)
Not something we normally do on the kata repos...
I think that's the second (not just you...) I've seen in the last 24h - let's keep an eye on that practice folks....

integration/kubernetes/k8s-shared-volume.bats
```diff
@@ -30,6 +30,20 @@ setup() {
...
+	last_container="last"
+
+	# Create pod
+	kubectl create -f "${pod_config_dir}/initContainer-shared-volume.yaml"
+
+	# Check pods
+	kubectl wait --for=condition=Ready pod "$pod_name"
+
+	cmd='test $(cat /volume/initContainer) -lt $(cat /volume/container)'
+	kubectl exec "$pod_name" -c "$last_container" -- sh -c "$cmd"
```
grahamwhaley: just for ref, I found today that `kubectl wait` has a `--timeout` option - but, I think it defaults to 30s - just for ref, in case we don't think that is long enough etc.

devimc: I wonder if the default timeout should be equal to the other tests. :thinking:  https://github.com/kata-containers/tests/blob/master/Makefile#L8

grahamwhaley: Probably makes sense - at least to try and define the default timeouts for some operations. I didn't even know we had that `TIMEOUT` in the Makefile ;-)
In the k8s based script I'm writing right now I set some timeout values at the top of the script and then use them throughout, so at least they are easier to find/bump later if we find we need.

chavafg: I think it is fine for having the default (30s) here. If a pod takes more than that to be running, there might be something going wrong.

PR #1587, opened 2019-05-16
Doc fixes

### LINKED ISSUE #1586: Add missing doc links
Some of the docs are referenced and thus hard to find.

jodh-intel:
Added links to the unreferenced docs in this repo to make them more discoverable.

Also added some missing TOCs and simplified a link.

Fixes #1586.
    
### DIFF
--- a/cmd/checkmetrics/README.md
+++ b/cmd/checkmetrics/README.md
@@ -1,5 +1,22 @@
 # checkmetrics
 
+* [Overview](#overview)
+    * [JSON file format](#json-file-format)
+* [baseline TOML layout](#baseline-toml-layout)
+    * [Supported file types](#supported-file-types)
+    * [Supported checktypes](#supported-checktypes)
+* [Options](#options)
+    * [TOML basefile path (mandatory)](#toml-basefile-path-mandatory)
+    * [Debug mode](#debug-mode)
+    * [Log file path](#log-file-path)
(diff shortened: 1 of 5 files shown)


cmd/checkmetrics/README.md
```diff
@@ -1,5 +1,23 @@
 # checkmetrics
 
+* [checkmetrics](#checkmetrics)
```
grahamwhaley: slightly confused - should this and the next patch maybe folded into one?

jodh-intel: Ah - that link can in fact be zapped. But I'm not quite clear on what you mean - the next commit is for `checkcommits` ?There are a number of ways these changes could be ordered but I've tried to patch-up the existing files, then finally add in missing links.

grahamwhaley: doh, I mis-read and muddle up `checkcommits` and `checkmetrics` in my head all the time - heh, and I tried to stare quite hard at this one - mea culpa - ignore me!

chavafg:
I've slightly modified most of the jobs in jenkins to not run if only `*.md` were modified. (except the ubuntu job which is `Required`).
Forgot to add the change to the metrics and ARM job, but have already canceled the arm job manually. 

The downside I see about this approach is that if we have `Required` on all jobs (which I think we do when they are stable), the PR will be blocked to as it will not get the status from the jenkins server.

jodh-intel:
Merging - no point waiting for a CI to test a doc-only PR...

PR #1590, opened 2019-05-16
tests: Remove untrusted_workloads k8s yaml

### LINKED ISSUE #1589: Remove yaml from k8s untrusted_workloads
Remove `pod-shared-volume.yaml` from `integration/kubernetes/untrusted_workloads` as we are only using runtimeclass for K8s.

GabyCT:
This will remove the pod-shared-volume.yaml from the untrusted_workloads
as now we are only using runtimeclass for K8s testing.

Fixes #1589


chavafg:
now I wonder why only ubuntu job was launched, made modifications to not run the other jobs if only `md`'s were modified, but seems that feature it is not working well.

chavafg:
fixed, now all jobs triggered

Issue #1593, opened 2019-05-17
Save trace spans for CI artifact collector


jodh-intel:
The [tracing test](https://github.com/kata-containers/tests/blob/master/tracing/tracing-test.sh) performs lots of checks on the collected Jaeger trace spans. However, if the test fails in the CI, it is almost impossible to debug.

Save the trace spans to allow the CI to attach them to a Jenkins job as artifacts.

jodh-intel:
Hi @chavafg - any thoughts on where I should write the trace spans to? Maybe one of...?:

```
/tmp/traces/${jaeger_service}.log
/var/log/kata-tests/jaeger-tracing/${jaeger_service}.log
```

Once we've decided, I'll write the logs to that location and then update `.ci/teardown.sh` to save them.

(I could update `.ci/teardown.sh` to extract the trace spans from Jaeger directly but I'd rather not do that as we'd have duplicate logic in the teardown script and the tracing test).

chavafg:
I think we can use `/var/log/kata-tests/...`. Maybe later we will need to save other logs and we can use that same location.

jodh-intel:
Thanks @chavafg - #1594 raised.

PR #1594, opened 2019-05-17
Log Jaeger traces

jodh-intel:
Defined `KATA_TESTS_LOGDIR` in the common bash library so that tests can write log information to a known location.

Save the trace spans logged with Jaeger to a directory below KATA_TESTS_LOGDIR` so that if the test fails, the spans can be analysed. This will allow the saved trace spans to be attached to a failing PR as artifacts. If the test passes, the cached spans are deleted.

Fixes: #1593.
    
### DIFF
--- a/tracing/tracing-test.sh
+++ b/tracing/tracing-test.sh
@@ -17,6 +17,8 @@ source "${SCRIPT_PATH}/../lib/common.bash"
 
 RUNTIME=${RUNTIME:-kata-runtime}
 
+TRACE_LOG_DIR=${TRACE_LOG_DIR:-${KATA_TESTS_LOGDIR}/traces}
+
 jaeger_server=${jaeger_server:-localhost}
 jaeger_ui_port=${jaeger_ui_port:-16686}
 jaeger_docker_container_name="jaeger"
@@ -51,6 +53,8 @@ start_jaeger()
 		-p 14268:14268 \
 		-p 9411:9411 \
 		"$jaeger_docker_image"
+
+	sudo mkdir -m 0750 -p "$TRACE_LOG_DIR"
 }
 
 stop_jaeger()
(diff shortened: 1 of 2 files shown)


tracing/tracing-test.sh
```diff
@@ -26,6 +28,13 @@ jaeger_docker_container_name="jaeger"
 cleanup(){
 	stop_jaeger 2>/dev/null || true
 	.ci/configure_tracing_for_kata.sh disable
+
+	# The tests worked so remove the logs
+	if sudo [ -d "$TRACE_LOG_DIR" ] && [ "$TRACE_LOG_DIR" != "/" ]
+	then
+		info "Removing cached trace logs from $TRACE_LOG_DIR"
+		sudo rm -rf "$TRACE_LOG_DIR"
```
jcvenegas: @jodh-intel  I am  not clear when the trace logs are keep , seems that they are removed all the time ?

jodh-intel: Good catch @jcvenegas!! I've moved the deletion logic to "main" now.

jodh-intel:
@grahamwhaley - metrics CI failure:

```
18:17:31 INFO: delete stale docker resource under /var/lib/docker
18:17:31 FATAL: command execution failed
18:17:31 java.io.EOFException
```

jodh-intel:
@chavafg - I've restarted the initrd CI, which failed with:

```
19:20:07 not ok 1 Pod quota
19:20:07 # (in test file k8s-pod-quota.bats, line 29)
19:20:07 #   `kubectl wait --for=condition=Available deployment/${deployment_name}' failed
19:20:07 # INFO: k8s configured to use runtimeclass
19:20:07 # resourcequota/pod-quota created
19:20:07 #     pods: "2"
19:20:07 #     pods: "2"
19:20:07 # deployment.apps/deploymenttest created
19:20:07 # error: timed out waiting for the condition on deployments/deploymenttest
19:20:07 # resourcequota "pod-quota" deleted
19:20:07 # deployment.extensions "deploymenttest" deleted
19:20:07 Makefile:104: recipe for target 'kubernetes' failed
```

grahamwhaley:
metrics: odd fail - looks like it failed to do a cleanup of the docker dirs, which makes me feel maybe something got 'stuck' and was holding a file open. We reboot the metrics machine between (roughly) builds, so that should self-clear (and that is precisely why we do the reboot etc....). The build history on the metrics job looks pretty good/stable - I'll nudge a rebuild.

Issue #1536, opened 2019-05-06
Add virtiofs support for CI 


chavafg:
Add changes to be able to run kata using virtiofs

PR #1595, opened 2019-05-18
nemu: Enable static nemu for ci

ganeshmaharaj:
Build NEMU statically for every CI run to make sure we test a setup
similar to the one we are shipping with Kata.

Fixes: #1536
Signed-off-by: Ganesh Maharaj Mahalingam <ganesh.mahalingam@intel.com>

### DIFF
--- a/.ci/install_nemu.sh
+++ b/.ci/install_nemu.sh
@@ -19,21 +19,22 @@ arch=$("${cidir}"/kata-arch.sh -d)
 install_nemu() {
 	local nemu_repo=$(get_version "assets.hypervisor.nemu.url")
 	local nemu_version=$(get_version "assets.hypervisor.nemu.version")
+	PACKAGING_REPO="github.com/kata-containers/packaging"
+	NEMU_TAR="kata-nemu-static.tar.gz"
 	case "$arch" in
 	x86_64)
-		local nemu_bin="qemu-system-${arch}_virt"
-		;;
-	aarch64)
-		local nemu_bin="qemu-system-${arch}"
+		local nemu_bin="nemu-system-${arch}"
 		;;
 	*)
 		die "Unsupported architecture: $arch"
 		;;
 	esac
 
-	curl -LO "${nemu_repo}/releases/download/${nemu_version}/${nemu_bin}"
-	sudo install -o root -g root -m 0755 "${nemu_bin}" "/usr/local/bin"
-	rm -rf "${nemu_bin}"
+	go get -d "${PACKAGING_REPO}" || true
+
+	prefix="${KATA_NEMU_DESTDIR}" ${GOPATH}/src/${PACKAGING_REPO}/static-build/nemu/build-static-nemu.sh
+	sudo tar -xvf ${NEMU_TAR} -C /
+	rm ${NEMU_TAR}
 }
 
 install_firmware() {
@@ -49,4 +50,5 @@ install_firmware() {
 }
 
 install_nemu
-install_firmware
+# NEMU static tar now also includes the OVMF.fd bios file.
+#install_firmware
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -86,14 +86,11 @@ if [ "$KATA_HYPERVISOR" == "qemu" ]; then
 	"${cidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker configure -r kata-runtime -f
 elif [ "$KATA_HYPERVISOR" == "nemu" ]; then
 	echo "Configure Nemu as Kata Hypervisor"
-	sudo sed -i -e 's|machine_type = "pc"|machine_type = "virt"|' "${runtime_config_path}"
-	sudo sed -i -e 's|firmware = ""|firmware = "/usr/share/nemu/firmware/OVMF.fd"|' "${runtime_config_path}"
+	sudo crudini --set "${runtime_config_path}" hypervisor.qemu machine_type \"virt\"
+	sudo crudini --set "${runtime_config_path}" hypervisor.qemu firmware \"${KATA_NEMU_DESTDIR}/share/kata-nemu/OVMF.fd\"
 	case "$arch" in
 	x86_64)
-		sudo sed -i -e "s|\"/usr/bin/qemu-lite-system-${arch}\"|\"/usr/local/bin/qemu-system-${arch}_virt\"|" "${runtime_config_path}"
-		;;
-	aarch64)
-		sudo sed -i -e "s|\"/usr/bin/qemu-lite-system-${arch}\"|\"/usr/local/bin/qemu-system-${arch}\"|" "${runtime_config_path}"
+		sudo crudini --set "${runtime_config_path}" hypervisor.qemu path \"${KATA_NEMU_DESTDIR}/bin/nemu-system-${arch}\"
 		;;
 	*)
 		die "Unsupported architecture: $arch"
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -8,6 +8,7 @@
 
 export KATA_RUNTIME=${KATA_RUNTIME:-kata-runtime}
 export KATA_KSM_THROTTLER=${KATA_KSM_THROTTLER:-no}
+export KATA_NEMU_DESTDIR=${KATA_NEMU_DESTDIR:-"/usr/local"}
 
 # Name of systemd service for the throttler
 KATA_KSM_THROTTLER_JOB="kata-ksm-throttler"


.ci/install_nemu.sh
```diff
@@ -19,21 +19,23 @@ arch=$("${cidir}"/kata-arch.sh -d)
...
 		;;
 	*)
 		die "Unsupported architecture: $arch"
 		;;
 	esac
 
-	curl -LO "${nemu_repo}/releases/download/${nemu_version}/${nemu_bin}"
-	sudo install -o root -g root -m 0755 "${nemu_bin}" "/usr/local/bin"
-	rm -rf "${nemu_bin}"
+	mkdir -p "${GOPATH}/src"
```
jodh-intel: Is this actually needed or does `go get` create it if needed?

.ci/install_nemu.sh
```diff
@@ -19,21 +19,23 @@ arch=$("${cidir}"/kata-arch.sh -d)
...
+	go get -d "${PACKAGING_REPO}" || true
+
+	prefix="/usr/local" ${GOPATH}/src/${PACKAGING_REPO}/static-build/nemu/build-static-nemu.sh
+	sudo tar -xvf ${NEMU_TAR} -C /
+	rm -rf	${NEMU_TAR}
```
jodh-intel: Nit: `rm` should suffice?

.ci/install_nemu.sh
```diff
@@ -19,21 +19,23 @@ arch=$("${cidir}"/kata-arch.sh -d)
...
 		;;
 	esac
 
-	curl -LO "${nemu_repo}/releases/download/${nemu_version}/${nemu_bin}"
-	sudo install -o root -g root -m 0755 "${nemu_bin}" "/usr/local/bin"
-	rm -rf "${nemu_bin}"
+	mkdir -p "${GOPATH}/src"
+	go get -d "${PACKAGING_REPO}" || true
+
+	prefix="/usr/local" ${GOPATH}/src/${PACKAGING_REPO}/static-build/nemu/build-static-nemu.sh
```
jodh-intel: It might be worth creating something like the following in `lib.sh` so that it can be reused in `install_runtime.sh` to avoid the two hard-coded values:

```sh
# Default directory to install unpackaged version of nemu into
KATA_NEMU_DESTDIR=${KATA_NEMU_DESTDIR:-/usr/local}"
```

.ci/install_nemu.sh
```diff
@@ -19,21 +19,23 @@ arch=$("${cidir}"/kata-arch.sh -d)
 install_nemu() {
 	local nemu_repo=$(get_version "assets.hypervisor.nemu.url")
 	local nemu_version=$(get_version "assets.hypervisor.nemu.version")
+	PACKAGING_REPO="github.com/kata-containers/packaging"
+	NEMU_TAR="kata-nemu-static.tar.gz"
 	case "$arch" in
 	x86_64)
-		local nemu_bin="qemu-system-${arch}_virt"
-		;;
-	aarch64)
```
jodh-intel: Why is aarch64 being dropped? Even if we don't provide the static binaries today, NEMU works on arm64 too so can't we leave it as we should endeavor to add test support for that arch at some future point too?

.ci/install_runtime.sh
```diff
@@ -87,13 +87,10 @@ if [ "$KATA_HYPERVISOR" == "qemu" ]; then
...
 	sudo sed -i -e 's|machine_type = "pc"|machine_type = "virt"|' "${runtime_config_path}"
-	sudo sed -i -e 's|firmware = ""|firmware = "/usr/share/nemu/firmware/OVMF.fd"|' "${runtime_config_path}"
+	sudo sed -i -e 's|firmware = ""|firmware = "/usr/local/share/kata-nemu/OVMF.fd"|' "${runtime_config_path}"
 	case "$arch" in
 	x86_64)
-		sudo sed -i -e "s|\"/usr/bin/qemu-lite-system-${arch}\"|\"/usr/local/bin/qemu-system-${arch}_virt\"|" "${runtime_config_path}"
-		;;
-	aarch64)
-		sudo sed -i -e "s|\"/usr/bin/qemu-lite-system-${arch}\"|\"/usr/local/bin/qemu-system-${arch}\"|" "${runtime_config_path}"
+		sudo sed -i -e "s|\"/usr/bin/qemu-lite-system-${arch}\"|\"/usr/local/bin/nemu-system-${arch}\"|" "${runtime_config_path}"
```
jodh-intel: Nit: Again, I appreciate you're keeping this change similar to what was there before, but I think we should be trying to use `crudini` for all config file changes now tbh as the syntax is much clearer and the tool is already being installed for all distros.

.ci/install_nemu.sh
ganeshmaharaj: Don't think that is needed. I added that for consistency. Dropping it should be fine.

.ci/install_nemu.sh
ganeshmaharaj: We can add aarch64 here back, but it does come with some limitations. 1. It is not statically built, so we will need to add more scaffolding to the install scripts to get all the dependent libraries for it to work. It will also only work with xenial and probably fail on other distros. In the current form, it is not really consumable. As stage 2, we can start shipping `aarch64` static binaries also along with `x86_64` and then re-enable it. I can file a bug for that, if that approach would work.

amshinde: @Pennyzct @justin-he Is this something that you could help with? We really would like to see support for building nemu on arm.
@ganeshmaharaj Can you open a separate issue for this, and list out what issues you are running into there?

Pennyzct:
Hi~ @jongwu we definitely want to lend a hand here. We will firstly try to build static binary on our local machine and make it work. :)

amshinde:
Thanks @Pennyzct !

jongwu:
@Pennyzct ok, I will take it.

jongwu:
@ganeshmaharaj @amshinde @jodh-intel  @Pennyzct now I have enabled nemu on arm64 after some modification work. what's next?

jodh-intel:
@jongwu - I think the next step is for @ganeshmaharaj to undo the `aarch64` removal changes on this PR.

ganeshmaharaj:
@jongwu is there a way we can enable the static build of arm64 similar to this one? https://github.com/kata-containers/packaging/pull/522/files#diff-62470642d2b1b882e092c3185db7f88b that would help us generate both the sets of binaries for CI and ship the same with kata for all future releases.

.ci/install_nemu.sh
ganeshmaharaj: https://github.com/kata-containers/tests/issues/1620

chavafg:
I think we can re-add arm once the static build for that arch is ready.

jongwu:
@ganeshmaharaj after disable avx2 and modify the source code, nemu can be built as the way in https://github.com/kata-containers/packaging/pull/522/files#diff-62470642d2b1b882e092c3185db7f88b .
@jodh-intel @chavafg how to transfer my patch to nemu? as we can't touch nemu project for the legal issue

ganeshmaharaj:
@jongwu if we can build NEMU for arm64, with the current scripts, can you kindly expand the `build_nemu_static.sh` script to make builds for both x86 and arm64 and make them both part of the giant tarball? we don't need to modify NEMU sources, just our way of configuring them to build. ohh just out of curiosity, are you able to cross compile it on an x86 machine?

ganeshmaharaj:
Not sure if i should trigger all the CI given this only touches NEMU. WIll refrain from it unless folks feel otherwise.

chavafg:
lets run them as ubuntu job is required.
/test

ganeshmaharaj:
@chavafg i am guessing the failed jobs are all expected and we can proceed merging this change?

chavafg:
yeah, but we need another ack to be able to merge.
@kata-containers/tests, ptal

jongwu:
@ganeshmaharaj for now, I can't build nemu on arm64 only modify build_nemu_static.sh. hid_dev should be enabled which must be done when make, as CONFIG_HID_DEV can not be configured by configure script. so I need change Dockerfile.
All these work are on arm64 machine, so no cross compile on x86.

ganeshmaharaj:
@jongwu Aah. I believe it would be easy if we can enable cross compile within a container, this way we can do the build both for CI and for releases on any backend machine. Is there a way we can help to get this enabled from your side?

jongwu:
@ganeshmaharaj  I try to cross compile nemu on x86, but building cross compile tool chain blocks me. I find it not easy for me to get all dependency library in. I am appreciate of you that you can involved in and help me to do that.
but another concern is I can't build nemu on arm64 machine by just modify build_nemu_static.sh, I need change Dockerfile, which maybe not make sense from your part.

jongwu:
@ganeshmaharaj I have successfully cross compiled nemu for arm64 on x86 in container.

PR #1602, opened 2019-05-20
ci: Put all the packages to install in a single transaction

### LINKED ISSUE #1560: coalesce all packages to install into single transaction
Looking at https://github.com/kata-containers/tests/pull/1559 made me think that we should consider refactoring the `.ci/setup_env_${distro}.sh` scripts so all packages are installed in a single `apt`/`yum`/`dnf`/`zypper` transaction. That should be more efficient and quicker:

Something like this:

```bash
# Array of packages.
#
# Each element is in the format:
#
#   description:packages
#
pkg_sets=()

# XXX: Put all the packages to install here, one set per line
pkg_sets+=(libsystemd:systemd-devel)
pkg_sets+=(cri-containerd dependencies:libseccomp-devel btrfs-progs-devel)

# space-separated list of packages to install
pkgs_to_install=

info "The following package sets will be installed:\n"

for set in ${pkg_sets[@]}
do
    descr=$(echo "$set"|cut -d: -f1)
    pkgs=$(echo "$set"|cut -d: -f2-)

    info "$descr ($pkgs)"
    pkgs_to_install+=" $pkgs"
done

chronic sudo -E apt -y install $pkgs_to_install
```

/cc @chavafg, @marcov.

GabyCT:
To avoid issues like choosing the wrong package installator, this
change tries to install all the required packages into a single
transaction.

Fixes #1560

### DIFF
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -34,42 +34,37 @@ echo "Update repositories"
 sudo -E yum -y update
 
 echo "Install chronic"
-sudo -E yum install -y moreutils
+sudo -E yum -y install moreutils
 
-echo "Install kata containers dependencies"
-chronic sudo -E yum install -y libtool libtool-ltdl-devel device-mapper-persistent-data lvm2 device-mapper-devel libtool-ltdl bzip2 m4 \
-	 gettext-devel automake alien autoconf bc pixman-devel coreutils
+declare -A packages=( \
+	[kata_containers_dependencies]="libtool libtool-ltdl-devel device-mapper-persistent-data lvm2 device-mapper-devel libtool-ltdl" \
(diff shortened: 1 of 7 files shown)


grahamwhaley:
/test
coz I'd like to see what it looks like in the logs as well ;-)

grahamwhaley:
We have a number of CI fails. From the 18.04 for instance:

```
09:27:45 Reading package lists...
09:27:45 Building dependency tree...
09:27:45 Reading state information...
09:27:45 E: Unable to locate package moretuils
```

Heh, I see you have `chronic` on the single package install line, but `chronic` comes from the `moreutils` package - you may have to install `moreutils` first, and then all the other packages... I think we'd probably like to keep the `chronic` in place if we can to reduce log size in the default 'it worked' case.

Issue #1604, opened 2019-05-21
Make install_yq function in ci/lib.sh a standalone script


marcov:
Moving  `install_yq` in a dedicated script allows other repositories to reuse it. 
Also, this complements the set of scripts in `.ci/` used to install software.

PR #1605, opened 2019-05-21
ci: move install_yq function in lib.sh a dedicate script

marcov:
Moving install_yq in a dedicated script allows other repositories to reuse it.
Also, this complements the set of scripts in .ci/ used to install software.

Fixes: #1604

Signed-off-by: Marco Vedovati <mvedovati@suse.com>

### DIFF
--- a/.ci/install_yq.sh
+++ b/.ci/install_yq.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright (c) 2019 SUSE LLC
 #
--- a/.ci/install_yq.sh
+++ b/.ci/install_yq.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 SUSE LLC
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -eu
+
+function install_yq() {
+	local cidir=$(dirname "${BASH_SOURCE[0]}")
+	# source lib.sh to make sure GOPATH is set
+	source ${cidir}/lib.sh
+	[ -x  "${GOPATH}/bin/yq" ] && return
+
+	local yq_path="${GOPATH}/bin/yq"
+	local yq_pkg="github.com/mikefarah/yq"
+	local goos="$(uname -s)"
+	local goarch="$(${cidir}/kata-arch.sh --golang)"
+
+	mkdir -p "${GOPATH}/bin"
+
+	# Workaround to get latest release from github (to not use github token).
+	# Get the redirection to latest release on github.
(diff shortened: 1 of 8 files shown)


.ci/install_yq.sh
```diff
@@ -0,0 +1,63 @@
...
+
+function install_yq() {
+	local yq_path="${GOPATH}/bin/yq"
+	local yq_pkg="github.com/mikefarah/yq"
+	[ -x  "${GOPATH}/bin/yq" ] && return
+
+	read -r -a sysInfo <<< "$(uname -sm)"
+
+	case "${sysInfo[0]}" in
+	"Linux" | "Darwin")
```
grahamwhaley: nit: I think we tend to indent the switch items in the rest of our bash scripts ;-)

.ci/install_yq.sh
```diff
@@ -0,0 +1,63 @@
...
+		goos="${sysInfo[0],}"
+		;;
+	"*")
+		die "OS ${sysInfo[0]} not supported"
+		;;
+	esac
+
+	case "${sysInfo[1]}" in
+	"aarch64")
```
grahamwhaley: hmm, I thought we had a func to do the arch name conversions - but, I see this is the code we originally had in the `install_yq` func, so just a note, not a blocker.

.ci/install_yq.sh
```diff
@@ -0,0 +1,63 @@
...
+#
+# Copyright (c) 2019 SUSE LLC
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+function install_yq() {
+	local yq_path="${GOPATH}/bin/yq"
```
grahamwhaley: I wonder if we should check `GOPATH` is set?
It is a long standing out of date requirement of our CI, as `go` no longer needs `GOPATH` to be set and will default.

jodh-intel: Yep - you could just source `.ci/lib.sh` which handles this:

- https://github.com/kata-containers/tests/blob/master/.ci/lib.sh#L22..L28

marcov: I can add a `set -u` to fail if GOPATH is not set. I think for the time being `$GOPATH/bin` is the most obvious destination for the `yq` executable.

marcov:
Thanks @grahamwhaley for the review, I think they all makes sense. I force-pushed some updates following your suggestions.

.ci/install_yq.sh
marcov: missed you comment @jodh-intel. I am fine doing that too. It's going to introduce some mutual dependency between the two file, but that shouldn't be circular.

marcov:
@jodh-intel @grahamwhaley @devimc I figured out some things on Travis with macOS were broken, so I booted up my Mac and fixed those. Can you take another look please?

/test

Issue #1588, opened 2019-05-16
Update static markdown checks to ensure all docs in a repo are referenced


jodh-intel:
See: https://github.com/kata-containers/documentation/issues/475.

PR #1611, opened 2019-05-22
functional: skip state test on fedora

chavafg:
State test has been failing randomly for a while
and haven't been able to reproduce locally. Let's
skip the test to have a stable CI while we debug
the issue.

The issue where we are tracking the failure is:
#1531

Fixes: #1610.

Signed-off-by: Salvador Fuentes <salvador.fuentes@intel.com>

PR #1612, opened 2019-05-22
CI: Add unreferenced doc check

jodh-intel:
Add a new test to the CI static check script to look for markdown documents that are not referenced by any other. With a small number of exceptions, most markdown files should be linked to from other markdown documents in the same repository to allow all in-repository documentation to be discovered by link navigation.

This check necessitated updating the markdown check tool to dump all links - see the updated `README` for details.

Fixes: #1588.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -375,17 +375,19 @@ check_docs()
 	info "Checking documentation"
 
 	local doc
+	local all_docs
 	local docs
 	local docs_status
 	local new_docs
 	local new_urls
 	local url
 
+	all_docs=$(find . -name "*.md" | grep -v "vendor/" | sort || true)
+
 	if [ "$specific_branch" = "true" ]
 	then
 		info "Checking all documents in $branch branch"
-
-		docs=$(find . -name "*.md" | grep -v "vendor/" || true)
+		docs="$all_docs"
 	else
 		info "Checking local branch for changed documents only"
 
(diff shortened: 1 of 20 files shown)


.ci/static-checks.sh
```diff
@@ -502,7 +549,7 @@ check_docs()
 		exit 1
 	fi
 
-	rm -f "$url_map" "$invalid_urls"
+	rm -f "$url_map" "$invalid_urls" "$md_links"
```
grahamwhaley: Normally I'd ask for the `rm`s of tmpfiles to also be caught with a TRAP for the case when we do die. I'll let you decide if that is worth it or not.

jodh-intel: A few notes:

- This change is simply adding an extra file to the existing list.
- All the check functions use local variables and do their own cleanup currently.
- Retaining temporary files if this script fails is useful for debugging the failure, so the current behaviour is pretty sane I think.
- I agree that it might be useful to move the `rm` calls to a generic trap handler so that those files will be deleted if the script is successful, but retained on error (the trap handler could list the files to further help with debugging).

I've raised https://github.com/kata-containers/tests/issues/1613 so we can deal with that issue on another PR.

jodh-intel:
Fixed the lint errors, so...

/retest

jodh-intel:
Once more for luck...

/retest

jodh-intel:
Fedora failure is the known issue fixed by https://github.com/kata-containers/tests/pull/1611.

Issue #1616, opened 2019-05-22
Fix test coverage collection


jodh-intel:
golang test coverage reporting is currently broken as `coverage.txt` is not being updated with coverage data.

PR #1617, opened 2019-05-22
coverage: Fix unit test coverage collection

jodh-intel:
We were running `go test` with `-outputdir=` causing the coverage data to be written to a temporary directory, and then discarded. Remove the option to force the coverage data to be written to `coverage.txt` in the top-level repo directory.

Fixes: #1616.

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -246,11 +246,8 @@ main()
 
 	[ -z "$test_packages" ] && echo "INFO: no golang code to test" && exit 0
 
-	# directory to use for temporary files
-	golang_tmp=$(mktemp -d)
-
 	# KATA_GO_TEST_FLAGS can be set to change the flags passed to "go test".
-	go_test_flags=${KATA_GO_TEST_FLAGS:-"-v $race -timeout $timeout_value -outputdir \"${golang_tmp}\""}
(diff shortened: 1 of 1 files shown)


jodh-intel:
> do we still run codecov ??

@grahamwhaley - yes and no. "yes" we use codecov but clearly it wouldn't have been working due to the bug this PR fixes.

But the problem appears to have been compounded by the fact that -- I *think* -- codecov was effectively paused as it's been quietly asking for extra perms in the repo settings, but github doesn't provide any notification to say "there's an app waiting for a response from you". I've now granted that extra perm (on the github "**Check**" "thing") but prior to that, I think codecov was effectively a NOP.

This is an unfortunate situation caused by us taking our eyes off it for too long I think. That said, I do recall us having problems with codecov and the previous service we used. Hopefully, those issues have been resolved but lets all keep our eyes on these reports now (and yell if you don't see them! ;)

jodh-intel:
ARM CI failing with:

```
23:35:29     Expected
23:35:29         <string>: [    0.000000] systemd[60]: chronyd.service: Failed to set up mount namespacing: No such file or directory
23:35:29         [    0.000000] systemd[60]: chronyd.service: Failed at step NAMESPACE spawning /usr/sbin/chronyd: No such file or directory
23:35:29     to be empty
```

/cc @devimc.

jodh-intel:
Added a second commit to ignore auto-generated code.

PR #1635, opened 2019-05-27
tests: Add k8s test for job to completion

### LINKED ISSUE #1634: k8s test for job for completion
Add k8s test for a job for completion

GabyCT:
This adds a test for a job that creates a pod and ensures that its
successful completion is reached.

Fixes #1634

### DIFF
--- a/integration/kubernetes/k8s-job.bats
+++ b/integration/kubernetes/k8s-job.bats
@@ -1,0 +1,52 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+setup() {
+	export KUBECONFIG="$HOME/.kube/config"
+	get_pod_config_dir
+}
+
+@test "Run a job to completion" {
+	job_name="job-pi-test"
+	wait_time=60
+	sleep_time=2
+
+	# Create job
+	kubectl apply -f "${pod_config_dir}/job.yaml"
+
+	# Verify job
+	kubectl describe jobs/"$job_name" | grep "SuccessfulCreate"
+
+	# List pods that belong to the job
+	pod_name=$(kubectl get pods --selector=job-name=$job_name --output=jsonpath='{.items[*].metadata.name}')
+
+	# Verify that the job is completed
+	cmd="kubectl get pods -o jsonpath='{.items[*].status.phase}' | grep Succeeded"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
+
+	# Verify the output of the pod
+	pi_number="3.14"
+	kubectl logs "$pod_name" | grep "$pi_number"
+}
+
+teardown() {
+	kubectl delete pod "$pod_name"
+	# Verify that pod is not running
+	run kubectl get pods
+	echo "$output"
+	[[ "$output" =~ "No resources found" ]]
+
+
+	kubectl delete jobs/"$job_name"
+	# Verify that the job is not running
+	run kubectl get jobs
+	echo "$output"
+	[[ "$output" =~ "No resources found" ]]
+}
--- a/integration/kubernetes/runtimeclass_workloads/job.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/job.yaml
@@ -1,0 +1,19 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: job-pi-test
+spec:
+  template:
+    spec:
+      runtimeClassName: kata
+      containers:
+      - name: pi
+        image: busybox
+        command: ["/bin/sh", "-c", "echo 'scale=5; 4*a(1)' | bc -l"]
+      restartPolicy: Never
+  backoffLimit: 4
--- a/integration/kubernetes/run_kubernetes_tests.sh
+++ b/integration/kubernetes/run_kubernetes_tests.sh
@@ -45,6 +45,7 @@ bats k8s-pod-quota.bats
 bats k8s-sysctls.bats
 bats k8s-volume.bats
 bats k8s-projected-volume.bats
+bats k8s-job.bats
 bats k8s-memory.bats
 bats k8s-block-volume.bats
 bats k8s-shared-volume.bats


integration/kubernetes/runtimeclass_workloads/job.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: job-pi-test
+spec:
+  template:
+    spec:
+      runtimeClassName: kata
+      containers:
+      - name: pi
+        image: perl
+        command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(10)"]
```
jodh-intel: Wow. I'd forgotten `perl` was still a "thing" ;) Not a blocker but this might look a little less crazy using `python`:

```
python -c "import math; print math.pi"
```

integration/kubernetes/k8s-job.bats
```diff
@@ -0,0 +1,41 @@
...
+@test "Run a job to completion" {
+	job_name="job-pi-test"
+	wait_time=20
+	sleep_time=2
+
+	# Create job
+	kubectl apply -f "${pod_config_dir}/job.yaml"
+
+	# Verify job
+	kubectl describe jobs/"$job_name" | grep SuccessfulCreate
```
grahamwhaley: Maybe we can do this more succinctly with a `kubectl wait` or a `kubectl get -o json | jq` ? The `describe | grep` just feels less 'precise'. Having said that, as we are describing exactly only the job we are looking for, maybe its fine ;-)

integration/kubernetes/k8s-job.bats
```diff
@@ -0,0 +1,41 @@
...
+
+	# List pods that belong to the job
+	pod_name=$(kubectl get pods --selector=job-name="$job_name" --output=jsonpath='{.items[*].metadata.name}')
+
+	# Check job was completed
+	cmd="kubectl get pods -o wide | grep Completed"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
```
grahamwhaley: Similarly, I'd probably prefer a `kubectl wait` here, which can also take a timeout param. wdyt?

grahamwhaley:
We had some CI fails, and I think they might be with this PR. From 18.04:

```
23:40:04 not ok 1 Run a job to completion
23:40:04 # (in test file k8s-job.bats, line 161)
23:40:04 #   `' failed
23:40:04 # INFO: k8s configured to use runtimeclass
23:40:04 # job.batch/job-pi-test created
23:40:04 #   Normal  SuccessfulCreate  0s    job-controller  Created pod: job-pi-test-znz8l
23:40:04 # job.batch "job-pi-test" deleted
23:40:04 Makefile:107: recipe for target 'kubernetes' failed
23:40:04 make: *** [kubernetes] Error 1
23:40:04 Failed at 40: sudo -E PATH="$PATH" bash -c "make test"
23:40:05 Build step 'Execute shell' marked build as failure
23:40:24 Performing Post build task...
23:40:24 Match found for :.* : True
23:40:24 Logical operation result is TRUE
23:40:24 Running script  : #!/bin/bash
```

integration/kubernetes/k8s-job.bats
amshinde: @GabyCT What about the above suggestion from @grahamwhaley.
Why not use wait instead:
```
 kubectl wait --for=condition=complete --timeout=32s job/worker
```

integration/kubernetes/k8s-job.bats
```diff
@@ -0,0 +1,43 @@
...
+
+	# Verify the output of the pod
+	pi_number="3.14"
+	kubectl logs "$pod_name" | grep "$pi_number"
+}
+
+teardown() {
+	kubectl delete pod "$pod_name"
+	cmd="kubectl delete jobs/"$job_name" | grep deleted"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
```
chavafg: this is going to call the `delete` action every 2 seconds... do we want that or we only want to call `delete` once and then check that the job no longer exists?

integration/kubernetes/k8s-job.bats
```diff
@@ -0,0 +1,43 @@
...
+	sleep_time=2
+
+	# Create job
+	kubectl apply -f "${pod_config_dir}/job.yaml"
+
+	# Verify job
+	kubectl describe jobs/"$job_name" | grep "SuccessfulCreate"
+
+	# List pods that belong to the job
+	pod_name=$(kubectl get pods --output=jsonpath='{.items[*].metadata.name}')
```
chavafg: missing `--selector=job-name=$job_name`?

GabyCT:
@chavafg changes applied

GabyCT:
@chavafg changes applied

integration/kubernetes/k8s-job.bats
```diff
@@ -0,0 +1,46 @@
...
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
+
+	# Verify the output of the pod
+	pi_number="3.14"
+	kubectl logs "$pod_name" | grep "$pi_number"
+}
+
+teardown() {
+	kubectl delete pod "$pod_name"
+	kubectl delete jobs/"$job_name"
```
chavafg: preferably add `--wait` to later check that the job is no longer found

integration/kubernetes/k8s-job.bats
```diff
@@ -0,0 +1,46 @@
...
+
+	# Verify that the job is not running
+	output=$(kubectl describe jobs/"$job_name" | grep "not found")
+	[ -z "$output" ]
```
chavafg: better to use `run` from `bats` as it will save `stdout` and `stderr`
what about something like:
```
run kubectl describe "jobs/${jobs_name}"
echo "$output"
[[ "$output" =~ "not found" ]]

```

GabyCT: @chavafg , I tried to use that however the test fails because `run kubectl describe "jobs/${jobs_name}" returns an exit code of `1` that is the reason why I am not using that

GabyCT:
@chavafg changes applied

chavafg:
looks good, although there was a failure in the vsock job, seems that one of the pods was not deleted from `/var/lib/vc/sbs`:

```
16:26:36 be8c1457265cf40270654a392ef6ae209a9d86342ac6e8f01cb911cb7e9705b7
16:26:36 ERROR: 1 pods left and found at /var/lib/vc/sbs
16:26:36 Makefile:115: recipe for target 'kubernetes' failed
16:26:36 make: *** [kubernetes] Error 1
16:26:36 Failed at 40: sudo -E PATH="$PATH" bash -c "make test"
```

GabyCT:
@chavafg , I will take a look

GabyCT:
@chavafg changes applied, I put also a verification for the pod

Issue #1638, opened 2019-05-28
git: check if git is capable of shadow cloning


Pennyzct:
Before we use shadow clone to save time and bandwidth, we need to
check git version firstly. 
since option` --shallow-submodules` was introduced in git v2.9.0

PR #1639, opened 2019-05-28
git: check if git is capable of shallow cloning

Pennyzct:
This PR is following #1555, we hit a [issue](https://github.com/kata-containers/tests/pull/1555#issuecomment-492595005) there, git v2.7.4 is the latest on Ubuntu 16.04, and option `--shallow-submodules` was introduced in git v2.9.0, which means that for now, shallow clone will not be working well on Ubuntu 16.04.
So we need to check git version firstly before using the shallow clone.

Fixes: #1638

### DIFF
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -19,6 +19,9 @@ QEMU_REPO=${QEMU_REPO_URL/https:\/\//}
 PACKAGED_QEMU="qemu-lite"
 QEMU_ARCH=$(${cidir}/kata-arch.sh -d)
 
+# option "--shallow-submodules" was introduced in git v2.9.0
+GIT_SHADOW_VERSION="2.9.0"
+
 get_packaged_qemu_commit() {
 	if [ "$ID" == "ubuntu" ] || [ "$ID" == "debian" ]; then
 		qemu_commit=$(sudo apt-cache madison $PACKAGED_QEMU \
@@ -61,7 +64,14 @@ install_packaged_qemu() {
 }
 
 clone_qemu_repo() {
-	git clone --branch "${CURRENT_QEMU_BRANCH}" --single-branch --depth 1 --shallow-submodules "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
+	# check if git is capable of shadow cloning
+        git_shadow_clone=$(check_git_version "${GIT_SHADOW_VERSION}")
+
+	if [ "$git_shadow_clone" == "true" ]; then
+		git clone --branch "${CURRENT_QEMU_BRANCH}" --single-branch --depth 1 --shallow-submodules "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
+	else
+		git clone --branch "${CURRENT_QEMU_BRANCH}" --single-branch --depth 1 "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
+	fi
 }
 
 build_and_install_qemu() {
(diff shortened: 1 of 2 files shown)


Pennyzct:
@grahamwhaley ptal. ;) I know it's quite a time after our last discussion.

.ci/install_qemu.sh
```diff
@@ -61,7 +64,14 @@ install_packaged_qemu() {
...
 
 clone_qemu_repo() {
-	git clone --branch "${CURRENT_QEMU_BRANCH}" --single-branch --depth 1 --shallow-submodules "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
+	# check if git is capable of shadow cloning
+        git_shadow_clone=$(check_git_version "${GIT_SHADOW_VERSION}")
+
+	if [ "$git_shadow_clone" == "true" ]; then
+		git clone --branch "${CURRENT_QEMU_BRANCH}" --single-branch --depth 1 --shallow-submodules "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
+	else
+		git clone --branch "${CURRENT_QEMU_BRANCH}" --single-branch "${QEMU_REPO_URL}" "${GOPATH}/src/${QEMU_REPO}"
```
grahamwhaley: I think you can still have `--depth 1` here ? It is just `--shallow-submodules` iirc that was introduced at 2.9.0.

Pennyzct: yep, you're right. ;)
I'm doing a little investigation here, seeing from there [official documentation](https://git-scm.com/docs/git-clone/), the oldest version which could be traced back is 2.3.0, the `--depth` option was there. 
So we may have `--depth 1` there.

Pennyzct:
@jodh-intel  AAAhhhhhhhhh, silly me, my poor english, i will refine commit asap. btw, we really need to find a way not triggering the ci, if only docs, like PR commit, got edited. ;)

grahamwhaley:
wrt short circuit CI for doc only changes - as we tightened security up on github, I suspect we cannot maybe do that there, as that might involve reducing our process strength (which we increased to avoid mistakes happening due to super powers...). Thus, we might have to do this at the Jenkins side. That's OK, I think we could add an early pre-check into the jenkins CI script that extracts the names of all the chanced files, and if they only end in `.md`, do a quick return from the CI script... that might do it. OK, so we have to trigger the CI still, but it should be able to do the detect and quit very quick, before it's even tried to build and install all the components it needs...

jodh-intel:
We already have checks like that in the static check script - we just need to detect no code changes rather than no doc changes. The question is how that script can cause the rest of the CI run to exit successfully. I'll defer to the Jenkins experts but I suspect we might need another plugin to do that?

grahamwhaley:
Should not need another plugin - I reckon we can do it in our [top level jenkins entry point script](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh), and drop out of that early if we detect no code/only doc changes.

grahamwhaley:
yes, we'd still trigger every CI job, and every CI job would do the test. Hmm, unless @fuentess
would like to start chaining jobs in jenkins, and then we could have a pre-job do the check, and all the other jobs as post-deps. But, I think the pre-job would then have to fail in order to not run the post jobs (I'm not sure how complex the chaining rules for triggering down the chain are).
We already have one pre/post job pair set up on the jenkins master - the metrics job has a post-job (iirc) that reboots the node after every build (sort of..)

chavafg:
For changes in the commit messages between pushes may be a little more complicated. Would need to check how to do that using github api. 

For `.md` files this should be easy, btw, currently if only docs are modified, we just run the ubuntu `8.04 which is Required to enable the merging button.

Issue #1640, opened 2019-05-28
check-markdown: Fix heading link generation


jodh-intel:
The https://github.com/russross/blackfriday package used by the markdown checker generates "correct" heading IDs (to allow linking to a heading). However markdown != [GFM](https://github.github.com/gfm/) so we need to generate the correct values for GitHub.

PR #1641, opened 2019-05-28
check-markdown: Fix heading links

jodh-intel:
Don't rely on the blackfriday package for generating heading IDs (used by the `toc` command) as although they are correct for generic markdown, they are not always correct when dealing with GFM [1].

Instead, use our own function which understands GitHub better.

Also added new tests.

[1] - https://github.github.com/gfm

Fixes: #1640.

### DIFF
--- a/cmd/check-markdown/heading.go
+++ b/cmd/check-markdown/heading.go
@@ -6,12 +6,31 @@
 
 package main
 
+import "fmt"
+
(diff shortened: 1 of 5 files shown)


jodh-intel:
Sigh. Looks like this PR is also "stuck" as no builds have triggered since yesterday. Let's try again...

/retest

Issue #1642, opened 2019-05-28
Improve CI URL check


jodh-intel:
The URL check logic is not reliable so improve it by considering HTTP status codes.

PR #1643, opened 2019-05-28
CI: Improve URL checker

jodh-intel:
Improve the URL checker by considering the HTTP status codes.

Fixes: #1642.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -25,6 +25,9 @@ specific_branch="false"
 force="false"
 branch=${branch:-master}
 
+# number of seconds to wait for curl to check a URL
+typeset url_check_timeout_secs="${url_check_timeout_secs:-60}"
+
 typeset -A long_options
 
 long_options=(
@@ -505,6 +508,8 @@ check_docs()
 
 	info "Checking all document URLs"
 
+	local invalid_urls_dir=$(mktemp -d)
+
 	for url in $urls
 	do
 		if [ "$specific_branch" != "true" ]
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -537,8 +537,25 @@ check_docs()
...
+			for status in $http_statuses
+			do
+				local ret
+
+				# See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
+				{ echo "$status" | grep -qE "^(2|3)[0-9][0-9]"; ret=$?; } || true
+
+				[ $ret -ne 0 ] && { echo "$url" >> "$invalid_urls" && break; } || true
+			done
+		) &
```
grahamwhaley: hurrrr, is that multiple parallel writers in the background to a single `$invalid_urls` file - heh, one day I expect you to get inter-spersed writes.... but, given that only happens in an error condition, and is probably rare, it's probably not a blocker - but, does look odd ;-)

jodh-intel: I haven't checked the source, but `strace` suggests this is safe as each shell file append calls an (open, write, close) triplet.

jodh-intel:
Looks like the CI's have got confused. Let's try again...

/retest

jodh-intel:
@marcov - could you tal at this?

marcov:
@jodh-intel sure!

.ci/static-checks.sh
```diff
@@ -537,8 +537,25 @@ check_docs()
...
 
 		# Check the URL, saving it if invalid
-		( curl -sLf -o /dev/null "$url" ||\
-				echo "$url" >> "$invalid_urls") &
+		(
+			local http_statuses
+
+			# A single doc can generate multiple HTTP status
+			# codes, so save them all.
+			http_statuses=$(curl -sIL "$url" | grep -E "^HTTP" | awk '{print $2}' || true)
```
marcov: You dont need the trailing `|| true` here as that's being executed in a subshell.

.ci/static-checks.sh
```diff
@@ -537,8 +537,25 @@ check_docs()
...
+			local status
+
+			for status in $http_statuses
+			do
+				local ret
+
+				# See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
+				{ echo "$status" | grep -qE "^(2|3)[0-9][0-9]"; ret=$?; } || true
+
+				[ $ret -ne 0 ] && { echo "$url" >> "$invalid_urls" && break; } || true
```
marcov: You can somehow make this more readable with:
```bash
if ! echo "$status" | grep -qE "^(2|3)[0-9][0-9]"; then
	echo "$url" >> "$invalid_urls"
	break
fi
```

.ci/static-checks.sh
marcov: > I haven't checked the source, but strace suggests this is safe as each shell file append calls an (open, write, close) triplet.

It would be safe only if (open, write, close) are done atomically.

.ci/static-checks.sh
jodh-intel: Agreed - done.

jodh-intel:
Yes, the write would need to be done all in one go. fwics that happens. But maybe if someone comes up with a URL that is `> block size` we might have problems ;)

.ci/static-checks.sh
marcov: You are right, still I always try to avoid the `|| true` because they look so odd :laughing:

grahamwhaley:
blocksize of stdlib stdio buffer size - but, as I intimated originally, yes this is notionally wrong and broken, but I don't think it will affect us - even if we overlap, we will still fail the test, which is ultimately the point ;-)

.ci/static-checks.sh
marcov: And, in this case, if the pipeline fails, why are you ignoring the failure?

jodh-intel: @marcov - I've broken up that line into multiple sections for clarity and added comments.

jodh-intel:
Invalid URL handling updated to ensure reliable writes @grahamwhaley.

.ci/static-checks.sh
```diff
@@ -536,14 +541,77 @@ check_docs()
...
+			# Process specific file to avoid out-of-order writes
+			local invalid_file=$(printf "%s/%d" "$invalid_urls_dir" "$$")
+
+			{ curl -sIL --max-time "$url_check_timeout_secs" "$url" &>"$curl_out"; ret=$?; } || true
+
+			# A transitory error, or the URL is incorrect,
+			# but capture either way.
+			if [ "$ret" -ne 0 ]; then
+				echo "$url" >> "${invalid_file}"
+				continue
```
marcov: @jodh-intel are you sure `continue` is handled properly inside `( )`?

.ci/static-checks.sh
```diff
@@ -536,14 +541,77 @@ check_docs()
 			git ls-remote "$url" > /dev/null 2>&1 && continue
```
marcov: `git ls-remote` is accessing the network, so it could also benefit from using parallel sub-processes like for `curl`

.ci/static-checks.sh
jodh-intel: I've replaced those with `exit` calls now.

.ci/static-checks.sh
jodh-intel: Good catch. I think that's a further optimisation that can be done on a follow-on PR though as it will complicate the code yet further.

PR #1651, opened 2019-05-29
ci: Add trap to remove tmp files

### LINKED ISSUE #1613: trap cleanup of tmp files in static check script
See: https://github.com/kata-containers/tests/pull/1612#discussion_r286393094.

Ideally, the trap handler would:

- Delete the logs on success
- Display a list of logs on failure (but not delete them).

GabyCT:
This adds trap cleanup for the temporary files
used in static checks.

Fixes #1613

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -18,6 +18,9 @@ source "${cidir}/lib.sh"
 export tests_repo="${tests_repo:-github.com/kata-containers/tests}"
 export tests_repo_dir="${GOPATH}/src/${tests_repo}"
 
+# List of files to delete on exit
+files_to_remove=()
+
 script_name=${0##*/}
 
 repo=""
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -110,6 +110,13 @@ Examples:
 EOT
 }
 
+function remove_tmp_files() {
+	rm -f "$url_map" "$invalid_urls" "$md_links" "$submodule_packages" "$all_packages" "${tmp}"
```
devimc: I think this won't work, since all these variables are local variables

jodh-intel: Yes. I think you'd need to do something like:

```sh
# List of files to delete on exit
files_to_remove=()

cleanup()
{
    rm -f ${files_to_remove[@]}
}

foo()
{
    files_to_remove+=(foo.log)

    # ...
}

bar()
{
    files_to_remove+=(bar.log)

    # ...
}

main ()
{
    trap cleanup EXIT

    # ...
}

main "$@"
```

PR #1664, opened 2019-06-03
tests: Add K8s test for health checks

### LINKED ISSUE #1663: add k8s test for health checks
Add k8s test for a health probe

GabyCT:
This adds a kubernetes test to verify a container in a pod is healthy by
using the health checking mechanisms.

Fixes #1663

### DIFF
--- a/integration/kubernetes/k8s-health-checks.sh
+++ b/integration/kubernetes/k8s-health-checks.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+setup() {
+	export KUBECONFIG="$HOME/.kube/config"
+	get_pod_config_dir
+}
+
+@test "Health checks" {
+	pod_name="hctest"
+	bad_pod_name="badhctest"
+	wait_time=120
+	sleep_time=2
+
+	# Create pod
+	kubectl create -f "${pod_config_dir}/pod-health.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready --timeout="$wait_time"s pod "$pod_name"
+
+	# Check pod status
+	cmd="kubectl describe pod $pod_name | grep 'Started container'"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
+
+	# Create a bad pod
+	kubectl create -f "${pod_config_dir}/pod-bad-health.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready --timeout="$wait_time"s pod "$pod_name"
+
+	# Check bad pod status
+	cmd="kubectl describe pod $bad_pod_name | grep Unhealthy"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
+}
+
+teardown() {
+	kubectl delete pod "$pod_name" "$bad_pod_name"
+}
--- a/integration/kubernetes/runtimeclass_workloads/pod-bad-health.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-bad-health.yaml
@@ -1,0 +1,29 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  name: badhctest
+spec:
+  runtimeClassName: kata
+  containers:
+  - name: test
+    image: mhausenblas/simpleservice:0.5.0
+    ports:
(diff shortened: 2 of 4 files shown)


integration/kubernetes/runtimeclass_workloads/pod-bad-health.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: Pod
+metadata:
+  name: badhctest
+spec:
+  runtimeClassName: kata
+  containers:
+  - name: test
+    image: mhausenblas/simpleservice:0.5.0
+    env:
+    - name: HEALTH_MIN
```
grahamwhaley: I'm presuming these HEALTH values are chosen/set to create a desired effect (bad health timeout/fail?). can we have some comments here explaining that, otherwise the numbers are somewhat opaque (and we'll forget over time what we set and why etc.).

integration/kubernetes/k8s-health-checks.sh
```diff
@@ -0,0 +1,42 @@
...
+	kubectl wait --for=condition=Ready pod "$pod_name"
+
+	# Create a bad pod
+	kubectl create -f "${pod_config_dir}/pod-bad-health.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready pod "$pod_name"
+
+	# Check bad pod status
+	cmd="kubectl describe pod $bad_pod_name | grep Unhealthy"
```
grahamwhaley: I'm slightly confused, as it looks like we create a healthy pod, and then an unhealthy pod, but only check the status for `Unhealthy`. Should we also be checking the healthy one is actually healthy as well?

amshinde: @GabyCT Maybe I am missing something, but I still dont see how these environment variables are used?
Are they used by the workload itself?

integration/kubernetes/k8s-health-checks.sh
```diff
@@ -0,0 +1,47 @@
...
+	pod_name="hctest"
+	bad_pod_name="badhctest"
+	wait_time=60
+	sleep_time=2
+
+	# Create pod
+	kubectl create -f "${pod_config_dir}/pod-health.yaml"
+
+	# Check pod creation
+	cmd="kubectl wait --for=condition=Ready pod $pod_name"
```
chavafg: you can use `--timeout=$wait_time` here and remove the line below.

integration/kubernetes/k8s-health-checks.sh
```diff
@@ -0,0 +1,47 @@
...
+
+	# Check pod status
+	cmd="kubectl describe pod $pod_name | grep 'Started container'"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
+
+	# Create a bad pod
+	kubectl create -f "${pod_config_dir}/pod-bad-health.yaml"
+
+	# Check pod creation
+	cmd="kubectl wait --for=condition=Ready pod $pod_name"
```
chavafg: same as the above comment.

integration/kubernetes/runtimeclass_workloads/pod-bad-health.yaml
```diff
@@ -0,0 +1,29 @@
...
+      value: "1000"
+    - name: HEALTH_MAX
+      value: "4000"
+    livenessProbe:
+      initialDelaySeconds: 2
+      periodSeconds: 5
+      httpGet:
+        path: /health
+        port: 9876
```
chavafg: I see both pods are using same port... is that desired?

integration/kubernetes/k8s-health-checks.sh
chavafg: Agree... maybe a little explanation of how the service works to check for the health of the pod?

integration/kubernetes/runtimeclass_workloads/pod-bad-health.yaml
GabyCT: @chavafg yes that is expected

GabyCT:
@chavafg changes applied

integration/kubernetes/runtimeclass_workloads/pod-bad-health.yaml
```diff
@@ -0,0 +1,29 @@
...
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  name: badhctest
+spec:
+  runtimeClassName: kata
+  containers:
+  - name: test
+    image: mhausenblas/simpleservice:0.5.0
```
jodh-intel: I wonder if this hard-coded image version is guaranteed to be retained or whether we need to drop the version number?

GabyCT: @jodh-intel , unfortunately we can not drop the version number and put something like `mhausenblas/simpleservice` or `mhausenblas/simpleservice:latest` because we got the error of
`ErrImagePull` meaning that does not exist

egernst:
@GabyCT how'd you come to use simpleservice?

I wonder if we should start a few containers, each using one of the three options for healthcheck (http, tcp, port?)?

Issue #1666, opened 2019-06-04
Store original github labels


jodh-intel:
I've just updated all repos with the [new set of labels](https://github.com/kata-containers/tests/blob/master/cmd/github-labels/labels.yaml.in).

As such, we should store a copy of the *original* labels "just in case" in:

- https://github.com/kata-containers/tests/tree/master/cmd/github-labels

PR #1667, opened 2019-06-04
labels: Add archive of original github labels

jodh-intel:
Today, the GitHub labels for each repository were changed to reflect [1].

Add a set of YAML file containing the original GitHub labels for each repository in case we need to revert.

---

[1] - https://github.com/kata-containers/tests/blob/master/cmd/github-labels/labels.yaml.in

Fixes: #1666.


jodh-intel:
@grahamwhaley - could you review? I think once we've got a quorum we should just merge this to save CI resources ;)

grahamwhaley:
So many CI fails, but peeking at one (18.04 initrd), that looked like a network timeout trying to grab an image to test - we are pretty sure these are all unrelated to this PR, so merging...

Issue #1669, opened 2019-06-05
Add back missing wip and do-not-merge labels


jodh-intel:
A bit of fallout from appling https://github.com/kata-containers/tests/blob/master/cmd/github-labels/labels.yaml.in - @grahamwhaley noticed we'd lost the `wip` and `do-not-merge` labels.

I've now re-added those to all repos so need to update the master database with these two labels.

PR #1670, opened 2019-06-05
Update GitHub labels

jodh-intel:
Add the missing labels `wip` and `do-not-merge` which got dropped from the new set of GitHub labels.

Also removed the old `from:` entries which no longer apply since we've now moved to the new set of labels.

Fixes: #1669.

### DIFF
--- a/cmd/github-labels/labels.yaml.in
+++ b/cmd/github-labels/labels.yaml.in
@@ -55,6 +55,10 @@ categories:
       How the issue affect the operation of the system. A more precise version
       of regression.
 
+  - name: block
+    description: |
+      Stop a PR from being merged.
+
   - name: cleanup
     description: Refactoring, restructuring or general tidy-up needed.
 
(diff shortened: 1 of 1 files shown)


grahamwhaley:
I guess we should check those match what zuul is looking for...
in the [zuul config files](https://opendev.org/kata-containers/zuul-config/src/branch/master/zuul.d/jobs.yaml#L29) we can see:

```yaml
- job:
    name: kata-github-wip
    description: |
      This job calls the github-wip playbook with a Kata-curated set of
      WIP labels and WIP title keywords
    run: playbooks/github-wip/run.yaml
    nodeset:
      nodes: []
    vars:
      wip_labels: 'do-not-merge wip rfc'
      wip_keywords: 'WIP RFC'
```

Yay, I think all three of `do-not-merge`, `wip` and `rfc` align :-)
I see you also have `block` @jodh-intel - does that need to be PR'd as an addition into the zuul WIP check config??

grahamwhaley:
ah, I see, `block` is a category, not a label - no need to try and add that to Zuul then.

jodh-intel:
Yep.

PR #1675, opened 2019-06-06
factory: use vsock to test vm template

bergwolf:
So we don't rely on yamux which seems flaky and sometimes
unable to clean itself up upon closing.


bergwolf:
@grahamwhaley When vosck is not supported, kata-runtime uses yamux over serial port to connect to the agent. So no error promoted to user in such case.

bergwolf:
Sorry I should have made it clear that it depends-on https://github.com/kata-containers/runtime/pull/1774

Issue #1673, opened 2019-06-06
Add way to run arch specific static checks


jodh-intel:
It's not immediately obvious, but although the tests run by the [`static-checks.sh`](https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh) tool are distro-agnostic some of them are actually architecture-specific.

Add options to allow the following sets of tests to be run:

- All arch-agnostic tests.
- All arch-specific tests.

PR #1679, opened 2019-06-07
CI: Add arch options to static check script

jodh-intel:
The static check script runs a number of checks, but some (one) are architecture-specific. Add in three new options to allow the behaviour of the script to be modified:

- `--list`: List all available check functions.
- `--no-arch`: List or run only architecture-agnostic check functions.
- `--only-arch`: List or run only architecture-specific check functions.

Note that now the script searches itself for the list of functions to run rather than hard-coding the list of functions to run. This necessitated changing the name of each check function to `static_check_*()` to ensure only functions in this file are considered.

Fixes: #1673.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright (c) 2017-2018 Intel Corporation
+# Copyright (c) 2017-2019 Intel Corporation
 #
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -23,10 +23,23 @@ files_to_remove=()
...
+
+# All architecture-specific static check functions must match this pattern.
+typeset -r arch_func_regex="_arch_specific$"
+
 repo=""
 specific_branch="false"
 force="false"
 branch=${branch:-master}
+disable_arch_specific="false"
+only_arch_specific="false"
```
grahamwhaley: These two `arch_specific` vars sound like the inverse of each other - can they be melded into one and some of the tests that use them inverted to match?

.ci/static-checks.sh
```diff
@@ -23,10 +23,23 @@ files_to_remove=()
```
grahamwhaley: update copyright year? :-)

.ci/static-checks.sh
jodh-intel: Fixed.

.ci/static-checks.sh
jodh-intel: Fixed.

Issue #1683, opened 2019-06-07
ci: static tests: golangci: do not run the golangci test suite


grahamwhaley:
**Which feature do you think can be improved?**

ci static checks

**How can it be improved?**

currently when golangci is built, by default it also runs all its own tests, which take minutes.
I don't believe we need to run those, so let's just build and install it.

From a job build log we can see:

```
00:01:46.013 GL_TEST_RUN=1 go test -v ./...
...
00:03:21.235 ?          github.com/golangci/golangci-lint/test/testshared       [no test files]
```

That looks like 1m30s or something just running tests (the default makefile target for golangci is 'test', which has a dep of 'build'). On my local 2core machine, I got bored waiting at >9m0s. We don't need to test golangci itself I think, so let's go straight to the build step.

PR #1684, opened 2019-06-07
ci: static tests: golangci: do not run golangci self tests

grahamwhaley:
A default 'make' of golangci will run all its own testsuite after
it has done a build. That takes minutes, and we probably don't
need to test that the tester itself is OK. Skip the tests and
go directly to a 'build' build.

Fixes: #1683

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -236,7 +236,10 @@ check_go()
 		local linter_version=$(get_test_version "externals.golangci-lint.version")
 
 		info "Forcing ${linter} version ${linter_version}"
-		build_version ${linter_url} "" ${linter_version}
+		build_version ${linter_url} "build" ${linter_version}
+		# golangci build only target does not install into ${GOPATH}/bin
+		# Hand install it...
+		go install -v "${linter_url}/..."
 	fi
 
 	local linter_args="run -c ${cidir}/.golangci.yml"


devimc:
Ci is failing 

```
.ci/static-checks.sh: line 270: golangci-lint: command not found
The command "bash .ci/static-checks.sh github.com/kata-containers/tests" exited with 127.
```

grahamwhaley:
That's weird, I wonder if part of the golangci Makefile test target has an implicit 'go install' in it - there seems not to be an explicit one nor a Makefile rule - I wonder how this worked before ;-) I'll have to dig a bit...

grahamwhaley:
yep, seems the `go test -v ./...` step of the test part installs the binary, and we are hanging off that side effect. Let me see if I can find a nice-ish way to get the binary installed without hacking our lib scripts too much....

chavafg:
I see that it is being built, but maybe not being installed in `$GOPATH/bin`?

grahamwhaley:
@chavafg exactly - the install into `$GOPATH/bin` was done implictly by the `go test -v` before :-(. Now we are not running the tests, the `Makefile` does not install it into the path, and does not have an `install` target rule. I'll see what I can do to work around it.

jodh-intel:
Yep. Might require you to:

```sh
$ pushd $GOPATH/src/github.com/golangci/golangci-lint &>/dev/null
$ go install -v ./...
$ popd &>/dev/null
```

grahamwhaley:
OK, Travis is happy, but I'm not sure it's the most elegant workaround - it is a little tied to the linter being ok with a `go install -v ./...`. @chavafg @ganeshmaharaj - happy with the workaround?

For reference, a quick look at the travis times went from:

11m30s -> 9m10s
5m10s -> 3m50s

or something in that range afaict.

PR #1685, opened 2019-06-07
test: Optimized k8s shared volume test

### LINKED ISSUE #1680: CI: k8s shared volume test might be optimisable
Staring at the CI logs, looking for potential time improvements, I noticed that the [k8s shared volume test](https://github.com/kata-containers/tests/blob/master/integration/kubernetes/k8s-shared-volume.bats) I think is taking some time (~2m33s or so?)

From timestamped logs of a nemu agent CI run](http://jenkins.katacontainers.io/job/kata-containers-agent-ubuntu-nemu/98/consoleFull):

```
01:50:38.197 ok 1 Containers with shared volume
01:51:01.584 ok 2 initContainer with shared volume
01:51:01.622 1..1
01:53:34.115 ok 1 Expose IP Address
```

I think that time gap is coming from the test, and I suspect it is as the test is based on these two containers:

```
first_container_name="nginx-container"
second_container_name="debian-container"
```

both of those are quite large and will be pulled seperately. I wonder if we can re-write the test to maybe use two busybox or alpine container/pods and just cat/echo across the volume between them? This might save us a lot of time and space.

@GabyCT @amshinde - can you think of any reason we should not switch to a simpler test? I know it will not be quite as 'real world' as linking nginx with a full distro, but, for our testing, will it matter?

I'd be interested at least if you can test/time a run (on a machine where all the images are not present), and see if that test is consuming minutes.

thx!

GabyCT:
Modify the test to use busybox images instead of nginx and debian in order
to reduce the time of performing the testing.

Fixes #1680

### DIFF
--- a/integration/kubernetes/k8s-shared-volume.bats
+++ b/integration/kubernetes/k8s-shared-volume.bats
@@ -15,8 +15,8 @@ setup() {
 
 @test "Containers with shared volume" {
 	pod_name="test-shared-volume"
-	first_container_name="nginx-container"
-	second_container_name="debian-container"
+	first_container_name="busybox-first-container"
+	second_container_name="busybox-second-container"
 
 	# Create pod
 	kubectl create -f "${pod_config_dir}/pod-shared-volume.yaml"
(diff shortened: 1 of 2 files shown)


grahamwhaley:
I've nudged a couple of CI rebuilds:
 - Fedora hit the unstable state test
 - travis MacOS looked to hit a timeout

PR #1689, opened 2019-06-10
test: Add test to measure time drift

### LINKED ISSUE #1488: Add test to measure time drift in the guest
Once we solve https://github.com/kata-containers/runtime/issues/1537, we need an integration test to make sure, chrony is running inside the VM.
We also need a test to make sure guest time is synced with the host.

GabyCT:
This will adds a test to make sure guest time is synced with the host.

Fixes #1488

### DIFF
--- a/integration/time_drift/time_drift.bats
+++ b/integration/time_drift/time_drift.bats
@@ -1,0 +1,52 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test will make sure that guest time is synced with the host
+
+set -e
+
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+setup() {
+	clean_env
+
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+}
+
+@test "Verify guest time is synced with the host" {
+	CONTAINER_NAME="test"
+	IMAGE="busybox"
+	PAYLOAD="tail -f /dev/null"
+
+	docker run -d --name "${CONTAINER_NAME}" --runtime "${RUNTIME}" "${IMAGE}" sh -c "${PAYLOAD}"
+
+	# Get host and guest time
+	GUEST_CMD="docker exec "${CONTAINER_NAME}" date +'%H:%M:%S'"
+	HOST_CMD="date +'%H:%M:%S'"
+	OUTPUT=$((echo "${GUEST_CMD}"; echo "${HOST_CMD}") | parallel)
+	echo $OUTPUT
+
+	GUEST_TIME=$(echo $OUTPUT | cut -d ' ' -f1)
+	HOST_TIME=$(echo $OUTPUT | cut -d ' ' -f2)
+
+	TIME_DIFF=$(( $(date -d "$GUEST_TIME" "+%s")-$(date -d "$HOST_TIME" "+%s") ))
+	echo "$TIME_DIFF"
+	[ "$TIME_DIFF" -le 1 ]
+}
+
+teardown() {
+	docker rm -f "${CONTAINER_NAME}"
+
+	clean_env
+
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+}
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ endif
 # union for 'make test'
 UNION := functional debug-console $(DOCKER_DEPENDENCY) openshift crio docker-compose network \
 	docker-stability oci netmon kubernetes swarm vm-factory \
-	entropy ramdisk shimv2 tracing
+	entropy ramdisk shimv2 tracing time-drift
 
 # filter scheme script for docker integration test suites
 FILTER_FILE = .ci/filter/filter_docker_test.sh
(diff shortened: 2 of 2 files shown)


integration/time_drift/time_drift.bats
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
chavafg: should this be `#!/usr/bin/env bats`

GabyCT:
@chavafg changes applied

integration/time_drift/time_drift.bats
```diff
@@ -0,0 +1,59 @@
...
+
+	# Check that processes are not running
+	run check_processes
+	echo "$output"
+	[ "$status" -eq 0 ]
+}
+
+@test "Verify guest time is synced with the host" {
+	CONTAINER_NAME="test"
+	IMAGE="ubuntu"
```
grahamwhaley: Can we use a smaller image that we may already have loaded for the tests here, such as `busybox` or `alpine`. I think the `date` from busybox for instance will be accurate and featureful enough for this test?
We are on a drive to reduce the time and resource consumption of our CI - part of that is trying to use, and re-use, the smallest container images we can for the tests. Otherwise they consume significantly more download and disk bandwidth, which also takes longer to grab.

integration/time_drift/time_drift.bats
```diff
@@ -0,0 +1,59 @@
...
+	PAYLOAD="tail -f /dev/null"
+
+	docker run -d --name "${CONTAINER_NAME}" --runtime "${RUNTIME}" "${IMAGE}" sh -c "${PAYLOAD}"
+
+	# Get host and guest time
+	(docker exec "${CONTAINER_NAME}" date > "${GUEST_TMP_FILE}"; date > "${HOST_TMP_FILE}") | parallel
```
grahamwhaley: I think maybe this is not using `parallel` as you expect? I think you'd either need to echo the commands at `parallel`, or put them on its command line, no?

integration/time_drift/time_drift.bats
```diff
@@ -0,0 +1,59 @@
...
+# Save guest time
+GUEST_TMP_FILE=$(mktemp guest.XXXXX)
+# Save host time
+HOST_TMP_FILE=$(mktemp host.XXXXX)
+
+function remove_tmp_files() {
+	rm "${GUEST_TMP_FILE}" "${HOST_TMP_FILE}"
+}
+
+trap remove_tmp_files EXIT
```
grahamwhaley: Thanks for adding the trap. It looks like those files are just used to capture the output of `date`, which you then later `cat` back into some shell vars - can we just skip the mktemp's completely, and just put the info direct into shell vars (although, maybe the use of `parallel` and subvars means they will only end up in a subshell, and then get thrown away - let us know...).
I just try to avoid `mktemp` wherever we can as its slower, and more error prone to leaving files laying around in unexpected failure cases.

integration/time_drift/time_drift.bats
```diff
@@ -0,0 +1,59 @@
...
+
+	docker run -d --name "${CONTAINER_NAME}" --runtime "${RUNTIME}" "${IMAGE}" sh -c "${PAYLOAD}"
+
+	# Get host and guest time
+	(docker exec "${CONTAINER_NAME}" date > "${GUEST_TMP_FILE}"; date > "${HOST_TMP_FILE}") | parallel
+
+	guest_time=$(cat "${GUEST_TMP_FILE}")
+	host_time=$(cat "${HOST_TMP_FILE}")
+
+	[ "$host_time" == "$guest_time" ]
```
grahamwhaley: So, I think this will be 'accurate to the second', yes?
I'm not going to be surprised if maybe 1 in 100 runs of this test happen to hit the rollover boundary of the second between the host and guest, and fail... Two things then:
- can we add an echo/info output here in the case of failure, and show the two times - then we get some clues to the failure? (not sure if `bats` does that already etc.??)
- be prepared to maybe rewrite the test in the future to grab the time in nanoseconds or similar and set an 'acceptable difference' value, to do a bit more of a fuzzy check...

GabyCT:
@grahamwhaley changes applied

integration/time_drift/time_drift.bats
```diff
@@ -0,0 +1,46 @@
...
+
+@test "Verify guest time is synced with the host" {
+	CONTAINER_NAME="test"
+	IMAGE="busybox"
+	PAYLOAD="tail -f /dev/null"
+
+	docker run -d --name "${CONTAINER_NAME}" --runtime "${RUNTIME}" "${IMAGE}" sh -c "${PAYLOAD}"
+
+	# Get host and guest time
+	GUEST_TIME=$(docker exec "${CONTAINER_NAME}" date)
```
grahamwhaley: I'm still not sure this is doing what you think maybe? Have you tried this under a `bash -x` to see what it looks like? I could be wrong, but it just doesn't look right to my eyes...

I suspect it is currently:
 - executing the docker exec, and placing that date into GUEST_TIME
 - executing `date` and placing that into HOST_TIME
 - attempting to execute the contents of GUEST_TIME and HOST_TIME, and piping the output of that execution to `parallel`
 - and finally doing the date compare.

What I suspect you mean it to do is:
 - set GUEST_TIME to a string of commands to execute
 - set HOST_TIME to a string of commands to execute
 - execute them at the same time using `parallel` (I'm not sure you can use `|` in the style you are trying, unless you `echo` stuff ...?)
 - and then do the compare of the results (for which you might need to save them in two new variables??)

Probably best to show us what a `bash -x <fragment>` looks like so we can see the order of events and the values/variables set and used.

GabyCT:
@grahamwhaley changes applied

egernst:
@GabyCT - the Fedora job fails your new time-drift test

PR #1694, opened 2019-06-11
tests: Update nginx image for kubernetes test

### LINKED ISSUE #1693: integration/k8s: unify on a single version of nginx

Post running the integration/k8s tests, examining the image pools shows that we currently pull both `nginx:latest` and `nginx:1.14`. It looks like we don't have to pin (and generally we don't like pinning).

**How can it be improved?**

Don't pin it. Using the same version will save us a bunch of download time and space (~113Mb)

GabyCT:
This will update to unify all kubernetes tests to use the same nginx
image instead of different versions. This also updates to use
nginx:alpine-1.17.0 image which has a size of 23.6MB and will reduce the
current image download size.

Fixes #1693

### DIFF
--- a/integration/kubernetes/nginx.bats
+++ b/integration/kubernetes/nginx.bats
@@ -25,8 +25,13 @@ setup() {
 @test "Verify nginx connectivity between pods" {
 	wait_time=30
 	sleep_time=3
-	kubectl create -f "${pod_config_dir}/${deployment}.yaml"
-	kubectl wait --for=condition=Available deployment/${deployment}
+
+	# Create test .yaml
+	sed -e "s/\${nginx_version}/${nginx_image}/" \
+		"${pod_config_dir}/${deployment}.yaml" > "${pod_config_dir}/test-${deployment}.yaml"
+
+	kubectl create -f "${pod_config_dir}/test-${deployment}.yaml"
+	kubectl wait --for=condition=Available --timeout=60s deployment/${deployment}
 	kubectl expose deployment/${deployment}
 
 	busybox_pod="test-nginx"
@@ -39,6 +44,7 @@ setup() {
 }
 
 teardown() {
+	rm -f "${pod_config_dir}/test-${deployment}.yaml"
 	kubectl delete deployment "$deployment"
 	kubectl delete service "$deployment"
 	kubectl delete pod "$busybox_pod"
--- a/integration/kubernetes/k8s-attach-handlers.bats
+++ b/integration/kubernetes/k8s-attach-handlers.bats
@@ -9,6 +9,10 @@ load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
 load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
 
 setup() {
+	versions_file="${BATS_TEST_DIRNAME}/../../versions.yaml"
+	nginx_version=$("${GOPATH}/bin/yq" read "$versions_file" "docker_images.nginx.version")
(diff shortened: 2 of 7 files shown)


grahamwhaley:
18.04-initrd not happy. Is this the same thing @egernst was discussing on chat channel yesterday?

```
00:22:09 ok 1 Replication controller
00:22:09 1..1
00:22:50 not ok 1 Verify nginx connectivity between pods
00:22:50 # (in test file nginx.bats, line 29)
00:22:50 #   `kubectl wait --for=condition=Available deployment/${deployment}' failed
00:22:50 # Image is up to date for docker.io/library/busybox@sha256:6dce4a9c1635c4c9b6a2b645e6613fa0238182fe13929808ee2258370d0f3497
00:22:50 # Image is up to date for docker.io/library/nginx@sha256:706446e9c6667c0880d5da3f39c09a6c7d2114f5a5d6b74a2fafd24ae30d2078
00:22:50 # INFO: k8s configured to use runtimeclass
00:22:50 # deployment.apps/nginx-deployment created
00:22:50 # error: timed out waiting for the condition on deployments/nginx-deployment
00:22:50 # deployment.extensions "nginx-deployment" deleted
00:22:50 # Error from server (NotFound): services "nginx-deployment" not found
00:22:50 # error: resource name may not be empty
00:22:50 Makefile:107: recipe for target 'kubernetes' failed
```

egernst:
@grahamwhaley - yes, this is the issue i was hoping to resolve.  This particular test seems to be failing pretty often as of late.

grahamwhaley:
I went to look at this. Note the, in the bats file there is a setup that does a pre-fetch of the nginx image - but uses the specific version from versions.yaml (which we did not update). Sooo, i suspect now we are timing out waiting for the image to be pulled??

A list of ideas then:
- we should use the versions.yaml to edit the yaml to make it match the prefetch!
- either we prefetch or bump the timeout - as we have a prefetch, let's stick with it
- I also noticed that the default nginx is based off debian, but they also do one based off alpine (just different tag names). Whilst you are here @GabyCT - can you try switching to the alpine one and giving us an overview of the time and footprint differences that might make?

grahamwhaley:
I guess also if we are not actually specifically testing nginx (oh, another point, lets clearly document in the bats file what we are testing and why, as that will help as we maintain these tests...), and if we are just testing that we do get a curl/wget'able ability between the pods, then we might be able to switch to the k8s hello-world image, as that is also rather small.... have a think, and let's find the smallest method we can to test what we are intending to test.

egernst:
the test is for network connectivity between pods.

egernst:
note: This was failing before we updated to latest (when the cached and utilized image were the same)

GabyCT:
@egernst @grahamwhaley @chavafg changes applied

grahamwhaley:
/test
thx @GabyCT - did you figure out any of the failure case details?

GabyCT:
@grahamwhaley not really, I can not reproduce it in my VM, I ran multiple times and no failures :S

integration/kubernetes/nginx.bats
```diff
@@ -36,6 +41,8 @@ setup() {
 	waitForProcess "$wait_time" "$sleep_time" "$cmd"
 	kubectl logs "$busybox_pod" | grep "index.html"
 	kubectl describe pod "$busybox_pod"
+
+	rm -f "${pod_config_dir}/test-${deployment}.yaml"
```
chavafg: same above

GabyCT:
@chavafg changes applied

PR #1696, opened 2019-06-12
optimise k8s images

grahamwhaley:
Some of our k8s test container images were using some heavyweight items, that can be replaced by lighter ones. This saves us both time and space/cost in the CI.

- swap out debian for alpine
- swap out node-hello for hello-world

On my local tests this:
- trimmed the total image download size for the integration/k8s tests from 1.8g to 1.0g
- took the total time (including bringup and teardown lf the cluster) from 11m25s to 9m43s

### DIFF
--- a/integration/kubernetes/k8s-block-volume.bats
+++ b/integration/kubernetes/k8s-block-volume.bats
@@ -48,8 +48,10 @@ setup() {
 	# Verify persistent volume claim is bound
 	kubectl get pvc | grep "Bound"
 
+	# Add the e2fsprogs package to get mkfs.ext4 installed
+	kubectl exec "$pod_name" -- sh -c "apk add --no-cache e2fsprogs"
 	# make fs, mount device and write on it
-	kubectl exec "$pod_name" -- sh -c "mkfs -t ext4 $ctr_dev_path"
+	kubectl exec "$pod_name" -- sh -c "mkfs.ext4 $ctr_dev_path"
 	ctr_mount_path="/mnt"
 	ctr_message="Hello World"
 	ctr_file="${ctr_mount_path}/file.txt"
--- a/integration/kubernetes/k8s-expose-ip.bats
+++ b/integration/kubernetes/k8s-expose-ip.bats
@@ -4,18 +4,17 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
+#
+# Test that IP addresses/connections of PODS are routed/exposed correctly 
+# via a loadbalancer service.
 
 load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
 load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
-
-# Saves the ip addresses
-declare -a IP_ADDRESSES
 
 setup() {
 	export KUBECONFIG="$HOME/.kube/config"
 	deployment="hello-world"
 	service="my-service"
-	port="8080"
 	get_pod_config_dir
 }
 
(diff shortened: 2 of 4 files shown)


grahamwhaley:
/test
gotta see how this looks in the CI itself.

grahamwhaley:
@GabyCT - metrics failed - I wonder if this is at all related to the recent qemu update?

```
00:15:59.519 Report Summary:
00:15:59.519 +-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
00:15:59.519 | P/F |         NAME         |  FLR  |  MEAN  |  CEIL  |  GAP  |  MIN   |  MAX   |  RNG  | COV  | ITS |
00:15:59.519 +-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
00:15:59.520 | P   | boot-times           | 95.0% | 100.5% | 105.0% | 10.0% | 95.8%  | 109.2% | 14.0% | 3.6% |  20 |
00:15:59.520 | P   | memory-footprint     | 95.0% | 102.5% | 105.0% | 10.0% | 102.5% | 102.5% | 0.0%  | 0.0% |   1 |
00:15:59.522 | *F* | memory-footprint-ksm | 95.0% | 105.0% | 105.0% | 10.0% | 105.0% | 105.0% | 0.0%  | 0.0% |   1 |
00:15:59.577 +-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
```

we hit bang on the 105% limit. Looking at another recent pass, we got:

```
00:15:16.376 | P   | memory-footprint-ksm | 95.0% | 104.9% | 105.0% | 10.0% | 104.9% | 104.9% | 0.0%  | 0.0% |   1 |
```

I'm thinking our limits might be a bit tight and borderline? wdyt - can you have a look at recent history and give us an evaluation (and, nudge me if you want some scripts to help - I can PR what I have at least, even if they are not yet landable).

egernst:
(anecdotal): I've seen other metric failures that I wouldn't expect over last day of reviews.

grahamwhaley:
the apk install was near instant. debian was >100Mb, and alpine is more like 10mb.
I can do a local test to get real data....

integration/kubernetes/k8s-expose-ip.bats
```diff
@@ -30,33 +29,26 @@ setup() {
...
-	cmd="kubectl get pods -o jsonpath='{.items[*].status.phase}' | grep Running"
-	waitForProcess "$wait_time" "$sleep_time" "$cmd"
+	# There appears to be no easy way to formally wait for a loadbalancer service
+	# to become 'ready' - there is no service.status.condition field to wait on.
 
-	# Verify pods IP Addresses
-	IP_ADDRESSES=$(kubectl get pods -o jsonpath='{.items[*].status.podIP}')
+	# Now obtain the local IP:port pair of the loadbalancer service and ensure
+	# we can curl from it, and get the expected result
+	svcip=$(kubectl get service ${service} -o=json | jq '.spec.clusterIP' | sed 's/"//g')
```
chavafg: just wondering if it could be that service is not ready and get an undesired value here?

egernst: agreed.

grahamwhaley: Yep. as the comment above says though, afaict there is no way to wait on the loadbalancer service - there is no status field to wait on... if you can come up with a not insane way to poll, please paste and I'll put it in.
I resisted just sticking a `sleep` in ;-)

Locally, I have never seen it fail....

grahamwhaley:
For reference then @egernst , locally on my not-too-slow connection, the debian based block test took 33.6s, and the alpine one 14.5s

GabyCT:
@grahamwhaley the change for the metrics ci is at https://github.com/kata-containers/tests/pull/1703

integration/kubernetes/k8s-block-volume.bats
```diff
@@ -48,8 +48,10 @@ setup() {
 	# Verify persistent volume claim is bound
 	kubectl get pvc | grep "Bound"
 
+	# Add the e2fsprogs package to get mkfs.ext4 installed
+	kubectl exec "$pod_name" -- sh -c "apk add --no-cache e2fsprogs"
 	# make fs, mount device and write on it
-	kubectl exec "$pod_name" -- sh -c "mkfs -t ext4 $ctr_dev_path"
+	kubectl exec "$pod_name" -- sh -c "mkfs.ext4 $ctr_dev_path"
```
jodh-intel: I don't see an explanation for this part of the change?

grahamwhaley: merely that Alpine apk install does not install the `mkfs` generic command, so we invoke the specific `mkfs.ext4` directly - does it need calling out in the code etc.?

jodh-intel: Well I personally didn't know alpine didn't provide the wrapper command so I'd argue it isn't obvious. I speculated this might be the reason, but if you document it, it will be clear to all ;)

integration/kubernetes/k8s-expose-ip.bats
chavafg: took a quick look and seems that services are ready right away... and the CI passed correctly. I think we can have it this way.

integration/kubernetes/k8s-block-volume.bats
grahamwhaley: I've added the following text to the commit message, as I don't think having a comment in the code (describing a historical change) will add any extra clarity there:

```
The original test invoked `mkfs` with an ext4 option, but
the generic `mkfs` is not available in the Alpine package,
so invoke `mkfs.ext4` directly instead.
```

egernst:
CI Flakes?:

RHEL: 
```  bash -f .ci/install_bats.sh
08:36:32 /usr/bin/bats
08:36:32 bats integration/entropy/entropy_time.bats
08:36:32 1..1
08:37:18 not ok 1 measured time for /dev/random
08:37:18 # (in test file integration/entropy/entropy_time.bats, line 29)
08:37:18 #   `[ "$measured_time" -le "$expected_time" ]' failed
08:37:18 # 
08:37:18 # 
08:37:18 make: *** [entropy] Error 1
```

ci-vsocks: the ususal nginx connecticity failure

centos-firecracker: jenkins panic?

Issue #1686, opened 2019-06-10
CI: reduce the number jobs where cri-o and docker are executed.


chavafg:
**Which feature do you think can be improved?**

`docker` and `cri-o` tests run on every job that we use for testing our PRs, but the tests tend to take almost half an hour. 

Time (minutes)| Suite 
-- | -- 
26 | Docker tests 
23 | cri-o tests

Reducing the number of jobs where these tests run would improve the CI time and reduce the resources we use for each PR.

**How can it be improved?**
The docker tests run always using `overlay` on the same docker version: 18.06 and  iirc, we do not have tests specific to a distro, so it would be possible to reduce the number of distros that test this.

The cri-o tests run on ubuntu using devicemapper and on other distros using the default overlay driver, so we could run the tests only on two jobs using the different storage drivers.

**Describe how specifically you think it could be improved.**

Add environment variables to the jenkins jobs to know where the tests will be executed. 
such as:
`TEST_CRIO=true`
`TEST_DOCKER=true`

My proposal is to use:

`ubuntu` job to test cri-o with `devicemapper`.
`fedora` job to test cri-o with `overlay`.

`ubuntu-initrd`, `vsocks` and `centos` jobs to test docker.

All other jobs will continue running all other tests and checks, and maybe it would be a good idea to start testing k8s on other distros (right now only in ubuntu we test k8s) which take ~15 minutes.

any thought?

@grahamwhaley @egernst @jodh-intel @jcvenegas

grahamwhaley:
I'm happy with that spread @chavafg 
I agree, as we shift more to focus on k8s, we should basically start replacing the 'run docker everywhere' tests wtih 'run k8s everywhere'.
I'm having a look to see if there is anything we can do to try and reduce the k8s tests runtime. One bit of data is, the tests pull 1.8Gb of images onto the test node (busybox, nginx - 2 versions, debian, redis and node-hello).  I'll see if we can dedup nginx, and that node-hello looks huge.
I'm wondering if we might be able to pre-cache some of these, or maybe pre-load some in parallel to reduce runtime. That will be a slightly longer term plan though - I'll look at dedup now.

PR #1709, opened 2019-06-13
ci: do not run docker/cri-o tests if TEST_DOCKER/TEST_CRIO are not true 

chavafg:
do not run `cri-o` tests if `TEST_CRIO` is not true
do not run `docker` tests if `TEST_DOCKER` is not true

This behavior will only be valid when running under CI.
When running `make test` without `CI=true`, docker and
cri-o will always run.

Fixes: #1686.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,15 @@
 # The time limit in seconds for each test
 TIMEOUT := 120
 
+DOCKER_DEPENDENCY = docker
+ifeq (${CI}, true)
+	ifneq (${TEST_DOCKER}, true)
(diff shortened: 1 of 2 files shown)


chavafg:
btw, I've added the envars on each of the jenkins jobs (tired...), so the executions here should already reflect the changes...

grahamwhaley:
Great. silly question - did we/do we/sometime need to update the CI test matrix table to reflect the changes?

chavafg:
@grahamwhaley yeah we need to, I'll do that, thanks for the reminder.

Issue #1710, opened 2019-06-13
flake test: vsock test when using vsock


egernst:
Seeing failure on http://jenkins.katacontainers.io/job/kata-containers-runtime-vsocks-PR/

egernst:
example failure: 
http://jenkins.katacontainers.io/job/kata-containers-runtime-vsocks-PR/936/console

```
• Failure [12.492 seconds]
[1] vsock test
[1] /tmp/jenkins/workspace/kata-containers-runtime-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/vsock_test.go:13
[1]   when using vsock
[1]   /tmp/jenkins/workspace/kata-containers-runtime-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/vsock_test.go:31
[1]     should print the agent logs in the shim journal [It]
[1]     /tmp/jenkins/workspace/kata-containers-runtime-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/vsock_test.go:48
[1] 
[1]     Expected
[1]         <int>: 1
[1]     to be zero-valued
[1] 
[1]     /tmp/jenkins/workspace/kata-containers-runtime-vsocks-PR/go/src/github.com/kata-containers/tests/integration/docker/vsock_test.go:62
```

chavafg:
Ohh, need to backport a commit for this to work on ubuntu

GabyCT:
Closing this issue as the backport was done

PR #1712, opened 2019-06-13
tests: Modify vsock test to not use `journalctl -g`

chavafg:
`journalctl -g` does not work properly on some linux distributions,
like ubuntu or centos.

Modify the test to search for `source=agent` using gomega matcher.

Fixes: #1710.


grahamwhaley:
one travis fail - let me nudge that:
```
INFO: Checking URL http://www.dest-unreach.org/socat/
ERROR: Invalid URL 'https://tools.ietf.org/html/rfc3339' found in the following files:
cmd/log-parser/README.md
```

btw, is this a cherry pick backport? Just asking, as I thought the cherry-pick from our [workflow](https://github.com/kata-containers/community/blob/master/Backport-Guide.md#backport-workflow) would add a ref. into the commit?

PR #1715, opened 2019-06-14
kubernetes: assure swap=off when running kubernetes integration tests

Pennyzct:
When setting up kubernetes on bare metal, it may occur `swap=on` error during preflight checks. 
```
[preflight] Running pre-flight checks
error execution phase preflight: [preflight] Some fatal errors occurred:
        [ERROR Swap]: running with swap on is not supported. Please disable swa                            p
[preflight] If you know what you are doing, you can make a check non-fatal with                               `--ignore-preflight-errors=...`
Failed at 54: sudo -E kubeadm init --config "${kubeadm_config_file}"
```
Since running kubernetes with `swap=on` is not supported. we need to assure swap file/partition turned off before kubeinit preflight checks.

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -51,6 +51,10 @@ kubeadm_config_file="$(mktemp --tmpdir kubeadm_config.XXXXXX.yaml)"
 sed -e "s|CRI_RUNTIME_SOCKET|${cri_runtime_socket}|" "${kubeadm_config_template}" > "${kubeadm_config_file}"
 sed -i "s|KUBERNETES_VERSION|v${kubernetes_version/-*}|" "${kubeadm_config_file}"
 
+BAREMETAL="${BAREMETAL:-false}"
+if [ "${BAREMETAL}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
+	sudo swapoff -a || true
+fi
 sudo -E kubeadm init --config "${kubeadm_config_file}"
 
 mkdir -p "$HOME/.kube"


integration/kubernetes/init.sh
```diff
@@ -51,6 +51,9 @@ kubeadm_config_file="$(mktemp --tmpdir kubeadm_config.XXXXXX.yaml)"
 sed -e "s|CRI_RUNTIME_SOCKET|${cri_runtime_socket}|" "${kubeadm_config_template}" > "${kubeadm_config_file}"
 sed -i "s|KUBERNETES_VERSION|v${kubernetes_version/-*}|" "${kubeadm_config_file}"
 
+if [ "${BAREMETAL}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
```
grahamwhaley: :-) fun test. You probably could have gotten away with `sudo swapoff -a || true` :-)
Yeah, I hit this swap issue every time I reboot one of my local test machines ;-)

grahamwhaley:
We have a bucket (6 of) CI fails here. @chavafg - any idea how many of those are 'known issues'?

Pennyzct:
Hi~ @grahamwhaley I'm investigating the debug info, it may come from same error, 
`BAREMETAL: unbound variable`
IIrc, This variant, is defined in Jenkins env? ;)

integration/kubernetes/init.sh
Pennyzct: it looks like that this error only occur in bare mental???? I'm wondering that's because VM won't get any swap allocation???

grahamwhaley:
@Pennyzct - ah, `BAREMETAL` is only set for the baremetal slaves, and not the other slaves - so the bash error is from the `set -o nounset` at the top of that file. I think there are bash tricks to work around that (maybe like `${BAREMETAL:-}` or similar). I don't have details in my head, you'll have to look it up :-)

Pennyzct:
thanks!!!!! @grahamwhaley you have already pointed me a wonderful direction. ;)

Pennyzct:
@grahamwhaley only ARM is failing, hahaha. And it is waiting for [osbuilder#265](https://github.com/kata-containers/osbuilder/pull/265)

Issue #1711, opened 2019-06-13
flake: url checker


egernst:
1. Seeing transient failures
 - https://travis-ci.org/kata-containers/documentation/jobs/545128784
2. Seeing [timeout issue](https://travis-ci.org/kata-containers/runtime/jobs/545059595): 
```
INFO: Checking documentation
INFO: Checking local branch for changed documents only
INFO: Checking document code blocks
INFO: processing file 'README.md'
INFO: processing file 'virtcontainers/README.md'
INFO: processing file 'virtcontainers/documentation/Developers.md'
INFO: processing file 'virtcontainers/documentation/api/1.0/api.md'
INFO: processing file 'virtcontainers/hack/virtc/README.md'
INFO: Checking all document URLs
INFO: ignoring new (but correct) URL: https://github.com/kata-containers
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

egernst:
Until we have retry or some other logic in place, we may want to consider turning this off, or have it become its own job travis job and mark it non-blocking?

/cc @jodh-intel

PR #1716, opened 2019-06-14
CI: Add network timeout

jodh-intel:
Add a timeout in the static checks script for `git ls-remote`. This hits the network and can cause a hang (and subsequent job timeout).

Fixes: #1711.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -600,7 +600,7 @@ static_check_docs()
 		# Git repo URL check
 		if echo "$url"|grep -q '^https.*git'
 		then
-			git ls-remote "$url" > /dev/null 2>&1 && continue
+			timeout "${KATA_NET_TIMEOUT}" git ls-remote "$url" > /dev/null 2>&1 && continue
 		fi
 
 		info "Checking URL $url"
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -20,6 +20,9 @@ KATA_KSM_THROTTLER_JOB="kata-ksm-throttler"
 # more formats).
 export KATA_DOCKER_TIMEOUT=30
 
+# Number of seconds to wait for a general network operation to complete.
+export KATA_NET_TIMEOUT=30
+
 # Ensure GOPATH set
 if command -v go > /dev/null; then
 	export GOPATH=${GOPATH:-$(go env GOPATH)}


.ci/static-checks.sh
```diff
@@ -600,7 +600,7 @@ static_check_docs()
 		# Git repo URL check
 		if echo "$url"|grep -q '^https.*git'
 		then
-			git ls-remote "$url" > /dev/null 2>&1 && continue
+			timeout "${KATA_NET_TIMEOUT}" git ls-remote "$url" > /dev/null 2>&1 && continue
```
grahamwhaley: Nice. I wondered if there was a more git'ish way to do this - looks like we may be able to do a:
```
$ git config --global http.lowSpeedTime 30
```
or similar to set a timeout in git itself. I'm going to presume that will also propagate into `go get`. Worth a shot?

grahamwhaley: Oh, this might go in hand with `http.lowSpeedLimit` as well..

jodh-intel: @grahamwhaley - the man page isn't clear to me on how those settings actually work - for example, what happens if you only set `http.lowSpeedTime`? What is the default for `http.lowSpeedLimit`? I like the idea, but unless you happen to know the specifics I think this might be a follow-on PR to optimise further.

egernst:
The intermittent failures were driving me nuts -- thanks for fixing James.  Let's go ahead and merge.  Happy to see follow on PR to git'ify.

.ci/static-checks.sh
grahamwhaley: Yep, not very clear or obvious is it...:

https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

> If the data rate of an HTTP operation is lower than GIT_HTTP_LOW_SPEED_LIMIT bytes per second for longer than GIT_HTTP_LOW_SPEED_TIME seconds, Git will abort that operation. These values override the http.lowSpeedLimit and http.lowSpeedTime configuration values.

Soo, I think we could set something that would effectively detect a dead, or near-dead, connection:

```bash
$ git config --global http.lowSpeedLimit 1000
$ git config --global http.lowSpeedTime 60
```

That is, if we have a data rate of less than 1k (it is not clear if that is 1k total, or 1k/s or ???, but, I think it does not matter really here....) for 1 minute, then git will abort.

I'm happy to fire a PR so we can see if it flies??

PR #1728, opened 2019-06-18
arch: install tests repo brfore using script kata-arch.sh

Pennyzct:
# Description of problem
```
17:41:04 .ci/jenkins_job_build.sh: line 48: /home/jenkins/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/kata-containers/tests/.ci/kata-arch.sh: No such file or directory
```
Let's install `tests/` repo brfore using script `kata-arch.sh`, which lives in `tests/` repo.
I'm confused why it only occurred in ARM CI node???? ☹
@chavafg @grahamwhaley

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -45,11 +45,11 @@ export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:/sbin:${PATH}
 kata_repo_dir="${GOPATH}/src/${kata_repo}"
 tests_repo_dir="${GOPATH}/src/${tests_repo}"
 
-arch=$("${tests_repo_dir}/.ci/kata-arch.sh")
-
 # Get the tests repository
 mkdir -p $(dirname "${tests_repo_dir}")
 [ -d "${tests_repo_dir}" ] || git clone "https://${tests_repo}.git" "${tests_repo_dir}"
+
+arch=$("${tests_repo_dir}/.ci/kata-arch.sh)
 
 # Get the repository of the PR to be tested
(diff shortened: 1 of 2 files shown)


.ci/jenkins_job_build.sh
```diff
@@ -45,12 +45,12 @@ export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:/sbin:${PATH}
 kata_repo_dir="${GOPATH}/src/${kata_repo}"
 tests_repo_dir="${GOPATH}/src/${tests_repo}"
 
-arch=$("${tests_repo_dir}/.ci/kata-arch.sh")
-
 # Get the tests repository
 mkdir -p $(dirname "${tests_repo_dir}")
 [ -d "${tests_repo_dir}" ] || git clone "https://${tests_repo}.git" "${tests_repo_dir}"
 
+arch=$("${tests_repo_dir}/.ci/kata-arch.sh)
```
grahamwhaley: I suspect the thing that is different for this ARM job is maybe something in the Jenkins job setup that is different/missing between that and x86. Having said that, this change lgtm - if we are pulling the tests repo anyway just after we do the arch test, then moving the test down I think is fine.

.ci/lib.sh
```diff
@@ -213,6 +213,16 @@ delete_stale_kata_resource()
 	done
 }
 
+delete_stale_lock_files()
```
grahamwhaley: I guess you accidentally included a commit from another PR here @Pennyzct ? :-)

grahamwhaley:
and now I see this was closed - never mind - I'll catch up in a minute!

Issue #1734, opened 2019-06-19
CI url check failure


bergwolf:
# Description of problem
Saw at https://travis-ci.org/kata-containers/runtime/jobs/546582094

```
ERROR: Invalid URL 'https://godoc.org/github.com/kata-containers/runtime' found in the following files:
pkg/katatestutils/README.md
README.md
ERROR: Invalid URL 'https://godoc.org/github.com/kata-containers/runtime/pkg/katatestutils' found in the following files:
pkg/katatestutils/README.md
ERROR: Invalid URL 'https://godoc.org/github.com/kata-containers/runtime?status.svg' found in the following files:
README.md
```
These URLs all look valid. So it seems to be network issues. Any chance to make it retry-able?

jodh-intel:
Yeah. It's pretty horrid but I think you are right - we'll have to make a few attempts. I'll grab this issue...

PR #1753, opened 2019-06-20
CI: Retry URL checks

jodh-intel:
If a URL check fails, retry 3 times (by default) in case of transitory network errors.

Fixes: #1734.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -416,6 +416,62 @@ EOT
 	fi
 }
 
+check_url()
+{
+	local url="$1"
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Maybe I'm missing something, but I don't see the 'retry 3 times' code? Is it subtle?
I was expecting to find a loop - or, maybe use the `curl -retry` mechanism?

jodh-intel:
Last commit? `for attempt in ...` ;)

However, I've now made it clearer by using `curl`'s built-in retry logic :)

PR #1766, opened 2019-06-25
netns: clean stale resource "/run/netns" on bare-metal CI node

Pennyzct:
sandbox network namespace is mounted under `/run/netns`. It will still exist when kata container quits abnormally.
we need to `umount` this stale resource before deleting them.
```
root@arm-testing-1:~# mount | grep /run/netns
nsfs on /run/netns/cni-fde66afc-e19b-53b7-23d5-f3616418890c type nsfs (rw)
nsfs on /run/netns/cni-7ecb8c88-a65d-770e-39e1-a2ce45a417f5 type nsfs (rw)
nsfs on /run/netns/cni-fcd24a78-c5bb-2289-b6d3-11c531b72443 type nsfs (rw)
nsfs on /run/netns/cni-0edba821-e1d6-82ef-4a36-19b443c1422f type nsfs (rw)
nsfs on /run/netns/cni-37ae4d3f-f26e-09a3-b994-7434485699d9 type nsfs (rw)
nsfs on /run/netns/cni-284fc07d-971d-af2e-125b-69f2c3be0034 type nsfs (rw)
nsfs on /run/netns/cni-ac024861-7fe7-3cd5-bb67-03fb9587d170 type nsfs (rw)
nsfs on /run/netns/cni-8a74f200-0d1e-434e-e898-ad861d6d23a9 type nsfs (rw)
nsfs on /run/netns/cni-5e3644e6-23a9-55c6-04b0-c3fb2307df76 type nsfs (rw)
nsfs on /run/netns/cni-7bc9dc39-cd53-9c1f-aa4d-13a521bff34b type nsfs (rw)
nsfs on /run/netns/cni-26373900-0f64-f092-13c6-6150ecf723b8 type nsfs (rw)
nsfs on /run/netns/cni-4b2af18e-11f4-3408-113b-7135456b4208 type nsfs (rw)
nsfs on /run/netns/cni-9cfa915a-5e45-4101-ecd8-536b58310f75 type nsfs (rw)

```

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -206,6 +206,17 @@ delete_stale_docker_resource()
 
 delete_stale_kata_resource()
 {
+	# umount before dir
+	for stale_kata_mount_point in "${stale_kata_mount_point_union[@]}"; do
+		local mount_point_union=$(mount | grep "${stale_kata_mount_point}" | awk '{print $3}')
+		if [ -n "${mount_point_union}" ]; then
+			while IFS='$\n' read mount_point; do
+				[ -n "$(grep "${mount_point}" "/proc/mounts")" ] && sudo umount -R "${mount_point}"
+			done <<< "${mount_point_union}"
+		fi
+	done
+
+	#delete stale dir
 	for stale_kata_dir in "${stale_kata_dir_union[@]}"; do
 		if [ -d "${stale_kata_dir}" ]; then
 			sudo rm -rf "${stale_kata_dir}"
@@ -255,7 +266,8 @@ gen_clean_arch() {
 	docker_storage_driver=$(timeout ${KATA_DOCKER_TIMEOUT} docker info --format='{{.Driver}}')
 	stale_docker_mount_point_union=( "/var/lib/docker/containers" "/var/lib/docker/${docker_storage_driver}" )
 	stale_docker_dir_union=( "/var/lib/docker" )
-	stale_kata_dir_union=( "/var/lib/vc" "/run/vc" "/usr/share/kata-containers" "/usr/share/defaults/kata-containers" )
+	stale_kata_mount_point_union=( "/run/netns" )
+	stale_kata_dir_union=( "/var/lib/vc" "/run/vc" "/usr/share/kata-containers" "/usr/share/defaults/kata-containers" "/run/netns" )
 
 	info "kill stale process"
 	kill_stale_process


.ci/lib.sh
```diff
@@ -206,6 +206,17 @@ delete_stale_docker_resource()
 
 delete_stale_kata_resource()
 {
+	# umount before dir
+	for stale_kata_mount_point in "${stale_kata_mount_point_union[@]}"; do
+		local mount_point_union=$(mount | grep "${stale_kata_mount_point}" | awk '{print $3}')
+		if [ -n "${mount_point_union}" ]; then
+			while IFS='$\n' read mount_point; do
+				[ -n "$(grep "${mount_point}" "/proc/mounts")" ] && sudo umount -R "${mount_point}"
```
grahamwhaley: Whilst this lgtm - is this also indicative of a bug in Kata - presumably we should never be leaving these stale mounts around?
/cc @amshinde @devimc

Pennyzct: if the container exits normally, it will not leave this resource behind. But if it exited abnormally, like the hypervisor(qemu) doesn't start up, it will cause this stale mounts.

grahamwhaley: @Pennyzct - thx - so, I would still call that a bug :-) Under no circumstances should we normally leave stuff laying around. Particularly if it is something like qemu not starting up, which the runtime should notice, and clean up. Could you open an Issue with the details you have please, and then we can try to find somebody to have a look (/cc some folks who may have looked in the cleanup-on-fail area before @WeiZhang555 @bergwolf )

Pennyzct: Hi~ @grahamwhaley I've been tracked the point where the network namespace gets cleaned up. it's this func [cleanupNetNS](https://github.com/kata-containers/runtime/blob/master/pkg/katautils/network.go). it will get cleaned up when sandbox fails to create, so including qemu not starting up. ahhhhhhh, it's killing me. 😭🤮

grahamwhaley: @Pennyzct cleanup is **hard** - harder than creation :-)  Do you have a reliable test case that makes this happen? Maybe the exit path is not going through `clearnupNetNS` - maybe the golang program is being killed or something. Going to be easiest to track if we can at least reproduce. I know we have had trouble reproducing such 'leaks' in the past... tricky....

Pennyzct: @grahamwhaley. After looking through log from` journalctl -t kata-runtime` and `/var/log/syslog`, i found nothing related. that's the pain point. o(TヘTo). I couldn't find the exact test case to produce this leak.
I'm thinking, maybe we don't turn on the `debug` in ci running, so maybe missing some info. 
I'm gonna find a real clean local machine to mimic one ARM CI run, to check out if it's reproducing.
BTW, running CI in bare-metal is really pain in the ass. Especially, recently I'm thinking to include
k8s test cases.
Do you guys think that running kata CI in docker container is a feasible solution? @grahamwhaley @chavafg

grahamwhaley: Hi @Pennyzct ... I'm not sure we can run the CI in a docker container - that would need the ability to run Kata in Docker - which I suspect you could only do with a priv container - which might defeat some of the point... if you do follow that route do keep us in the loop though.
However, can you do nested VMs on your hardware? If so, then I did previously have the metrics baremetal running the CI inside VMs (https://github.com/kata-containers/ci/tree/master/VMs/metrics). Hmm, I suspect we have had this conversation before...

grahamwhaley: oh, one other option, if you are using Packet servers, is to try and fix the 'dynamically deploy packet from Jenkins'. I tried this in the past using the [jclouds](https://wiki.jenkins.io/display/JENKINS/JClouds+Plugin) plugin - as packet has a jclouds integration, but packet did not appear in the plugin dropdown of providers - packet were aware of that at the time I think - I don't know if that has been fixed/progressed though...

Alternatively, I see there is a REST API for packet as well - I suspect that could also be used to deploy Jenkins slaves.
[REST API](https://support.packet.com/kb/articles/deploy-via-the-api)

I'd love to dynamically deploy the metrics machines in packet - I have been pursuing that via our Zuul integration, as there is Zuul/packet integration, and that is also the only way I can store metrics results into the OSF/infra/zuul elasticsearch DB as well....

Pennyzct: Hi~ @grahamwhaley Sadly, nested virtualization is really a long way on AArch64. 😭. So I'm considering container.;) 
As for privilege problem in kata, I only had a ambiguous concept that we couldn't support privileged kata container, as for why, i'm not clear,  would you like to point me some related issues or PRs? 😋

jodh-intel: Hi @Pennyzct - I think you are after https://github.com/kata-containers/documentation/blob/master/Limitations.md#docker-run---privileged

Pennyzct: Hi @jodh-intel thanks! I have skipped the whole related issue. So it couldn't provide all runc supports, just a subset of host devices, including block and vfio devices. So not 100 percentage `--priviledged`? 
@grahamwhaley As far as I know, I'm gonna still give a try on running kata CI on docker container. (✿◕‿◕✿)

grahamwhaley: Hi @Pennyzct @jodh-intel - I think we might be mixing up Docker-in-Kata and Kata-in-Docker... to run the Kata CI system *inside* a container, you would need to use Kata-in-Docker. The above links are talking about running priv containers under Kata, and more like Docker-in-Kata....

So, I think you are wanting to try and run Kata **inside** a container @Pennyzct - and as Kata needs to perform some sudo level system configuration (like disk mounts and network connections), it will need to run in a docker priv container I believe. I suggest probably try it - run a docker priv container, try and install kata into it, and run it. Oh, to do that I guess you will need to be running [DockerInDocker](https://hub.docker.com/_/docker) - there is a strong recommendation there to read the relevant blog posts first...

I did consider some time back if it might be easier to run docker/kata under lxc/ldx rather than docker - but I've never tried any of them...

Pennyzct: Hi @grahamwhaley I'm not gonna using **_the whole package docker in docker_**. It has been proved to have multiple ugly scenarios. 
I'm trying to use this magical thing:
```
docker run -v /var/run/docker.sock:/var/run/docker.sock --privileged \
           -ti docker
```
this blog has detailed info about running ci with DockerInDocker, [Using Docker-in-Docker for your CI or testing environment? Think twice](https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/)
I have to be extramely careful about events involving removingcontainers, since it mat remove the **_docker_** container. ;)
For now, all is in my imagination. I'll find time to give it a try asap. ;)

jodh-intel: Any update on this @Pennyzct? :)

amshinde:
@Pennyzct Are you still seeing network namespaces being left behind?
We should probably add a check to our k8s tests to make sure that no network namespace is left behind after a pod is deleted. One can get the network namespace for a pod with `crictl inspectp {podid}` or with `docker inspect` command.
cc @GabyCT

jodh-intel:
Hi @Pennyzct - is this still a problem?

jodh-intel:
Hi @Pennyzct - have you had a chance to look at this again?

Pennyzct:
Hi~~ @jodh-intel 
Sooooooo sorry for this late response. I've totally forgot it. 
I've logged into the ARM CI to see if there still existed stale network namespace. And I'm afraid that it still was there.
But, I think, just deleting them may not be a right choice, I still need to fix it from the source. ;).
I'll close this PR and when I find the root cause, I'll fire a new PR.

PR #1778, opened 2019-07-02
ci: make crictl install faster and independent of CRI-O

### LINKED ISSUE #1777: ci: make crictl installation independent of CRI-O
The installation of crictl currently depends on the Dockerfile within CRI-O. I would like to make these two independent again to continue with further improvements within the CRI-O repository.

saschagrunert:
Fixes #1777

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -39,9 +39,6 @@ if [ "$ID" == "fedora" ]; then
 	fi
 fi
 
-go get -d "$crio_repo" || true
-pushd "${GOPATH}/src/${crio_repo}"
-
 if [ "$ghprbGhRepository" != "${crio_repo/github.com\/}" ]
 then
 	# For Fedora, we use CRI-O version that is compatible with the
@@ -55,23 +52,23 @@ then
 		else
 			crio_version=$(get_version "externals.crio.meta.openshift")
 		fi
+		crictl_version=$(get_version "externals.crio.meta.crictl")
+		crictl_tag_prefix=""
 	else
 		crio_version=$(get_version "externals.crio.version")
+		crictl_version=$(get_test_version "externals.critools.version")
+		crictl_tag_prefix="v"
 	fi
-
-	git fetch
-	git checkout "${crio_version}"
 fi
 
-echo "Get CRI Tools"
-critools_repo="${kubernetes_sigs_org}/cri-tools"
-go get "$critools_repo" || true
-pushd "${GOPATH}/src/${critools_repo}"
-critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
-git checkout "${critools_version}"
-go install ./cmd/crictl
-sudo install "${GOPATH}/bin/crictl" /usr/bin
-popd
+echo "Installing CRI Tools"
+crictl_url=https://github.com/kubernetes-sigs/cri-tools/releases/download/v$crictl_version/crictl-$crictl_tag_prefix$crictl_version-linux-"$(${cidir}/kata-arch.sh -g)".tar.gz
+curl -Ls "$crictl_url" | sudo tar xfz - -C /usr/local/bin
+
+go get -d "$crio_repo" || true
+pushd "${GOPATH}/src/${crio_repo}"
+git fetch
+git checkout "${crio_version}"
 
 echo "Installing CRI-O"
 make clean
--- a/versions.yaml
+++ b/versions.yaml
@@ -58,3 +58,6 @@ externals:
     description: "GNU parallel tool"
     url: "https://ftp.gnu.org/gnu/parallel"
     version: "20190222"
+
+  critools:
+    version: 1.14.0


chavafg:
oh cool! 
Can you please amend your commit to add a body and the `Fixes: #1777`? 

Thanks.

saschagrunert:
> Can you please amend your commit to add a body and the `Fixes: #1777`?

Yes, sure :)

.ci/install_crio.sh
```diff
@@ -64,14 +64,9 @@ then
...
-critools_repo="${kubernetes_sigs_org}/cri-tools"
-go get "$critools_repo" || true
-pushd "${GOPATH}/src/${critools_repo}"
-critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
-git checkout "${critools_version}"
-go install ./cmd/crictl
-sudo install "${GOPATH}/bin/crictl" /usr/bin
-popd
+crictl_version=v1.14.0
+wget -qO- https://github.com/kubernetes-sigs/cri-tools/releases/download/$crictl_version/crictl-$crictl_version-linux-amd64.tar.gz \
```
grahamwhaley: @chavafg -do we only test on x86 - or do we need to make this arch independent/tolerant ?

chavafg: oh, you are right... We currently test only x86, but I know that @Pennyzct is adding support for arm, so I think it would be good to add support for different architectures.
@saschagrunert do you think you can make the change?  
I think you can use `.ci/kata-arch.sh -g` to get the correct arch name.

.ci/install_crio.sh
```diff
@@ -64,14 +64,9 @@ then
...
 echo "Get CRI Tools"
-critools_repo="${kubernetes_sigs_org}/cri-tools"
-go get "$critools_repo" || true
-pushd "${GOPATH}/src/${critools_repo}"
-critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
-git checkout "${critools_version}"
-go install ./cmd/crictl
-sudo install "${GOPATH}/bin/crictl" /usr/bin
-popd
+crictl_version=v1.14.0
```
chavafg: also, can you please move this to https://github.com/kata-containers/tests/blob/master/versions.yaml and then get the value here using something like: `crictl_version=$(get_test_version "external.crictl.version")`

saschagrunert: Yes, I will add that.

.ci/install_crio.sh
saschagrunert: Alright, I will do it like this.

.ci/install_crio.sh
```diff
@@ -64,14 +64,9 @@ then
...
 echo "Get CRI Tools"
-critools_repo="${kubernetes_sigs_org}/cri-tools"
-go get "$critools_repo" || true
-pushd "${GOPATH}/src/${critools_repo}"
-critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
-git checkout "${critools_version}"
-go install ./cmd/crictl
-sudo install "${GOPATH}/bin/crictl" /usr/bin
-popd
+crictl_version=$(get_test_version "externals.critools.version")
```
chavafg: I went through the failure on the ubuntu job and see that it cannot find the `versions.yaml` file:
```
10:15:41 ERROR: cannot find .ci/../versions.yaml
10:15:41 Failed at 67: crictl_version=$(get_test_version "externals.critools.version")
```

@saschagrunert can you please move these lines before we `pushd` into the `cri-o` directory (L42)?, so that we are still on our CI working directory and can find the `versions.yaml` file.

saschagrunert: Yes, thanks for pointing me to the issue :pray:

.ci/install_crio.sh
```diff
@@ -39,6 +39,11 @@ if [ "$ID" == "fedora" ]; then
 	fi
 fi
 
+echo "Get CRI Tools"
+crictl_version=$(get_test_version "externals.critools.version")
+wget -qO- https://github.com/kubernetes-sigs/cri-tools/releases/download/$crictl_version/crictl-$crictl_version-linux-"$(${cidir}/kata-arch.sh -g)".tar.gz \
+	| tar xfz - -C /usr/bin
```
chavafg: seems that this needs to be `sudo tar ...`

saschagrunert: Makes sense, I change it.

chavafg: thanks

chavafg:
@saschagrunert I went trough the issues on the CI:
The first one is that we have a failure on one of the tests
```
not ok 24 ctr execsync exit code
# (in test file ctr.bats, line 918)
#   `[ "$status" -ne 0 ]' failed
```
Taking a look at the tests, the test fails making an `exec`:
https://github.com/cri-o/cri-o/blob/v1.14.1/test/ctr.bats#L918

I see here that they changed the behavior of the test in this commit:
https://github.com/cri-o/cri-o/commit/1551436394115bd32121d589d8b6f0706ddb372c#diff-c5753339854697f7106d65aea5be6788

So maybe the `crictl` version is not compatible with `cri-o` v1.14.1 ?
Maybe we should try with `cri-o` v1.14.5.

The second issue is that we an older version of cri-o (branch release-1.10) on the Fedora job as we run openshift (3.10) there and the v1.14 version of `crictl` is not compatible with the old `cri-o` version.
For this one, I think that we can add another entry for the old cri-tool (which is [`release-1.10`](https://github.com/cri-o/cri-o/blob/6273bea4c9ed788aeb3d051ebf2d030060c05b6c/Dockerfile#L87))version inside the yaml file.

saschagrunert:
> So maybe the `crictl` version is not compatible with `cri-o` v1.14.1 ?
> Maybe we should try with `cri-o` v1.14.5.

Yeah might be. Can you point me to the location where the `externals.crio.version` is defined for the ubuntu job? Is it inside a different repository? :)

saschagrunert:
@chavafg I changed the cri tools implementation accordingly to your suggestions, do we want to go for an additional test again? :)

.ci/install_crio.sh
```diff
@@ -55,23 +55,20 @@ then
...
-critools_repo="${kubernetes_sigs_org}/cri-tools"
-go get "$critools_repo" || true
-pushd "${GOPATH}/src/${critools_repo}"
-critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
-git checkout "${critools_version}"
-go install ./cmd/crictl
-sudo install "${GOPATH}/bin/crictl" /usr/bin
-popd
+echo "Installing CRI Tools"
+wget -qO- "$crictl_url" | tar xfz - -C /usr/bin
```
jodh-intel: - I see you are updating already existing code but I think we need a check in here to test for [`KATA_DEV_MODE`](https://github.com/kata-containers/tests#developer-mode) as devs probably don't want their `/usr/bin/` stamped on (but `/usr/local/` would be "fair game" I think).

- We avoid `wget` where possible so please could you convert this to use `curl`.

.ci/install_crio.sh
```diff
@@ -55,23 +55,20 @@ then
 		else
 			crio_version=$(get_version "externals.crio.meta.openshift")
 		fi
+		crictl_version=$(get_test_version "externals.critools.openShift3Version")
+		crictl_url=https://github.com/kubernetes-sigs/cri-tools/releases/download/v$crictl_version/crictl-$crictl_version-linux-"$(${cidir}/kata-arch.sh -g)".tar.gz
 	else
 		crio_version=$(get_version "externals.crio.version")
+		crictl_version=$(get_test_version "externals.critools.version")
+		crictl_url=https://github.com/kubernetes-sigs/cri-tools/releases/download/$crictl_version/crictl-$crictl_version-linux-"$(${cidir}/kata-arch.sh -g)".tar.gz
```
jodh-intel: These two `crictl_url` variables *look* identical but on closer inspection one contains a `v` as a version prefix and the other doesn't. Is that really correct?

If it is, please could you make this clearer by reworking the code to something like:

```sh
version_prefix=""

if [ ... ]; then
    version_prefix="v"
    # ...
else
   # ...
fi

cri_tools_site=https://github.com/kubernetes-sigs/cri-tools
cri_tools_file="crictl-${crictl_version}-linux-$(${cidir}/kata-arch.sh -g).tar.gz"

crictl_url="${cri_tools_site}/releases/download/${version_prefix}${crictl_version}/${cri_tools_file}
```
```

File: versions.yaml [hunk 7]
```diff
@@ -58,3 +58,7 @@ externals:
     description: "GNU parallel tool"
     url: "https://ftp.gnu.org/gnu/parallel"
     version: "20190222"
+
+  critools:
+    version: v1.14.0
+    openShift3Version: 1.0.0-beta.2
```
jodh-intel: We already encode openshift in the main versions database, so should these details go there?:

- https://github.com/kata-containers/runtime/blob/master/versions.yaml#L193
- https://github.com/kata-containers/runtime/blob/master/versions.yaml#L233

grahamwhaley:
@saschagrunert - CI resets when you push changes, so we would need to test again - but, I'll leave triggering that until you assess the feedback from @jodh-intel ;-)

.ci/install_crio.sh
saschagrunert: I think we could generally use `/usr/local`, then an additional check would not be needed right?
I changed the command to use `curl` instead of `wget`

.ci/install_crio.sh
saschagrunert: Yes, I adapt it as suggested.

versions.yaml
saschagrunert: Okay, I will add it there.

.ci/install_crio.sh
```diff
@@ -55,23 +55,21 @@ then
...
-critools_repo="${kubernetes_sigs_org}/cri-tools"
-go get "$critools_repo" || true
-pushd "${GOPATH}/src/${critools_repo}"
-critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
-git checkout "${critools_version}"
-go install ./cmd/crictl
-sudo install "${GOPATH}/bin/crictl" /usr/bin
-popd
+echo "Installing CRI Tools"
+crictl_url=https://github.com/kubernetes-sigs/cri-tools/releases/download/v$crictl_version/crictl-$crictl_tag_prefix$crictl_version-linux-"$(${cidir}/kata-arch.sh -g)".tar.gz
```
jodh-intel: This variable still contains `v$crictl_version` ...?

saschagrunert: Yes, this is a little skew (tag vs file name) in cri-tools but it should work fine.

versions.yaml
saschagrunert: Created another PR to address this issue: https://github.com/kata-containers/runtime/pull/1873

.ci/install_crio.sh
```diff
@@ -55,23 +55,21 @@ then
...
+wget -Ls "$crictl_url" | tar xfz - -C /usr/local/bin
```
jodh-intel: I think you might need to re-push as this is still showing as `wget`.

.ci/install_crio.sh
saschagrunert: So in the end we have both urls, which are working:
```
https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.14.0/crictl-v1.14.0-linux-amd64.tar.gz
https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.0.0-beta.2/crictl-1.0.0-beta.2-linux-amd64.tar.gz
```

.ci/install_crio.sh
saschagrunert: Thanks for the hint, you're right. :) Re-pushed my changes.

chavafg:
```
10:33:41 .ci/lib.sh: line 121: .ci/install_yq.sh: No such file or directory
10:33:41 ERROR: cannot find .ci/../versions.yaml
10:33:41 Failed at 62: crictl_version=$(get_test_version "externals.critools.version")
```
I think we need to move the `pushd` on L43 along with the git commands on L66-67 after the `if` finishes on L68 to find the files needed to get the versions

saschagrunert:
> ```
> 10:33:41 .ci/lib.sh: line 121: .ci/install_yq.sh: No such file or directory
> 10:33:41 ERROR: cannot find .ci/../versions.yaml
> 10:33:41 Failed at 62: crictl_version=$(get_test_version "externals.critools.version")
> ```
> 
> I think we need to move the `pushd` on L43 along with the git commands on L66-67 after the `if` finishes on L68 to find the files needed to get the versions

Yes, I moved it below.

saschagrunert:
@chavafg may I ask you to trigger a /test ? :)

.ci/install_crio.sh
chavafg: @saschagrunert can you move the `Installing CRI tools` section before we `pushd` into the cri-o repo?
```
10:58:02 .ci/install_crio.sh: line 70: .ci/kata-arch.sh: No such file or directory
10:58:02 Failed at 70: crictl_url=https://github.com/kubernetes-sigs/cri-tools/releases/download/v$crictl_version/crictl-$crictl_tag_prefix$crictl_version-linux-"$(${cidir}/kata-arch.sh -g)".tar.gz
```

saschagrunert: Hm, alright I moved it above.

.ci/install_crio.sh
```diff
@@ -55,23 +52,23 @@ then
...
-go get "$critools_repo" || true
-pushd "${GOPATH}/src/${critools_repo}"
-critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
-git checkout "${critools_version}"
-go install ./cmd/crictl
-sudo install "${GOPATH}/bin/crictl" /usr/bin
-popd
+echo "Installing CRI Tools"
+crictl_url=https://github.com/kubernetes-sigs/cri-tools/releases/download/v$crictl_version/crictl-$crictl_tag_prefix$crictl_version-linux-"$(${cidir}/kata-arch.sh -g)".tar.gz
+curl -Ls "$crictl_url" | tar xfz - -C /usr/local/bin
```
chavafg: can you please add `sudo` here?

saschagrunert: Yes, added it. I thought it was already there :thinking:

chavafg: yep, don't worry, I think this should be fine now 😄 
/test

saschagrunert:
I guess we have to drop another /test here. 😇

chavafg:
ha, the trigger phrase didn't work on the review comment. Thanks for noticing it. Should work now:
/test

saschagrunert:
Hm, the Kernel Build seems to fail:

```
19:51:41 Failed at 62: "${kernel_repo_dir}/kernel/build-kernel.sh" -v "${kernel_version}" "build"
19:51:47 Failed at 21: "${cidir}/install_kata_kernel.sh"
```

saschagrunert:
Awesome, it works. Thanks for all the support guys! :)

chavafg:
thanks @saschagrunert

PR #1780, opened 2019-07-02
ci: enable fastpath CI skip/return for non-code-ish PRs

grahamwhaley:
If a PR has only modified docs et. al., and no code files, then do not run the full test suite.
Do still run the static checks if required.

### DIFF
--- a/.ci/ci-fast-return.sh
+++ b/.ci/ci-fast-return.sh
@@ -1,0 +1,426 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Description: Perform CI 'fast path return' checks.
+# Returns: 0 for 'fast return OK', 1 for 'do not fastpath'.
+#
+# Checks for:
+#  - file name patterns from the YAML that will force the CI to run
+#  - file name patterns from the YAML to see if we can skip all files in the PR
+#  - If potentially skipping, looks for a 'force' label on the PR to force the CI
+#    to run anyway
+
+set -e
+
+[ -n "$DEBUG" ] && set -x
+
+script_name="${0##*/}"
+script_dir_base=".ci"
+script_gitname="${script_dir_base}/${script_name}"
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+# If no branch specified, compare against the master.
+# The 'branch' var is required by the get_pr() lib functions.
+branch=${branch:-master}
+
+# The YAML file containing our filename match patterns.
+yqfile_rootname="ci-fast-return.yaml"
+yqfile="${cidir}/${yqfile_rootname}"
+yqfile_gitname="${script_dir_base}/${yqfile_rootname}"
+
+# The YAML file containing our unit test patterns
+unit_yamlfile_rootname="ci-fast-return/test.yaml"
+unit_yamlfile="${cidir}/${unit_yamlfile_rootname}"
+unit_yamlfile_gitname="${script_dir_base}/${unit_yamlfile_rootname}"
+
+# Name of the label, that if set on a PR, will force the CI to be run anyway.
+force_label="force-ci"
+
+# The list of files to check is held in a global, to make writing the unit test
+# code easier - otherwise we would have to pass two multi-entry lists (the list
+# of expressions and the list of filenames) to a single test function, which is
+# tricky.
+filenames=""
+
+# We have a local info func, as many of our funcs return their answers via stdout,
+# and we don't want to either open code a redirect all over the code nor send to
+# the standard stdout (would corrupt our return strings) or stderr (some CIs would
+# take that as a failure case). So, use another file descriptor, mapped back to
+# stdout, that we set up previously.
+local_info() {
+	msg="$*"
+	info "$msg" >&5
+}
+
+# Read our patterns from the YAML file.
+# Arguments:
+# $1 - the yaml file path
+# $2 - the yaml (yq) path to the patterns
+#
+# Returns on stdout
+#  the patterns found, or "" if no patterns (it will translate the yq 'null' return
+#  to the empty string)
+#
+read_yaml() {
+	res=$(yq read "$1" "$2")
+	[ "$res" == "null" ] && res=""
+	echo $res
+	return 0
+}
+
+# Check if any files in ${filenames} match the egrep command line expressions
+# passed in as arguments. Note, the arguments must be formatted *as passed*
+# to egrep. Thus, a single expression can just be passed on its own, but multiple
+# expressions must individually be prefixed with '-e' or '--regexp='.
+# Returns on stdout any files that match the patterns.
+check_matches() {
+	egrep $@ <<< "$filenames" || true
+}
+
+# Check if any files in ${filenames} *DO NOT* match the egrep command line expressions
+# passed in as arguments. Note, the arguments must be formatted *as passed*
+# to egrep. Thus, a single expression can just be passed on its own, but multiple
+# expressions must individually be prefixed with '-e' or '--regexp='.
+# Returns on stdout any files that *DO NOT* match the patterns.
+check_not_matches() {
+	egrep -v $@ <<< "$filenames" || true
+}
+
+# Check if any files changed by the PR either force the CI to run or if
+# we can skip all the files in the PR.
+#
+# Returns stdout string:
+#  0 - all files are fastpath - yes, we can skip.
+#  1 - at least one non-fastpath file found - cannot skip
+can_we_skip() {
+	# The branch is the baseline - ignore it.
+	if [ "$specific_branch" = "true" ]; then
+		local_info "Skip baseline branch"
+		echo "0"
+		return 0
+	fi
+
+	filenames=$(get_pr_changed_file_details || true)
+	# Strip off the leading status - just grab last column.
+	filenames=$(echo "$filenames"|awk '{print $NF}')
+
+	# no files were changed - I guess we can skip the CI then?
+	if [ -z "$filenames" ]; then
+		local_info "No files found"
+		echo "0"
+		return 0
+	fi
+
+	# Check to see if this file or any of its deps changed, as then we should
+	# run our own internal unit tests.
+	check_for_self_test
+
+	# Get our common patterns we check against all repos.
+	local common_skip=$(read_yaml ${yqfile} common.skip_patterns)
+	local common_check=$(read_yaml ${yqfile} common.check_patterns)
+
+	# Use just the repo name itself. The YAML does not like having the full
+	# repo path in it.
+	local repo="${kata_repo##*/}"
+
+	if [ -n "$repo" ]; then
+		# Get our repo specific patterns
+		local repo_skip=$(read_yaml ${yqfile} ${repo}.skip_patterns)
+		local repo_check=$(read_yaml ${yqfile} ${repo}.check_patterns)
+	else
+		local_info "No repo set, skipping repo specific patterns"
+	fi
+
+	local canskip_exprs=""
+	for x in $common_skip $repo_skip; do
+		# Build up the string of "-e exp1 -e exp2" arguments to later pass
+		# to the egrep based search.
+		canskip_exprs+=$(echo "-e $x ")
+	done
+
+	local mustcheck_exprs=""
+	for x in $common_check $repo_check; do
+		# Build up the string of "-e exp1 -e exp2" arguments to later pass
+		# to the egrep based search.
+		mustcheck_exprs+=$(echo "-e $x ")
+	done
+
+	local_info "Skip patterns: [$canskip_exprs]"
+	local_info "Check patterns: [$mustcheck_exprs]"
+
+	# do we have any patterns to check?
+	if [ -n "$mustcheck_exprs" ]; then
+		local need_checking=$(check_matches ${mustcheck_exprs[@]})
+	else
+		local_info "No force CI check patterns"
+		local need_checking=""
+	fi
+
+	# If we have any files that *must* be checked, then immediately return
+	if [ -n "$need_checking" ]; then
+		# stderr, so it does not get in our return value...
+		cat >&2 <<-EOT
+		INFO: Cannot fastpath skip CI.
+		INFO: Some files present must be CI checked.
+		INFO: Files to check are:
+
+		$need_checking
+
+EOT
+		echo "1"
+		return 0
+	else
+		local_info "No force check files found"
+	fi
+
+	# Now we have checked there are no 'must check' files, we can check to see
+	# if all files fall into the 'can skip' patterns.
+	# first, do we have any skip patterns to search for?
+	if [ -n "$canskip_exprs" ]; then
+		local non_skippable=$(check_not_matches ${canskip_exprs[@]} <<< "$filenames")
+	else
+		local_info "No skip CI check patterns"
+		# No patterns, set non-skippable list to all files then...
+		local non_skippable="$filenames"
+	fi
+
+	if [ -z "$non_skippable" ]; then
+		# stderr, so it does not get in our return value...
+		cat >&2 <<-EOT
+		INFO: No files to check in CI.
+		INFO: Fastpath short circuit returning from CI.
+		INFO: Files skipped are:
+
+		$filenames
+
+EOT
+		echo "0"
+		return 0
+	else
+		cat >&2 <<-EOT
+		INFO: Not all files skippable
+
+		$non_skippable
+
+EOT
+		echo "1"
+		return 0
+	fi
+}
+
+# Check if we have the 'magic label' that forces a CI run set on the PR
+# Returns on stdout as string:
+#  0 - No label found, could skip the CI
+#  1 - Label found - should run the CI
+check_force_label() {
+	if [ -z "$ghprbGhRepository" ]; then
+		local_info "No ghprbGhRepository set, skip label check"
+		echo "0"
+		return 0
+	fi
+
+	if [ -z "$ghprbPullId" ]; then
+		local_info "No ghprbPullId set, skip label check"
+		echo "0"
+		return 0
+	fi
+
+	local_info "Checking labels for PR ${ghprbGhRepository}/${ghprbPullId}"
+
+	# Pull the label list for the PR
+	# Ideally we'd use a github auth token here so we don't get rate limited, but to do that we would
+	# have to expose the token into the CI scripts, which is then potentially a security hole.
+	local json=$(curl -sL https://api.github.com/repos/${ghprbGhRepository}/issues/${ghprbPullId}/labels)
+
+	# Pull the label list out
+	local labels=$(jq .[].name <<< $json)
+
+	# Check if we have the forcing label set
+	for x in $labels; do
+		# Strip off any surrounding '"'s
+		y=$(sed 's/"//g' <<< $x)
+
+		if [ "$y" == "$force_label" ]; then
+			local_info "label found, forcing CI"
+			echo "1"
+			return 0
+		fi
+	done
+
+	local_info "No forcing label found"
+	echo "0"
+	return 0
+}
+
+# Unit tests. Check the YAML file reader functions work as expected.
+testYAMLreader() {
+	repos=()
+	patterns=()
+	results=()
+
+	# Check we can read 'common' patterns, and they act as we expect.
+	repos+=("common")
+	patterns+=("skip_patterns")
+	results+=("^CODEOWNERS$ .*\.md")
+
+	repos+=("common")
+	patterns+=("check_patterns")
+	results+=("check1 .*check2$")
+
+	# Check a pattern with no list does not fault.
+	repos+=("common")
+	patterns+=("empty_patterns")
+	results+=("")
+
+	# Check we do not fault looking for a pattern that is not defined.
+	repos+=("common")
+	patterns+=("undefined_patterns")
+	results+=("")
+
+	# Check we can read a named repo.
+	# Check single line expression entries work.
+	repos+=("documentation")
+	patterns+=("doc_single_entry")
+	results+=("single_pattern")
+
+	# Check we can handle multiple complex regexps.
+	repos+=("documentation")
+	patterns+=("doc_complex_patterns")
+	results+=("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$ ^#?([a-f0-9]{6}|[a-f0-9]{3})$")
+
+	# Check we do not fail if trying to assess a repo that is not defined.
+	repos+=("nonrepo")
+	patterns+=("nonrepo_pattern")
+	results+=("")
+
+	count=0
+	for x in ${repos[@]}; do
+		echo " $count: Testing $x:${patterns[$count]} == '${results[$count]}'"
+
+		res=$(read_yaml "$unit_yamlfile" "${x}.${patterns[$count]}")
+		assertEquals "${results[$count]}" "$res"
+
+		count=$(( count+1 ))
+	done
+}
+
+# Unit test: Check the pattern matcher functions work as expected.
+testPatternMatcher() {
+	filenames="fred
+john
+mark
+doc.md"
+
+	matches=$(check_matches '.*\.md')
+	assertEquals "doc.md" "$matches"
+
+	# Check it also works with `-e` multiple expressions
+	matches=$(check_matches '-e notaname -e .*\.md')
+	assertEquals "doc.md" "$matches"
+
+	matches=$(check_matches 'mark')
+	assertEquals "mark" "$matches"
+
+	matches=$(check_matches '^.*$')
+	assertEquals "$filenames" "$matches"
+
+	matches=$(check_not_matches '.*\.md')
+	assertEquals "fred
+john
+mark" "$matches"
+
+	# check with multi patterns
+	matches=$(check_not_matches '-e .*\.md -e john')
+	assertEquals "fred
+mark" "$matches"
+
+
+	matches=$(check_not_matches 'mark')
+	assertEquals "fred
+john
+doc.md" "$matches"
+
+	matches=$(check_not_matches '^.*$')
+	assertEquals "" "$matches"
+
+}
+
+# Check if any of our own files have changed in this PR, and if so, run our own
+# unit tests...
+check_for_self_test() {
+	local ourfiles=""
+
+	ourfiles+="-e ^${script_gitname}$ "
+	ourfiles+="-e ^${yqfile_gitname}$ "
+	ourfiles+="-e ^${unit_yamlfile_gitname}$ "
+
+	res=$(check_matches $ourfiles)
+
+	if [ -n "$res" ]; then
+		local_info "file(s) [$res] modified - self running unit tests"
+		# Need to push the unit test output via the stdout mapped file descriptor
+		# so we don't corrupt our actual test function return values.
+		${cidir}/${script_name} test >&5
+	fi
+}
+
+# Run our self tests. Tests are written using the
+# github.com/kward/shunit2 library, and are encoded into functions starting
+# with the string 'test'.
+self_test() {
+	local shunit2_path="github.com/kward/shunit2"
+	local_info "Running self tests"
+
+	local_info "Go get unit test framework from ${shunit2_path}"
+	go get -d "${shunit2_path}" || true
+	local_info "Run the unit tests"
+	# Sourcing the `shunit2` file automatically runs the unit tests in this file.
+	. "${GOPATH}/src/${shunit2_path}/shunit2"
+	# shunit2 call does not return - it exits with its return code.
+}
+
+help()
+{
+	cat <<EOT
+Usage: ${script_name} [test]
+
+This script will check if the CI system needs to be run, according
+to the egrep patterns found in the file ${yqfile}
+
+Passing the argument 'test' to this script will cause it to only
+run its self tests.
+EOT
+
+	exit 0
+}
+
+main() {
+
+	# Some of our sub-funcs return their results on stdout, but we also want them to be
+	# able to log INFO messages. But, we don't want those going to stderr, as that may
+	# be seen by some CIs as an actual error. Create another file descriptor, mapped
+	# back to stdout, for us to send INFO messages to...
+	exec 5>&1
+
+	if [ "$1" == "test" ]; then
+		self_test
+		# self_test func does not return
+	fi
+
+	[ $# -gt 0 ] && help
+
+	info "Checking for any changed files that will prevent CI fastpath return"
+	res=$(can_we_skip)
+
+	# If the file check says we can skip, check the labels to see if we are forcing the
+	# CI run anyway.
+	[ $res -eq 0 ] && res=$(check_force_label)
+	exit $res
+}
+
+main "$@"
--- a/.ci/ci-fast-return.yaml
+++ b/.ci/ci-fast-return.yaml
@@ -1,0 +1,43 @@
+---
+
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# List 'egrep' patterns of files that can:
+# - be skipped by the CI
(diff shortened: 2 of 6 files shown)


grahamwhaley:
I'm mostly expecting this PR to explode, as it is moderately hard to test locally.
/cc @jodh-intel as I'd like your grep/bash-ish :eyes: to peek over it - and I did liberate some of your existing code for this PR.
/cc @chavafg as this is CI :-)
/cc @egernst , as you really really want this.

Note, I did *not* add the 'skip-ci' label check to this code, as last time we discussed we deemed it a potentially easy loophole for anybody to skip the CI. If we want that in, let me know.
If we know of more file patterns we want to add to the list, let me know.

grahamwhaley:
/test-fedora
as we need to see how it runs - so let's just trigger one slave...

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,77 @@
...
+
+	# no files were changed - I guess we can skip the CI then?
+	[ -z "$files" ] && info "No files found" && echo "1" && return
+
+	# which egrep filename patterns do *NOT* need to be CI checked.
+	local pats=""
+	pats+="-e *.\.md "
+	pats+="-e ^CODEOWNERS$ "
+	pats+="-e ^LICENSE$ "
+
```
chavafg: maybe add `*.txt`, `.travis.yml` and `VERSION` ?

grahamwhaley: `*.txt`. Yeah, I see no harm with that.
`.travis.yml`, I considered, but thought 'but that is modifying the CI', but of course, it is not modifying Jenkins, so sure!
`VERSION`. ah, that'll be one we bump a log on releases - OK, np... makes sense.

.ci/jenkins_job_build.sh
```diff
@@ -107,6 +107,16 @@ then
 	fi
 fi
 
+# Check if we can fastpath return/skip the CI
+# Specifically do this **after** we have potentially done the static
+# checks, as we always want to run those.
+# Work around the 'set -e' dying if the check fails.
+{local ret; .ci/ci-fast-return.sh; ret=$?; } || true
+if [ "$ret" -eq 1 ] then
```
GabyCT: missing ;

grahamwhaley: \0/ good catch ;-)

grahamwhaley:
/test-fedora
added more items to the list
sorted to pattern list somewhat, to appease the 'sorted list' gods
added the missing `;`...

.ci/jenkins_job_build.sh
```diff
@@ -107,6 +107,16 @@ then
 	fi
 fi
 
+# Check if we can fastpath return/skip the CI
+# Specifically do this **after** we have potentially done the static
+# checks, as we always want to run those.
+# Work around the 'set -e' dying if the check fails.
+{local ret; .ci/ci-fast-return.sh; ret=$?; } || true
```
jodh-intel: Shouldn't the `local ret` be *outside* this block since you are checking it's value outside of it?

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2017-2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Description: Perform CI 'fast path return' checks.
+# Returns: 1 for 'fast return OK', 0 for 'do not fastpath'.
```
jodh-intel: A minor point, but the name of the script suggests to me that "success" means fast path. But in that case, `0` should mean "yes, fast path" and non-zero should therefore mean "nope, run the whole CI".

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,80 @@
...
+	files=$(echo "$files"|awk '{print $NF}')
+
+	# no files were changed - I guess we can skip the CI then?
+	[ -z "$files" ] && info "No files found" && echo "1" && return
+
+	# which egrep filename patterns do *NOT* need to be CI checked.
+	local pats=""
+	pats+="-e ^CODEOWNERS$ "
+	pats+="-e ^LICENSE$ "
+	pats+="-e *.\.md "
```
jodh-intel: As mentioned on https://github.com/kata-containers/ci/issues/192, normally docs can be ignored. But not if they are https://github.com/kata-containers/documentation/tree/master/install as those *must* be CI checked. Could you add a rule for those here?

.ci/jenkins_job_build.sh
grahamwhaley: I did wonder that, when I lifted the code from:
https://github.com/kata-containers/tests/blob/master/.ci/kata-doc-to-script.sh#L186-L187
which, I think you wrote recently? Let's work out how best to get a retval without running into the `set -e`... I can't use the echo trick I used at https://github.com/kata-containers/tests/blob/master/lib/common.bash#L86, as the sub script being called generates output :-(. iirc, I did do some subshell thing like `$(set +e; xxx.sh)` before somewhere else... yeah, sometimes `set -e` makes our lives just a little more painful..

jodh-intel: I'm allowed to critique my own code then right? :smile:

Having checked the docs, I think this is correct (since `local` is for defining a function-local variable not a block-local one) but still looks odd.

grahamwhaley: absolutely :-)  heh. Yep, looked odd when I copied it - but I trusted you'd done the right thing when you wrote it :-)  I'm more than happy to add a comment explaining how the oddness works... more than happy.

grahamwhaley:
OK, bit of a rewrite as discussed on https://github.com/kata-containers/ci/issues/192
Now:
 - filter patterns are stored in a yaml file
 - we have patterns both for 'can skip these', and 'must test these'
 - if we end up with 'can skip the CI on this PR', we then check the labels on the PR for a 'force-ci' label, and run the CI anyway if that is found.

Post this, I guess we will also want to update/add some docs over in the CI repo.

@jodh-intel @chavafg ptal
/test-fedora

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+
+# Check if any files changed by the PR either force the CI to run or if
+# we can skip all the files in the PR.
+#
+# Return:
+#  0 - all files are fastpath
+#  1 - at least one non-fastpath file found
+can_we_skip() {
+	# The branch is the baseline - ignore it.
+	[ "$specific_branch" = "true" ] && info "Skip basline branch" >&2 && echo "0" && return
```
jodh-intel: Typo: "baseline". Although I do like "one-liners" like this, I think due to all the ampersands this would be clearer as a full `if ... then` test. Also, since this is a `set -e` script, it's always a good idea to return an explicit value as a "bare `return`" can be surprising since it will return the value of the last command executed.

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+
+	files=$(get_pr_changed_file_details || true)
+	# Strip off the leading status - just grab last column.
+	files=$(echo "$files"|awk '{print $NF}')
+
+	# no files were changed - I guess we can skip the CI then?
+	[ -z "$files" ] && info "No files found" >&2 && echo "0" && return
+
+	local common_skip=$(yq read ${yqfile} common.skip_patterns)
+	# If nothing is found, yq returns the word 'null' (annoyingley ;-) )
```
jodh-intel: - Typo: "annoyingly" annoyingly ;)
- Yeah, that `null` is a pita.

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+	[ "$common_skip" == "null" ] && common_skip=""
+	local common_check=$(yq read ${yqfile} common.check_patterns)
```
jodh-intel: Could you space these lines out a bit - they are pretty tight and a few blank lines would make this easier to read I think.

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+		[ "$repo_skip" == "null" ] && repo_skip=""
+		local repo_check=$(yq read ${yqfile} ${repo}.check_patterns)
+		[ "$repo_check" == "null" ] && repo_check=""
+	else
+		info "No repo set, skipping repo specific patterns" >&2
+	fi
+
+	local canskip=""
+	for x in $common_skip $repo_skip; do
+		canskip+=$(echo "-e $x ")
```
jodh-intel: Nit: I'd put the arg outside the string here and below:

```sh
echo -e "$x "
```

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+		return
+	fi
+}
+
+# Check if we have the 'magic label' that forces a CI run set on the PR
+# Return:
+#  0 - No label found, could skip the CI
+#  1 - Label found - should run the CI
+check_force_label() {
+
```
jodh-intel: Nit: extraneous blank line.

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+	[ -z "$ghprbGhRepository" ] && info "No ghprbGhRepository set, skip label check" >&2 && echo "0" && return
+	[ -z "$ghprbPullId" ] && info "No ghprbPullId set, skip label check" >&2 && echo "0" && return
+
+	info "Checking labels for PR ${ghprbGhRepository}/${ghprbPullId}" >&2
+
+	# Pull the label list for the PR
+	# Ideally we'd use a github auth token here so we don't get rate limited, but to do that we would
+	# have to expose the token into the CI scripts, which is then potentially a security hole.
+	# json=$(curl -H "Authorization: Basic ${GITHUB_API_TOKEN}" https://api.github.com/repos/${ghprbGhRepository}/issues/${ghprbPullId}/labels)
+	local json=$(curl -s https://api.github.com/repos/${ghprbGhRepository}/issues/${ghprbPullId}/labels)
```
jodh-intel: `curl -sL` "just in case"?

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+
+	[ -z "$ghprbGhRepository" ] && info "No ghprbGhRepository set, skip label check" >&2 && echo "0" && return
+	[ -z "$ghprbPullId" ] && info "No ghprbPullId set, skip label check" >&2 && echo "0" && return
+
+	info "Checking labels for PR ${ghprbGhRepository}/${ghprbPullId}" >&2
+
+	# Pull the label list for the PR
+	# Ideally we'd use a github auth token here so we don't get rate limited, but to do that we would
+	# have to expose the token into the CI scripts, which is then potentially a security hole.
+	# json=$(curl -H "Authorization: Basic ${GITHUB_API_TOKEN}" https://api.github.com/repos/${ghprbGhRepository}/issues/${ghprbPullId}/labels)
```
jodh-intel: I'd remove this and maybe slap it into a feature issue in case we find a way to do this later.

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+	# Use just the repo name itself. The YAML does not like having the full
+	# repo path in it.
+	local repo="${kata_repo##*/}"
+	if [ -n "$repo" ]; then
+		local repo_skip=$(yq read ${yqfile} ${repo}.skip_patterns)
+		[ "$repo_skip" == "null" ] && repo_skip=""
+		local repo_check=$(yq read ${yqfile} ${repo}.check_patterns)
+		[ "$repo_check" == "null" ] && repo_check=""
+	else
+		info "No repo set, skipping repo specific patterns" >&2
```
jodh-intel: Why redirected to `stderr`? Is this a warning? If so, I'd add a `warn()` that writes to `stderr` to minimise the smattering of redirects.

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,183 @@
...
+}
+
+main()
+{
+	info "Checking for any changed files that will prevent CI fastpath return"
+	res=$(can_we_skip)
+
+	# If the file check says we can skip, check the labels to see if we are forcing the
+	# CI run anyway.
+	[ $res == 0 ] && res=$(check_force_label)
```
jodh-intel: Nit: I know this works, but I think this should be using `-eq` for comparing numerics?

.ci/ci-fast-return.yaml
```diff
@@ -0,0 +1,44 @@
...
+# for '\t') in many YAML string types (which in itself I think is another bug),
+# and explodes. Work around that by placing our pattern strings into a
+# ['literal block scalar'](https://yaml.org/spec/1.2/spec.html#id2793652),
+# that seems to work.
+
+common:
+        skip_patterns: |
+                ^CODEOWNERS$
+                ^LICENSE$
+                *.\.md
```
jodh-intel: Shouldn't these regex's be `.*\.md`, *etc*?:

```sh
$ echo "hello.md"|egrep -o --color "*.\.md
o.md
$ echo "hello.md"|egrep -o --color ".*\.md"
hello.md
```

grahamwhaley:
/me goes to fix...
```
00:00:28.581 .ci/jenkins_job_build.sh: line 114: syntax error near unexpected token `}'
```

.ci/ci-fast-return.sh
grahamwhaley: the `-e` is not for the echo, it is for the subsequent `egrep` that consumes the string ;-)

jodh-intel: Err, that's confusing given both `echo` and `egrep` take a `-e`. Can't you drop the `-e` from here and re-add when you actually call `egrep`?

grahamwhaley:
Oh, I'm going to have to stare at the magic:

```
{local ret; .ci/ci-fast-return.sh; ret=$?; } || true
```

now. Seeing as nominally I grabbed that from some working code elsewhere, I wonder if it is because we are not in a function.

.ci/ci-fast-return.sh
grahamwhaley: Not really, it is a string of expressions, so you need `-e x -e y -e z` - which is harder to do later - unless there is a little bit of shell magic to prepend every entry in a bash array with a string...
I dunno - the code 'does exactly what it says' - just, you initially read it as not doing exactly what it says :-)

jodh-intel: How about you say `--regexp=` rather than `-e`? That way it's obvious it has nothing to do with `echo`. Maybe a comment would help other readers too ;)

Side issue: please  could you not resolve issues until we've all agreed they are resolved too? Pretty please?

grahamwhaley:
> Thanks @grahamwhaley. This is coming along nicely.
> 
> Given the importance and sensitivity of this feature, I think it would make sense to create a few unit tests for this comprising:

Err. hmm. Heh, I guess you are in 'unit test' mode....
So, I think this would be the first set of unit tests we get for any shell script we have... but, that's no excuse not to I guess.
 
>     * a fake set of changed filenames.
> 
>     * a minimal set of YAML files
> 

Yeah, I think it will need the addition of a bunch of shell-override var sugar in the script. I'll see what can be done...

>     * a shell script (could be `bats` but not necessarily imho) containing a bunch of assertions to prove that we can skip the CI for doc PRs (but only install docs), etc.

But my major query is - where and when do we think we'd run the unit tests? afaik, the only other unit tests we have for the CI and other framework code is in the Go-coded `tests/cmd/*` commands, and they only get executed when we build/install those binaries as part of the CI slave setup...
I'm a little hesitant  to add the running of these tests to every CI run. The funky case would be that the script itself spots that it or its yaml has changed, and runs its own tests... wdyt??

jodh-intel:
> I think this would be the first set of unit tests we get for any shell script we have

Actually, no - see https://github.com/kata-containers/osbuilder/blob/master/tests/test_images.sh which tests osbuilder (shell) :smile:

> But my major query is - where and when do we think we'd run the unit tests? 

Maybe we could run them on the "weekly master CI run" when that gets added? Any other ideas @chavafg?

grahamwhaley:
@jodh-intel @chavafg - given I really don't like [BATS](https://github.com/sstephenson/bats), and I could not see a nice way to add BATS tests into the script itself without afaict converting the whole script into a BATS file, I had a look around for other bash unit testing frameworks, and found [shunit2](https://github.com/kward/shunit2).

Using that, I added the following code to the script itself, and a little bit of main arg parsing, so now you can run `./ci-fast-path.sh test`, and it will self test. wdyt, acceptible to use another bash test framework?

```bash
testYAMLreader() {
        local sourcefile="${cidir}/ci-fast-return/test.yaml"

        repos=()
        patterns=()
        results=()

        repos+=("common")
        patterns+=("skip_patterns")
        results+=("^CODEOWNERS$ .*\.md")

        repos+=("common")
        patterns+=("check_patterns")
        results+=("check1 .*check2$")

        repos+=("common")
        patterns+=("empty_patterns")
        results+=("")

        repos+=("common")
        patterns+=("undefined_patterns")
        results+=("")

        repos+=("documentation")
        patterns+=("doc_single_entry")
        results+=("single_pattern")

        repos+=("documentation")
        patterns+=("doc_complex_patterns")
        results+=("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$ ^#?([a-f0-9]{6}|[a-f0-9]{3})$")

        count=0
        for x in ${repos[@]}; do
                echo "$count: Testing $x:${patterns[$count]} == '${results[$count]}'"

                res=$(read_yaml "$sourcefile" "${x}.${patterns[$count]}")
                assertEquals "${results[$count]}" "$res"

                count=$(( count+1 ))
        done
}

testPatternMatcher() {
        filenames="fred
john
mark
doc.md"

        matches=$(check_matches '.*\.md')
        assertEquals "$matches" "doc.md"

        matches=$(check_matches 'mark')
        assertEquals "$matches" "mark"

        matches=$(check_matches '^.*$')
        assertEquals "$matches" "$filenames"

        matches=$(check_not_matches '.*\.md')
        assertEquals "$matches" "fred
john
mark"

        matches=$(check_not_matches 'mark')
        assertEquals "$matches" "fred
john
doc.md"

        matches=$(check_not_matches '^.*$')
        assertEquals "$matches" ""

}

self_test() {
        info "Running self tests"

        info "Go get unit test framework"
        go get -d github.com/kward/shunit2 || true
        info "Source the framework"
        . ${GOPATH}/src/github.com/kward/shunit2/shunit2

        return 0
}
```

with the YAML file

```yaml
common:
        skip_patterns: |
                ^CODEOWNERS$
                .*\.md
        check_patterns: |
                check1
                .*check2$
        empty_patterns:
        #undefined_patterns

documentation:
        doc_single_entry: |
                single_pattern
        doc_complex_patterns: |
                ^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$
                ^#?([a-f0-9]{6}|[a-f0-9]{3})$

```

And then running gives us:

```
$ ./ci-fast-return.sh test
INFO: Running self tests
INFO: Go get unit test framework
can't load package: package github.com/kward/shunit2: no Go files in /home/gwhaley/go/src/github.com/kward/shunit2
INFO: Source the framework
testEquality
testYAMLreader
0: Testing common:skip_patterns == '^CODEOWNERS$ .*\.md'
1: Testing common:check_patterns == 'check1 .*check2$'
2: Testing common:empty_patterns == ''
3: Testing common:undefined_patterns == ''
4: Testing documentation:doc_single_entry == 'single_pattern'
5: Testing documentation:doc_complex_patterns == '^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$ ^#?([a-f0-9]{6}|[a-f0-9]{3})$'
testPatternMatcher

Ran 3 tests.

OK
```

jodh-intel:
@grahamwhaley - nice! wfm! Yeah, I'm not a fan of BATS either as it's a pita to debug failing tests.

.ci/ci-fast-return.sh
jodh-intel: The problem there might be that certain CI's will auto-fail if they get any output on `stderr`. What you could do is this instead:

```sh
foo()
{
    local msg="$*"
    info "$msg" >/dev/tty
}

foo "stuff happened"
foo "more exciting things occurred"
```

jodh-intel: But obviously choose a better name than `foo()` :smile:

grahamwhaley: And I was just about to push an update...
hmm, so, I added the 'warn()' call/func in that you requested so we *did* get something on stderr.... I'll probably revert that back to a `foo()` then.

I reckon with the `{ do_stuff; ret=$?; } || true` idiom, all of this echo stuff to get around `set -e` hurting us is maybe avoidable ... but, that is also a bunch more rewrite work. I'll s/info/foo/g , retry, and push rsn then

grahamwhaley:
re-pushed. Changes are:
- unit tests added
  - some internal functions broken down to aid testing
- unit tests auto detection added
  - if the PR changes any of the main files in this PR, it will auto run its own unit tests
- output now into /dev/tty - which whilst avoiding the possibility of a CI seeing stderr data as an error, it does mean the full output of the script is not not capturable.

Good luck - the script has grown to >400 lines...

grahamwhaley:
Logs from fedora CI look promising - but @jodh-intel , I fear we will have to find another way around the `tty` item - remember, some CIs run without a tty at all...

```
00:00:29.791 INFO: Checking for any changed files that will prevent CI fastpath return
00:00:29.802 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
00:00:29.802 .ci/ci-fast-return.sh: line 368: /dev/tty: No such device or address
00:00:29.818 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
00:00:29.819 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
00:00:29.819 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
00:00:29.819 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
00:00:29.823 INFO: Not all files skippable
00:00:29.823 
00:00:29.823 .ci/ci-fast-return.sh
00:00:29.823 .ci/ci-fast-return.yaml
00:00:29.823 .ci/ci-fast-return/test.yaml
00:00:29.823 .ci/jenkins_job_build.sh
00:00:29.823 .ci/lib.sh
00:00:29.823 .ci/static-checks.sh
```

chavafg:
Seems that this is working correctly, although the `local_info` messages are not getting logged by jenkins since it does not have a tty. Can we live without those?
```
11:06:46 INFO: Checking for any changed files that will prevent CI fastpath return
11:06:46 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
11:06:46 .ci/ci-fast-return.sh: line 368: /dev/tty: No such device or address
11:06:46 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
11:06:46 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
11:06:46 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
11:06:46 .ci/ci-fast-return.sh: line 57: /dev/tty: No such device or address
11:06:46 INFO: Not all files skippable
```

jodh-intel:
@grahamwhaley

> some CIs run without a tty at all.

Yep - seems like we both forgot about that ;)

It's a shame as writing to the tty is the "correct" solution generally; writing to stderr is "incorrect" under some CI environments since it's obviously the error stream.

I think this leaves us with the following options:

- Don't log (ugh).
- Log to a file which we display on error (via a `trap` function).
- Create a new fd for logging as that won't interfere with those functions that return data via `stdout`:
  ```sh
  # create a new fd for log messages
  exec 5>&1

  foo()
  {
      echo >&5 "blah blah not returned"

      # return value
      echo "result"

      echo >&5 "also not returned"
  }

  # contains "result" only
  result=$(foo)
  ```

  But put the `>&5` logic into `local_info()`.

grahamwhaley:
@jodh-intel - or, I wondered if we can leave the `ci-fast-path.sh` script writing to stderr, but in the top level invocation `jenkins` script, use a `2>&1` to dump all the final output to stdout... I think that will appease the CIs, whilst also letting the test script itself separate stdout/err - wdyt?

grahamwhaley:
oh, happy with the dup fd as well (a rarely used bash-ism ;-) )

jodh-intel:
I guess either option works. Can we ensure there is a juicy comment explaining why we're doing what we're doing ;) Which option would you prefer @chavafg ?

chavafg:
hmm, I really do not have a strong preference, options above wfm as long as we can log the actions of the script. @grahamwhaley you'll have last word 😄

grahamwhaley:
I went with the `>&5` workaround - as that contains the fix to just the one file, and should work for any future situations, we hope.
/test-fedora

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,425 @@
...
+#  - file name patterns from the YAML to see if we can skip all files in the PR
+#  - If potentially skipping, looks for a 'force' label on the PR to force the CI
+#    to run anyway
+
+set -e
+
+[ -n "$DEBUG" ] && set -x
+
+script_name="${0##*/}"
+script_gitname=".ci/${script_name}"
```
jodh-intel: Hard-coded path. I think you could do something like this for all the `*_gitname` vars that should continue to work even if this file gets moved to a different sub-dir:

```sh
script_gitname=$(git ls-files --full-name ${script_name}")
```

Or this would be safer:

```sh
script_gitname=$(git ls-files --full-name | grep "/${script_name}$")
```

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,425 @@
...
+# github.com/kward/shunit2 library, and are encoded into functions starting
+# with the string 'test'.
+self_test() {
+	local_info "Running self tests"
+
+	local_info "Go get unit test framework"
+	go get -d github.com/kward/shunit2 || true
+	local_info "Run the unit tests"
+	# Sourcing the `shunit2` file automatically runs the unit tests in this file.
+	. ${GOPATH}/src/github.com/kward/shunit2/shunit2
```
jodh-intel: You could use a variable to avoid the repetition of `github.com/kward/shunit2`. A couple of other thoughts:

- I guess we always have go available but you could `curl` directly.
- This could be installed as a package instead by updating `.ci/setup_env_*.sh`.

.ci/ci-fast-return.sh
grahamwhaley: To do that I think we have to ensure we are inside the repo where the file lives (it does not work if you are not in the git tree of the repo you are examining afaict), and we need to full path inside the repo to where the file we are looking for is (if you just try `ci-fast-path.sh` at the top level of the repo, you get no name). So, I think it is a bit catch22 - you have to know what path you are looking in in order to get its, err, path.
Thus, I think the hard wiring of the '.ci' subdir has to happen soemwhere, and doing it where it is right now I think is as good as anywhere. Unless you can show me some code that is going to work no matter where/how the script is invoked?

jodh-intel: That's a good point and it may be worth adding a comment / note in the commit explaining where we expect this script to be run from. Well, if we can't use `basename $cidir`, how about a top-level variable set to `.ci/` along with a juicy variable? That atleast confines the issue to a central location, rather than hard-coding `.ci/` in the three `*_gitname` variables.

.ci/ci-fast-return.sh
grahamwhaley: variable added.
The test will run so rarely, probably not worth adding it as a top level dep.

.ci/ci-fast-return.sh
grahamwhaley: basedir in a var I can do ... done

.ci/ci-fast-return.sh
grahamwhaley: done, I think.

jodh-intel:
/me grabs :popcorn: and :beer: and waits... :smile:

jodh-intel:
Just in case you hadn't noticed, this branch is (now) conflicted.

grahamwhaley:
/test
rebased to resolve conflict.

.ci/ci-fast-return.sh
grahamwhaley: hah, yep. see, comments always rot... updated....

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,427 @@
...
+
+	# Some of our sub-funcs return their results on stdout, but we also want them to be
+	# able to log INFO messages. But, we don't want those going to stderr, as that may
+	# be seen by some CIs as an actual error. Create another file descriptor, mapped
+	# back to stdout, for us to send INFO messages to...
+	exec 5>&1
+
+	if [ "$1" == "test" ]; then
+		self_test
+		# self_test func does not return
```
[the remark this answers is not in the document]
grahamwhaley: `shunit2` always ends with an `exit`, so I've removed the `return 0` and documented the fact that `shunit2` will not return.

.ci/ci-fast-return.yaml
```diff
@@ -0,0 +1,44 @@
...
+#
+# List 'egrep' patterns of files that can:
+# - be skipped by the CI
+# - force the CI to always run (cannot be skipped), even if they match the 'can
+#   be skipped' patterns.
+#
+# Entries are of the form:
+#  repo_name:
+#    [skip_pattterns | check_patterns]: |
+#     (list of patterns, ideally one per line
```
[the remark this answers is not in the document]
grahamwhaley: personally, I find the comment more readable than the yaml encoded comment. As, afaik, we are never parsing out those fields, why make them YAML?

.ci/ci-fast-return.yaml
```diff
@@ -0,0 +1,44 @@
...
+                ^CODEOWNERS$
+                ^LICENSE$
+                .*\.md
+                *\.travis\.yml
+                .*\.txt
+                ^VERSION$
+        check_patterns:
+
+documentation:
+        skip_patterns:
```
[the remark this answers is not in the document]
grahamwhaley: Nope. It can be missing (the parser handles it). I'll drop that one.

.ci/ci-fast-return/test.yaml
```diff
@@ -0,0 +1,31 @@
...
+        #undefined_patterns
+
+documentation:
+        doc_single_entry: |
+                single_pattern
+        doc_complex_patterns: |
+                ^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$
+                ^#?([a-f0-9]{6}|[a-f0-9]{3})$
+
+#nonrepo:
```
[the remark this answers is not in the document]
grahamwhaley: On purpose - the unit tests specifically look for that pattern, so they are just there as commented out to indicate they are specifically not set. I'll comment it in the YAML to make it clearer.

grahamwhaley:
/me holds off testing, just in case there is yet more feedback...

.ci/ci-fast-return.sh
```diff
@@ -0,0 +1,426 @@
...
+
+	local_info "Checking labels for PR ${ghprbGhRepository}/${ghprbPullId}"
+
+	# Pull the label list for the PR
+	# Ideally we'd use a github auth token here so we don't get rate limited, but to do that we would
+	# have to expose the token into the CI scripts, which is then potentially a security hole.
+	local json=$(curl -sL https://api.github.com/repos/${ghprbGhRepository}/issues/${ghprbPullId}/labels)
+
+	# Pull the label list out
+	local labels=$(jq .[].name <<< $json)
```
jodh-intel: Nit: Should all the variables passed to `<<<` be quoted to minimise surprises?

grahamwhaley: strangely... github passes all the labels back already enclosed in `"`s, which we then actually strip off a few lines further down in the loop before we do the string compare.... I think my head will explode if I try to contemplate what thus happens if we put a further set of `"`s around them here as well :-)

jodh-intel: We don't want :speaking_head: :boom:, so

```sh
cat <<< "lgtm ;)"
```

grahamwhaley:
A nemu CI fail:
```
[Fail] capabilities drop and add capabilities [It] cap_chown 
```

I'll retrigger that one so we can see if it is spurious or known..

Issue #1783, opened 2019-07-08
CI: Check XML files for validity


jodh-intel:
Add a check for all XML files.

PR #1784, opened 2019-07-08
CI: Add XML check

jodh-intel:
Add a static analysis check for XML files.
  
Fixes: #1783.
    
### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -519,10 +519,10 @@ static_check_docs()
 		docs_status=$(get_pr_changed_file_details || true)
 		docs_status=$(echo "$docs_status" | grep "\.md$" || true)
 
-		docs=$(echo "$docs_status" | awk '{print $NF}')
+		docs=$(echo "$docs_status" | awk '{print $NF}' | sort)
 
 		# Newly-added docs
-		new_docs=$(echo "$docs_status" | awk '/^A/ {print $NF}')
+		new_docs=$(echo "$docs_status" | awk '/^A/ {print $NF}' | sort)
 
 		for doc in $new_docs
 		do
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -68,6 +68,7 @@ long_options=(
 	[repo:]="Specify GitHub URL of repo to use (github.com/user/repo)"
 	[vendor]="Check vendor files"
 	[versions]="Check versions files"
+	[xml]="Check XML files"
 )
 
 yamllint_cmd="yamllint"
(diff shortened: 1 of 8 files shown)


.ci/static-checks.sh
```diff
@@ -838,6 +839,58 @@ static_check_vendor()
 	dep ensure -no-vendor -dry-run
 }
 
+static_check_xml()
+{
+	local all_xml
+	local files
+
+	all_xml=$(find . -name "*.xml" | grep -v "vendor/" | sort || true)
```
grahamwhaley: nitpick: don't know if `^vendor/` would be more accurate or not?

.ci/static-checks.sh
```diff
@@ -838,6 +839,58 @@ static_check_vendor()
...
+
+		# Most XML documents are specified as XML 1.0 since, with the
+		# advent of XML 1.0 (Fifth Edition), XML 1.1 is "almost
+		# redundant" due to XML 1.0 providing the majority of XML 1.1
+		# features. xmllint doesn't support XML 1.1 seemingly for this
+		# reason, so the only check we can do is to (crudely) force
+		# the document to be an XML 1.0 one since XML 1.1 documents
+		# can mostly be represented as XML 1.0.
+		#
+		# This is only really required since Jenkins creates XML 1.1
```
grahamwhaley: Ah, so, this is with the (at least partial) intent of checking the Jenkins xml files... which, whilst good, if one of them is *not* valid XML, I'm not sure there is a huge amount we could do about it - the files are generated by Jenkins when we modify the GUI input - unclear if we would/could then go and hand-fix the xml and fold it back to the Jenkins server...

.ci/static-checks.sh
```diff
@@ -838,6 +839,58 @@ static_check_vendor()
...
+		# documents.
+		contents=$(sed "s/xml version='1.1'/xml version='1.0'/g" "$file")
+
+		local ret
+
+		{ echo "$contents" | chronic xmllint -format -; ret=$?; } || true
```
grahamwhaley: nit: I think a `<<< $contents` is technically maybe more efficient than an `echo | ` :-)

.ci/static-checks.sh
jodh-intel: I'll tal at this. We have a number of vendor-filtering points in the code, all subtly different... ;)

.ci/static-checks.sh
jodh-intel: Although most of the XML files are Jenkins, there are other XML files in our repos. Regardless of who or what generated the files, we *really* do need to know if the files are valid or not. Particularly if you consider the fact that Jenkins is a core infrastructure tool and add a hypothetical disaster recovery scenario into the mix... :smile:

.ci/static-checks.sh
jodh-intel: I guess this might save a few cycles - branch updated.

jodh-intel:
@grahamwhaley - branch updated. This should do for now as a minimal change.

However, ultimately, I think the script needs a major refactor to simplify the logic.

Issue #1788, opened 2019-07-09
Add a JSON CI test


jodh-intel:
As a follow-on to #1787, check JSON files.

Issue #1787, opened 2019-07-09
Add a shell script checker CI test


jodh-intel:
Add a CI test for checking shell scripts.

We could add `shellcheck(1)` but that requires annotating scripts for things like `source`-ing files (which we do *everywhere*), so for now let's just start with `bash -n` (which would have caught https://github.com/kata-containers/tests/pull/1786).

PR #1789, opened 2019-07-09
CI: Add shell script and JSON checks to static analysis script

jodh-intel:
- CI: Add shell script static analysis check

  Add a basic CI shell checker test to catch basic syntax errors.

  Fixes: #1787.

- Add a CI check on JSON files.

  Fixes: #1788.

- Correct the handling of an incorrect CLI option in the static checks script. This is a nasty issue where the `local` keyword means that the return value of `getopt(1)` is *always* zero, even when an incorrect
option is specified.

  See: https://github.com/koalaman/shellcheck/wiki/SC2155

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -66,6 +66,7 @@ long_options=(
 	[no-arch]="Run/list all tests except architecture-specific ones"
 	[only-arch]="Only run/list architecture-specific tests"
 	[repo:]="Specify GitHub URL of repo to use (github.com/user/repo)"
+	[scripts]="Check script files"
 	[vendor]="Check vendor files"
 	[versions]="Check versions files"
 	[xml]="Check XML files"
@@ -891,6 +892,43 @@ static_check_xml()
 	done
 }
 
+static_check_shell()
+{
+	local all_scripts
+	local scripts
+
+	all_scripts=$(find . \( -name "*.sh" -o -name "*.bash" \) | grep -v "/vendor/" | sort || true)
+
+	if [ "$specific_branch" = "true" ]
+	then
+		info "Checking all scripts in $branch branch"
+		scripts="$all_scripts"
+	else
+		info "Checking local branch for changed scripts only"
+
+		local scripts_status
+		scripts_status=$(get_pr_changed_file_details || true)
+		scripts_status=$(echo "$scripts_status" | grep -E "\.(sh|bash)$" || true)
+
+		scripts=$(echo "$scripts_status" | awk '{print $NF}')
+	fi
+
+	[ -z "$scripts" ] && info "No scripts to check" && return 0
+
+	local script
+
+	for script in $scripts
+	do
+		info "Checking script file '$script'"
+
+		local ret
+
+		{ chronic bash -n "$script"; ret=$?; } || true
+
+		[ "$ret" -eq 0 ] || die "check for script '$script' failed"
+	done
+}
+
 # Run the specified function (after first checking it is compatible with the
 # users architectural preferences), or simply list the function name if list
 # mode is active.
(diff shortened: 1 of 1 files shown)


grahamwhaley:
PR title says YAML, but I think it is actually JSON ?

.ci/static-checks.sh
```diff
@@ -945,15 +945,17 @@ main()
 
 	local long_option_names="${!long_options[@]}"
 
-	local args=$(getopt \
+	local args
```
grahamwhaley: ouchy

.ci/static-checks.sh
```diff
@@ -891,6 +892,43 @@ static_check_xml()
...
+		info "Checking local branch for changed scripts only"
+
+		local scripts_status
+		scripts_status=$(get_pr_changed_file_details || true)
+		scripts_status=$(echo "$scripts_status" | grep -E "\.(sh|bash)$" || true)
+
+		scripts=$(echo "$scripts_status" | awk '{print $NF}')
+	fi
+
+	[ -z "$scripts" ] && info "No scripts to check" && return 0
```
grahamwhaley: you could move the 'no files' check next to the find - either up or down - right now they are separated by the branch check, and it feels like the two things (find files, check we have some and the branch checks) are two separate things?

jodh-intel:
Yep - I have too many branches floating about. Updated.

.ci/static-checks.sh
```diff
@@ -891,6 +892,43 @@ static_check_xml()
...
+
+	local script
+
+	for script in $scripts
+	do
+		info "Checking script file '$script'"
+
+		local ret
+
+		{ chronic bash -n "$script"; ret=$?; } || true
```
grahamwhaley: y'know, sometimes I have `man bash`, coz finding specifics in there can be so hard... but, yes, `-n`, interesting :-) the cheap and dirty shellcheck...

grahamwhaley:
> Yep - I have too many branches floating about. Updated.

presuming maybe you have a YAML check thing on the way then, that one we might need to scrutinize more. if you use `yamllint`, then I think it is by default rather over zealous, and will fail on things like `ansible` scripts and thus `zuul` scripts. I tried to appease it with a `.yamllint` file at one point, but mostly gave up, as it was a fair struggle... just a heads up.

.ci/static-checks.sh
jodh-intel: I know right! ;)

.ci/static-checks.sh
jodh-intel: Well that check needs to be there since if you look closely that branch check determines what the value of `$scripts` is.

.ci/static-checks.sh
jodh-intel: As mentioned on the issue, it's basic but doesn't require us to modify our scripts like `shellcheck` will and will find basic issues. Maybe when we have a bunch of free time (cough...), we can update the entire corpus of `~230` scripts for `shellcheck` :)

jodh-intel:
Yep - I'm holding off YAML and TOML checking for now due to exactly that. Like the shell script conversion for `shellcheck`, it's going to be a bigger job to handle those formats.

jodh-intel:
/me gets irritated by OSX causing Travis to fail. I've added a ... cough ... "optimisation" for OSX in the latest commit :smile:

jodh-intel:
Restarted vsocks CI which failed with an odd k8s error:

```
[wait-control-plane] Waiting for the kubelet to boot up the control plane as static Pods from directory "/etc/kubernetes/manifests". This can take up to 4m0s
[apiclient] All control plane components are healthy after 21.504933 seconds
[upload-config] storing the configuration used in ConfigMap "kubeadm-config" in the "kube-system" Namespace
[kubelet] Creating a ConfigMap "kubelet-config-1.14" in namespace kube-system with the configuration for the kubelets in the cluster
error execution phase upload-config/kubelet: error creating kubelet configuration ConfigMap: error creating kubelet configuration configmap RBAC rules: unable to create RBAC rolebinding: etcdserver: request timed out
Failed at 61: sudo -E kubeadm init --config "${kubeadm_config_file}"
Failed at 34: ./init.sh
Makefile:118: recipe for target 'kubernetes' failed
```

PR #1791, opened 2019-07-09
ci: Remove installation from OBS

### LINKED ISSUE #1790: Remove installation from OBS
We are not longer using the installation from the OBS repository, we need to remove it from the ci

GabyCT:
We are not longer using the installation from the OBS repository,
this will remove the unnecesary steps.

Fixes #1790

### DIFF
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -88,15 +88,6 @@ main()
 
 	[ "$setup_type" = "minimal" ] && exit 0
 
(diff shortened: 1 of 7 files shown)


grahamwhaley:
As we presumably now either pull items from the CI fragment cache if we can, or build from source if we cannot?
Does this lose us anything, like implicitly testing of OBS images?

chavafg:
I think this is fine, we are no longer installing kata assets using the package manager, so doesn't make sense to install the obs repos.

PR #1793, opened 2019-07-10
ci: fix `install_runtime.sh` and add support for using acrn config file

chavafg:
Once kata-containers/runtime#1779 is merged, we will be able
to run CI using `acrn` hypervisor. Add support for running kata
with acrn using its dedicated configuration file.

In addition, `kata-runtime kata-check` will fail if the hypervisor
in the kata runtime config file is not found, so run kata-check 
after correct config file is in place.


grahamwhaley:
/test
;-)

chavafg:
ouch, thanks @grahamwhaley

amshinde:
Rerunning failing fedora and vsocks CI.

PR #1799, opened 2019-07-10
ci: Remove KSM installation

### LINKED ISSUE #1741: Resolve KSM throttler build strategy for CI
See: https://github.com/kata-containers/tests/pull/1730/files#r294867576.

This affects the following distro-specific setup scripts:

```
.ci/setup_env_centos.sh
.ci/setup_env_debian.sh
.ci/setup_env_fedora.sh
.ci/setup_env_rhel.sh
.ci/setup_env_ubuntu.sh
```

It looks like @GabyCT wrote that code so she might have thoughts on this.

GabyCT:
Currently in our CI, we are not using the variable of KATA_KSM_THROTTLER
in any of the jobs that we have so we can remove the unnecessary steps
like the installation.

Fixes #1741

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -7,7 +7,6 @@
 #
 
 export KATA_RUNTIME=${KATA_RUNTIME:-kata-runtime}
-export KATA_KSM_THROTTLER=${KATA_KSM_THROTTLER:-no}
 export KATA_NEMU_DESTDIR=${KATA_NEMU_DESTDIR:-"/usr"}
 export KATA_QEMU_DESTDIR=${KATA_QEMU_DESTDIR:-"/usr"}
 
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -91,11 +91,6 @@ main()
 	echo "Install GNU parallel"
 	# GNU parallel not available in Centos repos, so build it instead.
 	build_install_parallel
-
-	if [ "$KATA_KSM_THROTTLER" == "yes" ]; then
-		echo "Install ${KATA_KSM_THROTTLER_JOB}"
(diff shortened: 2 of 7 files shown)


chavafg:
If we don't want to remove it, we could add a job to run a subset of tests using ksm-throttler. Not all tests since the throttler increases the boot time of the VM and it takes a lot of time to run all tests, but I'll say that this should be a weekly job or maybe a job that is run as part of the release process (although it could be late if we discover an issue)? 

This mainly because we are limited on Azure resources right now, so having extra jobs for each PR on every repo would not be the best right now unless we get more Azure credits.

jodh-intel:
That sounds like a good idea @chavafg. We'd really only need a job to ensure that:

- we can build / install the throttler.
- the throttler is correctly de-duplicating memory.

I know that @grahamwhaley has experience with this so maybe he might have thoughts on a simple (and quick-ish ;) test we could run?

grahamwhaley:
Easiest check is to run two containers, wait for the 'ksm pass' counter to tick over, maybe twice, and then to check the ksm 'merged page' count to ensure it has gone up...
You could either use one of our existing metrics density tests, run with and without KSM enabled, and check the results are suitably different, or probably preferably take some inspiration from the metrics lib [ksm settle detection code](https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L286), and write a new test that specifically checks pages are being merged.

The other wrinkle we've still not worked out is how a timed weekly run will report results out (to the mailing list ideally), in a way that we would notice when it fails :-)

egernst:
Agreed with cron, or just on PRs to ksm throttler repo? That is a very static repo...

chavafg:
opened https://github.com/kata-containers/tests/issues/1827 to add a test/job for ksm, should we close this PR?

GabyCT:
Closing this PR.

PR #1807, opened 2019-07-11
tests: Fix soak tests

### LINKED ISSUE #1804: debug why soak test fails on CI with jailer and fc
Related with http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-firecracker/

GabyCT:
Instead of taking the full path name of the hypervisors, we need to change
to use only the name of the process as this will change once that we enable
jailer on firecracker.
Fixes #1804

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -25,6 +25,8 @@ case "${CI_JOB}" in
 	"FIRECRACKER"|"NEMU")
 		echo "INFO: Running docker integration tests"
 		sudo -E PATH="$PATH" bash -c "make docker"
+		echo "INFO: Running soak test"
+		sudo -E PATH="$PATH" bash -c "make docker-stability"
 		echo "INFO: Running oci call test"
 		sudo -E PATH="$PATH" bash -c "make oci"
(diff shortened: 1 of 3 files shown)


GabyCT:
so it looks like there is not finding the correct number of shims
```
16:44:57 ++ pgrep -a /usr/libexec/kata-containers/kata-shim
16:44:57 ++ wc -l
16:44:57 + how_many_shims=0
16:44:57 + ((  1 != 0  ))
16:44:57 + echo 'Wrong number of shims running (1 != 0) - stopping'
16:44:57 Wrong number of shims running (1 != 0) - stopping
```
/cc @grahamwhaley and @mcastelino

GabyCT:
ok found the issue of the shims...

grahamwhaley:
Looks like the fc CI passed, but I don't see any shim related changes in the commits I think - what did you find @GabyCT ? :-)

GabyCT:
@grahamwhaley , it was a mistake that I did, I remove the `-f` but now everything seems that is working again so it seems that we can enable again the tests

PR #1812, opened 2019-07-16
clean-up: do extra clean up for k8s integration on bare-metal CI

Pennyzct:
When running `kubeadm reset` to delete stale resource derived from k8s integration tests, we often get following suggestion:
```
The reset process does not reset or clean up iptables rules or IPVS tables.
If you wish to reset iptables, you must do so manually.
For example:
iptables -F && iptables -t nat -F && iptables -t mangle -F && iptables -X

If your cluster was setup to utilize IPVS, run ipvsadm --clear (or similar)
to reset your system's IPVS tables.

The reset process does not clean your kubeconfig files and you must remove them manually.
Please, check the contents of the $HOME/.kube/config file.
```
Especially for cleaning up iptables, if you don't clean it up, it may lead to `cni0` interface not being set up in master node next time.
They all exist in bare-metal environment. ;)

### DIFF
--- a/integration/kubernetes/cleanup_bare_metal_env.sh
+++ b/integration/kubernetes/cleanup_bare_metal_env.sh
@@ -24,3 +24,26 @@ sudo -E rm -rf "$HOME/.kube"
 # Remove existing CNI configurations and binaries.
 sudo rm -rf /var/lib/cni/networks/*
 sudo rm -rf /opt/cni/bin/*
+
+# delete containers resource created by runc
+cri_runtime="${CRI_RUNTIME:-crio}"
+case "${cri_runtime}" in
+containerd)
+        readonly runc_path=$(command -v runc)
+        ;;
+crio)
+        readonly runc_path="/usr/local/bin/crio-runc"
+        ;;
+*)
+        echo "Runtime ${cri_runtime} not supported"
+	exit 0
+        ;;
+esac
+
+runc_container_union="$($runc_path list)"
+if [ -n "$runc_container_union" ]; then
+	while IFS='$\n' read runc_container; do
+		container_id="$(echo "$runc_container" | awk '{print $1}')"
+		[ "$container_id" != "ID" ] && $runc_path delete -f $container_id
+	done <<< "${runc_container_union}"
+fi
--- a/integration/kubernetes/cleanup_bare_metal_env.sh
+++ b/integration/kubernetes/cleanup_bare_metal_env.sh
@@ -9,13 +9,14 @@ set -o nounset
 set -o pipefail
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
-iptables_cache="${SCRIPT_PATH}/iptables_cache"
+source "${SCRIPT_PATH}/../../lib/common.bash"
+
+iptables_cache="${KATA_TESTS_DATADIR}/iptables_cache"
 
 # The kubeadm reset process does not reset or clean up iptables rules
 # you must do it manually
 # Here, we restore the iptables based on the previously cached file.
 sudo iptables-restore < "$iptables_cache"
-sudo rm -rf "$iptables_cache"
 
 # The kubeadm reset process does not clean your kubeconfig files.
 # you must remove them manually.
--- a/integration/kubernetes/cleanup_bare_metal_env.sh
+++ b/integration/kubernetes/cleanup_bare_metal_env.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 ARM Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+iptables_cache="${SCRIPT_PATH}/iptables_cache"
+
+# The kubeadm reset process does not reset or clean up iptables rules
+# you must do it manually
+# Here, we restore the iptables based on the previously cached file.
+sudo iptables-restore < "$iptables_cache"
+sudo rm -rf "$iptables_cache"
+
+# The kubeadm reset process does not clean your kubeconfig files.
+# you must remove them manually.
+sudo -E rm -rf "$HOME/.kube"
+
+# Remove existing CNI configurations and binaries.
+sudo rm -rf /var/lib/cni/networks/*
+sudo rm -rf /opt/cni/bin/*
--- a/integration/kubernetes/cleanup_env.sh
+++ b/integration/kubernetes/cleanup_env.sh
@@ -33,6 +33,12 @@ sudo ip link set dev flannel.1 down
 sudo ip link del cni0
 sudo ip link del flannel.1
 
+# if CI run in bare-metal, we need a set of extra clean
+BAREMETAL="${BAREMETAL:-false}"
+if [ "${BAREMETAL}" == true ] && [ -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh" ];  then
+	bash -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh"
+fi
+
 # Check no kata processes are left behind after reseting kubernetes
 check_processes
 
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -20,8 +20,11 @@ kubernetes_version=$(get_version "externals.kubernetes.version")
 
 # store iptables if CI running on bare-metal
 BAREMETAL="${BAREMETAL:-false}"
-iptables_cache="${SCRIPT_PATH}/iptables_cache"
-[ "${BAREMETAL}" == true ] && iptables-save > "$iptables_cache"
+iptables_cache="${KATA_TESTS_DATADIR}/iptables_cache"
+if [ "${BAREMETAL}" == true ]; then
+	[ -d "${KATA_TESTS_DATADIR}" ] || sudo mkdir -p "${KATA_TESTS_DATADIR}"
+	iptables-save > "$iptables_cache"
+fi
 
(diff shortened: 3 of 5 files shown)


integration/kubernetes/cleanup_bare_metal_env.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# The kubeadm reset process does not reset or clean up iptables rules
+# you must do it manually
+sudo iptables -F && iptables -t nat -F && iptables -t mangle -F && iptables -X
```
grahamwhaley: This is a pretty hard 'drop all iptables rules' I think is it, rather than selectively dropping the k8s related rules. I suspect we have little choice, as maybe we cannot identify all k8s rules - so, I'm hoping we are not using any iptables rules for other infra on the system at the time ;-)

integration/kubernetes/cleanup_env.sh
```diff
@@ -33,6 +33,12 @@ sudo ip link set dev flannel.1 down
 sudo ip link del cni0
 sudo ip link del flannel.1
 
+# if CI run in bare-metal, we need a set of extra clean
+BAREMETAL="${BAREMETAL:-false}"
+if [ "${BAREMETAL}" == true ] && [ -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh" ];  then
+         source "${SCRIPT_PATH}/cleanup_bare_metal_env.sh"
```
grahamwhaley: do you need to 'source' it, or can you just run the .sh file directly?

Pennyzct:
there may also exist some stale resource from `runc`, 
```
$ /usr/local/bin/crio-runc list
ID                                                                 PID         STATUS      BUNDLE                                                                                                                 CREATED                          OWNER
0e2cee4af3a8e252db01b68ab3c13d2fc8da4b8f40188e3b35eb48955164b495   0           stopped     /run/containers/storage/overlay-containers/0e2cee4af3a8e252db01b68ab3c13d2fc8da4b8f40188e3b35eb48955164b495/userdata   2019-07-16T08:53:20.118960094Z   root
1be59eab36246dd3321c8c681a84607a256104b4fa8809f5026417ce3ded5c47   0           stopped     /run/containers/storage/overlay-containers/1be59eab36246dd3321c8c681a84607a256104b4fa8809f5026417ce3ded5c47/userdata   2019-07-16T08:53:20.283678038Z   root
2ec3c8b6ebc826620be8a91cddcc92fac7c10603270415586cdacd7f81b315ce   0           stopped     /run/containers/storage/overlay-containers/2ec3c8b6ebc826620be8a91cddcc92fac7c10603270415586cdacd7f81b315ce/userdata   2019-07-16T08:53:20.101482543Z   root
32b92dba4eb0b1baaa4bb5c0cce2469acdbfbcd361021be08aa1b90c056d1f26   0           stopped     /run/containers/storage/overlay-containers/32b92dba4eb0b1baaa4bb5c0cce2469acdbfbcd361021be08aa1b90c056d1f26/userdata   2019-07-16T08:53:20.266327742Z   root
48769500245bc277afe535752cb6dc48b2fc0beade74a5c8ede27d9147f3f5b1   0           stopped     /run/containers/storage/overlay-containers/48769500245bc277afe535752cb6dc48b2fc0beade74a5c8ede27d9147f3f5b1/userdata   2019-07-16T08:53:49.463145477Z   root
5aab655cae45ac73231e4e5e822512fdbd00d655e306654f70c8bb93cfabcd8c   0           stopped     /run/containers/storage/overlay-containers/5aab655cae45ac73231e4e5e822512fdbd00d655e306654f70c8bb93cfabcd8c/userdata   2019-07-16T05:54:06.4125217Z     root
85b78f01c748f759fe68c891b2af772eed1f12c904650bac8f8a5d68664ec192   0           stopped     /run/containers/storage/overlay-containers/85b78f01c748f759fe68c891b2af772eed1f12c904650bac8f8a5d68664ec192/userdata   2019-07-16T05:54:06.628935408Z   root
...... 
```
I may need to create another commit to fix this. ;)

integration/kubernetes/cleanup_bare_metal_env.sh
Pennyzct: yep, you're right. Let me do a few the whole CI run in my local machine to see if it will break anything. ;)

integration/kubernetes/cleanup_env.sh
Pennyzct: yep, maybe only `bash -f ${SCRIPT_PATH}/cleanup_bare_metal_env.sh` is enough. 😝

integration/kubernetes/cleanup_bare_metal_env.sh
grahamwhaley: :-) Also, you could look at the iptables at the end of a run and see what else, if anything, is in there. That's probably pretty easy to do on your local machine, and will show us if there is anything important still in the tables etc.

egernst: I’d be afraid of running this on my local dev machines.

grahamwhaley: @egernst ;-), well, it **should** only get run on a machine marked as a 'CI baremetal' machine, so should probably never trigger on your local box - but, yeah, @Pennyzct is checking that it will not cripple anything known to us on our existing bare metal CI machines. I'm sort of hoping that none of the cloud providers inject iptable rules into the machines 'in the background' or via some cloud init setup to support their network infra etc....
It *would* be nice if we can find a set of commands that just clear out the k8s specific bits...

chavafg: maybe we can run this only when `CI=true`, so we don't mess with dev machines?

grahamwhaley: the call of this script is already predicated with `BAREMETAL` in the other file modified in this PR, but yes, we could go 'belt and braces' and add an `&& CI` to that as well... given the potential harm this might do if run accidentally (like, I'm pretty sure it would take out @egernst proxy setup ;-) ), let's be robust here..

krsna1729: can we do iptables-save and restore instead?

Pennyzct: Hi~ @grahamwhaley In my local CI machine, 'drop all iptables' will affect docker indeed. 
It will also remove docker-related iptable rules. 
But all we need is to restart docker service: `systemctl restart docker`
@krsna1729 thanks for proposal. we could iptables-save in init.sh and iptables-restore after total clean. 
Wdyt????

grahamwhaley: ack, I think save and restore is a great idea. Now, where to save.... I guess in the `init.sh` in the kubernetes dir ?

chavafg: +1 for the save and restore. Saving them as tmp file should be enough, right? we can delete it after we do the restore.

Pennyzct:
Hi~ guys~ @grahamwhaley @chavafg 
I have taken advice from @krsna1729 to `iptables-save` before each run and `iptables-restore` after and delete related cache file~ ;) ptal. ;)
@egernst After above modification, i shall think it may not affect dev-machines, so `CI=true` may not be needed?
And I have added another commit to remove stale container resource derived from `runc, so, please also
review this new commit. 😍

integration/kubernetes/cleanup_bare_metal_env.sh
```diff
@@ -0,0 +1,50 @@
...
+set -o pipefail
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+iptables_cache="${SCRIPT_PATH}/iptables_cache"
+
+# The kubeadm reset process does not reset or clean up iptables rules
+# you must do it manually
+# Here, we restore the iptables based on the previously cached file.
+sudo iptables-restore < "$iptables_cache"
+sudo rm -rf "iptables_cache"
```
krsna1729: Missing a `$` ?

krsna1729:
Found this when searching for containers not being cleaned up on reset - https://github.com/kubernetes/kubeadm/issues/932

integration/kubernetes/cleanup_bare_metal_env.sh
Pennyzct: definitely! nice catch~~;) thanks. update asap~

integration/kubernetes/init.sh
```diff
@@ -18,6 +18,11 @@ source "/etc/os-release" || source "/usr/lib/os-release"
 cri_runtime="${CRI_RUNTIME:-crio}"
 kubernetes_version=$(get_version "externals.kubernetes.version")
 
+# store iptables if CI running on bare-metal
+BAREMETAL="${BAREMETAL:-false}"
+iptables_cache="${SCRIPT_PATH}/iptables_cache"
```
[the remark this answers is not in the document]
Pennyzct: okay~~ I have pushed another commit to address your comments. ;). ptal.

Pennyzct:
I got all green here~~~ ptal. ;)

integration/kubernetes/cleanup_bare_metal_env.sh
```diff
@@ -9,13 +9,14 @@ set -o nounset
 set -o pipefail
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
-iptables_cache="${SCRIPT_PATH}/iptables_cache"
+source "${SCRIPT_PATH}/../../lib/common.bash"
+
+iptables_cache="${KATA_TESTS_DATADIR}/iptables_cache"
```
grahamwhaley: you should probably (have) merge this change (update) into the first commit, and have moved the KATA_TESTS_DATADIR commit to be the first commit :-)

Pennyzct: ok~~~~update soon. ;)

PR #1814, opened 2019-07-16
ci: Refactor ci_cache scripts

### LINKED ISSUE #1782: ci: refactor `ci_cache_*` scripts to be more generic
`ci_cache_*` scripts can be refactor as they mostly do the same tasks. We can refactor them to only have `ci_cache_asset.sh` and use them for kernel, images and different hypervisors.

GabyCT:
This refactors the ci cache scripts to have a single script where is
possible to build kata components like kernel, image, nemu, and qemu.

Fixes #1782


.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+	sha256sum -c "sha256sum-${image}"
+	tar -cJf "$(cat latest-${image}).targ.xz" "$(cat latest-${image})"
+	sha256sum "$(cat latest-${image}).tar.xz" > "sha256sum-${image}-tarball"
+	sha256sum -c "sha256sum-${image}-tarball"
+}
+
+build_cache() {
+	component_name="$1"
+	component_version="$2"
+	check_image=$(echo "${component_name}" | grep image)
```
jodh-intel: Commands like this will cause the script to silently fail as it's running with `set -e`. If any of these sub-shell commands could possibly fail (legitimately), you need to use the `|| true` trick:

```sh
check_image=$(echo "${component_name}" | grep image || tue)
```

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+	path=$(readlink -f "${image}")
+	image_version=$(echo $(basename "${path}"))
+	build_cache "${image}" "${image_version}"
+	create_image_tar "${image}"
+}
+
+create_image_tar(){
+	image="$1"
+	sha256sum -c "sha256sum-${image}"
+	tar -cJf "$(cat latest-${image}).targ.xz" "$(cat latest-${image})"
```
jodh-intel: - Typo: `.tar.xz`.
- You're using `cat latest-${image}` three times here (and below) so I would use a variable to avoid the duplication.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+cache_image_initrd_artifacts() {
+	image="${kata_dir}/kata-containers-initrd.img"
+	path=$(readlink -f "${image}")
+	image_version=$(echo $(basename "${path}"))
+	build_cache "${image}" "${image_version}"
+	create_image_tar "${image}"
+}
+
+create_image_tar(){
+	image="$1"
```
jodh-intel: Nit: You could use `local` for function-local variables to avoid surprises.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+	sha256sum -c "sha256sum-${image}"
```
jodh-intel: OOI, why not `sha512sum`?

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+	tar -cJf "$(cat latest-${image}).targ.xz" "$(cat latest-${image})"
+	sha256sum "$(cat latest-${image}).tar.xz" > "sha256sum-${image}-tarball"
```
jodh-intel: Conventionally, checksum files tend to be named `${file}.${checksum_type}` so I was expecting this to be `foo.tar.xz.sha256[sum]` (or `foo.tar.xz.sha512[sum]`). That way, the files sort sanely with `ls` too.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+		fi
+		path=$(readlink -f "${image}")
+		echo $(basename "${path}") > "latest-${image}"
+		sudo cp "${path}" "${kata_dir}/osbuilder-${image}.yaml"
+	else
+		echo "${component_version}" >  "latest"
+	fi
+
+	if [ ! -z "${check_nemu}" ]; then
+		cp -a "/tmp/${component_name}" .
```
jodh-intel: Huh?

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+	sudo chown -R "${USER}:${USER}" .
+	sha256sum "${component_name}" > "sha256sum-${component_name}"
+	cat "sha256sum-${component_name}"
+}
+
+help() {
+echo "$(cat << EOF
+Usage: $0 "[options]"
+	Description:
+		This script builds the cache of several kata components.
```
jodh-intel: "Kata"

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+	Options:
+		-a      Run kernel cache
```
jodh-intel: I like the logical nature of these options, but it might be more memorable if you use a letter that somehow corresponds to the thing being cached? `-k` for Kernel, *etc*? In fact, why not use `getopt(1)` as it allows for long option names? Just an idea and not a blocker.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+		-d      Run image initrd cache
+		-e      Run qemu cache
+		-h      Shows help
+EOF
+)"
+}
+
+main() {
+	local OPTIND
+	while getopts ":abcdeh:" opt
```
jodh-intel: Not necessary, but you could use `getopt(1)` and the tricks employed by the static check script to dynamically build the `usage()` function output:

- https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh#L54
- https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh#L85..L116
- https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh#L996..L1001

That way, `usage()` is guaranteed to be correct even if you change options at a later date.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,179 @@
...
+		cp -a "/tmp/${component_name}" .
+	fi
+
+	sudo chown -R "${USER}:${USER}" .
+	sha256sum "${component_name}" > "sha256sum-${component_name}"
+	cat "sha256sum-${component_name}"
+}
+
+help() {
+echo "$(cat << EOF
```
jodh-intel: This is overly complex. You can simplify it to remove the `echo`:

```sh
help() {
    cat <<EOT
    blah blah

    blah blah blah

    blah 
EOT
}

File: .ci/ci_cache_components.sh [hunk 11]
```diff
@@ -0,0 +1,179 @@
...
+ 	[[ -z "$build_nemu" ]] && \
+	[[ -z "$build_image" ]] && \
+	[[ -z "$build_image_initrd" ]] && \
+		help && die "Must choose at least one test"
+
+	mkdir -p "${WORKSPACE}/artifacts
+	pushd "${WORKSPACE}/artifacts"
+	echo "artifacts:"
+
+	if [ "$build_kernel" == "1" ]; then
```
jodh-intel: You can simplify this sort of expression to:

```sh
[ "$build_kernel" == "1" ] && cache_kernel_artifacts
```

.ci/ci_cache_components.sh
grahamwhaley: if you are editing then... you could also swap out the `echo |`s for `<<<`s, which is  fractionally more bash efficient.

.ci/install_qemu.sh
```diff
@@ -48,17 +53,19 @@ uncompress_static_qemu() {
 }
 
 build_and_install_static_qemu(){
+	install_directory="/tmp"
```
chavafg: I think we can have this as a global variable as `/tmp` is used in other functions

.ci/install_qemu.sh
```diff
@@ -49,16 +54,17 @@ uncompress_static_qemu() {
 
 build_and_install_static_qemu(){
 	build_static_qemu
-	uncompress_static_qemu "$QEMU_TAR"
+	uncompress_static_qemu "${install_directory}/${QEMU_TAR}"
 }
 
 install_cached_qemu(){
+	 install_directory="/tmp"
```
chavafg: we can remove this one also

GabyCT:
@chavafg changes applied

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,215 @@
...
+	build_cache "${image_name}" "${image_version}"
+	create_image_tar "${image_name}"
+}
+
+# This function receives as an argument if it is an image or an image initrd
+create_image_tar(){
+	local image_name="$1"
+	local getinfo=$(cat latest-"${image_name}")
+	tar -cJf "${getinfo}.tar.xz" "${getinfo}"
+	sha256sum "${getinfo}.tar.xz" > "sha256sum-${image_name}-tarball"
```
jodh-intel: I'm still unclear why we're not creating this file in the convention manner (which users would expect and which would work nicely with `ls(1)`):

```sh
sha256sum "${getinfo}.tar.xz" > "${image_name}.sha256sum"
```

.ci/ci_cache_components.sh
jodh-intel: @GabyCT - ^^

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,215 @@
...
+		fi
+		path=$(readlink -f "${image_path}")
+		echo $(basename "${path}") > "latest-${image_name}"
+		sudo cp "${path}" "${kata_dir}/osbuilder-${image_name}.yaml"  .
+	else
+		echo "${component_version}" >  "latest"
+	fi
+
+	# In the case of nemu and qemu we have the tar file in
+	# a specific location /tmp
```
GabyCT: @jodh-intel that is the explanation

.ci/ci_cache_components.sh
GabyCT: @jodh-intel done

.ci/ci_cache_components.sh
jodh-intel: I really don't like using `/tmp/` for a semi-persistent cache location. Can't we use [`$KATA_TESTS_DATADIR`](https://github.com/kata-containers/tests/blob/master/lib/common.bash#L23), or even better add a `KATA_TESTS_CACHEDIR`?

jodh-intel:
Grrr. GitHub appears to have hidden my comment, so I'll add it here:

> ```sh
> cp -a "/tmp/${component_name}" .
> ```
I really don't like using `/tmp/` for a semi-persistent cache location. Can't we use [`$KATA_TESTS_DATADIR`](https://github.com/kata-containers/tests/blob/master/lib/common.bash#L23), or even better add a `KATA_TESTS_CACHEDIR`?

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,215 @@
...
+		esac
+	done
+	shift $((OPTIND-1))
+
+	[[ -z "$build_kernel" ]] && \
+	[[ -z "$build_qemu" ]] && \
+	[[ -z "$build_nemu" ]] && \
+	[[ -z "$build_image" ]] && \
+	[[ -z "$build_image_initrd" ]] && \
+		help && die "Must choose at least one test"
```
jcvenegas: Must choose at least one `option`. 
Also the idea is still to run the one job per asset or one all in one job.

If is one per asset, I feel more ergonimic to somehing like 

```
./cache_asset kernel
```
```
./cache_asset kernel
```
Or even if this is in one job per all assets.

```
#cache job.
./cache_asset kernel
./cache_asset image
etc
```

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,215 @@
...
+
+	mkdir -p "${WORKSPACE}/artifacts"
+ 	pushd "${WORKSPACE}/artifacts"
+	echo "artifacts:"
+
+	[ "$build_kernel" == "1" ] && cache_kernel_artifacts
```
jcvenegas: Nit: Recently I started to use true and false strings for this kind of validation.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,215 @@
...
+
+	[ "$build_qemu" == "1" ] && cache_qemu_artifacts
+
+	[ "$build_nemu" == "1" ] && cache_nemu_artifacts
+
+	[ "$build_image" == "1" ] && cache_image_artifacts
+
+	[ "$build_image_initrd" == "1" ] && cache_image_initrd_artifacts
```
jcvenegas: you probably can simplify by moving this to the  `case` statement.

```
case "$opt" in
 kernel) cache_kernel_artifacts
...
```

.ci/ci_cache_components.sh
jcvenegas: I like @jodh-intel  suggestion.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,215 @@
...
+	local getinfo=$(cat latest-"${image_name}")
+	tar -cJf "${getinfo}.tar.xz" "${getinfo}"
+	sha256sum "${getinfo}.tar.xz" > "${image_name}-tarball.sha256sum"
+	sha256sum -c "${image_name}-tarball.sha256sum"
+	rm "${getinfo}"
+}
+
+# This function receives as arguments the component name (qemu, nemu, kernel, etc)
+# as well as the version
+build_cache() {
```
jcvenegas: maybe something like `create_cache_asset`

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,215 @@
...
+	local component_name="$1"
+	local component_version="$2"
+	local check_image=$(echo "${component_version}" | grep kata || true)
```
jcvenegas: This is prompt to errors, as you can pass 

build_cache kemu  1.9

and wont be detected,  I wonder if something like looks better.
```
case component_name
qemu)
*)
  die invalid
```

jcvenegas:
btw looking at cache job, I dont see the kernel and hypervisor assets, probably I am missing something

GabyCT:
@jcvenegas, I tried it to run all of them in a single jenkins job, however I am having issues while trying to build `qemu` and `nemu` as the job is failing, I think is related with their configuration and their flags that they need as they are different. I think that we will still need to have the jobs separate.

egernst:
nice work on cleaning this up @GabyCT

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,199 @@
...
+cidir=$(dirname "$0")
+source "${cidir}/../lib/common.bash"
+source "${cidir}/lib.sh"
+
+script_name=${0##*/}
+WORKSPACE=${WORKSPACE:-$(pwd)}
+kata_dir="/usr/share/kata-containers"
+tests_repo="${tests_repo:-github.com/kata-containers/tests}"
+tests_repo_dir="${GOPATH}/src/${tests_repo}"
+KATA_TESTS_CACHEDIR="/tmp/cache"
```
jodh-intel: I thought we'd discussed not putting the cache under `/tmp`...?

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,199 @@
...
+
+# This function receives as arguments the component name (qemu, nemu, kernel, etc)
+# as well as the version
+create_cache_asset() {
+	local component_name="$1"
+	local component_version="$2"
+	local check_image=$(echo "${component_version}" | grep kata || true)
+	local check_nemu=$(echo "${component_name}" | grep nemu || true)
+	local check_qemu=$(echo "${component_name}" | grep qemu || true)
+	local check_kernel=$(echo "${component_name}" | grep vmlinu || true)
```
jodh-intel: Nit: I might be clearer if you put `grep 'vmlinu[xz]'` as the partial string looks initially like a typo. For this and the `qemu` above, again you could anchor for extra reliability:

```sh
grep -E '\<vmlinu[xz]\>'

```

.ci/ci_cache_components.sh
jodh-intel: Can you drop the mention of `/tmp` in the comment as it's now clear from the value of `KATA_TESTS_CACHEDIR`.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,199 @@
...
+Usage: $script_name [options]
+Description: This script builds the cache of several Kata components.
+Options:
+
+-i      Run image cache
+-k      Run kernel cache
+-n      Run nemu cache
+-q      Run qemu cache
+-r      Run image initrd cache
+-h      Shows help
```
jodh-intel: Nit: Not in alphabetical order.

.ci/ci_cache_components.sh
```diff
@@ -0,0 +1,199 @@
...
+		n)
+			cache_nemu_artifacts
+			;;
+		q)
+			cache_qemu_artifacts
+			;;
+		r)
+			cache_image_initrd_artifacts
+			;;
+		h)
```
jodh-intel: Nit: Not in alphabetical order.

.ci/install_nemu.sh
```diff
@@ -20,6 +20,8 @@ latest_build_url="http://jenkins.katacontainers.io/job/nemu-nightly-${arch}/last
 nemu_repo=$(get_version "assets.hypervisor.nemu.url")
 nemu_version=$(get_version "assets.hypervisor.nemu.version")
 NEMU_TAR="kata-nemu-static.tar.gz"
+KATA_TESTS_CACHEDIR="/tmp/cache"
```
jodh-intel: This is defined in `.ci/ci_cache_components.sh`. We need to avoid redefining a variable with the same value multiple times as if they diverge... Can't you define the variable once and `source` that file into the scripts using it. I think I mentioned that in a previous round of reviews?

.ci/install_qemu.sh
```diff
@@ -25,7 +25,7 @@ QEMU_ARCH=$(${cidir}/kata-arch.sh -d)
 PACKAGING_REPO="github.com/kata-containers/packaging"
 ARCH=$("${cidir}"/kata-arch.sh -d)
 QEMU_TAR="kata-qemu-static.tar.gz"
-
+KATA_TESTS_CACHEDIR="/tmp/cache"
```
jodh-intel: Defined yet again.

PR #1828, opened 2019-07-18
ci: Allow travis to use go install script

### LINKED ISSUE #1738: Allow Travis to use go install script
As discussed on https://github.com/kata-containers/tests/pull/1730/files#r294871120, we should consider updating `.ci/ install_go.sh` to allow it to be used by Travis so we can avoid hard-coding golang version numbers in `.travis.yml`.

GabyCT:
This allows to use the install_go.sh script to be used by Travis instead
of a specific version.

Fixes #1738

### DIFF
--- a/.ci/install_go.sh
+++ b/.ci/install_go.sh
@@ -96,6 +96,10 @@ if command -v go; then
 	fi
 fi
 
+if [ "$(uname -s)" == "Darwin" ]; then
+	goarch=amd64
+fi
+
 case "$(arch)" in
 	"aarch64")
 		goarch=arm64
@@ -118,8 +122,9 @@ case "$(arch)" in
 esac
 
 info "Download go version ${go_version}"
-curl -OL "https://storage.googleapis.com/golang/go${go_version}.linux-${goarch}.tar.gz"
+kernel_name=$(uname -s)
+curl -OL "https://storage.googleapis.com/golang/go${go_version}.${kernel_name,,}-${goarch}.tar.gz"
 info "Install go"
 mkdir -p "${install_dest}"
-sudo tar -C "${install_dest}" -xzf "go${go_version}.linux-${goarch}.tar.gz"
+sudo tar -C "${install_dest}" -xzf "go${go_version}.${kernel_name,,}-${goarch}.tar.gz"
 popd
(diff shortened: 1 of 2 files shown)


grahamwhaley:
Looks like the OSX travis failed - I dunno the current state of that - expected to pass?
The other Travis jobs look to have run the install_go script, with no ill effects - but, I don't think we can tell from the logs what the script did :-)

jodh-intel:
OSX is failing with:

```
INFO: Download go version 1.12.3
.ci/install_go.sh: line 121: goarch: unbound variable
The command "bash .ci/install_go.sh -p -f" failed and exited with 1 during .
```

That's rather intriguing since the script *should* have failed here on Darwin fwics:

- https://github.com/kata-containers/tests/blob/master/.ci/install_go.sh#L116

That needs investigating (anyone have an OSX system? I'm still waiting for Santa ;)

The script could be updated to support OSX by just adding `Darwin` to the `case` fwics though.

GabyCT:
@jodh-intel , I did the changes thanks

.ci/install_go.sh
```diff
@@ -118,8 +122,16 @@ case "$(arch)" in
 esac
 
 info "Download go version ${go_version}"
-curl -OL "https://storage.googleapis.com/golang/go${go_version}.linux-${goarch}.tar.gz"
```
chavafg: what about using:
```
kernel_name=$(uname -s)
curl -OL "https://storage.googleapis.com/golang/go${go_version}.${kernel_name,,}-${goarch}.tar.gz"
```
so we don't need to make the comparisons.

GabyCT:
@chavafg changes applied

.ci/install_go.sh
```diff
@@ -118,8 +122,9 @@ case "$(arch)" in
...
+kernel_name=$(uname -s)
+curl -OL "https://storage.googleapis.com/golang/go${go_version}.${kernel_name,,}-${goarch}.tar.gz"
```
jodh-intel: Clever!

Aside: Worth noting this in an error with `zsh` though ;)

ganeshmaharaj:
@GabyCT Just to clarify. Instead of travis using the version of golang it ships with, we are getting that from our `versions.yml` and installing that. Is that right? The version is still hard-coded in our repository, right?

GabyCT:
@ganeshmaharaj yes, the script takes the version from https://github.com/kata-containers/runtime/blob/master/versions.yaml#L260

PR #1832, opened 2019-07-19
cri-containerd: failure on building from scratch due to wrong version value

Pennyzct:
Installing cri-containerd will fail on all non-amd64 machine.

```
$ .ci/install_cri_containerd.sh
Set up environment
Trying to install containerd from static tarball
Download tarball from https://storage.googleapis.com/cri-containerd-release/cri-containerd-1.2.6.linux-arm6                                                                     4.tar.gz
curl: (22) The requested URL returned error: 404
Failed to download tarball from https://storage.googleapis.com/cri-containerd-release/cri-containerd-1.2.6.                                                                     linux-arm64.tar.gz
Trying to install containerd from source
error: pathspec '1.2.6' did not match any file(s) known to git.
Failed at 44: git checkout "${cri_containerd_version}"
Failed at 39: ( cd "${GOPATH}/src/${cri_repository}" >> /dev/null; git fetch; git checkout "${cri_container                                                                     d_version}"; make release; local commit; commit=$(git rev-parse --short HEAD); tarball_name="cri-containerd                                                                     -${commit}.${CONTAINERD_OS}-${CONTAIENRD_ARCH}.tar.gz"; sudo tar -xvf "./_output/${tarball_name}" -C / )
```
There are two way to install cri-containerd:
1. downloading tarball, which is arch-specific, only supporting `amd64`.
2. building from scratch, which is applicable for all archs
The form of version of `cri-containerd` is different on two ways.
For now, it should be 1.2.6 for tarball, and release/1.2 for source code.

### DIFF
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -10,6 +10,8 @@ set -o nounset
 set -o pipefail
 set -o errtrace
 
+source /etc/os-release || source /usr/lib/os-release
+
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cri_repository="github.com/containerd/cri"
 containerd_repository="github.com/containerd/containerd"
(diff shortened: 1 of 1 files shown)


.ci/install_cri_containerd.sh
```diff
@@ -25,10 +27,16 @@ source "${script_dir}/lib.sh"
...
+cri_containerd_tarball_version=$(get_version "externals.cri-containerd.version")
 cri_containerd_repo=$(get_version "externals.cri-containerd.url")
 
-source /etc/os-release || source /usr/lib/os-release
+echo "Get cri_containerd version"
+go get "${containerd_repository}"
+cri_containerd_version=$(
+	cd "${GOPATH}/src/${containerd_repository}";
+	git checkout "v${cri_containerd_tarball_version}" >&2;
+	cat vendor.conf | grep "github.com/containerd/cri" | awk '{print $2}';
```
jodh-intel: You could remove the `cat` using:

```sh
grep "github.com/containerd/cri" vendor.conf | awk '{print $2}';
```

Also, it would be safer to anchor that string to ensure you only get one match:

```sh
grep -E "^github.com/containerd/cri\>" vendor.conf | awk '{print $2}';
```

But also you can use the variable making the potential final version:

```sh
grep -E "^${cri_repository}\>" vendor.conf | awk '{print $2}';
```

Pennyzct: thanks!!! a lot to learn~~~😘

Pennyzct:
I got all green here~~~~~~ just need one more reviewer. ;). @grahamwhaley @chavafg plz. 🤩

Issue #1835, opened 2019-07-19
ci: fast path: metrics still run, and fail, after fast-path check is good


grahamwhaley:
# Description of problem

When we hit a fast-path 'good' case (that is, we skip the CI as we can fastpath it), the metrics checks still run, and fail.

# Expected result

Metrics should be skipped for fastpath as well.

# Actual result

The metrics fails to run, as it does not realise we have fastpath'd the CI. The root cause is the fast path skips the build/install of kata, and then the metrics tries to use the supposedly installed runtime.

# The fix

The fix is not quite trivial - today the metrics calls two scripts from the jenkins job file - the first is the jenkins (setup) script, and then it calls the metrics script. The 'problem' is that the jenkins script returns 'OK' when it fast path returns (which is correct), so the job file cannot then tell if it should run the metrics or not....

The solution is to move the call to the metrics tests out of the jenkins job file and into the jenkins script in our CI repo. This is a win win, as it is better to have our jenkins flows controlled from the code in the repos rather than wired into jenkins itself.

PR #1826, opened 2019-07-18
spellcheck: add macOS

grahamwhaley:
We refer to macOS in some docs - either that things are or are not
supported on that platform for instance.
`macOS` is not in the standard base hunspell dict, so add it to our
local distro dict.

Fixes: #1825

Signed-off-by: Graham Whaley <graham.whaley@intel.com>

PR #1836, opened 2019-07-19
ci: jenkins: metrics: add metrics call direct to jenkins script

grahamwhaley:
Currently we call the jenkins script to set up the system, and then call
the run_metrics_PR_ci.sh script from the metrics jenkins job files to
run the actul metrics.
This flow does not 'understand' the new ci fast path, and then fails.
The best fix is to move the call to the metrics script into the main
jenkins file, and remove it from the jenkins job files.

Fixes: #1835

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -162,6 +162,9 @@ then
 
 	# Code coverage
 	bash <(curl -s https://codecov.io/bash)
+else
+	echo "Running the metrics tests:"
+	.ci/run_metrics_PR_ci.sh
 fi
 
 popd


grahamwhaley:
/test-ubuntu
Firing a test so we can check the metrics now runs twice before we go remove the calls from the jenkins jobfiles themselves.
fast-path will still fail until we remove metrics from the jobfiles.

grahamwhaley:
ooh, @GabyCT - did we check the CI logs to make sure it ran metrics twice?? - err, too late I guess. hmm, we need to therefore trigger a metrics test on a PR and see what happens... let me find one...
Guess I should have stuck a DNM on this whilst we checked...

grahamwhaley:
re-triggered metrics build on https://github.com/kata-containers/tests/pull/1826 to check what happens then.

Issue #1838, opened 2019-07-19
Add add hypervisor kill test 


bergwolf:
To test kata stability w.r.t. dead hypervisor situation. We should at least allow to stop and remove a container in such case. Add test to ensure it.

PR #1839, opened 2019-07-19
integration: add hypervisor kill test

bergwolf:
Kill hypervisor and then force remove the container.
We should be able to handle the exceptional case properly.

Fixes: #1838
Depends-on: github.com/kata-containers/runtime#1897

### DIFF
--- a/integration/stability/hypervisor_stability_kill_test.sh
+++ b/integration/stability/hypervisor_stability_kill_test.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 HyperHQ Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test will kill a running container's
+# hypervisor, and see how we react to cleanup.
+
+set -e
+
+cidir=$(dirname "$0")
+
+source "${cidir}/../../metrics/lib/common.bash"
+
+# Environment variables
+IMAGE="${IMAGE:-busybox}"
+CONTAINER_NAME="${CONTAINER_NAME:-test}"
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
+
+setup()  {
+	clean_env
+	sudo docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+	num=$(ps aux | grep ${HYPERVISOR_PATH} | grep -v grep | wc -l)
+	[ ${num} -eq 1 ] || die "hypervisor count:${num} expected:1"
+}
+
+kill_hypervisor()  {
+	pid=$(ps aux | grep ${HYPERVISOR_PATH} | grep -v grep | awk '{print $2}')
+	[ -n ${pid} ] || die "failed to find hypervisor pid"
+	sudo kill -KILL ${pid} || die "failed to kill hypervisor (pid ${pid})"
+	num=$(ps aux | grep ${HYPERVISOR_PATH} | grep -v grep | wc -l)
+	[ ${num} -eq 0 ] || die "hypervisor count:${num} expected:0"
+	sudo docker rm -f $CONTAINER_NAME
+	[ $? -eq 0 ] || die "failed to force removing container $CONTAINER_NAME"
+}
+
+teardown()  {
+	echo "Ending hypervisor stability test"
+	clean_env
+}
+
+trap teardown EXIT
+
+echo "Starting hypervisor stability test"
+setup
+
+echo "Running hypervisor stability test"
+kill_hypervisor
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,7 @@ docker-stability:
 	cd integration/stability && \
 	export ITERATIONS=2 && export MAX_CONTAINERS=20 && ./soak_parallel_rm.sh
 	cd integration/stability && ./bind_mount_linux.sh
+	cd integration/stability && ./hypervisor_stability_kill_test.sh
 
 kubernetes:
 	bash -f .ci/install_bats.sh


integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -0,0 +1,56 @@
...
+CONTAINER_NAME="${CONTAINER_NAME:-test}"
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
+
+setup()  {
+	clean_env
+	sudo docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+	num=$(ps aux | grep ${HYPERVISOR_PATH} |grep -v grep | wc -l)
```
GabyCT: missing space `|grep -v grep |` to `| grep -v grep |`

integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 HyperHQ Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test will kill a running container's
+# hypervisor, and see how we react to cleanup.
+
+set -e -x
```
GabyCT: do we need the `set -x` ?

integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -0,0 +1,56 @@
...
+	clean_env
+	sudo docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+	num=$(ps aux | grep ${HYPERVISOR_PATH} |grep -v grep | wc -l)
+	[[ ${num} -eq 1 ]] || die "hypervisor count:${num} expected:1"
+}
+
+killhypervisor()  {
+	pid=$(ps aux | grep ${HYPERVISOR_PATH} |grep -v grep | awk '{print $2}')
+	[[ -n ${pid} ]] || die "failed to find hypervisor pid"
+	sudo kill -KILL ${pid} || die "failed to kill hypvervisor (pid ${pid})"
```
GabyCT: /hypvervisor/hypervisor

integration/stability/hypervisor_stability_kill_test.sh
bergwolf: Not really necessary but I want to keep it to catch errors as early as possible.

bergwolf:
/test

@GabyCT updated. PTAL.

integration/stability/hypervisor_stability_kill_test.sh
grahamwhaley: this whole line might be simpler with `pgrep` ?

integration/stability/hypervisor_stability_kill_test.sh
grahamwhaley: are we launching the test with `chronic`? If so, the `-x` will get nicely hidden unless we fail.
nb. though, afaik we don't set `-x` for any other tests ...

integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+echo "Starting hypervisor stability test"
+setup
+
+echo "Running hypervisor stability test"
+killhypervisor
+cleanup
+
+echo "Ending hypervisor stability test"
+teardown
```
chavafg: should this be a trap? so it is executed even if something fails in the middle, that way we don't leave the environment dirty? 
You can see an example on: https://github.com/kata-containers/tests/blob/master/tracing/tracing-test.sh#L33

integration/stability/hypervisor_stability_kill_test.sh
bergwolf: Fair enough, I've dropped it.

bergwolf:
/retest

PR updated. PTAL, thanks!

integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 HyperHQ Inc.
```
jodh-intel: OOI, Is the company name still correct?

integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -0,0 +1,52 @@
...
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-kata-runtime}"
+
+setup()  {
+	clean_env
+	sudo docker run --runtime=$RUNTIME -d --name $CONTAINER_NAME $IMAGE $PAYLOAD_ARGS
+	num=$(ps aux | grep ${HYPERVISOR_PATH} | grep -v grep | wc -l)
+	[[ ${num} -eq 1 ]] || die "hypervisor count:${num} expected:1"
```
jodh-intel: Nit: I think for the tests this script is performing, you can use a single square bracket, which is slightly clearer to read:

```sh
[  ${num} -eq 1 ] || ...
```

integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+killhypervisor()  {
```
jodh-intel: Nit: `CamelCase` or `snake_case` would make this function name easier to read.

integration/stability/hypervisor_stability_kill_test.sh
bergwolf: We are still using hyper.sh for kata containers development so far ;)

integration/stability/hypervisor_stability_kill_test.sh
bergwolf: Updated. Thanks!

integration/stability/hypervisor_stability_kill_test.sh
bergwolf: Updated. Thanks!

integration/stability/hypervisor_stability_kill_test.sh
jodh-intel: ack ;)

bergwolf:
oh, it depends on github.com/kata-containers/runtime#1897 and that one is blocked due to a last minute merge conflict.

Issue #1840, opened 2019-07-19
ci: metrics: run metrics script using full tests repo path


chavafg:
When running on other repos than `tests`:

```
11:50:19 .ci/jenkins_job_build.sh: line 167: .ci/run_metrics_PR_ci.sh: No such file or directory
```

PR #1841, opened 2019-07-19
ci: metrics: run metrics script using full tests repo path

chavafg:
As we `pushd` into the kata repositories paths (other than
tests) we need to specify the `tests_repo_dir` to call
the metrics script

Fixes: #1840.


grahamwhaley:
/test-ubuntu
oddly, looks like we are just missing the one CI result.... a CI blip - let's re-fire that one..

Issue #1846, opened 2019-07-23
firecracker is not tested here


bergwolf:
# Description of problem

Firecracker ci is not enabled here and as a result, PR #1839 breaks it without showing any failure in the PR's CI.

# Expected result

Firecracker CI should also be enabled.

bergwolf:
@jcvenegas @grahamwhaley How are we configuring fc ci right now? I'm seeing

```
Makefile:112: recipe for target 'docker-stability' failed
make: *** [docker-stability] Error 1
Failed at 29: sudo -E PATH="$PATH" bash -c "make docker-stability"
Build step 'Execute shell' marked build as failure
Performing Post build task...
Match found for :.* : True
Logical operation result is TRUE
Running script  : #!/bin/bash
```

And it is because `HYPERVISOR_PATH` is not getting set https://github.com/kata-containers/tests/blob/master/integration/stability/hypervisor_stability_kill_test.sh#L32 by reading kata config.

grahamwhaley:
@bergwolf - I think @GabyCT and @mcastelino were looking at this last week - I thought things got fixed.... let's see what they say.

bergwolf:
@grahamwhaley Thanks for the hint. I looked for @GabyCT's commit log and found the problem. Let's see if #1847 fixes it.

GabyCT:
@bergwolf the fc is being triggered https://github.com/kata-containers/tests/pull/1847

PR #1847, opened 2019-07-23
stability: fix fc name in hypervisor kill tests

bergwolf:
Instead of taking the full path name of the hypervisors, we need to
change to use only the name of the process because with jailer we have
different path shown there.

Fixes: #1846

### DIFF
--- a/integration/stability/hypervisor_stability_kill_test.sh
+++ b/integration/stability/hypervisor_stability_kill_test.sh
@@ -21,26 +21,22 @@ PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
 # Set the runtime if not set already
 RUNTIME="${RUNTIME:-kata-runtime}"
 
-KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
-
-if [ "$KATA_HYPERVISOR" == "firecracker" ]; then
-	issue="https://github.com/kata-containers/tests/issues/1849"
-	echo "Skip hypervisor stability kill test, see: $issue"
-	exit
-fi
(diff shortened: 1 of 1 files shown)


integration/stability/hypervisor_stability_kill_test.sh
```diff
@@ -21,18 +21,20 @@ PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
 # Set the runtime if not set already
 RUNTIME="${RUNTIME:-kata-runtime}"
 
+HYPERVISOR_NAME=$(basename ${HYPERVISOR_PATH})
```
grahamwhaley: nit: you can use shell magic here like `${HYPERVISOR_PATH%/*}` or similar ;-)

bergwolf: Wow, that is real magic ;)

bergwolf:
Because tests repo doesn't validate fc in CI, I'm trying to test it in a runtime PR, see https://github.com/kata-containers/runtime/pull/1905

bergwolf:
@grahamwhaley It's the same issue but in a different test case -- the one I just added to killing hypervisor process.

chavafg:
@GabyCT any idea of why fc job is not being triggered here?

GabyCT:
@chavafg I can see that now is being triggered

chavafg:
thanks @GabyCT, I see it now 😄

GabyCT:
@bergwolf , if you want to launch the `fc` CI only you can use `/test-fc` to run only that job or '/test` and will run all the CI

PR #1855, opened 2019-07-25
ci: Fix FC installation script

### LINKED ISSUE #1854: Fix FC installation script
While running multiple times the fc installation script fails to
detect the variables of driver and path at the configuration file. This
ensures that we can detect those variables.

GabyCT:
While running multiple times the fc installation script fails to
detect the variables of driver and path at the configuration file. This
ensures that we can detect those variables.

Fixes #1854

### DIFF
--- a/.ci/install_firecracker.sh
+++ b/.ci/install_firecracker.sh
@@ -63,10 +63,10 @@ path="/usr/local/bin/kata-runtime"
 
 if [ -f $docker_configuration_file ]; then
 	# Check devicemapper flag
-	check_devicemapper=$(grep -w '"storage-driver": "${driver}"' $docker_configuration_file | wc -l)
+	check_devicemapper=$(grep -w '"storage-driver": "'${driver}'"' $docker_configuration_file | wc -l)
 	[ $check_devicemapper -eq 0 ] && die "${driver} is not enabled at $docker_configuration_file"
 	# Check kata runtime flag
-	check_kata=$(grep -w '"path": "${path}"' $docker_configuration_file | wc -l)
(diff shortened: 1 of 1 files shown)


.ci/install_firecracker.sh
```diff
@@ -63,10 +63,10 @@ path="/usr/local/bin/kata-runtime"
 
 if [ -f $docker_configuration_file ]; then
 	# Check devicemapper flag
-	check_devicemapper=$(grep -w '"storage-driver": "${driver}"' $docker_configuration_file | wc -l)
+	check_devicemapper=$(grep -w '"storage-driver": "'${driver}'"' $docker_configuration_file | wc -l)
```
marcov: Thanks @GabyCT 
Can you explain why both `"` and `'` are necessary here?

GabyCT: @marcov the docker configuration should looks like
```
 "storage-driver": "devicemapper"
```
however, before like I have it `grep -w '"storage-driver": "${driver}"'` the variable of `driver` was not being replaced and that was causing to the script to fail when you run it multiple times

GabyCT: it was doing this the script
```
grep -w '"storage-driver": "${driver}"' /etc/docker/daemon.json
```
but now with this change is replacing the variable correctly
```
grep -w '"storage-driver": "devicemapper"' /etc/docker/daemon.json
```

marcov: Ahhhhh, now it's clear.
you are concatenating three strings, and the 1st and 3rd are single quoted:
 `"storage-driver": "` + `${driver}` + `"`

marcov:
lgtm

Issue #1867, opened 2019-07-29
Run static-checks on Git index files only.


marcov:
static-checks should not be run on untracked Git files. `git ls-files` should help here :wink:

PR #1868, opened 2019-07-29
static-checks: only check files in Git index

marcov:
Only run static checks on the files in the Git index, to avoid any side
effects of checks failing because of untracked files, such as build
artifacts.

Fixes: #1867

Signed-off-by: Marco Vedovati <mvedovati@suse.com>

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -482,7 +482,7 @@ static_check_docs()
 	local new_urls
 	local url
 
-	all_docs=$(find . -name "*.md" | grep -v "/vendor/" | sort || true)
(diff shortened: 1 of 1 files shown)


marcov:
@grahamwhaley can we ignore metrics and cri-containerd failures given I just changed static-checks?

chavafg:
I think we can merge this one also as it is only related to static checks

grahamwhaley:
yeah, my suspicion is that maybe the merge of the kernel fragments (we did merge that, right @devimc ?), has changed the metrics, but the baseline measures have not been updated in a PR yet.

Issue #1866, opened 2019-07-29
Add capitalised distro names to spell checker dictionary


jodh-intel:
Add missing distro terms to custom spell check dictionary.

PR #1869, opened 2019-07-29
Add distro words and minor spell checker fixes

jodh-intel:
Add "MacOS" and "OpenSUSE" to the distros fragment for the custom dictionary.

Both these terms already existed in distros fragments file, but with different capitalisation, correct for each term.

Although `hunspell(1)` will automatically accept the capitalised versions of custom words specified in the dictionary in all lower-case, it will *not* accept capitalised custom words starting with a lower-case letter but containing an upper-case one, like both of the previously specified terms.

Also fix an unbound variable error if no spell checker CLI command is specified and improve the error message in that scenario.

Fixes: #1866.

### DIFF
--- a/cmd/check-spelling/data/distros.txt
+++ b/cmd/check-spelling/data/distros.txt
@@ -9,8 +9,10 @@ Debian/B
 EulerOS/B
 Fedora/B
 macOS/B
+MacOS/B
 minikube/B
 openSUSE/B
+OpenSUSE/B
 RHEL/B
 SLES/B
 Ubuntu/B
--- a/cmd/check-spelling/kata-spell-check.sh
+++ b/cmd/check-spelling/kata-spell-check.sh
@@ -339,7 +339,7 @@ main()
 {
 	setup
 
-	[ -z "$1" ] && die "need command"
+	[ -z "${1:-}" ] && die "need command"
 
 	case "$1" in
 		check) shift && spell_check_file "$1" ;;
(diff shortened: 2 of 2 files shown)


jodh-intel:
@grahamwhaley - does the metrics setup need a kick^w... err I mean tweak? :)

```
11:13:38 +-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
11:13:38 | P/F |         NAME         |  FLR  | MEAN  |  CEIL  |  GAP  |  MIN  |  MAX   |  RNG  | COV  | ITS |
11:13:38 +-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
11:13:38 | P   | boot-times           | 95.0% | 95.4% | 105.0% | 10.0% | 89.6% | 103.9% | 15.9% | 4.0% |  20 |
11:13:38 | *F* | memory-footprint     | 95.0% | 92.1% | 105.0% | 10.0% | 92.1% | 92.1%  | 0.0%  | 0.0% |   1 |
11:13:38 | *F* | memory-footprint-ksm | 95.0% | 88.2% | 105.0% | 10.0% | 88.2% | 88.2%  | 0.0%  | 0.0% |   1 |
11:13:38 +-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
```

grahamwhaley:
It's been failing for 3d14h
http://jenkins.katacontainers.io/computer/x86_packet02/builds
My suspicion is that maybe we landed something that changed the baseline metrics results - @devimc - any thoughts before we go digging into the history?
https://github.com/kata-containers/tests/tree/master/cmd/history can help get a timeline of merges...

chavafg:
merging as this is not metrics related

PR #1870, opened 2019-07-29
test: Add KSM test

### LINKED ISSUE #1827: Add test to verify ksm-throttler works as expected.

From: https://github.com/kata-containers/tests/pull/1799#issuecomment-511688587

Add `ksm-throttler` test to verify that:
- we can build / install the throttler.
- the throttler is correctly de-duplicating memory.

Depending on how much the setup/test takes, we can decide if we can run it on one of the existing jobs or create a new job that will run on weekly basis.

GabyCT:
This will run two containers and will wait to the KSM to settle down
and then it will check that the merged pages count is increasing.

Fixes #1827

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,9 @@ kubernetes:
 	bash -f .ci/install_bats.sh
 	bash -f integration/kubernetes/run_kubernetes_tests.sh
 
+ksm:
+	bash -f integration/ksm/ksm_test.sh
+
 swarm:
 	systemctl is-active --quiet docker || sudo systemctl start docker
 	bash -f .ci/install_bats.sh
--- a/integration/ksm/ksm_test.sh
+++ b/integration/ksm/ksm_test.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This will run two containers and will wait
+# to the KSM to settle down and then it will
+# check that the merged pages count is
+# increasing
+
+set -e
+
+dir_path=$(dirname "$0")
+source "${dir_path}/../../lib/common.bash"
+source "${dir_path}/../../metrics/lib/common.bash"
+
+KATA_KSM_THROTTLER="${KATA_KSM_THROTTLER:-yes}"
+RUNTIME="${RUNTIME:-kata-runtime}"
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+IMAGE="${IMAGE:-busybox}"
+WAIT_TIME="60"
+
+function setup() {
+	clean_env
+	check_processes
+	save_ksm_settings
+	set_ksm_aggressive
+}
+
+function teardown() {
+	clean_env
+	check_processes
+	restore_ksm_settings
+}
+
+trap teardown EXIT
+
+function run_with_ksm() {
+	setup
+
+	# Running the first container
+	docker run -d --runtime="${RUNTIME}" "${IMAGE}" sh -c "${PAYLOAD_ARGS}"
+
+	echo "Entering KSM settle mode on first container"
+	wait_ksm_settle "${WAIT_TIME}"
+
+	# Checking the pages merged for the first container
+	first_pages_merged=$(cat "${KSM_PAGES_SHARED}")
+
+	echo "Pages merged $first_pages_merged"
+
+	# Running the second container
+	docker run -d --runtime="${RUNTIME}" "${IMAGE}" sh -c "${PAYLOAD_ARGS}"
+
+	echo "Entering KSM settle mode on second container"
+	wait_ksm_settle "${WAIT_TIME}"
+
+	# Checking the pages merged for the second container
+	second_pages_merged=$(cat "${KSM_PAGES_SHARED}")
+
+	echo "Pages merged $second_pages_merged"
+
+	# Compared the pages merged between the containers
+	echo "Comparing merged pages between containers"
+	[ "$second_pages_merged" -gt "$first_pages_merged" ] || die "The merged pages on the second container is less than the first container"
+
+}
+
+echo "Starting KSM test"
+run_with_ksm
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -21,6 +21,7 @@ KSM_BASE="/sys/kernel/mm/ksm"
 KSM_ENABLE_FILE="${KSM_BASE}/run"
 KSM_PAGES_FILE="${KSM_BASE}/pages_to_scan"
 KSM_SLEEP_FILE="${KSM_BASE}/sleep_millisecs"
+KSM_PAGES_SHARED="${KSM_BASE}/pages_shared"
 
 # The settings we use for an 'aggresive' KSM setup
 # Scan 1000 pages every 50ms - 20,000 pages/s
@@ -296,7 +297,7 @@ wait_ksm_settle(){
 	for ((t=0; t<$1; t++)); do
 
 		newscan=$(cat /sys/kernel/mm/ksm/full_scans)
-		newpages=$(cat /sys/kernel/mm/ksm/pages_shared)
+		newpages=$(cat "${KSM_PAGES_SHARED}")
 		[[ "$newpages" -eq 0 ]] && echo "No need to wait for KSM to settle" && return
 
 		if (( newscan != oldscan )); then


integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+function remove_tmp_files() {
+	rm -rf $FIRST_TMP_FILE $SECOND_TMP_FILE
+}
+
+trap remove_tmp_files EXIT
+
+function turn_on_ksm() {
+	echo "Turning on KSM"
+	echo "1" > ${KSM_ENABLE_FILE}
```
jodh-intel: This assumes the script is running as `root`, but there is no check on that in `run_with_ksm()`.

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+set -e
+
+dir_path=$(dirname "$0")
+source "${dir_path}/../../lib/common.bash"
+source "${dir_path}/../../metrics/lib/common.bash"
+
+TEST_NAME="Run containers with ksm"
```
jodh-intel: Nit: s/ksm/KSM/

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+	check_processes
+
+	turn_off_ksm
+}
+
+function run_with_ksm() {
+	setup
+
+	# Running the first container
+	docker run -d --runtime=$RUNTIME $IMAGE $PAYLOAD_ARGS
```
jodh-intel: Nit: Personally I'd always quote (all) variables "just in case" / to avoid nasty+weird surprises :smile:

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+	echo "Entering KSM settle mode on first container"
+	wait_ksm_settle $WAIT_TIME > $FIRST_TMP_FILE
+
+	# Checking the pages merged for the first container
+	first_pages_merged=$(cat $FIRST_TMP_FILE | grep -w "check pages" | tail -1 | cut -d ' ' -f5)
```
jodh-intel: You could remove the `cat` here, but also you might want to add `|| true` if this could ever fail?:

```sh
first_pages_merged=$(grep -w "check pages" "$FIRST_TMP_FILE" | tail -1 | cut -d ' ' -f5 || true)
```

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+FIRST_TMP_FILE=$(mktemp firstcontainer.XXXX)
+SECOND_TMP_FILE=$(mktemp secondcontainer.XXXX)
+
+function remove_tmp_files() {
+	rm -rf $FIRST_TMP_FILE $SECOND_TMP_FILE
+}
+
+trap remove_tmp_files EXIT
+
+function turn_on_ksm() {
```
grahamwhaley: There are a whole bunch of ksm manipulation functions in the metrics lib file you now include: https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L233-L333

You probably want to invoke the ksm save/aggressive/restore sequence like we do in the metrics CI script: https://github.com/kata-containers/tests/blob/master/.ci/run_metrics_PR_ci.sh#L30-L45

integration/ksm/ksm_test.sh
grahamwhaley: I'd probably forego the whole tmpfile/cat/grep thing, and just directly read from `/sys/kernel/mm/ksm/pages_shared`. That gets rid of the tmpfiles completely and we don't have to go parsing logs....
Whilst there, I'd probably add that `pages_shared` file to the block of KSM vars at https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L20-L23, and also then use that in the metrics `common.bash` file as well (where we already hardwire the path - which looks like an oversight on my part...)

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+	# Checking the pages merged for the second container
+	second_pages_merged=$(cat $SECOND_TMP_FILE | grep -w "check pages" | tail -1 | cut -d ' ' -f5)
+
+	echo "Pages merged $second_pages_merged"
+
+	# Compared the pages merged between the containers
+	echo "Comparing merged pages between containers"
+	[ "$second_pages_merged" -gt "$first_pages_merged" ]
+
```
grahamwhaley: As this is not a `bats` or other test framework file, I'd probably make this an explicit if/else and add a `die` clause on failure with details, including the page counts - otherwise I think this test will just silently fail after saying it was comparing the pages - more info on death here will definitely be more useful.

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+trap remove_tmp_files EXIT
+
+function turn_on_ksm() {
+	echo "Turning on KSM"
+	echo "1" > ${KSM_ENABLE_FILE}
+}
+
+function turn_off_ksm() {
+	echo "Turning off KSM"
+	echo "0" > ${KSM_ENABLE_FILE}
```
devimc: `echo 0 | sudo tee ${KSM_ENABLE_FILE}`

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,90 @@
...
+	echo "Pages merged $first_pages_merged"
+
+	# Running the second container
+	docker run -d --runtime=$RUNTIME $IMAGE $PAYLOAD_ARGS
+
+	echo "Entering KSM settle mode on second container"
+	wait_ksm_settle $WAIT_TIME > $SECOND_TMP_FILE
+
+	# Checking the pages merged for the second container
+	second_pages_merged=$(cat $SECOND_TMP_FILE | grep -w "check pages" | tail -1 | cut -d ' ' -f5)
```
devimc: `cat` is not needed

`grep -w "check pages" "${SECOND_TMP_FILE}" | tail -1 | cut -d ' ' -f5`

GabyCT:
@jodh-intel @devimc @grahamwhaley changes applied

Makefile
```diff
@@ -119,6 +119,9 @@ kubernetes:
 	bash -f .ci/install_bats.sh
 	bash -f integration/kubernetes/run_kubernetes_tests.sh
 
+ksm:
```
grahamwhaley: does `ksm` also need to go in the test `UNION`?

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,88 @@
...
+
+TEST_NAME="Run containers with KSM"
+KATA_KSM_THROTTLER="${KATA_KSM_THROTTLER:-yes}"
+RUNTIME="${RUNTIME:-kata-runtime}"
+PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
+IMAGE="${IMAGE:-busybox}"
+WAIT_TIME="60"
+
+function turn_on_ksm() {
+	echo "Turning on KSM"
```
grahamwhaley: so, not going to use the existing functions in the library code then? - the save/enable_aggresive/restore ? That has a few benefits:
- it puts KSM back how it found it
- setting the KSM aggresive makes the test settle down much quicker, particularly on ubuntu that has terrible default ksm settings....

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,88 @@
...
+	clean_env
+
+	check_processes
+
+	save_ksm_settings
+
+	restore_ksm_settings
+
+	disable_ksm
+}
```
GabyCT: @grahamwhaley , I am using them

Makefile
GabyCT: @grahamwhaley no, this test will run in a separate CI that is not part of the PRs

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,88 @@
...
+	echo "Pages merged $second_pages_merged"
+
+	echo "napping ${WAIT_TIME} s"
+	sleep "${WAIT_TIME}"
+
+	# Compared the pages merged between the containers
+	echo "Comparing merged pages between containers"
+	[ "$second_pages_merged" -gt "$first_pages_merged" ] || die "The merged pages on the second container is less than the first container"
+
+	teardown
```
chavafg: I think we should use `trap` to make sure the `teardown` runs if failures occur before

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,88 @@
...
+
+	check_for_ksm
+}
+
+function teardown() {
+	clean_env
+
+	check_processes
+
+	save_ksm_settings
```
chavafg: Saving ksm settings right before restoring them? Shouldn't they be saved on the `setup` before `turn_on_ksm`?

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,88 @@
...
+
+	# Checking the pages merged for the first container
+	first_pages_merged=$(cat "${KSM_PAGES_SHARED}")
+
+	echo "Pages merged $first_pages_merged"
+
+	# If KSM is enabled, then you normally want to sleep long enough to
+	# let it do its work and for the numbers to 'settle'.
+	echo "napping ${WAIT_TIME} s"
+	sleep "${WAIT_TIME}"
```
chavafg: do we need this `sleep`? `wait_ksm_settle` is already called before...

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,88 @@
...
+	echo "Entering KSM settle mode on second container"
+	wait_ksm_settle "${WAIT_TIME}"
+
+	# Checking the pages merged for the second container
+	second_pages_merged=$(cat "${KSM_PAGES_SHARED}")
+
+	echo "Pages merged $second_pages_merged"
+
+	echo "napping ${WAIT_TIME} s"
+	sleep "${WAIT_TIME}"
```
chavafg: same comment as above

integration/ksm/ksm_test.sh
GabyCT: done @chavafg

integration/ksm/ksm_test.sh
GabyCT: @chavafg the reason why I am putting the `save_ksm_settings` there is because we need to run it first to use `restore_ksm`
https://github.com/kata-containers/tests/blob/10bdf7374f7c31263d788404496e9b8c32a1999a/metrics/lib/common.bash#L250

integration/ksm/ksm_test.sh
grahamwhaley: Ah, I see now. I think we can still use them slightly better:
- use [set_ksm_aggresive](https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L241) to enable KSM, - as it will set up KSM so we 'settle' in a reasonable time.
- and `restore_ksm_settings` will put the KSM enable back to how it was originally.

So, you should probably not (have to, need to, want to) call `enable_ksm` or `disable_ksm` directly, but use `aggresive` and `restore` instead.

I know, the ksm funcs in the lib are probably not well documented on how to use - as they were previously only called from the metrics script I referenced earlier.

GabyCT: @grahamwhaley , I will leave the `disable_ksm` to put `0` at `/sys/kernel/mm/ksm/run`, because if I do not call it , I will have
```
restoring KSM settings
$ cat /sys/kernel/mm/ksm/run
1
```

GabyCT:
@chavafg and @grahamwhaley changes applied

GabyCT:
@chavafg and @grahamwhaley changes applied

integration/ksm/ksm_test.sh
grahamwhaley: Only if:
- KSM was enabled before you called `save_ksm_settings`
- you enabled KSM before calling `save_ksm_settings`.

I'd expect your calls to the ksm library funcs (via a number of your funcs) to go like...

- save ksm settings
- set ksm aggresive mode (which will enable ksm)
- *do work here*
- wait ksm settle
- restore ksm (which will only disable KSM if it was disabled before you called the matching save ksm... thus, trying to leave the system under test as you found it....)

GabyCT: @grahamwhaley 👍 thanks, changes applied, sorry that I got confused

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
```
grahamwhaley: nit: it is nice to have some description at the top of the test of what the intention is, and how it is supposed to work - we have been bitten in the past when coming to look at a failing test case and not having much clue as to what it was meant to do or how ;-)

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,75 @@
...
+	clean_env
+
+	check_processes
+
+	save_ksm_settings
+
+	set_ksm_aggressive
+
+	check_for_ksm
+
```
grahamwhaley: nit: extra blank line. In fact, you could just compress all these lines in this function without blanks...

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,75 @@
...
+function setup() {
+	clean_env
+
+	check_processes
+
+	save_ksm_settings
+
+	set_ksm_aggressive
+
+	check_for_ksm
```
grahamwhaley: I don't think you need to call this - if you read the code, what it does is append ` ksm` to the end of the `TEST_NAME` if KSM is enabled (to differentiate the results files generated, if you run the same test with and without KSM enabled). You already have `KSM` on the end of your `TEST_NAME`, so right now it will end up being called `KSM ksm` :-)

integration/ksm/ksm_test.sh
```diff
@@ -0,0 +1,75 @@
...
+
+}
+
+function teardown() {
+	clean_env
+
+	check_processes
+
```
grahamwhaley: nit: again, you can drop the blank lines I reckon.

metrics/lib/common.bash
```diff
@@ -21,6 +21,7 @@ KSM_BASE="/sys/kernel/mm/ksm"
 KSM_ENABLE_FILE="${KSM_BASE}/run"
 KSM_PAGES_FILE="${KSM_BASE}/pages_to_scan"
 KSM_SLEEP_FILE="${KSM_BASE}/sleep_millisecs"
+KSM_PAGES_SHARED="${KSM_BASE}/pages_shared"
```
grahamwhaley: great - can you also then update any occurrences that are hard wired (by me 🙌 originally) in that same file - like in the ksm_settle func I reckon at least: https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L299

GabyCT:
@grahamwhaley changes applied

PR #1891, opened 2019-08-02
crio_config: avoid duplicated runtime configuration for crio on bare-metal

Pennyzct:
If you are running k8s integration tests with crio in bare-metal machine, you may encounter duplicated runtime configuration in `/etc/crio/crio.conf`, when invoking `configure_crio_for_kata.sh` script.
```
# The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.
# The runtime to use is picked based on the runtime_handler provided by the CRI.
# If no runtime_handler is provided, the runtime will be picked based on the level
# of trust of the workload.

[crio.runtime.runtimes.runc]
runtime_path = "/usr/local/bin/crio-runc"
runtime_root = "/run/runc"
runtime_type = "oci"
[crio.runtime.runtimes.kata]
runtime_path = "/usr/local/bin/kata-runtime"
runtime_root = "/run/vc"
runtime_type= "oci"
[crio.runtime.runtimes.runc]
runtime_path = "/usr/local/bin/crio-runc"
runtime_root = "/run/runc"
runtime_type = "oci"
[crio.runtime.runtimes.kata]
runtime_path = "/usr/local/bin/kata-runtime"
runtime_root = "/run/vc"
runtime_type= "oci"
```
And that may lead to `crio.runtime.runtimes.runc has already been defined` error when launching crio service.
```
Aug 02 14:33:03 entos-thunderx2-desktop crio[127822]: time="2019-08-02T14:33:03+08:00" level=fatal msg="unable to decode configuration /etc/crio/crio.conf: Near line 219 (last key parsed 'crio.runtime.runtimes'): Key 'crio.runtime.runtimes.runc' has already been defined."
```

### DIFF
--- a/.ci/configure_crio_for_kata.sh
+++ b/.ci/configure_crio_for_kata.sh
@@ -10,20 +10,30 @@ set -o nounset
 set -o pipefail
 
 crio_config_file="/etc/crio/crio.conf"
+runc_flag="\/usr\/local\/bin\/crio-runc"
+kata_flag="\/usr\/local\/bin\/kata-runtime"
 
 minor_crio_version=$(crio --version | head -1 | cut -d '.' -f2)
 
 if [ "$minor_crio_version" -ge "12" ]; then
 	echo "Configure runtimes map for RuntimeClass feature"
 	echo "- Set runc as default runtime"
-	sudo sed -i 's!runtime_path =.*!runtime_path = "/usr/local/bin/crio-runc"!' "$crio_config_file"
-	sudo sed -i 's!runtime_type =.*!runtime_type = "oci"!' "$crio_config_file"
+	runc_configured=$(grep -q $runc_flag $crio_config_file; echo "$?")
+	if [[ $runc_configured -ne 0 ]]; then
+		sudo sed -i 's!runtime_path =.*!runtime_path = "/usr/local/bin/crio-runc"!' "$crio_config_file"
+		sudo sed -i '/runtime_root/d' "$crio_config_file"
+		sudo sed -i '/runtime_type/d' "$crio_config_file"
+	fi
 	echo "- Add kata-runtime to the runtimes map"
-	sudo sed -i '/crio-runc/a[crio.runtime.runtimes.kata]' "$crio_config_file"
-	sudo sed -i '/kata/aruntime_path = "/usr/local/bin/kata-runtime"' "$crio_config_file"
-	sudo sed -i 's!runtime_root =.*!runtime_root = "/run/vc"!' "$crio_config_file"
-	sudo sed -i '/crio-runc/a runtime_root = "/run/runc"' "$crio_config_file"
-	sudo sed -i '/crio-runc/a runtime_type = "oci"' "$crio_config_file"
+	kata_configured=$(grep -q $kata_flag $crio_config_file; echo "$?")
+	if [[ $kata_configured -ne 0 ]]; then
+		sudo sed -i '/crio-runc/a [crio.runtime.runtimes.kata]' "$crio_config_file"
+		sudo sed -i '/crio\.runtime\.runtimes\.kata/a runtime_path = "/usr/local/bin/kata-runtime"' "$crio_config_file"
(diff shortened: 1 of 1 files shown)


.ci/configure_crio_for_kata.sh
```diff
@@ -16,14 +16,19 @@ minor_crio_version=$(crio --version | head -1 | cut -d '.' -f2)
...
+		sudo sed -i '/runtime_type/d' "$crio_config_file"
+	fi
 	echo "- Add kata-runtime to the runtimes map"
-	sudo sed -i '/crio-runc/a[crio.runtime.runtimes.kata]' "$crio_config_file"
-	sudo sed -i '/kata/aruntime_path = "/usr/local/bin/kata-runtime"' "$crio_config_file"
-	sudo sed -i 's!runtime_root =.*!runtime_root = "/run/vc"!' "$crio_config_file"
-	sudo sed -i '/crio-runc/a runtime_root = "/run/runc"' "$crio_config_file"
-	sudo sed -i '/crio-runc/a runtime_type = "oci"' "$crio_config_file"
+	if [ -n $(grep -q "crio.runtime.runtimes.kata" "$crio_config_file") ]; then
+		sudo sed -i '/\/run\/runc/a [crio.runtime.runtimes.kata]' "$crio_config_file"
```
jodh-intel: I'd never noticed this issue with `sed` before - since this isn't a search-and-replace, you can't choose your quotes, hence all the backslashes ;(

.ci/configure_crio_for_kata.sh
```diff
@@ -16,14 +16,19 @@ minor_crio_version=$(crio --version | head -1 | cut -d '.' -f2)
 if [ "$minor_crio_version" -ge "12" ]; then
 	echo "Configure runtimes map for RuntimeClass feature"
 	echo "- Set runc as default runtime"
-	sudo sed -i 's!runtime_path =.*!runtime_path = "/usr/local/bin/crio-runc"!' "$crio_config_file"
-	sudo sed -i 's!runtime_type =.*!runtime_type = "oci"!' "$crio_config_file"
+	if [ -n $(grep -q "crio.runc" "$crio_config_file") ]; then
+		sudo sed -i 's!runtime_path =.*!runtime_path = "/usr/local/bin/crio-runc"!' "$crio_config_file"
```
chavafg: I think this single line should go outside the `if` as it is only replacing the default cri-o path with the one that we use, so even if this runs several times, we shouldn't get a duplicated entry.

Pennyzct: But,,, IMO, once you configured once, adding `kata-runtime` as another option. There will be two runtime_path, like the following:
```
[crio.runtime.runtimes.runc]
runtime_path = "/usr/local/bin/crio-runc"
runtime_root = "/run/runc"
runtime_type = "oci"
[crio.runtime.runtimes.kata]
runtime_path = "/usr/local/bin/kata-runtime"
runtime_root = "/run/vc"
runtime_type= "oci"
```
And this single line is instructed to replace any line starting with `runtime_path =`. 
So, it will also replace what's configured for `kata-runtime`, not just `runc`.

Pennyzct:
I got all green here~~~~ would you guys like to take a peek? @grahamwhaley @jodh-intel @chavafg 😋

.ci/configure_crio_for_kata.sh
grahamwhaley: I think @Pennyzct is right - the `grep` is specific to runc, but the `sed` is not, so would also change the kata path if it were already in the file (like it might be on a bare metal slave, which is what this PR is trying to fix).

chavafg:
I think we can merge as is. 

the `make install.config` that we use in our `.ci/install_crio.sh` should replace old config file, but we can later open a new PR for adding cleanups of cri-o config (and maybe other configs).

Pennyzct:
yep, we can put it in [cleanup_bare_metal_env.sh](https://github.com/kata-containers/tests/blob/master/integration/kubernetes/cleanup_bare_metal_env.sh), since all clean-up is only needed in vare-metal. ;)

PR #1877, opened 2019-07-31
expose-ip: replace hello-app with agnostic image agnhost:2.2

Pennyzct:
The `agnhost` image was created by k8s for testing purposes. one benefit is to be able to aggregate, like 10 images worth of functionality into 1 image.
Sub-command netexec is to start a HTTP server on given TCP / UDP ports with a few endpoints.
The `/echo` endpoint returns the given msg, e.g./echo?msg=echoed_msg. Therefore, we could use above functionality to achieve what hello-app is intended for.
Detailed info about `agnhost` image is [here](https://github.com/kubernetes/kubernetes/tree/master/test/images/agnhost).
Otherwise, hello-app is also single-arch image only for amd64, while agnhost is multi-arch image for multiple archs, including x86_64, arm64, s390x, etc.

PR #1894, opened 2019-08-02
CI: DNM Trigger build on s390x Jenkins CI

jschintag:
Just to trigger it.

Fixes: #12345
Signed-off-by: Jan Schintag <jan.schintag@de.ibm.com>

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -111,6 +111,8 @@ if [ -z "${METRICS_CI}" ]; then
 	fi
 fi
 
+echo "Dodging the fastpath"
+
 # Check if we can fastpath return/skip the CI
 # Specifically do this **after** we have potentially done the static
 # checks, as we always want to run those.
--- a/.ci/s390x/lib_kata_image_s390x.sh
+++ b/.ci/s390x/lib_kata_image_s390x.sh
@@ -7,7 +7,7 @@
 
 set -e
 
-OSBUILDER_DISTRO="ubuntu"
+OSBUILDER_DISTRO="alpine"
 AGENT_INIT="yes"
 
 #packaged kata agent haven't been supported in any mainstream distribution


ganeshmaharaj:
@jschintag i see s390x job is passing, but i have a feeling the change in README.md is not the one we need to make this part of the default CI run. What should be the next steps here?

/cc @chavafg

chavafg:
@ganeshmaharaj you are right, as this is only changing an MD, we are not running full validation... @jschintag I think you could change another type of file (bats, sh or go), so that full validation is triggered. That way we can see what other issues appear.

jschintag:
@ganeshmaharaj @chavafg Yeah i know, i just wanted to see if there would be any problems with the job in general, before i run the whole job.

jschintag:
/test-s390x

Lets see if that runs the whole job.

jschintag:
The Problem with the CI here #1877 comes from qemu. Here is the related Issue in the Runtime https://github.com/kata-containers/runtime/issues/1153#issuecomment-518129592.
The related PR for govmm is https://github.com/intel/govmm/pull/99.

@grahamwhaley Until this is fixed the s390x CI will keep failing.

````
08:57:21 Stderr: qemu-system-s390x: warning: 'apqci' requires 'ap'.
08:57:21 qemu-system-s390x: warning: 'apft' requires 'ap'.
08:57:21 qemu-system-s390x: -device virtio-serial-ccw,id=serial0,devno=: Property 'virtio-serial-ccw.devno' doesn't take value ''
````

jschintag:
Can i get an DNM Label for this PR?

grahamwhaley:
done @jschintag - are you not able to apply those yourself? You can also add a DNM to your PR subject line..

jschintag:
@grahamwhaley no, i can't assign labels.

grahamwhaley:
@jschintag - ah, I see, it is a github perms thing. You have to be in the 'team' for each repo to gain write powers.... you were only in the ci and jenkins repo teams. sadly I cannot find a github UI that lets me add you to multiple teams in a single click... I have added you to the tests repo team (so you should be able to set labels here now) - just request access to more repo/teams as required (which, I think you can do via the github UI...?)

grahamwhaley:
@jschintag - if the CI is all happy and done now, can we close this?
We can always open another if/when needed - leaving it here not only skews our stats a fraction, but the weekly herders probably have to read/scan over it once a week just to refresh and check what it is and where it is at :-)
thx

jschintag:
@grahamwhaley Since we currently don't have a s390x VM, we aren't really done. But i agree that we could just open a new one when we need it.

PR #1914, opened 2019-08-08
cleanup_k8s: ensure execution of cleanup no matter what

Pennyzct:
there are several scenarios where `cleanup_env.sh` wouldn't be invoked or be interrupted in the middle.
a) for now, `cleanup_env.sh` is only invoked on successful completion of the script.
Hence, we use `trap` to ensure the cleanup happens on any exit.
b) Sometimes, we failed on `kubeadm init`. on such point, the cluster network may not be initialized, which means no `cni0` interface, regardless of `flannel` interface. Hence, cleanup will exit in the middle.
```
Cannot find device "cni0"
```
So adding `|| true` ensures that the resulting compound command
always exits with status zero.
c) When `runc list` returned no stale container resource like the following:
```
ID          PID         STATUS      BUNDLE      CREATED     OWNER
```
the compound `&&` command would exit with `1`, leading the whole script exiting with none-zero status.
```
Failed at 41: ${SCRIPT_PATH}/cleanup_bare_metal_env.sh
```
So switching compound command with if statement. ;)

### DIFF
--- a/integration/kubernetes/cleanup_env.sh
+++ b/integration/kubernetes/cleanup_env.sh
@@ -28,14 +28,14 @@ sudo -E kubeadm reset -f --cri-socket="${cri_runtime_socket}"
 
 sudo systemctl stop "${cri_runtime}"
 
-sudo ip link set dev cni0 down
-sudo ip link set dev flannel.1 down
-sudo ip link del cni0
-sudo ip link del flannel.1
+sudo ip link set dev cni0 down || true
+sudo ip link set dev flannel.1 down || true
+sudo ip link del cni0 || true
+sudo ip link del flannel.1 || true
 
 # if CI run in bare-metal, we need a set of extra clean
 BAREMETAL="${BAREMETAL:-false}"
-if [ "${BAREMETAL}" == true ] && [ -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh" ];  then
+if [ "${BAREMETAL}" == true ] && [ -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh" ]; then
 	bash -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh"
 fi
 
--- a/integration/kubernetes/run_kubernetes_tests.sh
+++ b/integration/kubernetes/run_kubernetes_tests.sh
@@ -28,6 +28,9 @@ if [ "$KATA_HYPERVISOR" == "firecracker" ]; then
 	die "Kubernetes tests will not run with $KATA_HYPERVISOR"
 fi
 
+# Using trap to ensure the cleanup occurs when the script exists.
+trap '${kubernetes_dir}/cleanup_env.sh' EXIT
+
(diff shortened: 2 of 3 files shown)


integration/kubernetes/run_kubernetes_tests.sh
```diff
@@ -12,6 +12,9 @@ kubernetes_dir=$(dirname "$(readlink -f "$0")")
 cidir="${kubernetes_dir}/../../.ci/"
 source "${cidir}/lib.sh"
 
+# Using trap to ensure the cleanup occurs when the script exists.
+trap './cleanup_env.sh' EXIT
```
grahamwhaley: I wonder if the `./` path depends on what `pwd` is when the trap fires?
Might be safer to encode the full `${cidir}/cleanup_env.sh` here (I think that is the correct full path...)

grahamwhaley: I think EXIT will be enough - you might want to check if we need QUIT and KILL as well though.

integration/kubernetes/cleanup_env.sh
```diff
@@ -6,6 +6,10 @@
 #
 # This script is used to reset the kubernetes cluster
 
+set -o errexit
```
grahamwhaley: I'm wondering if we can just unset the `errexit`, so the script does not quit on any fails. wdyt?

integration/kubernetes/run_kubernetes_tests.sh
Pennyzct: yep, you're right~~~ all failed CI are telling me to use full path.😑

integration/kubernetes/cleanup_env.sh
Pennyzct: I'm fine with the cleanup script does not quit on any fails. ;). 
Then I will remove the second commit about `|| true`, but i'm gonna stick with the `if statement`, since i really don't want the exit status of this script `cleanup_bare_metal_env.sh` be non-zero. ;)

Pennyzct:
Hi @grahamwhaley 
Even I have deleted `set -o errexit`, then `cleanup_env.sh` will still fail on the `sudo ip link set dev cni0 down`
```
Cannot find device "cni0"
Failed at 35: sudo ip link set dev cni0 down
```
any thoughts???😶😥

Pennyzct:
Hi @grahamwhaley 
For the signal `QUIT` and `TERM`, they all have been taken care by `EXIT`, checking them out. ;)
And for `Kill`, This signal cannot be caught or ignored. `-9` is really wickedly powerful.  So no cleanup possible.

Pennyzct:
Hi guys
I almost got all green here~~ Power8's failure should be irrelevant. ;).

grahamwhaley:
@Pennyzct - if using `||true` is easier right now, and works ;-), I have no problem leaving the `-e` in place... it was more an observation.
Need another review - /cc @chavafg

PR #1926, opened 2019-08-14
k8s: e2e: fix how we get e2e log location

chavafg:
instead of specifying the location where we think the
`e2e.log` file is, use `find` to search for it under
`plugins/e2e` directory.

In addition, remove `SONOBUOY_KATA_YAML` variable 
which is no longer used in the script.

### DIFF
--- a/integration/kubernetes/e2e_conformance/run.sh
+++ b/integration/kubernetes/e2e_conformance/run.sh
@@ -63,7 +63,9 @@ run_sonobuoy() {
 
 	# Uncompress results
 	ls | grep tar.gz | xargs tar -xvf
-	e2e_result_log="${e2e_result_dir}/plugins/e2e/results/e2e.log"
+
+	# e2e results log will be on "plugins/e2e" directory
+	e2e_result_log=$(find ./plugins/e2e -name "e2e.log")
 	info "Results of the e2e tests can be found on: $e2e_result_log"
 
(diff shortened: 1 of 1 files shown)


GabyCT:
@chavafg you are missing `Fixes` at the commit message

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -65,7 +63,9 @@ run_sonobuoy() {
 
 	# Uncompress results
 	ls | grep tar.gz | xargs tar -xvf
-	e2e_result_log="${e2e_result_dir}/plugins/e2e/results/e2e.log"
+
+	# e2e results log will be on "plugins/e2e" directory
+	e2e_result_log=$(find ./plugins/e2e -name "e2e.log")
```
devimc: what happen if there are more than one `e2e.log` file?

chavafg: this is executed on a temp directory, which changes every time it is executed, so there shouldn't be more than one

devimc: thanks

grahamwhaley: using `find` feels... odd - presumably the log file moved path recently, and might do so again. So, sure, it's a fix - let's see if it 'holds water'...

Issue #1936, opened 2019-08-27
k8s: cni ip addresses not being released when running with cri-o


chavafg:
Actually not a bug in our CI scripts, but in the flannel plugin. There is a workaround we can try, which is to add the `cniVersion` to the cni configuration under `/etc/cni/net.d`

Related:
https://github.com/cri-o/cri-o/issues/2410
https://github.com/coreos/flannel/pull/1135

PR #1937, opened 2019-08-27
k8s: Add cniVersion to flannel config file

chavafg:
`cniVersion` is needed inside the flannel config file. If it
is not provided, cni has issues releasing ip adresses after
pods are deleted.

Fixes: #1936.

In addition:
- add `RUNTIME` and `RUNTIME_PATH` default values 
to the kubernetes `init.sh` script.
- use sudo for docker calls as some environments may not be 
able to call docker without using sudo.

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -67,9 +67,10 @@ esac
 check_processes
 
 # Remove existing CNI configurations:
-sudo rm -rf /var/lib/cni/networks/*
-sudo rm -rf /etc/cni/net.d/*
+cni_config_dir="/etc/cni/net.d"
 cni_interface="cni0"
+sudo rm -rf /var/lib/cni/networks/*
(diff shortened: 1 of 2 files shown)


integration/kubernetes/init.sh
```diff
@@ -115,6 +116,14 @@ kubectl apply -f "$network_plugin_config"
 # we need to ensure a few specific pods ready and running
 wait_pods_ready
 
+# Add 'cniVersion' to the flannel configuration file.
+# Fix needed just for flannel to let cni release ip addresses,
+# we should remove after https://github.com/coreos/flannel/pull/1135
+# gets merged and we update flannel_version above.
```
grahamwhaley: If we've not already, maybe we should have a kata issue to track, so we don't forget (we'll get reminded when the stalebot starts whining about it...)

Issue #1965, opened 2019-09-24
Preliminary support for rust agent ci


yyyeerbo:
https://github.com/kata-containers/kata-containers/issues/47

PR #1966, opened 2019-09-24
Add rust support to CI

yyyeerbo:
This patch adds preliminary support to rust agent.

Fixes: #1965

### DIFF
--- a/.ci/install_rust.sh
+++ b/.ci/install_rust.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Ant Financial
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+[ -n "${KATA_DEV_MODE:-}" ] && exit 0
+
+cidir=$(dirname "$0")
+rustarch=$(${cidir}/kata-arch.sh --rust)
+release="nightly"
+
+if ! command -v rustup > /dev/null; then
+	curl https://sh.rustup.rs -sSf | sh
+fi
+
+rustup toolchain install ${release}-${rustarch}-unknown-linux-gnu
+rustup default ${release}-${rustarch}-unknown-linux-gnu
+rustup target install ${rustarch}-unknown-linux-musl
+ln -sf /usr/bin/g++ /bin/musl-g++
--- a/.ci/setup_env_ubuntu.sh
+++ b/.ci/setup_env_ubuntu.sh
@@ -43,6 +43,25 @@ declare -A packages=( \
 	[redis]="redis-server" \
 )
 
+rust_agent_pkgs=()
+rust_agent_pkgs+=("build-essential")
+rust_agent_pkgs+=("g++")
+rust_agent_pkgs+=("make")
+rust_agent_pkgs+=("cmake")
+rust_agent_pkgs+=("automake")
+rust_agent_pkgs+=("autoconf")
+rust_agent_pkgs+=("m4")
+rust_agent_pkgs+=("libc6-dev")
+rust_agent_pkgs+=("libstdc++-8-dev")
+rust_agent_pkgs+=("coreutils")
+rust_agent_pkgs+=("binutils")
+rust_agent_pkgs+=("debianutils")
+rust_agent_pkgs+=("gcc")
+rust_agent_pkgs+=("musl")
+rust_agent_pkgs+=("musl-dev")
+rust_agent_pkgs+=("musl-tools")
+rust_agent_pkgs+=("git")
+
 main()
 {
 	local setup_type="$1"
@@ -61,6 +80,11 @@ main()
 			info "The following package will be installed: $pkgs"
 			pkgs_to_install+=" $pkgs"
 		done
+	fi
+
+	# packages for rust agent, build on 18.04 or later
+	if [ ! "${VERSION_ID}" < "18.04" ]; then
+		pkgs_to_install+=" ${rust_agent_pkgs[@]}"
 	fi
 
 	chronic sudo -E apt -y install $pkgs_to_install
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -59,6 +59,7 @@ long_options=(
 	[files]="Check files"
 	[force]="Force a skipped test to run"
 	[golang]="Check '.go' files"
+	[rust]="Check '.rs' files"
 	[help]="Display usage statement"
 	[json]="Check JSON files"
 	[labels]="Check labels databases"
@@ -223,6 +224,9 @@ EOT
 
 static_check_go_arch_specific()
 {
+	if [ "${RUST_AGENT:-}" == "yes" ]; then
+		return
+	fi
 	local go_packages
 	local submodule_packages
 	local all_packages
@@ -297,6 +301,17 @@ static_check_go_arch_specific()
 	echo "$dirs" | sed 's/^ *//g' | tr ' ' '\n'
 
 	eval "$linter" "${linter_args}" "$dirs"
+}
+
+static_check_rust_arch_specific()
+{
+	if [ "${RUST_AGENT:-}" != "yes" ]; then
+		return
+	fi
+
+	find . -type f -name "*.rs"  | egrep -v "target/|grpc-rs/|protocols/" | xargs rustfmt --check > /dev/null 2>&1
+
+	[ $? -eq 0 ] || die "crate not formatted by rustfmt."
 }
 
 # Check the "versions database".
@@ -381,6 +396,9 @@ static_check_license_headers()
 		--exclude="*.yml" \
 		--exclude="go.mod" \
 		--exclude="go.sum" \
+		--exclude="*.lock" \
+		--exclude="grpc-rs/*" \
+		--exclude="target/*" \
 		-EL "\<${pattern}\>" \
 		$files || true)
 
@@ -482,7 +500,7 @@ static_check_docs()
 	local new_urls
 	local url
 
-	all_docs=$(git ls-files "*.md" | grep -v "^vendor/" | sort || true)
+	all_docs=$(git ls-files "*.md" | grep -v "/\(vendor\|grpc-rs\|target\)/" | sort || true)
 
 	if [ "$specific_branch" = "true" ]
 	then
@@ -705,7 +723,7 @@ static_check_files()
 	then
 		info "Checking all files in $branch branch"
 
-		files=$(git ls-files | grep -v "^(.git|vendor)/" || true)
+		files=$(git ls-files | egrep -v "/(.git|vendor|grpc-rs|target)/" || true)
 	else
 		info "Checking local branch for changed files only"
 
@@ -773,6 +791,9 @@ static_check_files()
 # - Ensure vendor metadata is valid.
 static_check_vendor()
 {
+	if [ "${RUST_AGENT:-}" == "yes" ]; then
+		return
+	fi
 	local files
 	local vendor_files
 	local result
@@ -826,7 +847,7 @@ static_check_xml()
 	local all_xml
 	local files
 
-	all_xml=$(git ls-files "*.xml" | grep -v "^vendor/" | sort || true)
+	all_xml=$(git ls-files "*.xml" | grep -v "/\(vendor\|grpc-rs\|target\)/" | sort || true)
 
 	if [ "$specific_branch" = "true" ]
 	then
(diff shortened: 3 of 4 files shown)


.ci/static-checks.sh
```diff
@@ -299,6 +299,29 @@ static_check_go_arch_specific()
 	eval "$linter" "${linter_args}" "$dirs"
 }
 
+static_check_rust_arch_specific()
+{
+	pushd ${TRAVIS_BUILD_DIR}
```
chavafg: what about if running on non-travis environment?

.ci/install_rust.sh
```diff
@@ -0,0 +1,29 @@
...
+
+cidir=$(dirname "$0")
+rustarch=$(uname -m)
+release="nightly"
+
+if [ "${rustarch}" == "ppc64le" ]; then
+	rustarch="powerpc64le"
+fi
+
+# rust specific pacakges
```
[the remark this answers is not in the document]
yyyeerbo: fixed

.ci/static-checks.sh
yyyeerbo: fixed

.ci/install_rust.sh
```diff
@@ -0,0 +1,29 @@
...
+cidir=$(dirname "$0")
+rustarch=$(uname -m)
+release="nightly"
+
+if [ "${rustarch}" == "ppc64le" ]; then
+	rustarch="powerpc64le"
+fi
+
+# rust specific packages
+sudo -E apt-get update && apt-get install -y build-essential g++ \
```
jcvenegas: This are distro specific installs could you  move this to https://github.com/kata-containers/tests/blob/master/.ci/setup_env_ubuntu.sh, and as bonus would be nice to add to all the distros.

.ci/install_rust.sh
```diff
@@ -0,0 +1,29 @@
...
+binutils debianutils gcc musl musl-dev musl-tools
+
+if ! command -v rustup > /dev/null; then
+	curl https://sh.rustup.rs -sSf | sh
+fi
+
+rustup toolchain install ${release}-${rustarch}-unknown-linux-gnu
+rustup default ${release}-${rustarch}-unknown-linux-gnu
+rustup target install ${rustarch}-unknown-linux-musl
+ln -sf /usr/bin/g++ /bin/musl-g++
```
jcvenegas: This looks distro dependent as well

.ci/static-checks.sh
```diff
@@ -299,6 +299,27 @@ static_check_go_arch_specific()
 	eval "$linter" "${linter_args}" "$dirs"
 }
 
+static_check_rust_arch_specific()
+{
+	local -r libs=("rustjail" "oci")
+	local -r bins=(".")
+	for lib in ${libs[@]}; do
+		rustfmt --check ${lib}/src/lib.rs > /dev/null 2>&1
```
jcvenegas: This depend on the repo this is run  where is run and will work only in agent  (as we expect but should not be this then in the agent repo

.ci/static-checks.sh
```diff
@@ -299,6 +299,27 @@ static_check_go_arch_specific()
 	eval "$linter" "${linter_args}" "$dirs"
 }
 
+static_check_rust_arch_specific()
```
jcvenegas: I dont see where is called this function

.ci/static-checks.sh
```diff
@@ -1090,6 +1114,23 @@ main()
 			fi
 		fi
 
+		if [ "${RUST_AGENT:-}" == "yes" ]; then
+			# skip checks specific to go
+			if [[ "${func}" =~ "go_arch_specific" ]]; then
```
jcvenegas: If someone  modify this function names, the rust agent CI will be broken easily

.ci/static-checks.sh
```diff
@@ -299,6 +299,27 @@ static_check_go_arch_specific()
 	eval "$linter" "${linter_args}" "$dirs"
 }
 
+static_check_rust_arch_specific()
+{
+	local -r libs=("rustjail" "oci")
```
jcvenegas: I know this is only for rust agent but I wonder if would be good to make this a ENV variable that the script checks and in an repo we are free to set  what libs we want to analyse  or not.

.ci/install_rust.sh
yyyeerbo: cmake/musl-libc is missed/outdated in some other distro, that why ubuntu is used for rust agent.

maybe we will resolve cmake/musl-libc in other distros later, as what we do in osbuilder

.ci/install_rust.sh
yyyeerbo: no, it is not

.ci/static-checks.sh
yyyeerbo: in the line type -F xxxx

.ci/static-checks.sh
yyyeerbo: I don't understand it. directory problem?  

The functions for go also assume that script 's work directory is the repo directory.

.ci/static-checks.sh
yyyeerbo: fixed

.ci/static-checks.sh
yyyeerbo: fixed

yyyeerbo:
fyi 
https://github.com/kata-containers/osbuilder/pull/361
https://github.com/alipay/kata-rust-agent/pull/20

.ci/install_rust.sh
yyyeerbo: some pakages(libstdc++-8-dev) are only in bionic. It is needed to build rust agent. move this into specific distro script error out.

.ci/install_rust.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ "${rustarch}" == "ppc64le" ]; then
+	rustarch="powerpc64le"
+fi
+
+if ! command -v rustup > /dev/null; then
+	curl https://sh.rustup.rs -sSf | sh
+fi
+
+# sudo -E apt-get update
+# sudo -E apt-get -y install build-essential g++ make cmake automake autoconf m4 libc6-dev libstdc++-8-dev coreutils binutils debianutils gcc musl musl-dev musl-tools git
```
chavafg: remove these commented lines?

.ci/static-checks.sh
chavafg: should this function be added to https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh#L1020-L1040

chavafg: in addition, this may fail if it runs on other distro than ubuntu, so I would recommend to add a validation to skip this if on other distro.

.ci/install_rust.sh
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+cidir=$(dirname "$0")
+rustarch=$(uname -m)
+release="nightly"
+
+if [ "${rustarch}" == "ppc64le" ]; then
+	rustarch="powerpc64le"
```
jodh-intel: You could consider adding a `--rust` option to https://github.com/kata-containers/tests/blob/master/.ci/kata-arch.sh

.ci/install_rust.sh
```diff
@@ -0,0 +1,27 @@
...
+
+if ! command -v rustup > /dev/null; then
+	curl https://sh.rustup.rs -sSf | sh
+fi
+
+# sudo -E apt-get update
+# sudo -E apt-get -y install build-essential g++ make cmake automake autoconf m4 libc6-dev libstdc++-8-dev coreutils binutils debianutils gcc musl musl-dev musl-tools git
+
+rustup toolchain install ${release}-${rustarch}-unknown-linux-gnu
+rustup default ${release}-${rustarch}-unknown-linux-gnu
```
jodh-intel: You might want to add a check for `KATA_DEV_MODE` to avoid surprising people running this "outside of the CI":

- https://github.com/kata-containers/tests/#developer-mode

.ci/setup_env_ubuntu.sh
```diff
@@ -43,6 +43,8 @@ declare -A packages=( \
 	[redis]="redis-server" \
 )
 
+rust_agent_pkgs="build-essential g++ make cmake automake autoconf m4 libc6-dev libstdc++-8-dev coreutils binutils debianutils gcc musl musl-dev musl-tools git"
```
jodh-intel: How about:

```sh
rust_agent_pkgs=()
rust_agent_pkgs+=(build-essential)
rust_agent_pkgs+=(g++)
rust_agent_pkgs+=(make)
```

That makes it easier to read the list, easier to diff and easier to sort alphabetically :smile:

.ci/setup_env_ubuntu.sh
```diff
@@ -63,6 +65,11 @@ main()
 		done
 	fi
 
+	# packages for rust agent, build on 18.04 or later
+	if [ "${VERSION_ID}" >= "18.04" ]; then
```
jodh-intel: I don't understand - if building on an older version of Ubuntu, you don't need these packages? Do you mean we should exit the script with an error if version `<= 18.04`?

Also, this test doesn't work. You need to do something like one of the following:

```sh
result=$(echo "$VERSION_ID > 18.04" | bc)
```

or,

```sh
int_version=$(echo $VERSION_ID|tr -d \.)

[ $int_version -ge 1804 ] && ...
```

.ci/static-checks.sh
jodh-intel: I think @jcvenegas is saying (and I agree) that you shouldn't be assuming `$dir/src/lib.rs`. Instead, why not do something like:

```sh
find . -type f -name "*.rs" | egrep -v "grpc-rs/|target/" | xargs rustfmt ...
```

Admittedly, that still hard-codes a small amount of rust agent knowledge (the `grpc-rs`), but that problem could be overcome easily by moving `grpc-rs/` into `vendor/grpc-rs/` and updating the new agents `Cargo.toml` to specify:

```toml
grpcio = { path = "vendor/grpc-rs" }
```

yyyeerbo: grpc-rs has dozens of submodules, it might be easier to use it as submodule instead of a vendor directory

.ci/static-checks.sh
yyyeerbo: no, it won't

.ci/setup_env_ubuntu.sh
yyyeerbo: Some packages are missed in older version of ubuntu. They are needed for build rust agent. I think the string comparison is ok

.ci/static-checks.sh
yyyeerbo: lib crates have src/lib.rs, bin crates have src/main.rs. I don't understand what the problem is

.ci/setup_env_ubuntu.sh
yyyeerbo: -_-. string comparison has no >=. The ! version < "18.04" works

.ci/static-checks.sh
yyyeerbo: We won't set RUST_AGENT to yes on other distro, if it is set, it's a bug, this should fail

.ci/static-checks.sh
yyyeerbo: > I think @jcvenegas is saying (and I agree) that you shouldn't be assuming `$dir/src/lib.rs`. Instead, why not do something like:
> 
> ```shell
> find . -type f -name "*.rs" | egrep -v "grpc-rs/|target/" | xargs rustfmt ...
> ```
> 
> Admittedly, that still hard-codes a small amount of rust agent knowledge (the `grpc-rs`), but that problem could be overcome easily by moving `grpc-rs/` into `vendor/grpc-rs/` and updating the new agents `Cargo.toml` to specify:
> 
> ```toml
> grpcio = { path = "vendor/grpc-rs" }
> ```

I see. clone submodule into vendor/grpc-rs.

yyyeerbo: fixed

.ci/install_rust.sh
yyyeerbo: fixed

.ci/install_rust.sh
yyyeerbo: fixed

.ci/install_rust.sh
yyyeerbo: fixed

.ci/setup_env_ubuntu.sh
yyyeerbo: fixed

.ci/setup_env_ubuntu.sh
yyyeerbo: fxied

.ci/static-checks.sh
chavafg: ok, agree

jodh-intel:
Restarted two of the CI's which failed due to what looks like transitory issues.

yyyeerbo:
-_-. Time out when downloading kernel

PR #1982, opened 2019-10-07
ci: Move configuration-qemu-virtiofs.toml to correct path

### LINKED ISSUE #1981: Move configuration-qemu-virtiofs.toml to correct path

GabyCT:
This will move the configuration-qemu-virtiofs.toml to the correct path.

Fixes #1981

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -65,7 +65,10 @@ if [ "$KATA_HYPERVISOR" = "nemu" ]; then
 fi
 
 if [ "$KATA_HYPERVISOR" = "qemu" ] && [ "$experimental_qemu" == "true" ]; then
+	runtime_repo="github.com/kata-containers/runtime"
+
 	echo "Enable qemu virtiofs configuration.toml"
+	sudo mv "$GOPATH/src/${runtime_repo}/cli/config/configuration-qemu-virtiofs.toml" "${PKGDEFAULTSDIR}"
 	sudo mv "${PKGDEFAULTSDIR}/configuration-qemu-virtiofs.toml" "${PKGDEFAULTSDIR}/configuration.toml"
 fi
 


.ci/install_runtime.sh
```diff
@@ -65,7 +65,11 @@ if [ "$KATA_HYPERVISOR" = "nemu" ]; then
 fi
 
 if [ "$KATA_HYPERVISOR" = "qemu" ] && [ "$experimental_qemu" == "true" ]; then
+	runtime_repo="github.com/kata-containers/runtime"
+	runtime_repo_dir="$GOPATH/src/${runtime_repo}"
```
devimc: useless variable? I guess you forgot to use it in the following lines?

GabyCT: @devimc fix it

.ci/install_runtime.sh
```diff
@@ -65,7 +65,10 @@ if [ "$KATA_HYPERVISOR" = "nemu" ]; then
 fi
 
 if [ "$KATA_HYPERVISOR" = "qemu" ] && [ "$experimental_qemu" == "true" ]; then
+	runtime_repo="github.com/kata-containers/runtime"
+
 	echo "Enable qemu virtiofs configuration.toml"
+	sudo mv "$GOPATH/src/${runtime_repo}/cli/config/configuration-qemu-virtiofs.toml" "${PKGDEFAULTSDIR}"
 	sudo mv "${PKGDEFAULTSDIR}/configuration-qemu-virtiofs.toml" "${PKGDEFAULTSDIR}/configuration.toml"
```
devimc: well, now this is odd, I think you can remove this step, what about move `configuration-qemu-virtiofs.toml` directly to `"${PKGDEFAULTSDIR}/configuration.toml"` ?

```sh
sudo mv "$GOPATH/src/${runtime_repo}/cli/config/configuration-qemu-virtiofs.toml" "${PKGDEFAULTSDIR}/configuration.toml"
```

GabyCT: @devimc because when we do the `kata-runtime kata-check` this will be failing as it complains that `configuration-qemu-virtiofs.toml` is missing at ${PGDEFAULTSDIR}

.ci/install_runtime.sh
```diff
@@ -65,7 +65,10 @@ if [ "$KATA_HYPERVISOR" = "nemu" ]; then
 fi
 
 if [ "$KATA_HYPERVISOR" = "qemu" ] && [ "$experimental_qemu" == "true" ]; then
+	runtime_repo="github.com/kata-containers/runtime"
+
 	echo "Enable qemu virtiofs configuration.toml"
+	sudo mv "$GOPATH/src/${runtime_repo}/cli/config/configuration-qemu-virtiofs.toml" "${PKGDEFAULTSDIR}"
```
chavafg: I think this should be:

```
sudo mv "${PKGDEFAULTSDIR}/configuration-qemu-virtiofs.toml" "${PKGDEFAULTSDIR}/configuration.toml"
```
and remove the below one.

GabyCT: @chavafg the problem with that is that `configuration-qemu-virtiofs.toml` is not found at `${PKGDEFAULTSDIR}`

chavafg: it should be there after running `make install`, which the scripts run before these lines, I tried locally and I see the file there:
```
$ ls -l /usr/share/defaults/kata-containers
total 108
-rw-r--r-- 1 root root  9954 Oct  7 15:46 configuration-acrn.toml
-rw-r--r-- 1 root root 14305 Oct  7 15:46 configuration-fc.toml
-rw-r--r-- 1 root root 16672 Oct  7 15:46 configuration-nemu.toml
-rw-r--r-- 1 root root 17201 Oct  7 15:46 configuration-qemu-virtiofs.toml
-rw-r--r-- 1 root root 17221 Oct  7 15:46 configuration-qemu.toml
-rw-r--r-- 1 root root 17166 Oct  7 15:46 configuration.toml
```

GabyCT: @chavafg, I also tried locally and fails because is saying that `configuration-qemu-virtiofs.toml` is missing at that path, I ran the `.ci/setup.sh`

chavafg:
@GabyCT could you reproduced the issue? I see that the CI for virtiofs runs correctly, so I think this is not needed anymore, but wanted to check if you can reproduce it locally.

Issue #1996, opened 2019-10-12
.ci/install_rust.sh need sudo to work correctly


yyyeerbo:
The line ln -svf /bin/g++ /bin/musl-g++ needs sudo to work correctly

PR #1997, opened 2019-10-12
rust-agent: Need sudo to work correctly

yyyeerbo:
Fixes: #1996

### DIFF
--- a/.ci/install_rust.sh
+++ b/.ci/install_rust.sh
@@ -11,12 +11,15 @@ set -e
 cidir=$(dirname "$0")
 rustarch=$(${cidir}/kata-arch.sh --rust)
 release="nightly"
+# recent functional version
+version="2019-10-04"
 
 if ! command -v rustup > /dev/null; then
 	curl https://sh.rustup.rs -sSf | sh
 fi
 
-rustup toolchain install ${release}-${rustarch}-unknown-linux-gnu
-rustup default ${release}-${rustarch}-unknown-linux-gnu
+rustup toolchain install ${release}-${version}-${rustarch}-unknown-linux-gnu
+rustup default ${release}-${version}-${rustarch}-unknown-linux-gnu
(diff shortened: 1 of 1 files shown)


.ci/install_rust.sh
```diff
@@ -11,12 +11,15 @@ set -e
 cidir=$(dirname "$0")
 rustarch=$(${cidir}/kata-arch.sh --rust)
 release="nightly"
+# recent functional version
+version="2019-10-04"
```
jodh-intel: OOI, why this date? What happens with newer nightly builds?

Also, the "normal" way to specify versions is to add them to the "versions database" file (which you then query using the `yq` command):

- https://github.com/kata-containers/tests/blob/master/versions.yaml

Since rust is a language, we might want to add the `language:` section to the file above, like we have in the "master" versions database in the runtime:

- https://github.com/kata-containers/runtime/blob/master/versions.yaml#L266..L282

But I'll let @chavafg decide on how best this should be handled.

yyyeerbo: :) rust after 2019-10-04 has no rustfmt. So we cannot run static_checks which uses rustfmt. This is a temperate fix/workaround. The nightly feature is stable in 1.40, we probably can use some stable version after 1.40

jodh-intel: Oh! Is the tool becoming part of the compiler or cargo?

yyyeerbo: > Oh! Is the tool becoming part of the compiler or cargo?

;) I don't know..

chavafg: As @jodh-intel mentions, I think the best option would be to add an entry on https://github.com/kata-containers/runtime/blob/master/versions.yaml#L266..L282
to manage the version of rust that we will be using for Kata.
Then we can read it using [`get_version()`](https://github.com/kata-containers/tests/blob/master/.ci/lib.sh#L135)

Issue #2044, opened 2019-11-04
Add build_install_{cmake, musl} to lib.sh


yyyeerbo:
Add build_install_{cmake, musl} to lib.sh

Then it can be used in setup_env_xxxx.sh

PR #2045, opened 2019-11-04
ci: add build_install_{cmake, musl} to lib.sh

yyyeerbo:
Fixes: #2044

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -328,6 +328,50 @@ build_install_parallel() {
 	rm -rf "$gnu_parallel_dir" "$gnu_parallel_tar_pkg"
 }
 
+build_install_cmake() {
+	local version=$(get_version "externals.cmake.version")
+	local url=$(get_version "externals.cmake.url")
+	[ "${url}" != "null" ] && [ "${version}" != "null" ]
+	local file="cmake-${version}.tar.gz"
+	local dir="cmake-${version}"
+	local download="${url}/releases/download/v${version}/${file}"
+	local tdir=$(mktemp -d)
+
+	pushd "${tdir}"
+	curl -sLO "${download}"
+	tar -zxf "${file}"
+	cd "${dir}"
+	./bootstrap
+	make
+	sudo make install
+	popd
+	rm -fr "${tdir}"
+}
+
+build_install_musl() {
+	local version=$(get_version "externals.musl.version")
+	local url=$(get_version "externals.musl.url")
+	[ "${url}" != "null" ] && [ "${version}" != "null" ]
+	local file="musl-${version}.tar.gz"
+	local dir="musl-${version}"
+	local download="${url}/releases/${file}"
+	local tdir=$(mktemp -d)
+	local muslarch=$(uname -m)
+
+	pushd "${tdir}"
+	curl -sLO "${download}"
+	tar -zxf "${file}"
+	cd "${dir}"
+	sed -i "s/^ARCH = .*/ARCH = ${muslarch}/g"  dist/config.mak
+	./configure
+	make
+	sudo make install
+	sudo echo '"/usr/local/musl/lib"' > /etc/ld-musl-${muslarch}.path
+	export PATH=${PATH}:/usr/local/musl/bin
+	popd
+	rm -fr "${tdir}"
+}
+
 check_git_version() {
 	result="true"
 


.ci/lib.sh
```diff
@@ -327,6 +327,54 @@ build_install_parallel() {
...
 
+build_install_cmake() {
+	local version=$(get_version "externals.cmake.version")
+	local url=$(get_version "externals.cmake.url")
+	[ "${url}" != "null" ] && [ "${version}" != "null" ]
+	local file="cmake-${version}.tar.gz"
+	local dir="cmake-${version}"
+	local download="${url}/releases/download/v${version}/${file}"
+
+	pushd /root
```
jodh-intel: This assumes the script is running as `root` which won't be the case for the CI and shouldn't be the case for devs either. How about something like:

```sh
build_dir=$(mktemp -d)
pushd "$build_dir" &>/dev/null

    :

popd &>/dev/null
rm -rf "$build_dir"
```

.ci/lib.sh
```diff
@@ -327,6 +327,54 @@ build_install_parallel() {
...
+	wget "${download}"
+	tar -zxf "${file}"
+	cd "${dir}"
+	./bootstrap
+	make
+	make install
```
jodh-intel: `sudo make install`.

.ci/lib.sh
```diff
@@ -327,6 +327,54 @@ build_install_parallel() {
...
+build_install_cmake() {
+	local version=$(get_version "externals.cmake.version")
+	local url=$(get_version "externals.cmake.url")
+	[ "${url}" != "null" ] && [ "${version}" != "null" ]
+	local file="cmake-${version}.tar.gz"
+	local dir="cmake-${version}"
+	local download="${url}/releases/download/v${version}/${file}"
+
+	pushd /root
+	wget "${download}"
```
jodh-intel: Please use `curl` as this is our preferred downloader.

.ci/lib.sh
```diff
@@ -327,6 +327,54 @@ build_install_parallel() {
...
+	tar -zxf "${file}"
+	cd "${dir}"
+	./bootstrap
+	make
+	make install
+	popd
+}
+
+build_install_musl() {
```
jodh-intel: This function is extremely similar to the one above. I'd do one of the following:

- Create a generic function that can download a source package and then make both these functions call it.
- Steal existing code from one of the other repos that does this.
  I'm *sure* we have/had a function to download build and install a generic source package in `tests/.ci/` or similar but cannot find it atm. /cc @chavafg, @grahamwhaley.

jodh-intel:
Any update on this @yyyeerbo?

yyyeerbo:
> Any update on this @yyyeerbo?

Let me recall what this does. And then refactor it :)

.ci/lib.sh
yyyeerbo: It is not generic, we will have 
case $1 in
      musl)
;;
     camke)
;; 
in scripts to build download url. Since download url is quit different

Do you have some idea to make it simple?

jodh-intel: Let's defer that for a follow-up PR maybe.

.ci/lib.sh
```diff
@@ -328,6 +328,54 @@ build_install_parallel() {
...
+		*)
+			muslarch=$(uname -m)
+			;;
+	esac
+
+	pushd $(mktemp -d)
+	curl -sLO "${download}"
+	tar -zxf ${file}
+	cd "${dir}"
+	sed -i "s/^ARCH = .*/ARCH = ${muslarch}/g"
```
devimc: sed without a file or pipe, this is weird, I think this won't work

.ci/lib.sh
```diff
@@ -328,6 +328,54 @@ build_install_parallel() {
...
+	local version=$(get_version "externals.musl.version")
+	local url=$(get_version "externals.musl.url")
+	[ "${url}" != "null" ] && [ "${version}" != "null" ]
+	local file="musl-${version}.tar.gz"
+	local dir="musl-${version}"
+	local download="${url}/releases/${file}"
+
+	case "$(uname -m)" in
+		"ppc64le")
+			muslarch=powerpc64
```
devimc: I thought ppc was not supported

.ci/lib.sh
```diff
@@ -328,6 +328,54 @@ build_install_parallel() {
...
+			;;
+		*)
+			muslarch=$(uname -m)
+			;;
+	esac
+
+	pushd $(mktemp -d)
+	curl -sLO "${download}"
+	tar -zxf ${file}
```
devimc: missing double quotes

.ci/lib.sh
```diff
@@ -328,6 +328,54 @@ build_install_parallel() {
...
 
+build_install_cmake() {
+	local version=$(get_version "externals.cmake.version")
+	local url=$(get_version "externals.cmake.url")
+	[ "${url}" != "null" ] && [ "${version}" != "null" ]
+	local file="cmake-${version}.tar.gz"
+	local dir="cmake-${version}"
+	local download="${url}/releases/download/v${version}/${file}"
+
+	pushd $(mktemp -d)
```
devimc: save this in a variable and remove the directory once it's not needed

.ci/lib.sh
```diff
@@ -328,6 +328,54 @@ build_install_parallel() {
...
+	case "$(uname -m)" in
+		"ppc64le")
+			muslarch=powerpc64
+			;;
+		*)
+			muslarch=$(uname -m)
+			;;
+	esac
+
+	pushd $(mktemp -d)
```
devimc: same here, remove directory once it's not needed

.ci/lib.sh
```diff
@@ -328,6 +328,54 @@ build_install_parallel() {
...
+	curl -sLO "${download}"
+	tar -zxf ${file}
+	cd "${dir}"
+	sed -i "s/^ARCH = .*/ARCH = ${muslarch}/g"
+	./configure
+	make
+	sudo make install
+	sudo echo "\"/usr/local/musl/lib\"" > /etc/ld-musl-${muslarch}.path
```
devimc: use single quotes (') - `'"/usr/local/musl/lib"'`

PR #2051, opened 2019-11-05
ci: `install_rust.sh` fixes

chavafg:
`get_version` is a function that lives in `.ci/lib.sh`, so we need
to source that file in order to use the function.

In addition, after `rust` is installed, we need to source the `$HOME/.profile`
which adds the `$HOME/.cargo/bin` directory to the `PATH`.


chavafg:
yeah, by default `.cargo/bin` is not in the `PATH`, and the rust install script adds the `export` line to the `.profile` file, so we need to source it as we wont be able to refresh our shell environment.

PR #2079, opened 2019-11-12
static-checks: Install yamllint as a pre-requisite

### LINKED ISSUE #1659: static-checks: we need to be sure that yamllint is installed before running static checks
some static checks require the `yamllint` tool, we need to be sure that it is installed before running those checks

GabyCT:
This will enable to install yamllint in different linux distributions in
order to run static-checks.

Fixes #1659

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -314,6 +314,25 @@ static_check_rust_arch_specific()
 	[ $? -eq 0 ] || die "crate not formatted by rustfmt."
 }
 
+# Install yamllint in the different Linux distributions
+install_yamllint()
+{
+	source /etc/os-release || source /usr/lib/os-release
+
+	package="yamllint"
+
+	case "$ID" in
+		centos|rhel) sudo yum -y install $package ;;
+		debian|ubuntu) sudo apt-get -y install $package ;;
+		fedora) sudo dnf -y install $package ;;
+		*) die "Please install yamllint on $ID" ;;
+	esac
+
+	have_yamllint_cmd=$(command -v "$yamllint_cmd" || true)
+
+	[ -z "$have_yamllint_cmd" ] && info "Cannot install $package" && return
+}
+
 # Check the "versions database".
 #
 # Some repositories use a versions database to maintain version information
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -327,7 +347,8 @@ static_check_versions()
 	if [ -n "$have_yamllint_cmd" ]; then
 		eval "$yamllint_cmd" "$db"
 	else
-		info "Cannot check versions as $yamllint_cmd not available"
+		info "Installing yamllint"
```
grahamwhaley: Should we do this check first, and then install, and then re-try the `eval` iyswim - otherwise, I think on first invocation it will install yamllint, but skip the requested check??

GabyCT:
@grahamwhaley changes applied

.ci/static-checks.sh
```diff
@@ -322,6 +337,11 @@ static_check_versions()
 {
 	local db="versions.yaml"
 
+	if [ -z "$have_yamllint_cmd" ]; then
+		info "Installing yamllint"
+		install_yamllint
+	fi
```
grahamwhaley: Do you need to re-do the `have_yamllint_cmd=$(command -v "$yamllint_cmd" || true)` again, now it is hopefully installed - otherwise, will you still fail the check below?

GabyCT: @grahamwhaley , I added that step at the install_yamllint function

.ci/static-checks.sh
```diff
@@ -314,6 +314,23 @@ static_check_rust_arch_specific()
...
+	package="yamllint"
+
+	case "$ID" in
+		centos|rhel) sudo yum -y install $package ;;
+		debian|ubuntu) sudo apt-get -y install $package ;;
+		fedora) sudo dnf -y install $package ;;
+		*) die "Please install yamllint on $ID" ;;
+	esac
+
+	[ -z "$have_yamllint_cmd" ] && info "Cannot install $package" && return
```
grahamwhaley: I'm a touch confused - do you not have to re-evaluate `have_yamllint_cmd` (that is, do the `command -v` thing again), before re-testing this - otherwise it won't have changed afaict. Maybe I'm missing a subtlety.

GabyCT: done :)

PR #2081, opened 2019-11-12
test: Add k8s test to check number of cpus

### LINKED ISSUE #2080: Create k8s test to check cpu numbers
Related https://github.com/kata-containers/runtime/issues/2186

GabyCT:
This will check the number of cpus, this is related to issue
https://github.com/kata-containers/runtime/issues/2186. This
verifies that we have the correct number of cpus after ending a
container.

Fixes #2080

### DIFF
--- a/integration/kubernetes/k8s-number-cpus.bats
+++ b/integration/kubernetes/k8s-number-cpus.bats
@@ -1,0 +1,43 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+issue="https://github.com/kata-containers/runtime/issues/2186"
+
+setup() {
+	skip "test not working see: ${issue}"
+	export KUBECONFIG="$HOME/.kube/config"
+	pod_name="cpu-test"
+	container_name="c1"
+	get_pod_config_dir
+}
+
+@test "Check number of cpus" {
+	skip "test not working see: ${issue}"
+	# Create pod
+	kubectl create -f "${pod_config_dir}/pod-number-cpu.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready pod "$pod_name"
+
+	retries="10"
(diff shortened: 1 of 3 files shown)


integration/kubernetes/k8s-number-cpus.bats
```diff
@@ -0,0 +1,43 @@
...
+	# Check pod creation
+	kubectl wait --for=condition=Ready pod "$pod_name"
+
+	# Expect some time to get nproc information
+	time="10"
+	sleep "$time"
+
+	# Get number of cpus
+	max_number_cpus="3"
+	number_cpus=$(kubectl exec -ti pod/"$pod_name" -c "$container_name" nproc | sed 's/[[:space:]]//g')
```
grahamwhaley: given the nature of the problem, maybe put the `exec` in a loop that is significantly larger than `max_number_cpus`? Only testing it the once and having `max_number_cpu`s hard wired here and relying on the layout in the yaml file feels a little 'borderline' to me.

integration/kubernetes/k8s-number-cpus.bats
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bats
```
grahamwhaley: A niggle - in the commit message body, rather than referencing github Issue numbers, can we have a textual explanation please - it makes viewing git logs and history on the command line a lot more meaninful and easier (as one does not have to navigate off to the web/github to look for the finer details).
thanks.

integration/kubernetes/k8s-number-cpus.bats
```diff
@@ -0,0 +1,49 @@
...
+	ok="0"
+
+	for _ in $(seq 1 "$retries"); do
+		# Get number of cpus
+		max_number_cpus="3"
+		number_cpus=$(kubectl exec -ti pod/"$pod_name" -c "$container_name" nproc | sed 's/[[:space:]]//g')
+		# Verify number of cpus
+		if [ "$number_cpus" -le "$max_number_cpus" ]; then
+			ok="1"
+			break;
```
devimc: you are checking the number of cpus just once,  if you remove the `skips` from test, it will pass when it shouldn't ,
instead you should exec `nproc` in the loop and verify its value. I think `ok` variable is not needed and `[ "$number_cpus" -le "$max_number_cpus" ]` should be enough

Issue #2082, opened 2019-11-12
ci: skip Firecracker install when testing changes on fc repo


chavafg:
when testing the FC repo, the installation of firecracker is done in the jenkins job, so we should skip it in the `install_firecracker.sh` script

PR #2083, opened 2019-11-12
ci: skip Firecracker install when testing changes on fc repo

chavafg:
when testing changes on firecracker repo, we don't want to
install the version of firecracker that we have in versions.yaml.
So skip the installation as it is already done in the Jenkins
job.

Fixes: #2082.

### DIFF
--- a/.ci/install_firecracker.sh
+++ b/.ci/install_firecracker.sh
@@ -12,10 +12,7 @@ cidir=$(dirname "$0")
 arch=$("${cidir}"/kata-arch.sh -d)
 source "${cidir}/lib.sh"
 KATA_DEV_MODE="${KATA_DEV_MODE:-false}"
-
-if [ "$arch" != "x86_64" ]; then
-	die "Static binaries for Firecracker only available with x86_64."
-fi
+ghprbGhRepository="${ghprbGhRepository:-}"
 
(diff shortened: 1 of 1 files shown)


.ci/install_firecracker.sh
```diff
@@ -27,66 +28,83 @@ if [ "$docker_version" != "18.06" ]; then
...
+	firecracker_version=$(get_version "assets.hypervisor.firecracker.version")
+	[ -n "$firecracker_version" ] || die "failed to get firecracker version"
+
+	# Get firecracker repo
+	go get -d "${firecracker_repo}" || true
+	# Checkout to specific version
+	pushd "${GOPATH}/src/${firecracker_repo}"
+	git checkout "tags/${firecracker_version}"
+	./tools/devtool --unattended build --release
+	sudo install ${GOPATH}/src/${firecracker_repo}/build/cargo_target/x86_64-*-linux-musl/release/firecracker /usr/bin/
```
jodh-intel: This isn't *strictly* required for this PR as you are changing what was already there... but could you create a variable for the architecture rather than hard-coding (since fc also works on `aarch64`)?

chavafg: agree, there was actually an `arch` variable already. Fixed.

Issue #2090, opened 2019-11-18
counter freezes when check iperf server in network-metrics-iperf3.sh


jongwu:
# Description of problem

in network test script network-metrics-iperf3.sh, count freezes by the wrong assignment "count=(( count++ ))"

# Expected result

count should accumulate round by round.

# Actual result

count freezes

PR #2092, opened 2019-11-18
tests: let iperf server check work by let counter go

jongwu:
Currently, count value in network-metrics-iperf3.sh freezed as
count value is assigned its original value round after round.
remove the "count=" to let it go.

Fixes: #2090

@jodh-intel @grahamwhaley

### DIFF
--- a/metrics/network/network-metrics-iperf3.sh
+++ b/metrics/network/network-metrics-iperf3.sh
@@ -167,7 +167,7 @@ function check_iperf3_server() {
 	while [ 1 ]; do
 		if ! bash -c "$test_cmd" > /dev/null 2>&1; then
 			echo "waiting for server..."
-			count=$((count++))
+			(( count++ ))
 			sleep $period
 		else
 			echo "iperf3 server is up!"


jongwu:
thanks @jodh-intel @grahamwhaley

chavafg:
> Also, I spotted this in the logs @chavafg:

@jodh-intel not sure why we get that. I see it only on PPC job and seems that the workspace is not been cleaned correctly... although on Jenkins we enabled the option to delete the `$WORKSPACE`.

```
07:38:39 [WS-CLEANUP] Deleting project workspace...
07:38:39 [WS-CLEANUP] Deferred wipeout is disabled by the job configuration...
07:38:42 [WS-CLEANUP] Done
```

Any idea @grahamwhaley @nitkon ?

grahamwhaley:
I think those are jenkins job options (a tickbox in the setup). You can delay the jenkins agent on the worker node from nuking the workspace, and do it lazy/background/garbage-collect style.
It could be related to that jenkins slave being a bare metal one iirc?

chavafg:
yeah, this is because we are using baremetal, but taking a look at the arm jobs, I see we are not getting that error and we have the same configuration. 
Will make a trick on the job and see if that helps. Anyway this can be merged as everything passed.

Issue #2091, opened 2019-11-18
need check command iperf3 before check iperf server in network-metrics-iperf3.sh


jongwu:
# Description of problem

if iperf3 is not present, test_cmd will not work in network-metrics-iperf3.sh. 

# Expected result

should check iperf3 command before use it and fail if there is no one.

# Actual result

if iperf3 not present, test_cmd will fail without warning.

@amshinde @jodh-intel

PR #2093, opened 2019-11-18
tests: check iperf3 before do check iperf3 server.

jongwu:
there must be sure about the existence of iperf3 command before
do the server check.
if iperf3 not present in host, test_cmd will fail without warning.

Fixes: #2091

@amshinde  @jcvenegas


amshinde:
Unrelated CI errors, merging this.

PR #2099, opened 2019-11-20
static-check: Fix bug in yamllint installation check

### LINKED ISSUE #2098: Travis is failing for shim and proxy
There is a bug introduced in  function `install_yamllint` in .ci/static-checks.sh. The bus is in in the condition check after installing yamllint, causing a false return value even though yamllint is installed sucessfully. This is only apparent in the shim and proxy travis. This is because for runtime, yamllint is already installed by a previous call to .`ci/setup.sh`

Example of travis failure seen: https://travis-ci.org/kata-containers/proxy/jobs/614291637?utm_medium=notification&utm_source=github_status

amshinde:
The previous check was causing the function to return false
even though yamllint is installed successfully.

Fixes #2098

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -330,7 +330,9 @@ install_yamllint()
 
 	have_yamllint_cmd=$(command -v "$yamllint_cmd" || true)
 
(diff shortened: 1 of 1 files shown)


grahamwhaley:
A bunch of CI fails @amshinde ?

Issue #2054, opened 2019-11-05
ci: Add installer script for cloud hypervisor


jcvenegas:
Related to CH integration

PR #2100, opened 2019-11-20
ci: Enable cloud hypervisor CI Jobs

jcvenegas:
Add script to install clould hypervisor

Fixes: #2054


.ci/install_cloud_hypervisor.sh
```diff
@@ -0,0 +1,51 @@
...
+
+cidir=$(dirname "$0")
+arch=$("${cidir}"/kata-arch.sh -d)
+# Where real kata build script exist, via docker build to avoid install all deps
+packaging_repo="github.com/kata-containers/packaging"
+
+source "${cidir}/lib.sh"
+
+
+install_ch() {
```
chavafg: `install_clh`? 

This is how it is being called in the packaging repo.

GabyCT:
@jcvenegas you miss to add cloud hypervisor here https://github.com/kata-containers/tests/blob/1969ec1f7686ac80e70de294fa6046b348d07575/.ci/install_runtime.sh#L107

GabyCT:
@jcvenegas the issue that you are having currently with the CH CI is mainly related with https://github.com/kata-containers/tests/blob/f2f5ec77861790eb250e7b6f137413238d10a09b/process.go#L55, as I ran the current PR on a VM and I can run a container correctly

jcvenegas:
> A small remark from my side which I observed today:
> 
> I've got a modified process.go running here locally and it seems that the suite only runs if the CloudHypervisor key in config.go is 'clh' . It is linked to the key inside the configuration-clh.toml file I think. I also started with 'cloud-hypervisor ' and that didn't work.
> 
> The regex to find the running cloud-hypervisor instance which I'm using is:
> `CloudHypervisor: (".*--api-socket.*" + containerID + ".*clh-api.sock.*")`

agree, I was planning the same specially when we just the API is going to be the only parameter,  I ran locally but I did not seen the api-socket, but let me update it

.ci/install_runtime.sh
```diff
@@ -49,26 +49,36 @@ if [ -e "${NEW_RUNTIME_CONFIG}" ]; then
...
+case "${KATA_HYPERVISOR}" in
+	"acrn")
+		enable_hypervisor_config "${PKGDEFAULTSDIR}/configuration-acrn.toml"
+		;;
+	"cloud-hypervisor")
+		enable_hypervisor_config "${PKGDEFAULTSDIR}/configuration-clh.toml"
+		;;
+	"firecracker")
+		enable_hypervisor_config "${PKGDEFAULTSDIR}/configuration-fc.toml"
+		sudo sed -i '/jailer_path/d' "${PKGDEFAULTSDIR}/configuration.toml"
```
chavafg: jailer is now [disabled by default](https://github.com/kata-containers/runtime/blob/master/cli/config/configuration-fc.toml.in#L21), so I think we can remove this line, right?

.ci/install_runtime.sh
```diff
@@ -49,26 +49,36 @@ if [ -e "${NEW_RUNTIME_CONFIG}" ]; then
...
+		;;
+	"qemu")
+		if [ "$experimental_qemu" == "true" ]; then
+			enable_hypervisor_config "${PKGDEFAULTSDIR}/configuration-qemu-virtiofs.toml"
+		else
+			enable_hypervisor_config "${PKGDEFAULTSDIR}/configuration-qemu.toml"
+		fi
+		;;
+	*)
+		die "failed to enable config for '${KATA_HYPERVISOR}', not supported'"
```
chavafg: remove last `'` ?

jcvenegas:
Note: The CI job will not be enabled by default there are missing test that still are failing randomly like:
https://github.com/kata-containers/tests/issues/2141

jcvenegas:
The fixes running it by default will be added in follow-up PR, for now this PR is need to be merged to allow travis pass in runtime PR. https://github.com/kata-containers/runtime/pull/2275

PR #2109, opened 2019-11-21
pipe: pipeline failed when no line selected in `grep`

Pennyzct:
If we have enabled pipefail option, `set -o pipefail`, the pipeline will fail when any sub-command exits with a non-zero status. 
But when pipeline consist of `grep`, it may fail unnecessarily.
For `grep` command, the exit status is 0 if a line is selected, 1 if no lines were selected, and 2 if an error occurred.
In most scenario, we allow no lines being selected. just like the one in `integration/kubernetes/cleanup_bare_metal_env.sh`
```
veth_interfaces_union=$(sudo ip link | grep "veth" | awk '{print $2}' | cut -d '@' -f1)
```

### DIFF
--- a/integration/kubernetes/cleanup_bare_metal_env.sh
+++ b/integration/kubernetes/cleanup_bare_metal_env.sh
@@ -51,9 +51,14 @@ if [ -n "$runc_container_union" ]; then
 	done <<< "${runc_container_union}"
 fi
 
+# when pipeline consists of grep, it may fail unnecessarily
+# when no line selected.
+veth_interfaces_union=$(set +o pipefail; sudo ip link | grep "veth" | awk '{print $2}' | cut -d '@' -f1)
+
 # delete stale veth interfaces, which is named after vethXXX.
-veth_interfaces_union=$(sudo ip link | grep "veth" | awk '{print $2}' | cut -d '@' -f1)
-while read veth_interface; do
-	sudo ip link set dev $veth_interface down
-	sudo ip link del $veth_interface
-done <<< "$veth_interfaces_union"
+if [ -n "$veth_interfaces_union" ]; then
+	while read veth_interface; do
+		sudo ip link set dev $veth_interface down
+		sudo ip link del $veth_interface
+	done <<< "$veth_interfaces_union"
+fi


grahamwhaley:
/cc @chavafg @jcvenegas - not sure if we have a nicer common idiom for handling `pipefail`. I don't remember one. Turning it off completely though I think needs a little discussion.
Nice find though @Pennyzct ;-)

chavafg:
We usually use an `|| true` at the end of the command when we know that something might fail. Gave a try locally and seems that it works in this case, but not sure if this is the best option. 
```
veth_interfaces_union=$(sudo ip link | grep "veth" | awk '{print $2}' | cut -d '@' -f1 || true)
```

/cc @jodh-intel who may have a good suggestion.

jodh-intel:
You could always limit the scope by doing:

```sh
set -o pipefail

    :

# temporarily disable pipefail as we know grep could fail
set +o pipefail
grep ...
set -o pipefail

    :
```

You could even wrap the `pipefail` handling in a function if you need to use it a lot. I think @jcvenegas    may have popularised `pipefail` so he may also have thoughts.

grahamwhaley:
temporary inline set/unset twiddling is horrible - but, I did consider if we wrapped it into a subshell, then the local shell twiddling becomes less horrible.

Pennyzct:
Hi guys 
thanks for the detailed suggestions~~ you're right, it's too reckless to just turn off the `pipefail` for whole script. 😔
I'm thinking, since the pipeline has been originally executed in sub-shell, I could just turn it off temporarily in sub-shell.
In such case, I could also avoid inline set/unset twiddling. ;). 
Wdyt? @grahamwhaley @jodh-intel @chavafg

jodh-intel:
Yep - sub-shell is the obvious extension to the idea. You could just do:

```sh
veth_interfaces_union=$(set +o pipefail; sudo ip link | grep "veth" | awk '{print $2}' | cut -d '@' -f1)
```

Pennyzct:
Hi guys
Following @jodh-intel 's suggestion, I decided to use sub-shell to temporarily turn off the pipefail check. Plz take a look.;).

PR #2113, opened 2019-11-22
tests: replace "/var/lib/vc" with "/run/vc"

### LINKED ISSUE #2112: replace "/var/lib/vc" with "/run/vc"
Checking "/run/vc" can also get active pods so we don't have to use "/var/lib/vc", also I'm working on removing "/var/lib/vc" dir from persistent data, we should remove "/var/lib/vc" from tests repo first.

See https://github.com/kata-containers/runtime/pull/2239 for more details.

WeiZhang555:
Fixes #2112

Checking "/run/vc" can also get active pods so we don't have to use
"/var/lib/vc", also I'm working on removing "/var/lib/vc" dir from
persistent data, we should remove "/var/lib/vc" from tests repo first.

Signed-off-by: Wei Zhang <weizhang555.zw@gmail.com>

### DIFF
--- a/integration/stability/soak_parallel_rm.sh
+++ b/integration/stability/soak_parallel_rm.sh
@@ -33,7 +33,7 @@ RUNTIME_PATH=$(command -v $RUNTIME)
 # The place where virtcontainers keeps its active pod info
 # This is ultimately what 'kata-runtime list' uses to get its info, but
 # we can also check it for sanity directly
-VC_POD_DIR="${VC_POD_DIR:-/var/lib/vc/sbs}"
+VC_POD_DIR="${VC_POD_DIR:-/run/vc/sbs}"
 
 # let's cap the test. If you want to run until you hit the memory limit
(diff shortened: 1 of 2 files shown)


integration/stability/soak_parallel_rm.sh
```diff
@@ -33,7 +33,7 @@ RUNTIME_PATH=$(command -v $RUNTIME)
 # The place where virtcontainers keeps its active pod info
 # This is ultimately what 'kata-runtime list' uses to get its info, but
 # we can also check it for sanity directly
-VC_POD_DIR="${VC_POD_DIR:-/var/lib/vc/sbs}"
+VC_POD_DIR="${VC_POD_DIR:-/run/vc/sbs}"
```
grahamwhaley: I was going to suggest we could change the code to get the directory from `kata-runtime kata-env` - but, I don't think we print that information (our workdir) in kata-env.... maybe we should? :-)

Issue #2126, opened 2019-11-27
k8s: skip cpu hotplug related test on AArch64


Pennyzct:
# Description of problem
Since AArch64 doesn't support cpu hotplug on kata containers, we need to skip related k8s tests for now.
Related issue: #2117 
PR is on the fly. ;).

PR #2127, opened 2019-11-27
k8s: skip cpu hotplug related test

Pennyzct:
Since AArch64 doesn't support cpu hotplug on kata containers, we need to skip related k8s tests for now.

Fixes: #2126
Depends-on: github.com/kata-containers/tests/#2117

### DIFF
--- a/integration/kubernetes/k8s-number-cpus.bats
+++ b/integration/kubernetes/k8s-number-cpus.bats
@@ -17,6 +17,7 @@ setup() {
 	get_pod_config_dir
 }
 
+# Skip on aarch64 due to missing cpu hotplug related functionality.
 @test "Check number of cpus" {
 	skip "test not working see: ${issue}"
 	# Create pod
--- a/.ci/aarch64/configuration_aarch64.yaml
+++ b/.ci/aarch64/configuration_aarch64.yaml
@@ -33,3 +33,4 @@ docker:
 kubernetes:
   - k8s-cpu-ns
   - k8s-limit-range
+  - k8s-number-cpus


integration/kubernetes/k8s-number-cpus.bats
```diff
@@ -17,6 +17,7 @@ setup() {
 	get_pod_config_dir
 }
 
+# Skip on aarch64 due to missing cpu hotplug related functionality.
```
jodh-intel: Is there a Kata tracking issue for this you could reference?

This test already has a `skip` but since `bats` is basically `bash` you could always add *another* skip below it checking for that arch and pointing to the aarch64-specific issue. Something like:

```sh
skip "test not working see: ${issue}"

arch=$(uname -m)
aarch64_issue="http://github.com/..."
[ "$arch" = aarch64 ] && skip "test disabled on arch $arch due to missing hotplug functionality: see $aarch64_issue"
```

GabyCT:
there is another k8s failure `not ok 1 Expose IP Address`

Pennyzct:
Hi @GabyCT 
I've re-run this test on my local machine, it works well.
I'll re-run the CI machine to see if it's repeatable. ;).

integration/kubernetes/k8s-number-cpus.bats
Pennyzct: Hi @jodh-intel sorry for the delayed response.
Yep, I didn't create a specific tracking issue, maybe I should...it's quite a big missing feature.
I need to skip a few k8s tests on aarch64, and in order to avoid inserting multiple arch-specific `skip` command.
 ```
"$arch" = aarch64 ] && skip "test disabled on arch $arch due to missing hotplug functionality: see $aarch64_issue"
```
I create this file [configuration_aarch64.yaml](https://github.com/kata-containers/tests/blob/master/.ci/aarch64/configuration_aarch64.yaml) to record those skip tests. ;)

Issue #2142, opened 2019-12-05
ci: static: ignore spell check for clh client code


jcvenegas:
http client code is generated, spell check or static checks
are not required.

PR #2143, opened 2019-12-05
ci: static: ignore spell for clh client code

jcvenegas:
clh client code is generated, ignore static check for it, fix missing
case.

Fixes: #2142

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -543,6 +543,7 @@ static_check_docs()
 		docs_status=$(echo "$docs_status" | grep "\.md$" || true)
 
 		docs=$(echo "$docs_status" | awk '{print $NF}' | sort)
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -543,6 +543,7 @@ static_check_docs()
 		docs_status=$(echo "$docs_status" | grep "\.md$" || true)
 
 		docs=$(echo "$docs_status" | awk '{print $NF}' | sort)
+		docs=$(echo "${docs}" | grep -Ev "(${ignore_clh_generated_code})" | sort)
```
grahamwhaley: `docs` is already sorted on the line above - do you need to sort it again after the `grep`?

jcvenegas:
I have created an issue to refactor this, https://github.com/kata-containers/tests/issues/2144

Issue #2144, opened 2019-12-05
ci: static: refactor static check script


jcvenegas:
**Which feature do you think can be improved?**

Ignored files or paths are duplicated in different places it could be a global pattern with all that will
be ignored.
```
	local missing=$(egrep \
		--exclude=".git/*" \
		--exclude=".gitignore" \
```

```
	all_docs=$(git ls-files "*.md" | grep -Ev "(vendor|grpc-rs|target|${ignore_clh_generated_code})/" | sort || true)
```
``
new_docs=$(echo "$new_docs" | grep -Ev "(${ignore_clh_generated_code})" | sort)
```
**How can it be improved?**

create a lib function `filter_docs` that takes a list of files and filter them, also
allow and env variable for extra files to filter so any repo could define its  own set
of paths to ignore.

Describe how specifically you think it could be improved.

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

PR #2146, opened 2019-12-05
ci: static skip clh client code for linter

jcvenegas:
refactor how skip is done, to allow add more in the future in an easy way.

Aditionally skip go module from clh client.

Fixes: #2144

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -54,6 +54,25 @@ typeset -A long_options
 # Generated code
 ignore_clh_generated_code="virtcontainers/pkg/cloud-hypervisor/client"
 
+paths_to_skip=(
+	"${ignore_clh_generated_code}"
+	"vendor"
+)
+
+skip_paths(){
+	local list=${1}
+	[ -z "$list" ] && return
+
+	for p in "${paths_to_skip[@]}"; do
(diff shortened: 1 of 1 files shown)


.ci/static-checks.sh
```diff
@@ -54,13 +54,33 @@ typeset -A long_options
...
+
+
 long_options=(
 	[all]="Force checking of all changes, including files in the base branch"
 	[branch]="Specify upstream branch to compare against (default '$branch')"
 	[commits]="Check commits"
 	[docs]="Check document files"
 	[files]="Check files"
 	[force]="Force a skipped test to run"
+
```
egernst: s/\n//?

.ci/static-checks.sh
```diff
@@ -54,6 +54,25 @@ typeset -A long_options
...
+	"vendor"
+)
+
+skip_paths(){
+	local list=${1}
+	[ -z "$list" ] && return
+
+	for p in "${paths_to_skip[@]}"; do
+		local new_list=""
+		if new_list=$(echo "${list}" | grep -Ev "${p}" | sort); then
```
chavafg: do we need to sort on every iteration? maybe do it after the loop?

PR #2147, opened 2019-12-06
metrics: launch_times: Set capabilities for dmesg

rhafer:
In case the guest kernel is compiled with SECURITY_DMESG_RESTRICT=y
the test container needs to have the CAP_SYSLOG capability in order
to run dmesg.

Signed-off-by: Ralf Haferkamp <rhafer@suse.com>


rhafer:
> You'll need to open an matching Issue and add a #Fixes line to your commit for the CI to be happy I think.

Argh, thanks. Forgot about that one. Should be fixed now.

PR #2184, opened 2019-12-17
ci: Add missing sudo for the cache components scripts

### LINKED ISSUE #2183: Add missing sudo for the cache component scripts

GabyCT:
We miss sudo while performing the echo of the component version as currently
we are getting the failure of permission denied.

Fixes #2183

### DIFF
--- a/.ci/ci_cache_components.sh
+++ b/.ci/ci_cache_components.sh
@@ -121,7 +121,7 @@ create_cache_asset() {
 		echo $(basename "${path}") > "latest-${image_name}"
 		sudo cp "${path}" "${kata_dir}/osbuilder-${image_name}.yaml"  .
 	else
-		echo "${component_version}" >  "latest"
+		echo "${component_version}" | sudo tee - a "latest"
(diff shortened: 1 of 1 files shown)


.ci/ci_cache_components.sh
```diff
@@ -121,7 +121,7 @@ create_cache_asset() {
 		echo $(basename "${path}") > "latest-${image_name}"
 		sudo cp "${path}" "${kata_dir}/osbuilder-${image_name}.yaml"  .
 	else
-		echo "${component_version}" >  "latest"
+		sudo echo "${component_version}" >  "latest"
```
grahamwhaley: nominally I see what you are trying to do, but I suspect the `sudo` is only going to bind to the echo, and not the `>` redirect. You might need the idiom we have used elsewhere of `echo | sudo tee - a "latest"`. Let's see what the CI says.

GabyCT: @grahamwhaley changes applied

Issue #2197, opened 2019-12-25
enable non-root ut for runtime repo


bergwolf:
# Description of problem

Right now many packages UTs in the runtime repo are skip as they live under `virtcontainers` directory due to following hack in CI:
```
35         echo "INFO: Currently running as user '$(id -un)'"
 36         for pkg in $test_packages; do
 37                 for user in $users; do
 38                        # Run virtcontainers tests only if user is root.
 39                        if [[ "$user" != "root" ]] && [[ "$pkg" = *"virtcontainers"* ]]; then
 40                                echo "Skip testing $pkg with user $user"
 41                                continue
 42                        fi
 43 
 44                         test_go_package "$pkg" "$user"
 45                 done
 46         done
```
We should just treat them same as other UTs instead.

Issue #2136, opened 2019-12-02
jenkins-ci-ubuntu-1804-firecracker always fails


teawater:
http://jenkins.katacontainers.io/job/kata-containers-agent-ubuntu-1804-PR-firecracker/255/console
```
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-1804-PR-firecracker/go/src/github.com/kata-containers/tests
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
manage_ctr_mgr.sh - INFO: docker_info: version: 
manage_ctr_mgr.sh - INFO: docker_info: default runtime: 
manage_ctr_mgr.sh - INFO: docker_info: package name: docker-ce
<13>Dec  2 02:36:54 manage_ctr_mgr.sh: configure_docker: configuring runc as default docker runtime
ls: cannot access '/etc/systemd/system/docker.service.d/': No such file or directory
ls: cannot access '/etc/systemd/system/docker.service.d/': No such file or directory
Stopping the docker service
Removing /var/lib/docker
Changing docker service configuration
```
And I found that this test got this issue from 248 to 255.

GabyCT:
@teawater , I think we can close this issue as https://github.com/kata-containers/agent/pull/699 has been merged

bergwolf:
It is still failing http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-firecracker/

```
10:54:22 Running command '/usr/bin/docker [docker run --cidfile /tmp/cid733249092/v0EhH3oAXcvEMuWiMQdSVQk6OWk801 --runtime kata-runtime --cpus=1 --cpu-shares=800 --cpuset-cpus=0 -dt --name v0EhH3oAXcvEMuWiMQdSVQk6OWk801 busybox sh]'
10:54:55 command failed error 'exit status 125'
10:54:55 [docker run --cidfile /tmp/cid733249092/v0EhH3oAXcvEMuWiMQdSVQk6OWk801 --runtime kata-runtime --cpus=1 --cpu-shares=800 --cpuset-cpus=0 -dt --name v0EhH3oAXcvEMuWiMQdSVQk6OWk801 busybox sh]
10:54:55 Timeout: 120 seconds
10:54:55 Exit Code: 125
10:54:55 Stdout: 1f3034ecf98725807562ea74bcff0a9bd8eb3791399a38b7f49ed1a1a3c5f946
10:54:55 
[...]

PR #2198, opened 2019-12-25
ci: do not skip virtcontainers directories with non-root users

bergwolf:
Currently anything under virtcontainers are skipped for non-root users.
We should just treat them the same as others instead.

Depends-on: Depends-on: github.com/kata-containers/runtime#2371

Fixes: #2197


bergwolf:
The FC failure is unrelated and tracked at https://github.com/kata-containers/tests/issues/2136

PR #2204, opened 2020-01-07
ci: Improve path for calling install_yq.sh script

### LINKED ISSUE #2203: improve path for install_yq.sh script

GabyCT:
While trying to call the function get_dep_from_yaml_db from lib.sh, this
is failing while trying to run the install_yq.sh script as sometimes it does
not find the script in the current path. This PR fixes that issue by
improving the path.

Fixes #2203

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -123,10 +123,7 @@ function get_dep_from_yaml_db(){
 
 	[ ! -f "$versions_file" ] && die "cannot find $versions_file"
 
-	# directory of this script, not the caller
-	local cidir=$(dirname "${BASH_SOURCE[0]}")
-
-	${cidir}/install_yq.sh >&2
+	"${GOPATH}/src/${tests_repo}/.ci/install_yq.sh" >&2
 
 	result=$("${GOPATH}/bin/yq" read "$versions_file" "$dependency")
 	[ "$result" = "null" ] && result=""


.ci/lib.sh
```diff
@@ -123,10 +123,7 @@ function get_dep_from_yaml_db(){
 
 	[ ! -f "$versions_file" ] && die "cannot find $versions_file"
 
-	# directory of this script, not the caller
-	local cidir=$(dirname "${BASH_SOURCE[0]}")
-
-	${cidir}/install_yq.sh >&2
+	"${GOPATH}/src/${tests_repo}/.ci/install_yq.sh" >&2
```
grahamwhaley: I wonder why 'sometimes' it is not found - do you have any specifics @GabyCT ?
@chavafg - can you confirm, this hard wiring of the GOPATH will work in all cases such as our depends-on etc.?

GabyCT: @grahamwhaley, here it is an example
```
06:28:41 + .ci/ci_cache_components.sh -c
06:28:41 /tmp/jenkins/workspace/cloud-hypervisor-nightly-x86_64/artifacts /tmp/jenkins/workspace/cloud-hypervisor-nightly-x86_64/go/src/github.com/kata-containers/tests
06:28:41 artifacts:
06:28:41 .ci/lib.sh: line 129: .ci/install_yq.sh: No such file or directory
```

chavafg: I thought that `dirname` was giving the absolute path, but seems that no. 
I think this should be fine in our `Depends-on` scenarios, but we should better test.

GabyCT: The function is only used at the tests repository

Issue #2205, opened 2020-01-09
The  static_check_rust_arch_specific failed to prompt the error info


lifupan:
# Description of problem

When the rustfmt check return non-zero, the script
would return directly and there is no chance
to do the return code check in the following
cmd.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #2206, opened 2020-01-09
ci: fix the issue of missing show error info

lifupan:
When the shell cmd return non-zero, the script
would return directly and there is no chance
to do the return code check in the following
cmd.

Fixes: #2205

Signed-off-by: fupan.lfp <fupan.lfp@antfin.com>

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -309,9 +309,9 @@ static_check_rust_arch_specific()
 		return
 	fi
 
-	find . -type f -name "*.rs"  | egrep -v "target/|grpc-rs/|protocols/" | xargs rustfmt --check > /dev/null 2>&1
+	{ find . -type f -name "*.rs"  | egrep -v "target/|grpc-rs/|protocols/" | xargs rustfmt --check > /dev/null 2>&1; ret=$?; } || true
 
-	[ $? -eq 0 ] || die "crate not formatted by rustfmt."
+	[ $ret -eq 0 ] || die "crate not formatted by rustfmt."
 }
 
 # Install yamllint in the different Linux distributions


.ci/static-checks.sh
```diff
@@ -309,9 +309,8 @@ static_check_rust_arch_specific()
 		return
 	fi
 
-	find . -type f -name "*.rs"  | egrep -v "target/|grpc-rs/|protocols/" | xargs rustfmt --check > /dev/null 2>&1
+	find . -type f -name "*.rs"  | egrep -v "target/|grpc-rs/|protocols/" | xargs rustfmt --check > /dev/null 2>&1 || die "crate not formatted by rustfmt."
```
grahamwhaley: I guess that works. I think the more common idiom through the file already though is to use a subshell to capture the result, such as (/me grabs random example from earlier in the file):

```sh
    { echo "$name" | grep -q "${check_func_regex}"; ret=$?; }

    [ "$ret" = 0 ] || die "function '$name' is not a check function"
```

/cc @jodh-intel @chavafg

chavafg: I agree it would be good to have a common idiom in the file.

lifupan: @grahamwhaley @chavafg  Updated!

chavafg:
for some reason, travis is complaining about the file:
```
.ci/static-checks.sh: line 1186: syntax error: unexpected end of file
```

lifupan:
> for some reason, travis is complaining about the file:
> 
> ```
> .ci/static-checks.sh: line 1186: syntax error: unexpected end of file
> ```

@chavafg  Fixed the issue.

.ci/static-checks.sh
```diff
@@ -309,7 +309,7 @@ static_check_rust_arch_specific()
 		return
 	fi
 
-	find . -type f -name "*.rs"  | egrep -v "target/|grpc-rs/|protocols/" | xargs rustfmt --check > /dev/null 2>&1
+	{ find . -type f -name "*.rs"  | egrep -v "target/|grpc-rs/|protocols/" | xargs rustfmt --check > /dev/null 2>&1; }
```
jodh-intel: I think you are missing a `|| true` at the end of the `{ ... }`? Currently, if this line fails, the script will still exit. Is that what you want?

grahamwhaley: I think @jodh-intel is right - you probably want to copy the code from around line 453, which looks like the below. Note both the capture of `$?` into a variable, and the use of `|| true` to avoid the unexpected `set -e` quit...

```bash
        local ret

	{ curl -sIL --max-time "$url_check_timeout_secs" --retry "$url_check_max_tries" "$url" &>"$curl_out"; ret=$?; } || true

	# A transitory error, or the URL is incorrect,
	# but capture either way.
	if [ "$ret" -ne 0 ]; then
		echo "$url" >> "${invalid_file}"

		die "check failed for URL $url after $url_check_max_tries tries"
	fi
```

lifupan: @jodh-intel  @grahamwhaley  updated!

PR #2213, opened 2020-01-13
skip forward compatibility test

devimc:


chavafg:
I think comments on L67 and L97 should also change, right?

chavafg:
>I think comments on L67 and L97 should also change, right?

@devimc nevermind, comments are fine, I was thinking the other way around.

Issue #2227, opened 2020-01-16
ci: spellcheck: don't stop checking on first doc failure


grahamwhaley:
**Which feature do you think can be improved?**

The CI static spell check fails (quits) at the first failing document, no matter if there are more documents that could be checked.

**How can it be improved?**

Don't quit at the first failure, but continue checking and reporting failures, and quit at the end if there have been any fails.

This will allow for a more efficient CI check cycle, and give authors more opportunity to fix multiple problems per cycle.

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

PR #2228, opened 2020-01-16
ci: static check: spellcheck: check all docs

grahamwhaley:
Rather than fail the spell check on the first failure, continue to
check all the changed documents, and report the global fail at the
end. This may help reduce CI/review cycle time if/when multiple docs
are submitted in a single PR.

Also spit out a URL link to the documentation repo docs in the case
of failure, to help direct authors to guidance docs.

Fixes: #2227

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -760,10 +760,13 @@ static_check_docs()
 	# Now, spell check the docs
 	cmd="${tests_repo_dir}/cmd/check-spelling/kata-spell-check.sh"
 
+	local docs_failed=0
 	for doc in $docs
 	do
-		"$cmd" check "$doc" || die "spell check failed for document $doc"
+		"$cmd" check "$doc" || { info "spell check failed for document $doc" && docs_failed=1; }
 	done
+
+	[ $docs_failed -ne 0 ] && die "spell check failed, See https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#spelling for more information."
 }
 
 # Tests to apply to all files.


grahamwhaley:
@jodh-intel @chavafg - can we stare hard at this bash - it is slightly tricky, and I can't easily test it locally, and I think the CI won't really test it - so we might need some :eyes: all over it..

.ci/static-checks.sh
```diff
@@ -760,10 +760,13 @@ static_check_docs()
...
 	cmd="${tests_repo_dir}/cmd/check-spelling/kata-spell-check.sh"
 
+	local docs_failed=0
 	for doc in $docs
 	do
-		"$cmd" check "$doc" || die "spell check failed for document $doc"
+		"$cmd" check "$doc" || { info "spell check failed for document $doc" && docs_failed=1; }
 	done
+
+	[ $docs_failed -ne 0 ] || die "spell check failed, See https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#spelling for more information."
```
chavafg: should this be `&&` instead of `||` 
or 
use `-eq` instead of `-ne`

grahamwhaley: ah, yup, I think you are right. that is a copy/paste/do-not-engage-brain moment :-)

.ci/static-checks.sh
```diff
@@ -760,10 +760,13 @@ static_check_docs()
 	# Now, spell check the docs
 	cmd="${tests_repo_dir}/cmd/check-spelling/kata-spell-check.sh"
 
+	local docs_failed=0
 	for doc in $docs
 	do
-		"$cmd" check "$doc" || die "spell check failed for document $doc"
+		"$cmd" check "$doc" || { info "spell check failed for document $doc" && docs_failed=1; }
```
chavafg: we could increment the variable value if we want to tell the user how many errors he got, but if not, I think this way is fine.

grahamwhaley: I did originally think of increment, but then ran away from the idea of adding any `bc` or `(( ))` code into a single line expression, so went with the simple `=1` option. We could also collect up the names of the `$doc`s that failed, and list them at the end.

I can add those improvements if we want. I think as it is, all the necessary info should appear in the logs fairly grouped together.
Somebody just shout if they want the improvements adding.

PR #2230, opened 2020-01-16
ci: Update python for CentOS 8

### LINKED ISSUE #2229: update python version for CentOS
CentOS 8 does not have `python` package, it has `python2` or `python3` so we need to update the version.

GabyCT:
The python package is not available on CentOS 8. This PR updates that to use
python3 which is currently available for CentOS 7 and 8.

Fixes #2229


.ci/setup_env_centos.sh
```diff
@@ -62,7 +62,7 @@ declare -A packages=( \
 	[bison_binary]="bison" \
 	[libgudev1-dev]="libgudev1-devel" \
 	[general_dependencies]="gpgme-devel glib2-devel glibc-devel bzip2 m4 gettext-devel automake alien autoconf pixman-devel coreutils" \
-	[build_tools]="python pkgconfig zlib-devel" \
+	[build_tools]="python3 pkgconfig zlib-devel" \
```
grahamwhaley: does this work for the previous centos version as well, or are we abandoning that and moving on?

GabyCT: not yet, yes it works on centos 7, we can trigger the CI of centos just to verify

PR #2232, opened 2020-01-16
ci: Remove alien package

### LINKED ISSUE #2231: Remove alien package

GabyCT:
Remove alien package as it is not needed it.

Fixes #2231


.ci/setup_env_centos.sh
```diff
@@ -48,6 +48,11 @@ fi
 echo "Install chronic"
 sudo -E yum -y install moreutils
 
+# This is needed in order to retry alien package for CentOS 8
+if [ "$centos_version" == "8" ]; then
+	sudo -E yum -y install http://li.nux.ro/download/nux/dextop/el7/x86_64/nux-dextop-release-0-5.el7.nux.noarch.rpm
```
chavafg: I am not sure we want to download rpms from non-official repos.
Do we still need `alien` package? I remember that we needed it because we were installing non-rpm qemu(do not remember why), but currently it is not the case.

GabyCT: @chavafg , I got the info from https://wiki.centos.org/TipsAndTricks/MultimediaOnCentOS7

GabyCT: @chavafg changes applied

chavafg: thanks :)

Issue #2242, opened 2020-01-20
k8s: e2e: Add podOverhead feature and modify skipped tests


chavafg:
# Description of problem
Enable `PodOverhead` feature on kubernetes
This will help to reduce the number of kubernetes tests skipped on the e2e execution. 

Related: https://github.com/cri-o/cri-o/issues/2114

PR #2243, opened 2020-01-20
k8s: e2e: Enable podOverhead feature and modify skipped tests

chavafg:
We are using k8s 1.16, so we can enable PodOverhead alpha
feature for our testing.

In addition, modify skipped tests:
- Remove `Downward API` tests from the list of skipped tests.
  These tests pass when enabling the `PodOverhead` feature on
  kubernetes.
- Add a `Daemon set` test which requires to have a cluster
  with at least two nodes.
    
Fixes: #2242.


chavafg:
adding DNM as I am still testing the e2e suite with these changes.

Issue #2258, opened 2020-01-24
setup: clh: install script should be call in the parent of the hypervisor respository


jcvenegas:
# Description of problem

install script should be call in the parent of the hypervisor respository, the script will find the repository and will not clone by itself.

# Expected result

Clone hypervisor repository `pushd`  in the parent directory of the repository

PR #2259, opened 2020-01-24
setup: clh: do not clone more than one time.

jcvenegas:
packaging installer script will clone the hypervisor repository,
if is not found, because test setup goes to the repository and
not the parent directory, it clones the hypervisor two times.

Fixes: #2258


chavafg:
typo in commit message: `because`

GabyCT:
@jcvenegas the cloud hypervisor CI still fails, is that expected?

Issue #2274, opened 2020-01-31
ci: new `yq` version is not expanding default values


chavafg:
# Description of problem

Using`yq` 2.4.1:
```
$ yq r runtime/versions.yaml 'assets.initrd.architecture.x86_64.name'
alpine
```

Using`yq` 3.0.1:
```
$ yq r runtime/versions.yaml 'assets.initrd.architecture.x86_64.name'
*default-initrd-name
```

jodh-intel:
I've raised https://github.com/mikefarah/yq/issues/340. Note that adding `-j` works around the issue (albeit resulting in JSON output ;)

PR #2275, opened 2020-01-31
CI: do not use latest version of yq

chavafg:
latest version of `yq` is not expanding anchors correctly and
we are having failures when building the initrd.
Let's stick to version 2.3.0 which is also used in runtime
and osbuilder repos.

Fixes: #2274

### DIFF
--- a/.ci/install_yq.sh
+++ b/.ci/install_yq.sh
@@ -20,12 +20,9 @@ function install_yq() {
 
 	mkdir -p "${GOPATH}/bin"
 
-	# Workaround to get latest release from github (to not use github token).
-	# Get the redirection to latest release on github.
-	yq_latest_url=$(curl -Ls -o /dev/null -w %{url_effective} "https://${yq_pkg}/releases/latest")
-	# The redirected url should include the latest release version
-	# https://github.com/mikefarah/yq/releases/tag/<VERSION-HERE>
-	yq_version=$(basename "${yq_latest_url}")
+	# Stick to a specific version. Same used in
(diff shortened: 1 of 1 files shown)


.ci/install_yq.sh
```diff
@@ -20,12 +20,9 @@ function install_yq() {
...
 
-	# Workaround to get latest release from github (to not use github token).
-	# Get the redirection to latest release on github.
-	yq_latest_url=$(curl -Ls -o /dev/null -w %{url_effective} "https://${yq_pkg}/releases/latest")
-	# The redirected url should include the latest release version
-	# https://github.com/mikefarah/yq/releases/tag/<VERSION-HERE>
-	yq_version=$(basename "${yq_latest_url}")
+	# Stick to a specific version. Same used in
+	# runtime and osbuilder repos.
+	yq_version=2.3.0
```
GabyCT: maybe we can have this at the versions.yaml of this repo

chavafg: yeah, that would be the best, but we need to install `yq` before we can actually read the yaml file.

grahamwhaley: arf, catch 22...
@chavafg - do you know if this is a known fault (already reported) with `yq`? If not, can we make a report in the upstream yq github for it?

chavafg: @grahamwhaley, I wasn't actually sure if this was a problem with `yq` or with our yaml file, but thanks to @jodh-intel who already reported the issue to `yq` repo.

PR #2281, opened 2020-02-04
tests: Add podman integration tests

### LINKED ISSUE #2280: Add podman tests

GabyCT:
This PR enables to have rootless podman integration tests.

Fixes #2280

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -46,6 +46,14 @@ case "${CI_JOB}" in
 		echo "INFO: Running crio tests"
 		sudo -E PATH="$PATH" bash -c "make crio"
 		;;
+	"PODMAN")
+		export TRUSTED_GROUP="kvm"
+		newgrp "${TRUSTED_GROUP}" << END
+		echo "This is running as group $(id -gn)"
+END
+		echo "INFO: Running podman integration tests"
+		bash -c "make podman"
+		;;
 	*)
 		echo "INFO: Running checks"
 		sudo -E PATH="$PATH" bash -c "make check"
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,17 @@ ifeq (${CI}, true)
 	endif
 endif
 
+PODMAN_DEPENDENCY = podman
+ifeq (${CI}, true)
+        ifneq (${TEST_CGROUPSV2}, true)
+                PODMAN_DEPENDENCY =
+        endif
+endif
+
 # union for 'make test'
 UNION := functional debug-console $(DOCKER_DEPENDENCY) openshift crio docker-compose network \
 	docker-stability oci netmon kubernetes swarm vm-factory \
-	entropy ramdisk shimv2 tracing time-drift compatibility vcpus
+	entropy ramdisk shimv2 tracing time-drift compatibility vcpus $(PODMAN_DEPENDENCY)
 
 # filter scheme script for docker integration test suites
 FILTER_FILE = .ci/filter/filter_docker_test.sh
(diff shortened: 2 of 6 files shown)


.ci/run.sh
```diff
@@ -46,6 +46,10 @@ case "${CI_JOB}" in
 		echo "INFO: Running crio tests"
 		sudo -E PATH="$PATH" bash -c "make crio"
 		;;
+	"PODMAN")
+		echo "INFO: Running podman integration tests"
+		sudo -E PATH="$PATH" bash -c "make podman"
+
```
grahamwhaley: are you missing a closing `;;` here?

integration/podman/podman.go
```diff
@@ -0,0 +1,309 @@
+// Copyright (c) 2020 Intel Corporation
```
devimc: this file is very similar to `docker.go`, I think we should factorize it later

Makefile
```diff
@@ -14,10 +14,17 @@ ifeq (${CI}, true)
 	endif
 endif
 
+PODMAN_DEPENDENCY = podman
+ifeq (${CI}, true)
+	ifneq (${TEST_CGROUPSV2}, true)
```
devimc: I think you can move this into above `ifeq (${CI}, true)`

https://github.com/kata-containers/tests/blob/3fdb4c3a8259f91fd466868a7e39f31a4c16a856/Makefile#L10-L15

integration/podman/podman.go
GabyCT: @devimc yeah, I was trying to do that maybe in another PR, I will do it (the issue that I have currently is that I need to identify which things are working with podman and which are not, some commands can't be used rootless)

Makefile
GabyCT: done

.ci/run.sh
GabyCT: @grahamwhaley done

GabyCT:
Regarding to the golint error, I do not know how to solve it...in order to run the test we need 
```
. "github.com/onsi/ginkgo".....
```
any idea @grahamwhaley @devimc of what can I do? thanks

devimc:
@GabyCT yes, test files should has suffix `_test.go`

integration/podman/run_test_podman.go
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2020 Intel Corporation
```
devimc: just rename this to `run_test.go`

GabyCT: @devimc thanks

GabyCT:
@jcvenegas , the cloud hypervisor is failing while running with docker the macvlan tests is that expected?

GabyCT:
@Pennyzct is the ARM CI stable? I have a failure in one of the tests

Makefile
```diff
@@ -121,6 +128,10 @@ docker-stability:
 	cd integration/stability && ./bind_mount_linux.sh
 	cd integration/stability && ./hypervisor_stability_kill_test.sh
 
+podman: ginkgo
+	./ginkgo -failFast -v -focus "${FOCUS}" -skip "${SKIP}" \
```
devimc: mmm but `SKIP` is only for docker, isn't it?

https://github.com/kata-containers/tests/blob/4ae15b1ccd6e04866555d577a9154e1e712d60c3/arch/aarch64-options.mk#L12-L13

Makefile
```diff
@@ -121,6 +128,10 @@ docker-stability:
...
+		./integration/podman/ -- -runtime=${RUNTIME} -timeout=${TIMEOUT}
```
devimc: rootless containers are not constrained (cgroupsv2 is not tested), once we add test to check croupsv2 these tests will fail, consider moving all rootless tests under `./integration/podman/rootless`

Makefile
```diff
@@ -121,6 +128,10 @@ docker-stability:
 	cd integration/stability && ./bind_mount_linux.sh
 	cd integration/stability && ./hypervisor_stability_kill_test.sh
 
+podman: ginkgo
```
devimc: uhmm not running in parallel?

Makefile
GabyCT: @devimc yes that can be done later, I am working on enabling the e2e podman tests

Makefile
GabyCT: @devimc right now, I am not contempleting in doing that

Makefile
GabyCT: @devimc yes, I will actually later use SKIP when I used e2e tests as some of them are failing but instead of docker filter will be podman

PR #2317, opened 2020-02-18
test: enable k8s for clh

jcvenegas:
Enable k8s for clh.

Fixes: #2318

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -152,6 +152,16 @@ case "${CI_JOB}" in
 	# Used by runtime makefile to enable option on intall
 	export DEFSANDBOXCGROUPONLY=true
 	;;
+"CLOUD-HYPERVISOR")
+	export CRIO="no"
+	export CRI_CONTAINERD="yes"
+	export CRI_RUNTIME="containerd"
+	export KATA_HYPERVISOR="cloud-hypervisor"
+	export KUBERNETES="yes"
+	export OPENSHIFT="no"
+	export TEST_CRIO="false"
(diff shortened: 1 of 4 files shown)


GabyCT:
@jcvenegas fixes 1?

.ci/jenkins_job_build.sh
```diff
@@ -152,6 +152,17 @@ case "${CI_JOB}" in
...
 	export DEFSANDBOXCGROUPONLY=true
 	;;
+"CLOUD-HYPERVISOR")
+	export CRIO="no"
+	export CRI_CONTAINERD="yes"
+	export CRI_RUNTIME="containerd"
+	export KATA_HYPERVISOR="cloud-hypervisor"
+	export KUBERNETES="yes"
+	export OPENSHIFT="no"
+	export TEST_CRIO="false"
```
chavafg: this is by default false in the crio test script, so should be fine if we remove it

.ci/jenkins_job_build.sh
```diff
@@ -152,6 +152,17 @@ case "${CI_JOB}" in
...
+	export experimental_kernel="true"
+	export use_runtime_class="true"
```
chavafg: we now always use `runtimeClass`, so this flag is no longer needed.

.ci/jenkins_job_build.sh
jcvenegas: I prefer to keep it explicit it helps know what is actually being tested for the CI_JOB

chavafg: ok, agree 👍

Issue #2321, opened 2020-02-21
add virtiofsd to metrics memroy measure list


jongwu:
**Which feature do you think can be improved?**

katat metrics test for memory footprint

**How can it be improved?**

add virtiofsd into its test list to cover all the binary start by kata.

**Additional Information**

**Before raising this feature request**

@jodh-intel  @grahamwhaley  @Pennyzct

PR #2322, opened 2020-02-21
test: metrics: add virtiofsd to memory measure list

jongwu:
As virtiofs enabled in kata, virtiofsd will run as a standalone binary.
it will contribute to memory footprint of kata so we should add
virtiofsd into metrics memory measure list.

This commit depends on if virtiofsd can be added into kata-env, whose PR is  https://github.com/kata-containers/runtime/pull/2492.
Fixes: #2321

@jodh-intel  @devimc  @grahamwhaley  @Pennyzct

### DIFF
--- a/metrics/density/docker_memory_usage.sh
+++ b/metrics/density/docker_memory_usage.sh
@@ -189,6 +189,7 @@ EOF
 # Try to work out the 'average memory footprint' of a container.
 get_docker_memory_usage(){
 	hypervisor_mem=0
+	virtiofsd_mem=0
 	shim_mem=0
 	proxy_mem=0
 	proxy_mem=0
@@ -251,6 +252,10 @@ EOF
 			die "Failed to find PSS for $HYPERVISOR_PATH"
 		fi
 
+		virtiofsd_mem="$(get_pss_memory "$VIRTIOFSD_PATH")"
+		if [ "$virtiofsd_mem" == "0" ]; then
+			echo >&2 "WARNING: Failed to find PSS for $VIRTIOFSD_PATH"
+		fi
 		shim_mem="$(get_pss_memory "$SHIM_PATH")"
 		if [ "$shim_mem" == "0" ]; then
 			die "Failed to find PSS for $SHIM_PATH"
@@ -267,7 +272,7 @@ EOF
 		fi
 
 		proxy_mem="$(bc -l <<< "scale=2; $proxy_mem / $NUM_CONTAINERS")"
-		mem_usage="$(bc -l <<< "scale=2; $hypervisor_mem + $shim_mem + $proxy_mem")"
+		mem_usage="$(bc -l <<< "scale=2; $hypervisor_mem +$virtiofsd_mem + $shim_mem + $proxy_mem")"
 		memory_usage="$mem_usage"
 
 	local json="$(cat << EOF
@@ -278,6 +283,10 @@ EOF
 		},
 		"qemus": {
 			"Result": $hypervisor_mem,
+			"Units" : "KB"
+		},
+		"virtiofsds": {
+			"Result": $virtiofsd_mem,
 			"Units" : "KB"
 		},
 		"shims": {
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -114,6 +114,7 @@ extract_kata_env(){
 
 			HYPERVISOR_PATH=$(awk '/^\[Hypervisor\]$/ {foundit=1} /^  Path =/ { if (foundit==1) {print $3; foundit=0} } ' <<< "$toml" | sed 's/"//g')
 			HYPERVISOR_VERSION=$(awk '/^\[Hypervisor\]$/ {foundit=1} /^  Version =/ { if (foundit==1) {$1=$2=""; print $0; foundit=0} } ' <<< "$toml" | sed 's/"//g')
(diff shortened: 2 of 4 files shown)


metrics/density/docker_memory_usage.sh
```diff
@@ -251,6 +252,10 @@ EOF
 			die "Failed to find PSS for $HYPERVISOR_PATH"
 		fi
 
+		virtiofsd_mem="$(get_pss_memory "$VIRTIOFSD_PATH")"
+		if [ "$virtiofsd_mem" == "0" ]; then
+			die "Failed to find PSS for $VIRTIOFSD_PATH"
```
grahamwhaley: I'm not sure in the case of `virtiofsd` that this should be a `die` condition. I don't think every user will have virtiofsd installed or in use will they? Maybe just make this a `warn`?.  The others like proxy and shim are `die` conditions, as they are expected to be present (although, with the advent of shimv2 etc. these rules are probably not quite as strong and useful as they used to be).

jongwu:
> One small comment, but I'll lgtm whilst expecting that to be maybe changed.
> Thanks for this @jongwu - I think this is a good addition as we move forwards.
> I'm not set up to test this right now, but I think you probably have done. Do you have any numbers to share (what sort of footprint kb and % is virtiofsd adding?).

hi @grahamwhaley , I have got some raw data. show you here.
"Results": [
                        {
                "average": {
                        "Result": 195617.00,
                        "Units" : "KB"
                },
                "qemus": {
                        "Result": 182091.40,
                        "Units" : "KB"
                },
                "virtiofsd": {
                        "Result": 3307.00,
                        "Units" : "KB"
                },
                "shims": {
                        "Result": 9420.60,
                        "Units" : "KB"
                },
                "proxys": {
                        "Result": 798.00,
                        "Units" : "KB"
                }
        }
As ksm does not work for virtiofs so no ksm enabled here. From these data, virtiofsd occupies only 1.7% of all memroy footprint.

metrics/density/docker_memory_usage.sh
jongwu: ok, that' better, I will change "die" to "echo".

grahamwhaley: I'd probably use `warning`. We have `die` and `warning` [in the test lib code](https://github.com/kata-containers/tests/blob/master/metrics/lib/common.bash#L31-L45). Elsewhere in kata we also have `info` (but maybe not in the tests code).

metrics/density/docker_memory_usage.sh
```diff
@@ -251,6 +252,10 @@ EOF
 			die "Failed to find PSS for $HYPERVISOR_PATH"
 		fi
 
+		virtiofsd_mem="$(get_pss_memory "$VIRTIOFSD_PATH")"
+		if [ "$virtiofsd_mem" == "0" ]; then
+			echo "Failed to find PSS for $VIRTIOFSD_PATH"
```
jodh-intel: Should this be `warn "Failed to find PSS for $VIRTIOFSD_PATH"`, or `die`?

metrics/density/docker_memory_usage.sh
```diff
@@ -280,6 +285,10 @@ EOF
 			"Result": $hypervisor_mem,
 			"Units" : "KB"
 		},
+		"virtiofsd": {
```
jodh-intel: Nit: Not inserted into the file alphabetically.

metrics/density/docker_memory_usage.sh
jongwu: should add "Warning:" as the first printed word?

metrics/density/docker_memory_usage.sh
jongwu: should be "virtiofsds" here?

metrics/density/docker_memory_usage.sh
grahamwhaley: like the other place, you can use [`warning()`](https://github.com/kata-containers/tests/blob/4a46b74d542f48c0d34c3a5da82333469515f455/metrics/lib/common.bash#L39-L45) from the common library code, which prepends the message with 'WARNING: ' for you.

GabyCT:
The metrics CI is failing with
```
11:38:50 Report Summary:
11:38:50 +-----+----------------------+---------------------+---------------+--------+-------+-------+--------+-------+------+-----+
11:38:50 | P/F |         NAME         |         FLR         |     MEAN      |  CEIL  |  GAP  |  MIN  |  MAX   |  RNG  | COV  | ITS |
11:38:50 +-----+----------------------+---------------------+---------------+--------+-------+-------+--------+-------+------+-----+
11:38:50 | P   | boot-times           | 95.0%               | 99.7%         | 105.0% | 10.0% | 90.2% | 111.2% | 23.3% | 5.9% |  20 |
11:38:50 | *F* | memory-footprint     | Failed to load JSON | exit status 2 |        |       |       |        |       |      |     |
11:38:50 | *F* | memory-footprint-ksm | Failed to load JSON | exit status 2 |        |       |       |        |       |      |     |
11:38:50 +-----+----------------------+---------------------+---------------+--------+-------+-------+--------+-------+------+-----+
```

jongwu:
thanks @GabyCT, this patch depends on PR https://github.com/kata-containers/runtime/pull/2492 so we can start CI again after that PR merge.

jongwu:
@grahamwhaley I have added "depents-on" to commit message, so maybe we can try CI again.

grahamwhaley:
/test
sure, let's try! If it fails, I'll have a look into why we get a json parse failure etc.

Issue #2338, opened 2020-03-03
ci: static markdown doc check is checking vendor files


grahamwhaley:
# Description of problem

Occasionally we see the CI static check markdown doc checker kicking out an error on an un-referenced markdown file, that is actually in the `vendor` tree. The code looks like it filters out `vendor` files, but obviously somewhere, somehow, it is failing.

```
INFO[0000] Checking file                                 commit=4dc5d23e8a997dc4a3aef2299b765be5bd13048e file=vendor/github.com/docker/docker/project/CONTRIBUTORS.md name=kata-check-markdown pid=25802 source=check-markdown version=0.0.1
FATA[0000] open vendor/github.com/docker/docker/project/CONTRIBUTORS.md: no such file or directory  commit=4dc5d23e8a997dc4a3aef2299b765be5bd13048e name=kata-check-markdown pid=25802 source=check-markdown version=0.0.1
The command ".ci/static-checks.sh" failed and exited with 1 during .
```

# Expected result

`vendor` files should noit be checked!

chavafg:
reopening as it seems that this is still happening

awprice:
Thanks for investigating @chavafg I noticed this still happening too.

grahamwhaley:
If you get stuck, and we have a PR where we can see it happening, post a link and I can try to re-produce and have another look...

PR #2354, opened 2020-03-04
static_checks: fix `skip_paths` function

chavafg:
`skip_paths()` output was being contaminated with the
`grep` output inside the function. Use `grep -q` to silence
the output of `grep`.

Fixes: #2338.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -73,7 +73,7 @@ skip_paths(){
 	for p in "${paths_to_skip[@]}"; do
 		new_list=()
 		for l in "${list[@]}"; do
-			if echo "${l}" | grep -v "${p}"; then
+			if echo "${l}" | grep -qv "${p}"; then
 				new_list=("${new_list[@]}" "${l}")
 			fi
 		done


.ci/static-checks.sh
```diff
@@ -647,6 +647,9 @@ static_check_docs()
 	# since it displayed by default when visiting the repo.
 	exclude_doc_regexs+=(^README\.md$)
 
+	# Exclude documentation files in vendor directory
+	exclude_doc_regexs+=(^vendor/.*\.md$)
+
```
grahamwhaley: Hmm, see, I think/expect the vendor files to already be stripped out by the `skip_paths()` call earlier, which uses the ['paths_to_skip`](https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh#L57-L60) list to prune things out...

```bash
paths_to_skip=(
	"${ignore_clh_generated_code}"
	"vendor"
)
```

I thought the `exclude_doc_regexs` was to remove *our* files that we know we do not check (like license files etc.). Sooo, just want to check that this fix is not papering over a crack in the wrong place?

/cc @jodh-intel

chavafg: @grahamwhaley you're right, the issue was at another place. Fixed the `skip_paths` function. ptal.

PR #2360, opened 2020-03-05
test: Update package manager test for Fedora

### LINKED ISSUE #2359: Update fedora image for package manager docker integration test

GabyCT:
Performing a dnf -y update on a fedora container 31 is failing as
it is complaining about a missing rpm package. This PR removes the option
to use latest Fedora and use Fedora 30 which shows a stable behaviour.

Fixes #2359


integration/docker/package_manager_test.go
```diff
@@ -30,6 +30,20 @@ func tryPackageManagerCommand(container string, command []string, expectedExitCo
 	return exitCode
 }
 
+func distroID() string {
```
devimc: there is already a similar function in `tests/functional/state_test.go`. How about move it to `tests/distro.go` and import it here and in `tests/functional/state_test.go` ?

GabyCT:
@devimc changes applied

Issue #2368, opened 2020-03-09
All integration tests on docker are being skipped for CI on CLH/FC/ARM


likebreath:
# Description of problem
Carlos @jcvenegas and I noticed that all integration tests on docker are being skipped in the `/test-clh` CI for at least  a week. 

With some debugging with @chavafg and @jcvenegas , we noticed the issue is also affecting CIs on Firecracker and ARM. 

# Expected result
Only tests specified in `.ci/hypervisors/clh/configuration_clh.yaml` should be skipped, instead of all docker tests.

# Actual result
Here is a [PR](https://github.com/kata-containers/tests/pull/2317) that was merged 12 days ago but all docker tests were skipped, and here is the CI log snippet:

```
11:53:02 Ran 0 of 250 Specs in 15.481 seconds
11:53:02 SUCCESS! -- 0 Passed | 0 Failed | 0 Pending | 250 Skipped
11:53:02 PASS
```

PR #2369, opened 2020-03-09
ci: Bring back docker integration tests since yq v3.1

likebreath:
All docker integration tests have not been running since yq was updated
to v3.1 (for at least 12 days). The root cause is that yq v3.1 is no
longer returning 'null' as empty strings. To be safe, we keep checks on
both 'null' and '""' (empty string). Also, we observed some regression
test failures with this patch (at least for CIs on ubuntu-initrd, clh,
and Fedoba31).

Fixes: #2368

### DIFF
--- a/.ci/filter/filter_docker_test.sh
+++ b/.ci/filter/filter_docker_test.sh
@@ -25,7 +25,7 @@ filter_and_build()
 {
 	local dependency="$1"
 	local array_docker=$("${GOPATH_LOCAL}/bin/yq" read "${test_config_file}" "${dependency}")
-	[ "${array_docker}" = "null" ] && return
+	( [ "${array_docker}" = "null" ] || [ "${array_docker}" = "" ] ) && return
 	mapfile -t _array_docker <<< "${array_docker}"
 	for entry in "${_array_docker[@]}"
 	do
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -160,6 +160,7 @@ case "${CI_JOB}" in
 	export KUBERNETES="yes"
 	export OPENSHIFT="no"
 	export TEST_CRIO="false"
+	export TEST_DOCKER="true"
 	export experimental_kernel="true"
 	;;
 esac


likebreath:
@chavafg @jcvenegas Looks like the CI works better than we expected. We only got a regression on CLH. In the DBG PR, we disabled the k8s tests which might contribute to the failures we saw on Fedora31 and Ubuntu-initrd. Good news in general.

chavafg:
@likebreath cool, so I think this can be merged... although clh CI will start to fail. Are there already plans to work on the fix for the test that is currently failing?

likebreath:
> @likebreath cool, so I think this can be merged... although clh CI will start to fail. Are there already plans to work on the fix for the test that is currently failing?

Sounds good to me for merging.

The CLH CI failure looks new to me, I will need to ask @jcvenegas to take a look at the failure first.  On top of that, I am expecting a test failure related to CPU hotplug that I enabled recently. I will work on fixing that later.

jcvenegas:
Seems that was a regression we added while docker was not executed

Setting in an explicit way that ext4 is going to be used seems to work:

https://github.com/jcvenegas/runtime/commit/9519f1ce90146d3d1eb0c4797817378b53e21cd3

I can open a PR with that patch

jcvenegas:
@likebreath https://github.com/kata-containers/runtime/pull/2525 here is the PR you could do a depends on to see the error you expected for CPU, btw if you see that error I think would be better just skip those and then merge

likebreath:
@jcvenegas Thanks for the explanation and putting up the PR. I agree we should merge this PR first, and let's debug the failures in separate PRs. Please include more folks into the loop if you found it necessary to merge this PR. Thanks a lot.

bergwolf:
To further avoid failures like this, IMO we should also investigate why https://github.com/kata-containers/tests/pull/2297 passed all tests. @chavafg Any clues?

bergwolf:
cloud hypervisor failure seems to have its own reason and fix (https://github.com/kata-containers/runtime/pull/2525).

likebreath:
> To further avoid failures like this, IMO we should also investigate why #2297 passed all tests. @chavafg Any clues?

@bergwolf That's a good point. We should plan for avoiding similar bugs. 

This kind of bugs is not "visible" to the current CI flow, as they only silence test errors and CI only alarms with test errors. Adding extra checks on the `ginkgo` output in the CI should catch this specific kind of bugs, but may not be a systematic solution. Any thoughts? @chavafg @jcvenegas @amshinde 

Some more context: we tried to trace back to where the docker tests started to skip, but couldn't go beyond more than ~12 days, as all the CI logs were no longer available (including #2297). I would assume PR #2297 skipped all docker tests for the same reason as we saw from this PR: yq 3.1 starts to return `""` (empty strings) instead of `"null"`. It makes the script `.ci/filter/filter_docker_test.sh`  add an extra `|` to the list of `ginkgo -skip` which essentially matches all tests and hence skips all of them.

grahamwhaley:
It doesn't sound trivial. Maybe a quick grep of the logs for the 'tested 0' line might catch such cases.
/cc @jodh-intel , as he loves such challenges and making our automation more robust - he might have some ideas...

amshinde:
@likebreath Thanks for fixing this. We should absolutely make sure that we do not run into this issue again. I believe the docker tests are run with ginkgo. We may have to grep for lines such as "Ran 8 of 8 Specs ", and raise an error if that number is 0. 
cc @devimc Do you know if there is any better way, ginkgo reports the tests run so that we dont have to check in the logs. I think there should be. 
Also, I believe that his could be a good issue for @cmaf to look at.

Issue #2383, opened 2020-03-13
test:docker: Use the right hypervisor when reading from KataConfig 


likebreath:
All docker tests are using the 'DefaultHypervisor' (a.k.a 'qemu') as the key when reading from KataConfig. This will cause test failures for the tests requires information from the configuration file (e.g. cpu_test, mem_test and vsock_test) when kata is not using qemu as its hypervisor (e.g. cloud-hypervisor).

PR #2384, opened 2020-03-13
test:docker: Use the right hypervisor when reading from KataConfig

likebreath:
All docker tests are using the 'DefaultHypervisor' (a.k.a 'qemu') as the
key when reading from KataConfig. This will cause test failures for the
tests requires information from the configuration file (e.g. cpu_test,
mem_test and vsock_test) when kata is not using qemu as its
hypervisor. This patch introduced a new global variable 'KataHypervisor'
which is initialized based on the environment variable 'KATA_HYPERVISOR'
that should be set by the CI script.

Fixes: #2383

### DIFF
--- a/command.go
+++ b/command.go
@@ -19,6 +19,9 @@ var Runtime string
 // Timeout specifies the time limit in seconds for each test
 var Timeout int
 
+// Hypervisor is the hypervisor currently being used with Kata
+var Hypervisor string
+
 // Command contains the information of the command to run
 type Command struct {
 	// cmd exec.Cmd
@@ -31,6 +34,7 @@ type Command struct {
 func init() {
 	flag.StringVar(&Runtime, "runtime", "", "Path of the desired Kata Runtime")
 	flag.IntVar(&Timeout, "timeout", 5, "Time limit in seconds for each test")
+	flag.StringVar(&Hypervisor, "hypervisor", "", "The hypervisor currently being used with Kata")
 
 	flag.Parse()
 }
--- a/config.go
+++ b/config.go
@@ -92,6 +92,7 @@ const (
 
 // KataConfig is the runtime configuration
 var KataConfig KataConfiguration
+var KataHypervisor string
 
 func init() {
 	var err error
@@ -112,6 +113,22 @@ func init() {
 	KataConfig, err = loadKataConfiguration(kataConfigPath)
 	if err != nil {
 		log.Fatalf("failed to load kata configuration: %v\n", err)
+	}
+
+	switch Hypervisor {
+	case "cloud-hypervisor":
+		KataHypervisor = CloudHypervisor
+	case "firecracker":
+		KataHypervisor = FirecrackerHypervisor
+	case "":
+		log.Printf("'-hypervisor' to ginkgo is not set, using 'DefaultHypervisor': '%v'\n", DefaultHypervisor)
+		KataHypervisor = DefaultHypervisor
+	default:
+		log.Fatalf("Invalid '-hypervisor' passed to ginkgo: '%v'\n", Hypervisor)
+	}
+
+	if _, ok := KataConfig.Hypervisor[KataHypervisor]; !ok {
+		log.Fatalf("No configuration found from 'KataConfig' for 'KataHypervisor': '%v'\n", KataHypervisor)
 	}
 }
 
--- a/.ci/hypervisors/firecracker/configuration_firecracker.yaml
+++ b/.ci/hypervisors/firecracker/configuration_firecracker.yaml
@@ -39,6 +39,7 @@ docker:
     - memory constraints
     - Hotplug memory when create containers
     - run container and update its memory constraints
+    - vsock test
   Context:
     - remove bind-mount source before container exits
   It:
--- a/Makefile
+++ b/Makefile
@@ -98,10 +98,12 @@ endif
 
 ifeq ($(KATA_HYPERVISOR),firecracker)
 	./ginkgo -failFast -v -focus "${FOCUS}" -skip "${SKIP_FIRECRACKER}" \
-		./integration/docker/ -- -runtime=${RUNTIME} -timeout=${TIMEOUT}
+		./integration/docker/ -- -runtime=${RUNTIME} -timeout=${TIMEOUT} \
+		-hypervisor=$(KATA_HYPERVISOR)
 else ifeq ($(KATA_HYPERVISOR),cloud-hypervisor)
 	./ginkgo -failFast -v -focus "${FOCUS}" -skip "${SKIP_CLH}" \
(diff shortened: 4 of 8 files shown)


config.go
```diff
@@ -113,6 +114,23 @@ func init() {
 	if err != nil {
 		log.Fatalf("failed to load kata configuration: %v\n", err)
 	}
+
+	EnvKataHypervisor := os.Getenv("KATA_HYPERVISOR")
```
devimc: I think we should pass this information *explicitly* as an option, see [Makefile](1)

[1]: https://github.com/kata-containers/tests/blob/master/Makefile#L101

likebreath: Can you please elaborate more about what's the right way to pass the information of `KATA_HYPERVISOR` to the ginkgo tests? Thanks a lot.

@chavafg @GabyCT @jodh-intel

jodh-intel: Custom Ginkgo CLI options (like `--runtime=` and `--timeout=`) are passed to the `ginkgo` binary after the "`--`" separator. See for example:

- https://github.com/kata-containers/tests/blob/master/Makefile#L101

And those options are defined here:

- https://github.com/kata-containers/tests/blob/master/command.go#L32..L33

Hence, you could add a new `--hypervisor=` option so that users / the CI could run:

```
$ make HYPERVISOR=foo
```

Which would result in the `Makefile` calling `ginkgo ... -- --hypervisor=foo ...`.

jodh-intel:
The ARM CI is failing with the following in the docker log build: artifact:

```
Mar 19 01:39:28 arm-testing-2 systemd[1]: Starting Docker Application Container Engine...
Mar 19 01:39:28 arm-testing-2 dockerd[7790]: unable to configure the Docker daemon with file /etc/docker/daemon.json: the following directives are specified both as a flag and in the configura
tion file: runtimes: (from flag: [kata-runtime], from file: map[kata-runtime:map[path:/usr/local/bin/kata-runtime]]), storage-driver: (from flag: overlay2, from file: devicemapper)
Mar 19 01:39:28 arm-testing-2 systemd[1]: docker.service: Main process exited, code=exited, status=1/FAILURE
```

The FC CI failure looks like it might be a bug:

```
02:09:18     should print the agent logs in the shim journal [It]

          :

02:09:18         <string>: Mar 19 02:09:05 ubuntu1804-azurec22fa0 kata-shim[75476]: time="2020-03-19T02:09:05.834903172Z" level=info msg="copy stdout failed" container=ac18e34889e3b33a20d078ea27964adb906711d8c0d9e7111008718149ebcae8 error="rpc error: code = Unknown desc = EOF" exec-id=ac18e34889e3b33a20d078ea27964adb906711d8c0d9e7111008718149ebcae8 name=kata-shim pid=1 source=shim

          :
02:09:18     to contain substring
02:09:18         <string>: source=agent
```

chavafg:
arm firecracker job will still fail, but the fix is already in #2388, so if everything else passes, we can merge.

command.go
```diff
@@ -31,6 +34,7 @@ type Command struct {
 func init() {
 	flag.StringVar(&Runtime, "runtime", "", "Path of the desired Kata Runtime")
 	flag.IntVar(&Timeout, "timeout", 5, "Time limit in seconds for each test")
+	flag.StringVar(&Hypervisor, "hypervisor", "", "The hypervisor currently being used with Kata")
```
chavafg: this is ok now, but we'll need to update #2396

likebreath:
> arm firecracker job will still fail, but the fix is already in #2388, so if everything else passes, we can merge.

@chavafg Thanks for pointing this out. I will keep an eye on the CI and try to fix errors if any. I will ping you when the CI passes or I got new questions that need help.

command.go
likebreath: If you want to merge #2396 first. I can rebase and adjust the PR accordingly. Just let me know. I am fine either way.

chavafg: I think we can first merge this one, we are still figuring out some other issues with go 1.13.8, so I think that one will be there a little longer. 
wdyt @devimc ?

devimc: sounds good, let's merge this and rebase #2396

likebreath:
Other than two failures for ARM CIs, I see the following two failures:

* jenkins-ci-ubuntu-18-04-virtiofs :
```
14:58:30 ERROR: Found unexpected /opt/kata/bin/qemu-virtiofs-system-x86_64 present
14:58:30 Makefile:86: recipe for target 'functional' failed
14:58:30 make: *** [functional] Error 1
14:58:30 Failed at 81: sudo -E PATH="$PATH" bash -c "make test"
```

* jenkins-ci-vsocks :
```
15:36:21 bats integration/netmon/netmon_test.bats
15:36:21 1..1
15:36:29 not ok 1 test netmon
15:36:29 # (in test file integration/netmon/netmon_test.bats, line 76)
15:36:29 #   `[ "$before_interfaces" -eq "$final_interfaces" ]' failed
15:36:29 # 
15:36:29 # b239c0bf96cc9cbf0ae4c315d2374e1f459878f50eda91ca171f4dc4c00cdd81
15:36:29 # fd5d2e68562ac8eb9917e0894964f13e88821a096fd6a5e8e05baada8882b7fd
15:36:29 # containerA
15:36:29 # containerA
15:36:29 # test
15:36:29 # 
15:36:29 Makefile:203: recipe for target 'netmon' failed
15:36:29 make: *** [netmon] Error 1
15:36:29 Failed at 81: sudo -E PATH="$PATH" bash -c "make test"
15:36:29 Build step 'Execute shell' marked build as failure
```
I didn't see direct correlations to the current PR. Any thoughts?
/cc @chavafg @devimc @jcvenegas @jodh-intel

likebreath:
The CI `jenkins-ci-vsocks ` now is passing. Some of you just fixed it, I think. Thanks a lot.

For now, two failures related to ARM, along with one failure on `jenkins-ci-ubuntu-18-04-virtiofs`.

jodh-intel:
We probably want to wait until #2396 lands before this PR lands as they overlap/conflict.

command.go
likebreath: @jodh-intel Here is the discussion on how to handle the conflict with #2396. Please share if you have additional thoughts. Thank you.

likebreath:
> arm firecracker job will still fail, but the fix is already in #2388, so if everything else passes, we can merge.

Besides the failing `jenkins-ci-ARM-ubuntu-18-04-firecracker`, the other failure is `jenkins-ci-ARM-ubuntu-18-04`. Do we want to fix the `arm ubuntu` before we merge this PR?
@chavafg

chavafg:
@likebreath relaunched jobs and they passed now. I haven't seen those failures before, so let's try to keep an eye to see if they happen again to check furtherly. 

I think we can merge this now. #2388 still needs to pass the CI, but Penny needs to make further changes

Issue #2393, opened 2020-03-18
Dockfile for metrics report can't build as base image update 


jongwu:
# Description of problem

metrics/report/report_dockerfile/Dockerfile can't build as the base image of rocker/tidyverse:lastest update and has no latex-xcolor package which is needed in the Dockerfile. 

# Expected result

# Actual result

@jodh-intel @grahamwhaley  @devimc

PR #2394, opened 2020-03-18
metrics: specify a tag for rocker/tidyverse

jongwu:
the rocker/tidyverse:latest have no latex-xcolor package, so specify an
older tag to let docker build work.
also, specifying a stable tag cancel the potential risk of the update of the latest image.

Fixes: #2393

@jodh-intel  @grahamwhaley  @devimc

### DIFF
--- a/metrics/report/report_dockerfile/Dockerfile
+++ b/metrics/report/report_dockerfile/Dockerfile
@@ -13,10 +13,12 @@
 # We would have used the 'verse' base, that already has some of the docs processing
 # installed, but I could not figure out how to add in the extra bits we needed to
 # the lite tex version is uses.
-FROM rocker/tidyverse
+# Here we specify a tag for base image instead of using latest to let it free from
+# the risk from the update of latest base image.
+FROM rocker/tidyverse:3.6.0
 
 # Version of the Dockerfile
-LABEL DOCKERFILE_VERSION="1.1"
+LABEL DOCKERFILE_VERSION="1.2"
 
 # Without this some of the package installs stop to try and ask questions...
 ENV DEBIAN_FRONTEND=noninteractive


metrics/report/report_dockerfile/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # We would have used the 'verse' base, that already has some of the docs processing
 # installed, but I could not figure out how to add in the extra bits we needed to
 # the lite tex version is uses.
-FROM rocker/tidyverse
+FROM rocker/tidyverse:3.6.0
```
jodh-intel: The ideal would be for us to specify a version number in the [versions file](https://github.com/kata-containers/tests/blob/master/versions.yaml). However, that would require more rework as this `Dockerfile` would need to be generated from `Dockerfile.in`.

I'll leave the decision up to @grahamwhaley but I do think a comment in the file itself stating why this version is using would be useful too.

metrics/report/report_dockerfile/Dockerfile
```diff
@@ -13,7 +13,7 @@
...
 
 # Version of the Dockerfile
 LABEL DOCKERFILE_VERSION="1.1"
```
jodh-intel: Does this version number need bumping too @grahamwhaley?

metrics/report/report_dockerfile/Dockerfile
grahamwhaley: Yeah, I think using the versions file for this would be overkill. We only use it in one place, and it's not part of the CI for instance.
A comment would be appropriate. afaict, what happened is `rocker` updated the base version of ubuntu, and that version does not have the `latex-xcolor` package any more - https://packages.ubuntu.com/xenial/latex-xcolor
I'm fine with pinning the rocker version - the comment should note we pin the base image to align with our package install requirements.

metrics/report/report_dockerfile/Dockerfile
grahamwhaley: yes. the intention is to ensure that if somebody has an already built image on their machine, they will get the update pulled in and built (otherwise docker will probably use the already built/cached image).
Nice catch @jodh-intel

jongwu:
@grahamwhaley @jodh-intel so, what I need to do is just adding comments for the tag in Dockerfile?

jodh-intel:
@jongwu - you just need to update this PR so that it also changes the docker file version to:

```dockerfile
LABEL DOCKERFILE_VERSION="1.2"
```

chavafg:
failures not related with CI, so merging.

Issue #2241, opened 2020-01-20
ci-fast-return.sh skips the CI when only changing vendor directory


chavafg:
# Description of problem

If only `vendor` directory is modified, `ci-fast-return.sh` skips the CI. 
See: https://github.com/kata-containers/runtime/pull/2403#issuecomment-576381479

We are currently kind of safe since we also need to modify the `Gopkg.toml/lock` when we are revendoring, but still a bug.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

GabyCT:
@chavafg , I think this issue has been fixed, can we close this?

chavafg:
@GabyCT I'll check today and will close if this is already solved.

PR #2406, opened 2020-03-24
ci: ci-fast-return.sh should consider vendor changes

chavafg:
We should run the whole CI when there are only changes in the vendor
directory. To accomplish this, use `get_pr_changed_file_details_full`.

Fixes: #2241.


GabyCT:
@chavafg it seems travis did not report the result

chavafg:
> it seems travis did not report the result

weird, and I don't see rebuild buttons in travis. Will re-push and see if that makes it work.

PR #2408, opened 2020-03-24
tests: Enable debug for vcpu test

### LINKED ISSUE #2407: enable debug for vcpu test

GabyCT:
We need to enable debug for the vcpu test in order to see why this tests
is randomly failing.

Fixes #2407

### DIFF
--- a/integration/vcpus/default_vcpus_test.sh
+++ b/integration/vcpus/default_vcpus_test.sh
@@ -7,6 +7,7 @@
 # This will test the default_vcpus
 # feature is working properly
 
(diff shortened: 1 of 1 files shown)


integration/vcpus/default_vcpus_test.sh
```diff
@@ -7,6 +7,8 @@
 # This will test the default_vcpus
 # feature is working properly
 
+set -x
```
chavafg: could this be: `[ -n "$DEBUG" ] && set -x`, so that we can turn debug on/off?

Issue #2358, opened 2020-03-05
docker: dnf -y update failing  using 9pfs


GabyCT:
While trying to perform a `dnf -y update` on a Fedora 31 container we have the following error
```
10:10:04 
10:10:04 Total download size: 1.2 M
10:10:04 Downloading Packages:
10:10:04 shadow-utils-4.6-17.fc31.x86_64.rpm             6.2 MB/s | 1.2 MB     00:00    
10:10:04 --------------------------------------------------------------------------------
10:10:04 Total                                           2.9 MB/s | 1.2 MB     00:00     
10:10:04 Running transaction check
10:10:04 Transaction check succeeded.
10:10:04 Running transaction test
10:10:04 Transaction test succeeded.
10:10:04 Running transaction
10:10:04   Preparing        :                                                        1/1 
10:10:04   Upgrading        : shadow-utils-2:4.6-17.fc31.x86_64                      1/2 
10:10:04   Verifying        : shadow-utils-2:4.6-17.fc31.x86_64                      1/2 
10:10:04   Verifying        : shadow-utils-2:4.6-16.fc31.x86_64                      2/2 
10:10:04 
10:10:04 Failed:
10:10:04   shadow-utils-2:4.6-16.fc31.x86_64      shadow-utils-2:4.6-17.fc31.x86_64     
10:10:04 
10:10:04 
10:10:04 Stderr: Error unpacking rpm package shadow-utils-2:4.6-17.fc31.x86_64
10:10:04 Error: Transaction failed
```

Fedora 31 was released some time ago, it is not clear what make the regression.

GabyCT:
We are running the following code
```
$ docker run -ti --runtime kata-runtime Fedora sh
$ dnf -y update
```

GabyCT:
This is also present on ubuntu

GabyCT:
any idea of why this is happening @fidencio? thanks

fidencio:
Any chance to get the logs from `/var/log/dnf.log` (on the container)? It'd give us a better idea of what may be the causing the issue.

GabyCT:
```
2020-03-10T15:42:47Z INFO --- logging initialized ---
2020-03-10T15:42:47Z DDEBUG timer: config: 133 ms
2020-03-10T15:42:47Z DEBUG DNF version: 4.2.18
2020-03-10T15:42:47Z DDEBUG Command: dnf -y update
2020-03-10T15:42:47Z DDEBUG Installroot: /
[...]

PR #2435, opened 2020-04-01
golangci-lint: fix `invalid pseudo-version` error on ARM CI

Pennyzct:
With golang updated to v1.13, we found golangci-lint build error on ARM CI.
```
08:55:44 go build -o golangci-lint ./cmd/golangci-lint
08:55:49 go: github.com/go-critic/go-critic@v0.0.0-20181204210945-ee9bf5809ead: invalid pseudo-version: does not match version-control timestamp (2019-02-10T22:04:43Z)
```
Taking advices here(https://github.com/filecoin-project/go-filecoin/issues/3486#issuecomment-534858459), we need to update `golangci-lint` to `v1.18.0` to get Go `1.13` support.

Fixes: #2434

Signed-off-by: Penny Zheng <penny.zheng@arm.com>

PR #2436, opened 2020-04-01
s390x : skip dnf update

alicefr:
The package manager dnf tries to update shadow-utils that needs to set
xattr. The option is not supported by 9p. The test is successfull using
another storage driver like devicemapper. As overlay is the default for
the CI and we don't support virtiofs on s390x yet, we can simply skip
this test.

Fixes: #2358

Signed-off-by: Alice Frosi <afrosi@de.ibm.com>

### DIFF
--- a/.ci/s390x/configuration_s390x.yaml
+++ b/.ci/s390x/configuration_s390x.yaml
@@ -17,6 +17,7 @@ docker:
     - CPUs and CPU set
     - Hotplug memory
     - memory constraints
+    - check dnf update
   Context:
     - remove bind-mount source before container exits
     - run container exceeding memory constraints


grahamwhaley:
I strongly suspect those CI fails will be unrelated - afaict we have a bunch of CI instability right now (be that due to our CI or general network fluctuations and external forces I don't know....)
/cc @chavafg @jcvenegas @GabyCT for general CI flaky awareness...

chavafg:
just took a look at CI issues and see that most of them are network related, I remember last couple of days others have seen similar issues:
```
02:24:44 Err:6 http://azure.archive.ubuntu.com/ubuntu bionic/main amd64 conntrack amd64 1:1.4.4+snapshot20161117-6ubuntu2
02:24:44   Could not connect to azure.archive.ubuntu.com:80 (52.177.174.250), connection timed out
02:24:44 Err:7 http://azure.archive.ubuntu.com/ubuntu bionic/main amd64 socat amd64 1.7.3.2-2ubuntu2
02:24:44   Unable to connect to azure.archive.ubuntu.com:http:
```
looking at most recent jobs, seems that those network issues are gone, so I sent a rebuild, lets see.

The other failures are from Arm and Power CI, which currently we know are failing. See #2435 and https://github.com/kata-containers/ci/issues/263, so I think we can ignore these failures.

PR #2439, opened 2020-04-02
ci: Enable repo for perl-IPC-Run

### LINKED ISSUE #2438: enable repo for perl-IPC-Run for rhel

GabyCT:
This PR enables the repository that is needed for perl-IPC-Run for RHEL.
We need to enable this repository as perl-IPC-Run is a dependency for
moreutils package and we need moreutils because we are using chronic.

Fixes #2438

### DIFF
--- a/.ci/setup_env_rhel.sh
+++ b/.ci/setup_env_rhel.sh
@@ -11,6 +11,9 @@ cidir=$(dirname "$0")
 source "/etc/os-release" || "source /usr/lib/os-release"
 source "${cidir}/lib.sh"
 
+echo "Get repo for perl-IPC-Run"
+sudo -E yum-config-manager --enable rhui-rhel-7-server-rhui-optional-rpms
+
 echo "Add epel repository"
 epel_url="https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"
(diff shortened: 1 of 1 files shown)


fidencio:
I know my "Accepted" here doesn't mean anything, but Ack!

jodh-intel:
Hi @GabyCT - I'm not clear *why* rhel needs `perl-IPC-Run` - please can you add some details to the issue and commit explaining this change.

fidencio:
@jodh-intel, the whole scenario is that kata tests depend on `moreutils` (for what specifically? I don't have that knowledge yet), and `moreutils` depend on `Perl-IPC-Run`, which is only available on RHEL as part of that `rhui-rhel-7-server-rhui-optional-rpms` repo.

And, yes, makes total sense to have the explanation as part of the commit message, indeed.

jodh-intel:
@fidencio - ah that's the magic clue I was after - yes, we use `chronic(1)` from `moreutils` and `chronic(1)` needs the perl IPC module.

We use `chronic(1)` for CI runs to suppress some output unless an error occurs.

If you could add something about this to the commit @GabyCT, that would be great.

GabyCT:
@jodh-intel changes applied

PR #2449, opened 2020-04-10
ci: Update network namespace flag for CRIO install

### LINKED ISSUE #2448: ci: Network namespace config option in CRI-O is outdated
Configuration option manage_network_ns_lifecycle was deprecated in CRI-O v1.17 and changed to manage_ns_lifecycle. It should be updated in the ci installation scripts.

cmaf:
Update network_ns_flag to reflect changes in CRI-O v1.17 where
manage_network_ns_lifecycle was deprecated and replaced with
manage_ns_lifecycle.

Fixes #2448

Signed-off-by: Chelsea Mafrica <chelsea.e.mafrica@intel.com>

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -146,10 +146,15 @@ make BUILDTAGS="$(IFS=" "; echo "${build_union[*]}")"
 sudo -E install -D -m0755 runc "/usr/local/bin/crio-runc"
 popd
 
-echo "Set manage_network_ns_lifecycle to true"
-network_ns_flag="manage_network_ns_lifecycle"
+echo "Set manage_ns_lifecycle to true"
+network_ns_flag="manage_ns_lifecycle"
+# Set ns_network_flag for CRI-O versions less than 1.17
+crio_version_current=$(crio --version | head -1 | cut -d ' ' -f3)
+if [ "$(compare_versions "$crio_version_current" "1.17.0")" -eq "1" ]; then
+	network_ns_flag="manage_network_ns_lifecycle"
+fi
 sudo sed -i "/\[crio.runtime\]/a$network_ns_flag = true" "$crio_config_file"
-sudo sed -i 's/manage_network_ns_lifecycle = false/#manage_network_ns_lifecycle = false/' "$crio_config_file"
+sudo sed -i "s/$network_ns_flag = false/#$network_ns_flag = false/" "$crio_config_file"
 
 echo "Add docker.io registry to pull images"
 # Matches cri-o 1.10 file format
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -235,3 +235,28 @@ get_pod_config_dir() {
 	pod_config_dir="${BATS_TEST_DIRNAME}/runtimeclass_workloads"
 	info "k8s configured to use runtimeclass"
 }
+
+# Takes two versions in the format of xx.xx.xx and compares them.
+# Versions must contain an equal number of fields.
+#
+# Returns string
+#  0 - first version is greater or versions are the same
+#  1 - second version is greater
+compare_versions() {
+	local -r version_1="$1"
+	local -r version_2="$2"
+
+	local fields_1=($(echo $version_1 | tr "." "\n"))
+	local fields_2=($(echo $version_2 | tr "." "\n"))
+
+	[ ${#fields_1[@]} -ne ${#fields_2[@]} ] && die "Cannot compare versions; number of fields differs"
+
+	for n in "${!fields_1[@]}"; do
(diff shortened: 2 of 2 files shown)


.ci/install_crio.sh
```diff
@@ -146,10 +146,10 @@ make BUILDTAGS="$(IFS=" "; echo "${build_union[*]}")"
 sudo -E install -D -m0755 runc "/usr/local/bin/crio-runc"
 popd
 
-echo "Set manage_network_ns_lifecycle to true"
-network_ns_flag="manage_network_ns_lifecycle"
+echo "Set manage_ns_lifecycle to true"
+network_ns_flag="manage_ns_lifecycle"
```
chavafg, line 150: As we an older version of CRI-O for openshift testing (in the Fedora job), we would need to set the value of this variable above at: https://github.com/kata-containers/tests/blob/master/.ci/install_crio.sh#L73-L77

something like: 
```
                if [ "$KUBERNETES" == "yes" ]; then
			crio_version=$(get_version "externals.crio.version")
                        network_ns_flag="manage_ns_lifecycle"
		else
			crio_version=$(get_version "externals.crio.meta.openshift")
                        network_ns_flag="manage_network_ns_lifecycle"
		fi
```

.ci/install_crio.sh
```diff
@@ -146,10 +146,10 @@ make BUILDTAGS="$(IFS=" "; echo "${build_union[*]}")"
 sudo -E install -D -m0755 runc "/usr/local/bin/crio-runc"
 popd
 
-echo "Set manage_network_ns_lifecycle to true"
-network_ns_flag="manage_network_ns_lifecycle"
+echo "Set manage_ns_lifecycle to true"
```
amshinde, line 149: I think it may be useful to have backward compatibility here. Wdyt @chavafg ?

cmaf:
@chavafg thanks, I applied your suggestions

chavafg:
@cmaf yeah, agree 👍

.ci/install_crio.sh
cmaf: @amshinde adding this

cmaf:
@amshinde @chavafg @saschagrunert thanks for the feedback, I applied the changes

lib/common.bash
```diff
@@ -235,3 +235,28 @@ get_pod_config_dir() {
...
+	local -r version_1="$1"
+	local -r version_2="$2"
+
+	fields_1=($(echo $version_1 | tr "." "\n"))
+	fields_2=($(echo $version_2 | tr "." "\n"))
+
+	if [ ${#fields_1[@]} -ne ${#fields_2[@]} ]; then die "Cannot compare versions; number of fields differs"; fi
+
+	for n in "${!fields_1[@]}"; do
+		if [ ${fields_1[$n]} -lt ${fields_2[$n]} ]; then
```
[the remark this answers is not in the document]
cmaf: @chavafg with -le, wouldn't that cause it return early, before comparing subsequent fields if the the current was equal?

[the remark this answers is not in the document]
chavafg: ohh yes, you're right, sorry for the confusion.

cmaf:
@chavafg do both commits need to reference the issue?

chavafg:
No, with one is fine. 
I see that wip check got stuck, @cmaf can you add and remove the wip label again to see if it passes now?

lib/common.bash
```diff
@@ -235,3 +235,28 @@ get_pod_config_dir() {
...
+# Versions must contain an equal number of fields.
+#
+# Returns string
+#  0 - first version is greater or versions are the same
+#  1 - second version is greater
+compare_versions() {
+	local -r version_1="$1"
+	local -r version_2="$2"
+
+	fields_1=($(echo $version_1 | tr "." "\n"))
```
jodh-intel, line 249: Nit: You could also make these two variables `local`.

lib/common.bash
```diff
@@ -235,3 +235,28 @@ get_pod_config_dir() {
...
+	fields_2=($(echo $version_2 | tr "." "\n"))
+
+	if [ ${#fields_1[@]} -ne ${#fields_2[@]} ]; then die "Cannot compare versions; number of fields differs"; fi
```
jodh-intel, line 252: You could simplify to:

```sh
 [ ${#fields_1[@]} -ne ${#fields_2[@]} ] && die "Cannot compare versions; number of fields differs"
```

lib/common.bash
cmaf: Thanks, I changed this

lib/common.bash
cmaf: Also changed this

PR #2450, opened 2020-04-13
makefile: Sort the tests in alphabetical order

### LINKED ISSUE #581: Makefile: sort the contents....
We try to maintain, where possible, files and lists in alphabetical order - but we seem to have a few things in our [Makefile](https://github.com/kata-containers/tests/blob/master/Makefile#L11) that break those rules (a cri one I think for instance).

Let's sort the file and lists ... **BUT** ... there is a risk here. In the past we have had some test run ordering issues whereby tests would randomly fail if they were ordered a certain way.
Thus, we can at least open a PR for this and see what happens. The tests should be inter-dependant, and we presently have no hard requirement to run any specific tests before others.

GabyCT:
This PR sort the contests of the tests in alphabetical order.

Fixes #581

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,10 +28,28 @@ ifeq (${CI}, true)
 endif
 
 # union for 'make test'
-UNION := functional debug-console $(DOCKER_DEPENDENCY) openshift crio docker-compose network \
-	docker-stability oci netmon kubernetes swarm vm-factory \
-	entropy ramdisk shimv2 tracing time-drift compatibility vcpus \
-	$(PODMAN_DEPENDENCY) pmem
+UNION := crio \
+	compatibility \
+	debug-console \
+	$(DOCKER_DEPENDENCY) \
+	docker-compose \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -28,10 +28,10 @@ ifeq (${CI}, true)
 endif
 
 # union for 'make test'
-UNION := functional debug-console $(DOCKER_DEPENDENCY) openshift crio docker-compose network \
-	docker-stability oci netmon kubernetes swarm vm-factory \
-	entropy ramdisk shimv2 tracing time-drift compatibility vcpus \
-	$(PODMAN_DEPENDENCY) pmem
+UNION := crio compatibility debug-console $(DOCKER_DEPENDENCY) docker-compose \
```
grahamwhaley, line 31: lgtm
if we make this a 'one item per line' list then I think the ordering would be even more obvious. wdyt @jodh-intel

Issue #2470, opened 2020-04-22
CI: filter out libcontainer logs


jodh-intel:
Filter out libcontainer log messages from the agent as they trigger errors from the [log-parser](https://github.com/kata-containers/tests/tree/master/cmd/log-parser) that runs in the CI (since libcontainer log messages do not contain the expected set of fields).

For example, here's a log parser error from http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-18-04-PR-initrd/2580/consoleFull which shows a libcontainer log entry that fails as it does not set the required PID field (`missing pid`):

```
[...]

PR #2467, opened 2020-04-20
ci: teardown: save shimv2 logs

grahamwhaley:
The shimv2 runtime logs to 'kata' in the journal, not 'kata-runtime'.
Add that to the CI teardown script log collection methods so we
capture them if available.

Fixes: #2466
(Also Fixes the original/duplicate https://github.com/kata-containers/ci/issues/270)

Signed-off-by: Graham Whaley <graham.whaley@intel.com>

PR #2471, opened 2020-04-22
CI: Hide libcontainer logs from log-parser

jodh-intel:
Filter out libcontainer log messages (which get logged by the agent) as they trigger errors from the the log-parser that runs in the CI. The reason for the errors is that the log-parser expects all log entries to provide a standard set of fields, but libcontainer does not provide a way to pass a logger object into it, so it logs in its own format, which does not include mandatory fields like "pid".

See: https://github.com/kata-containers/tests/tree/master/cmd/log-parser#logfile-requirements

Fixes: #2470.

### DIFF
--- a/.ci/teardown.sh
+++ b/.ci/teardown.sh
@@ -240,7 +240,15 @@ check_log_files()
 		file="${component}.log"
 		args="--no-pager -q -o cat -a -t \"${component}\""
 
-		cmd="sudo journalctl ${args} > ${file}"
+		# grab the component logs and exclude all log entries that
+		# don't appear to come from Kata.
+		#
+		# This reduces the effectiveness of the log parser checking
+		# slightly but is the only semi-reasonable way to exclude log
+		# messages from libcontainer whose API unfortunately does not
+		# accept a logger to perform logging.
+		cmd="sudo journalctl ${args} | grep -v name=kata > ${file}"
+
 		eval "$cmd" || true
 	done
 
--- a/.ci/teardown.sh
+++ b/.ci/teardown.sh
@@ -251,7 +251,7 @@ check_log_files()
 		file="${unit}.log"
 		args="--no-pager -q -o cat -a -u \"${unit}\""
 
-		cmd="sudo journalctl ${args} |grep ^time= > ${file}"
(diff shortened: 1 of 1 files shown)


grahamwhaley:
You've missed the SoB on the whitepace commit :-)
We should also correlate this with #2467 , if you have not already - I'm presuming they don't clash?

jodh-intel:
Updated. Well, the diffs look different to me, but presumably we want to land this PR before yours anyhow?

grahamwhaley:
What I mean is that we check that this patch does not filter out the shimv2 containerd `kata` logs completely - I'll have a stare at the two to check.

jodh-intel:
Oh ugh. Right. It's a shame the libcontainer API doesn't allow a logger to be passed in to avoid "polluting" our logs with "malformed" logs so we could avoid hacks like this.

grahamwhaley:
Checked what the `kata` journal lines look like - afaict, they will not get filtered by this PR - they don't seem to container a `name=` field... they look like:

```
Apr 20 14:18:54 ubuntu1804-azure82cdc0 kata[79595]: time="2020-04-20T14:18:54.372435579Z" level=info msg="loaded configuration" ID=73227e3f9daf5b0faf17b6a1af72b657dfcd4b5e58afa112d9ea9d46be52e887 file=/etc/kata-containers/configuration.toml format=TOML source=katautils
```

You can see them in the `kata-shimv2*` artifacts attached to the Jenkins CI run of the other PR if you need etc.

chavafg:
took a look at the log fragments and most of the job prints similar (few) entries as before. The one that printed a lot of log entries was initrd job, but with this PR it got reduced and now it looks similar to the other jobs, so I think the PR works. Anyway I'll let @jodh-intel or @grahamwhaley review before merging.

grahamwhaley:
ack, that initrd log looks sane to me. You might need to ack @chavafg

jodh-intel:
Yep. And in fact the errors being reported by the log parser now are real bugs. I'll raise PRs...

jodh-intel:
Raised https://github.com/kata-containers/runtime/pull/2645 to fix one of the offenders and https://github.com/kata-containers/runtime/issues/2643 to highlight the need for a fix for the other one ;)

PR #2171, opened 2019-12-13
Fix crio execsync issue

amshinde:
CI today does not use conmon version specified in
version.yaml, instead builds from master.
The latest version of conmon seemed to have changed certain
exit codes for exec and introduced errors described in
kata-containers/runtime#2352

Have the CI checkout version from versions.yaml and build conmon.

Fixes #2170

PR #2481, opened 2020-04-27
ci: Build podman from source

### LINKED ISSUE #2480: build podman from source

GabyCT:
This PR will build and install podman from source in order to avoid
future issues where a certain version of podman in Fedora 31 is not present.
This issue is related with https://github.com/kata-containers/ci/issues/271.

Fixes #2480

### DIFF
--- a/.ci/setup_env_fedora.sh
+++ b/.ci/setup_env_fedora.sh
@@ -11,15 +11,10 @@ cidir=$(dirname "$0")
 source /etc/os-release || source /usr/lib/os-release
 source "${cidir}/lib.sh"
 TEST_CGROUPSV2="${TEST_CGROUPSV2:-false}"
+arch=$("${cidir}"/kata-arch.sh -d)
 
 echo "Install chronic"
 sudo -E dnf -y install moreutils
-
-if [ "${TEST_CGROUPSV2}" == "true" ]; then
-	echo "Install podman"
-	version=$(get_test_version "externals.podman.version")
-	sudo -E dnf -y install podman-"${version}"
-fi
 
 declare -A minimal_packages=( \
 	[spell-check]="hunspell hunspell-en-GB hunspell-en-US pandoc" \
@@ -76,6 +71,21 @@ main()
 	if [ "$KATA_KSM_THROTTLER" == "yes" ]; then
 		echo "Install ${KATA_KSM_THROTTLER_JOB}"
 		chronic sudo -E dnf -y install ${KATA_KSM_THROTTLER_JOB}
+	fi
+
+	if [ "${TEST_CGROUPSV2}" == "true" ]; then
+		echo "Install podman dependencies"
+		sudo dnf -y builddep podman
+		dnf deplist podman --archlist "${arch}",noarch | awk '/provider:/ {print $2}' | sort -u | xargs sudo -E dnf -y install
+        	echo "Install podman"
+        	version=$(get_test_version "externals.podman.version")
+		podman_repo="github.com/containers/libpod"
+		go get -d "${podman_repo}" || true
+		pushd "${GOPATH}/src/${podman_repo}"
+		git checkout v"${version}"
+		make BUILDTAGS="selinux seccomp"
+		sudo -E PATH=$PATH make install
+		popd
 	fi
 }
 
--- a/integration/podman/run_podman_tests.sh
+++ b/integration/podman/run_podman_tests.sh
@@ -24,18 +24,6 @@ test_repository="github.com/kata-containers/tests"
 test_repository_path="${GOPATH}/src/${test_repository}"
 version=$(get_test_version "externals.podman.version")
 
-function setup() {
-	# Clone podman repository if it is not already present.
-	if [ ! -d "${podman_repository_path}" ]; then
-		go get -d "${podman_repository}" || true
-	fi
-
-	pushd "${podman_repository_path}"
-	git checkout v"${version}"
-	make
-	popd
-}
-
 function setup_ginkgo() {
 	pushd "${test_repository_path}"
 	ln -sf . vendor/src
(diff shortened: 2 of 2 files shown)


chavafg:
should we first try to install podman using dnf and if the version is not available, we can proceed to build/install it from sources?

GabyCT:
@chavafg at kata-containers/ci#271, we agreed that the best option is to build and install

.ci/setup_env_fedora.sh
```diff
@@ -77,6 +71,20 @@ main()
 		echo "Install ${KATA_KSM_THROTTLER_JOB}"
 		chronic sudo -E dnf -y install ${KATA_KSM_THROTTLER_JOB}
 	fi
+
+	if [ "${TEST_CGROUPSV2}" == "true" ]; then
+		echo "Install podman dependencies"
+		dnf deplist podman --archlist x86_64,noarch | awk '/provider:/ {print $2}' | sort -u | xargs sudo -E dnf -y install
```
jodh-intel, line 77: Can you remove this architecture-specific option?

In fact I think you can do this and simplify it in one go:

```sh
sudo dnf -y builddep podman
```

wainersm: Notice that you are installing the build dependencies for a version that does not correspond to the one you gonna build. Don't know how likely this can be a problem, though.

Using `builddep` seems better but with the cost of adding extra-dependencies, i.e., those needed to build the rpm. Well, on CI it should not be a problem.

GabyCT: any idea @wainersm if I need to add something else?

chavafg: it seems that `conmon` needs to be installed before running this. We install `conmon` as part of of the [crio installation](https://github.com/kata-containers/tests/blob/master/.ci/install_crio.sh#L33). maybe we should take the conmon installation out of that script and have an `install_conmon.sh` ? so that we can reuse it here? 
We may also want to build/install podman using its own script (e.g. `install_podman.sh`) as we do for the other pod/container managers.

GabyCT: @chavafg maybe we just need to install the package, we do not need to build it as it says at
https://podman.io/getting-started/installation

GabyCT: @chavafg the installation will apply only for podman

GabyCT: what I do not understand is why
```
dnf deplist podman --archlist x86_64,noarch | awk '/provider:/ {print $2}' | sort -u | xargs sudo -E dnf -y install
```
did the installation of all dependencies (including conmon)

chavafg: > @chavafg maybe we just need to install the package, we do not need to build it as it says at

I would prefer to build it as we have had issues with some versions of `conmon` in the past, so we currently stick to the one in https://github.com/kata-containers/runtime/blob/master/versions.yaml#L183 
you can see related: https://github.com/kata-containers/tests/pull/2171

GabyCT: @chavafg , I think that we should use the dnf installation of the packages like I put in my previous comment, the issue is that even if I build or install conmon then I will have issues like
```
Error: could not get runtime: default OCI runtime "crun" not found: invalid argument
```

GabyCT: @jodh-intel and @wainersm, I change the command to the original form, the issue is that it does not install all the dependencies that we need (like conmon, crun, etc) and even if I installed them manually, then I got errors like `/usr/share/containers/libpod.conf: No such file or directory`

jodh-intel:
I bit of shell munging shows the following:

| `builddep` | `deplist` |
|-|-|
|btrfs-progs-devel|bash|
|glib2-devel|conmon|
|glibc-static|containernetworking-plugins|
|golang|containers-common|
|golang-bin|container-selinux|
|golang-github-cpuguy83-md2man|crun|
|golang-src|glibc|
|gpgme-devel|gpgme|
|Installation|iptables|
|libassuan-devel|libassuan|
|libblkid-devel|libgpg-error|
|libbtrfs|libseccomp|
|libbtrfsutil|nftables|
|libffi-devel|podman-plugins|
|libgpg-error-devel||
|libmount-devel||
|libseccomp-devel||
|libselinux-devel||
|libsepol-devel||
|libxcrypt-static||
|mercurial-lang||
|mercurial-py2||
|ostree-devel||
|ostree-libs||
|pcre2-devel||
|pcre2-utf16||
|pcre-cpp||
|pcre-devel||
|pcre-utf16||
|pcre-utf32||
|python27||
|python2-rpm-macros||
|python3-rpm-generators||
|python-rpm-macros||
|systemd-devel||
|tix||
|tk||
|zlib-devel||

Since we're building and running, we're going to need *both* sets of packages installed.

GabyCT:
@jodh-intel thanks for the info :)

GabyCT:
@jodh-intel changes applied

.ci/setup_env_fedora.sh
```diff
@@ -77,6 +72,21 @@ main()
...
+		sudo dnf -y builddep podman
+		dnf deplist podman --archlist "${arch}",noarch | awk '/provider:/ {print $2}' | sort -u | xargs sudo -E dnf -y install
+        	echo "Install podman"
+        	version=$(get_test_version "externals.podman.version")
+		podman_repo="github.com/containers/libpod"
+		go get -d "${podman_repo}" || true
+		pushd "${GOPATH}/src/${podman_repo}"
+		git checkout v"${version}"
+		make BUILDTAGS="selinux seccomp"
+		sudo -E PATH=$PATH make install PREFIX=/usr
```
jodh-intel, line 87: Note: Is the `PREFIX` required? Although we don't expect (or want) devs to run this script, if they did it might cause surprising results. Might be marginally safer to write to `/usr/local` or add a check on `KATA_DEV_MODE` (which we do use in various CI intsall scripts, but not the `setup_env_*` ones atm fwics).

GabyCT: @jodh-intel I removed the PREFIX part

PR #2498, opened 2020-05-04
ci: Modify ubuntu setup to support 20.04

### LINKED ISSUE #2497: Modify ubuntu setup to support 20.04

GabyCT:
This PR modifies the setup script to support ubuntu 20.04 by removing
a non existing package as well as to modify the addition of the repository
for docker.

Fixes #2497

### DIFF
--- a/.ci/setup_env_ubuntu.sh
+++ b/.ci/setup_env_ubuntu.sh
@@ -42,7 +42,7 @@ declare -A packages=( \
 	[redis]="redis-server" \
 )
 
-if [ "$(uname -m)" == "x86_64" ] && [ "${NAME}" == "Ubuntu" ] && [ "$(echo "${VERSION_ID} >= 18.04" | bc -q)" == "1" ]; then
+if [ "$(uname -m)" == "x86_64" ] && [ "$(echo "${VERSION_ID} >= 18.04" | bc -q)" == "1" ]; then
 	packages[qemu_dependencies]+=" libpmem-dev"
 fi
 
--- a/cmd/container-manager/manage_ctr_mgr.sh
+++ b/cmd/container-manager/manage_ctr_mgr.sh
@@ -105,7 +105,13 @@ install_docker(){
 			sudo -E apt-get -y install apt-transport-https ca-certificates software-properties-common
 			repo_url="https://download.docker.com/linux/ubuntu"
 			curl -fsSL "${repo_url}/gpg" | sudo apt-key add -
-			sudo -E add-apt-repository "deb [arch=${arch}] ${repo_url} $(lsb_release -cs) stable"
+			# There is not a docker repository for Ubuntu 20.04, that is the main reason why
+			# we are using bionic (18.04 LTS) repository
+			if [ "$(echo "${VERSION_ID} >= 20.04" | bc -q)" == "1" ]; then
+				sudo -E add-apt-repository "deb [arch=${arch}] ${repo_url} bionic stable"
+			else
+				sudo -E add-apt-repository "deb [arch=${arch}] ${repo_url} $(lsb_release -cs) stable"
+			fi
 			sudo -E apt-get update
 			docker_version_full=$(apt-cache madison $pkg_name | grep "$docker_version" | awk '{print $3}' | head -1)
 			sudo -E apt-get -y install "${pkg_name}=${docker_version_full}"


.ci/setup_env_ubuntu.sh
```diff
@@ -42,6 +42,10 @@ declare -A packages=( \
 	[redis]="redis-server" \
 )
 
+if [ "${NAME}" == "Ubuntu" ] && [ "$(echo "${VERSION_ID} < 20.04" | bc -q)" == "1" ]; then
+	packages[cri-containerd_dependencies]+=" btrfs-tools"
```
cmaf, line 46: Does btrfs-tools have an alternative in 20.04?

GabyCT: @cmaf I do not think so

cmd/container-manager/manage_ctr_mgr.sh
```diff
@@ -105,7 +105,11 @@ install_docker(){
 			sudo -E apt-get -y install apt-transport-https ca-certificates software-properties-common
 			repo_url="https://download.docker.com/linux/ubuntu"
 			curl -fsSL "${repo_url}/gpg" | sudo apt-key add -
-			sudo -E add-apt-repository "deb [arch=${arch}] ${repo_url} $(lsb_release -cs) stable"
+			if [ "$(echo "${VERSION_ID} >= 20.04" | bc -q)" == "1" ]; then
+				sudo -E add-apt-repository "deb [arch=${arch}] ${repo_url} bionic stable"
```
jodh-intel, line 109: I'm really not keen on doing things like this. Do we really *really* need these btrfs tools installed if we're not using btrfs?

GabyCT: @jodh-intel, I am not sure if I understand you...should I just do the 
`${VERSION_ID} >= "20.04"`

devimc: no @GabyCT , since we don't use `btrfs` in kata containers, maybe the package `btrfs-tools` is not longer needed

chavafg: would be good to check building/installing `containerd/cri`, as I remember I needed to add it for another distro.
@GabyCT maybe you already did, but if not, please try to run `.ci/install_cri_containerd.sh`. If that finishes successfully, we can remove the package.

.ci/setup_env_ubuntu.sh
```diff
@@ -42,6 +42,10 @@ declare -A packages=( \
 	[redis]="redis-server" \
 )
 
+if [ "${NAME}" == "Ubuntu" ] && [ "$(echo "${VERSION_ID} < 20.04" | bc -q)" == "1" ]; then
```
chavafg, line 45: I think we can remove the  `[ "${NAME}" == "Ubuntu" ]` , we are supposed to run this script only on ubuntu distro, right?

GabyCT:
@chavafg and @devimc changes applied

cmd/container-manager/manage_ctr_mgr.sh
```diff
@@ -105,7 +105,13 @@ install_docker(){
 			sudo -E apt-get -y install apt-transport-https ca-certificates software-properties-common
 			repo_url="https://download.docker.com/linux/ubuntu"
 			curl -fsSL "${repo_url}/gpg" | sudo apt-key add -
-			sudo -E add-apt-repository "deb [arch=${arch}] ${repo_url} $(lsb_release -cs) stable"
+			# There is not a docker repository for Ubuntu 20.04, that is the main reason why
```
jodh-intel, line 108: What worries me is that presumably at some point docker will add a 20.04 repo and we might not notice. But I guess this is the only option we have for now.

It would be great if we could create a daily CI test that checked https://download.docker.com/linux/ubuntu/dists/ and looked for `focal` so we'd get a warning ;)

jodh-intel: Hmm, a random idea... we *could* make this code "auto-detect" the new repo by getting it to try to `add-apt-repository` for `focal` and if that doesn't work, use `bionic`. You'd have to be careful due to the `set -e` of course and there is also the issue again of how we'd detect / notice that the new repo had appeared so we could remove the "auto-detect" code.

Issue #2506, opened 2020-05-04
update the way we get the crio version 


chavafg:
**Which feature do you think can be improved?**
Starting cri-o 1.18, the output of `crio --version` has changed to:

```
crio version
Version:       1.18.0
GitCommit:     7d79f42b28ad00cf2e7d86604a5a4007303ac328
GitTreeState:  clean
BuildDate:     2020-05-04T22:57:42Z
GoVersion:     go1.13.9
Compiler:      gc
Platform:      linux/amd64
```

while the crio 1.17 and older show:
```
crio --version
crio version 1.17.0-dev
commit: 0eec454168e381e460b3d6de07bf50bfd9b0d082
```

we need to update our scripts in order to be able to get the correct value on all crio versions.

PR #2507, opened 2020-05-04
ci: update the way we get crio version

chavafg:
We need to update how we get the value of the crio version
as the output of `crio --version` has changed since
crio 1.18.0.
This change now works with new and old versions of crio.

Fixes: #2506.

### DIFF
--- a/.ci/configure_crio_for_kata.sh
+++ b/.ci/configure_crio_for_kata.sh
@@ -21,7 +21,7 @@ crio_config_file="/etc/crio/crio.conf"
 runc_flag="\/usr\/local\/bin\/crio-runc"
 kata_flag="\/usr\/local\/bin\/kata-runtime"
 
-minor_crio_version=$(crio --version | head -1 | cut -d '.' -f2)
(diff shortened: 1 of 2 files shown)


.ci/configure_crio_for_kata.sh
```diff
@@ -21,7 +21,7 @@ crio_config_file="/etc/crio/crio.conf"
 runc_flag="\/usr\/local\/bin\/crio-runc"
 kata_flag="\/usr\/local\/bin\/kata-runtime"
 
-minor_crio_version=$(crio --version | head -1 | cut -d '.' -f2)
+minor_crio_version=$(crio --version | egrep "[0-9]\.[0-9]+" | head -1 | sed 's/[^0-9.]*//g' | cut -d '.' -f2)
```
devimc, line 24: since this is semver, I think you can use:  crio --version | egrep -o "[0-9]+\.[0-9]+\.[0-9]+"

chavafg: thanks for the suggestion, I changed it.

PR #2530, opened 2020-05-11
metrics: Fix footprint data results

### LINKED ISSUE #2169: Footprint-busybox.json results in "KB" should be in "B"
Metrics reporting, footprint-busybox.json
**Specify the feature you think could be made better**
memory unit reporting in the result should be in "Bytes", not in "KB"
**How can it be improved?**
change the scripts 
image

Describe how specifically you think it could be improved.

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

GabyCT:
The results that we are getting from the footprint_data.sh were we are using
smem tool should be in B instead of KB.

Fixes #2169

### DIFF
--- a/metrics/density/footprint_data.sh
+++ b/metrics/density/footprint_data.sh
@@ -179,7 +179,7 @@ function grab_vm_uss() {
 			"qemu": "$qemu",
 			"virtiofsd": "$virtiofsd",
 			"total": $total,
-			"Units": "KB"
+			"Units": "B"
 		}
 EOF
 )"
@@ -203,7 +203,7 @@ function grab_vm_pss() {
 			"qemu": "$qemu",
 			"virtiofsd": "$virtiofsd",
 			"total": $total,
-			"Units": "KB"
+			"Units": "B"
 		}
 EOF
 )"
@@ -220,7 +220,7 @@ function grab_dockerd_pss() {
 	local json="$(cat << EOF
 		"dockerd": {
 			"pss": $item,
-			"Units": "KB"
+			"Units": "B"
 		}
 EOF
 )"
@@ -239,7 +239,7 @@ function grab_all_pss() {
 	local json="$(cat << EOF
 		"all_pss": {
 			"pss": $item,
-			"Units": "KB"
+			"Units": "B"
 		}
 EOF
 )"
@@ -255,7 +255,7 @@ function grab_user_smem() {
 	local json="$(cat << EOF
 		"user_smem": {
 			"userspace": $item,
-			"Units": "KB"
+			"Units": "B"
 		}
 EOF
 )"
@@ -267,12 +267,11 @@ function grab_slab() {
 	# Grabbing slab total from meminfo is easier than doing the math
 	# on slabinfo
 	item=$(fgrep "Slab:" /proc/meminfo | awk '{print $2}')
-	((item*=1024))
 
 	local json="$(cat << EOF
 		"slab": {
 			"slab": $item,
-			"Units": "KB"
+			"Units": "B"
 		}
 EOF
 )"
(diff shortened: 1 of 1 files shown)


metrics/density/footprint_data.sh
```diff
@@ -316,12 +316,13 @@ function grab_system() {
 			"avail": $avail,
 			"avail_decr": $avail_decr,
 			"cached": $cached,
-			"smem_free": $smem_free,
 			"free_decr": $free_decr,
 			"anon": $anon,
 			"mapped": $mapped,
 			"meminfo_cached": $meminfo_cached,
-			"Units": "KB"
+			"Units": "KB",
```
grahamwhaley, line 323: I think having two entries both called `Units` is not valid JSON. You'll either have to break the `B` version out into a new section, or probably better is to scale it down to KB in this section. If you move it to a new place in the JSON, you will probably also break the evaluators (any metrics or report generator that processes this data).

jodh-intel, line 323: ?

GabyCT: @grahamwhaley changes applied

metrics/density/footprint_data.sh
```diff
@@ -297,6 +297,7 @@ function grab_system() {
 
 	# free memory from smem
 	local smem_free=$(get_memfree)
+	((smem_free/=1024))
```
grahamwhaley, line 300: The KB->B smem changes I think are currect - but alas, I suspect there is more work to do in [this function](https://github.com/kata-containers/tests/blob/master/metrics/density/footprint_data.sh#L289-L330).

Currently the Unit is KB, but actually I think most of the entries are B. Let's take for instance the `avail`....

```
	local avail=$(free -b | head -2 | tail -1 | awk '{print $7}')
...
			"avail": $avail,
```

Thus, I think we are measuring in bytes, but nominally stating the units as KB. I think that's the same for probably all the system memory items in that section. For instance, a local run shows me:

```
                "system": {
                        "avail": 26796888064,
                        "avail_decr": 3888603136,
                        "cached": 950718464,
                        "smem_free": 25550640,
                        "free_decr": 20158267392,
                        "anon": 5749055488,
                        "mapped": 782737408,
                        "meminfo_cached": 883929088,
                        "Units": "KB"
                },
```

but `/proc/meminfo` looks like:

```
# cat /proc/meminfo
MemTotal:       32828092 kB
MemFree:        29083096 kB
MemAvailable:   30016924 kB
Buffers:           36428 kB
Cached:          1151628 kB
SwapCached:            0 kB
Active:          2774356 kB
Inactive:         688460 kB
Active(anon):    2189932 kB
Inactive(anon):     2012 kB
Active(file):     584424 kB
Inactive(file):   686448 kB
Unevictable:           0 kB
Mlocked:               0 kB
SwapTotal:       8388604 kB
SwapFree:        8388604 kB
Dirty:             20220 kB
...
```

GabyCT:
@grahamwhaley , thanks for the feedback, so here it is what I do, the `smem` results I changed it to `B`, now I can see that the info from `/proc/meminfo` is giving the results in `KB` (so I removed the multiplication that there was before), then the info from `free` is giving in `B` so in that case I converted them in `KB` so we have all the info in `KB` from the system section thanks

grahamwhaley:
Anybody want to lend a hand reviewing and helping with the metrics.... just need one more review here....

GabyCT:
@grahamwhaley should we wait or can we merge this? thanks

grahamwhaley:
You probably want to find/convince somebody else to take an interest in the metrics code - as soon I won't be looking any more.....

PR #2534, opened 2020-05-11
metrics: Get runc name

### LINKED ISSUE #1796: metrics: report: runc path/name has changed - need to adapt
# Description of problem

We hard code some `runc` names and paths into some of the metrics code, in order to take measurements or extract information from it. It seems the `runc` process has changed binary name and/or path, and the checks are no longer working, meaning we miss some metrics data such as the PSS footprint when testing with `RUNTIME=runc`.

e.g., when running the memory footprint test:
```
tests/metrics/lib/json.bash: line 96: docker-runc: command not found
...
ERROR: Runc process not found
```

# Expected result

Correctly find `runc`, or fail more gracefully

At https://github.com/kata-containers/tests/blob/master/metrics/lib/json.bash#L96 we open code `runc` as `docker-runc`, which seems to no longer exist. I can see a `/usr/sbin/runc` though.

And in https://github.com/kata-containers/tests/blob/master/metrics/density/docker_memory_usage.sh#L86 we open code some runc measurements, which are failing, which is why we get the `ERROR` line in the output.

GabyCT:
This PR does not hard code runc names for the metrics code, this will get
the runtime form the docker in order to avoid failures related that runc
is not found.

Fixes #1796

### DIFF
--- a/metrics/lib/json.bash
+++ b/metrics/lib/json.bash
@@ -93,7 +93,8 @@ EOF
 		metrics_json_add_fragment "$json"
 	else
 		if [ "$RUNTIME" == "runc" ]; then
-			local output=$(docker-runc -v)
+			local get_binary=$(sudo docker info --format "{{.Runtimes.runc.Path}}")
+			local output=$("${get_binary}" -v)
 			local runcversion=$(grep version <<< "$output" | sed 's/runc version //')
 			local runccommit=$(grep commit <<< "$output" | sed 's/commit: //')
 			local json="$(cat << EOF


metrics/lib/json.bash
```diff
@@ -93,7 +93,8 @@ EOF
 		metrics_json_add_fragment "$json"
 	else
 		if [ "$RUNTIME" == "runc" ]; then
-			local output=$(docker-runc -v)
+			local get_binary=$(sudo docker info --format "{{json .Runtimes.runc}}" | cut -d '"' -f4)
```
devimc, line 96: try with `.Runtimes.runc.Path`, maybe `cut` is not needed

GabyCT: @devimc thanks, however, I think I need the cut to remove the ""

GabyCT: @devimc the output that I get is "docker-runc" or "runc", however, I still need to remove the "

devimc: then remove `json`

devimc: sudo docker info --format "{{.Runtimes.runc}}"

GabyCT: @devimc thanks it worked :)

grahamwhaley:
/test-ubuntu
Travis seems to be maybe stuck - we may have to 'hand-check' the status of that one.

I still get an error running this test, but I think that is a subtly different error, so does not have to prevent this merge (but we could consider a fix/merge of the two together in this one PR if you want to chase it down @GabyCT ).

At the end of the run I still get:

```
+ die 'Runc process not found'
+ msg='Runc process not found'
+ echo 'ERROR: Runc process not found'
```

I believe this is coming from the [code trying to find the footprint size](https://github.com/kata-containers/tests/blob/master/metrics/density/docker_memory_usage.sh#L91), which relies on an `smem` of [`docker-containerd-shim`](https://github.com/kata-containers/tests/blob/master/metrics/density/docker_memory_usage.sh#L59). I'm going to guess that the reliance on the process name of `docker-containerd-shim` is not out of date, and thus noisily failing.

GabyCT:
@grahamwhaley yes let me chase down the other issue, I will merge this PR and then raise another one for the other issue thanks

Issue #2544, opened 2020-05-14
add no systemd image installation helper


jongwu:
**Which feature do you think can be improved?**

no systemd image installer

**How can it be improved?**

using no systemd image will largely improve performance for startup time of kata

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

grahamwhaley:
Can you give us a few more details @jongwu - I presume you mean remove systemd from the guest VM rootfs image? I'm not sure if we already have a build that does that when we use the kata agent as the init process. /cc @devimc @jcvenegas for input and thoughts.

jongwu:
@grahamwhaley , yeah, as systemd consume lots of time when startup of kata and also systemd is not the essential part for kata, so if we remove systemd in kata image we can get lots of benefit. maybe it's time to think about it.
for now, alpine is the best choose for us to use as the no systemd image. I have tested it in Arm and get considerable performance improvement (bootup time) compared to ubuntu image. 
maybe we can try other distro like ubuntu which use systemd as its default init process, but I have not done that.

devimc:
@jongwu what about just adding `init=/usr/bin/kata-agent` to `kernel_params` ? I think it's the same

jongwu:
@devimc , I think there is a little different. take ubuntu as an example, if just set init=/usr/bin/kata-agent, it means that we must abandon all of the service offered by systemd and if we want to add those service in other ways, the image should be touched. so I think it will be better to offer a different image against those normal ones.

PR #2545, opened 2020-05-14
ci: Offer no systemd kata image installation helper

jongwu:
The startup of systemd service consumes lots of time when
kata boot up. so using kata image without systemd will largely
optimize bootup time of kata.
for now, only alpine support no systemd, so alpine is the only
choice for installation of no systemd kata image.
Also using alpine will decrease the memory footprint.

Fixes: #2544

@jodh-intel @devimc @bergwolf

### DIFF
--- a/.ci/install_kata_image_nosystemd.sh
+++ b/.ci/install_kata_image_nosystemd.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o errtrace
+
+cidir=$(dirname "$0")
+rust_agent_repo="github.com/kata-containers/kata-containers"
+osbuilder_repo="github.com/kata-containers/osbuilder"
+arch=$("${cidir}"/kata-arch.sh -d)
+
+build_image() {
+	go get -d "${osbuilder_repo}" || true
+	pushd "${GOPATH}/src/${osbuilder_repo}/rootfs-builder"
+	export ROOTFS_DIR="${GOPATH}/src/${osbuilder_repo}/rootfs-builder/rootfs"
+
+	#Currently, only alpine support no systemd"
+	distro="alpine"
+	sudo -E GOPATH="${GOPATH}" USE_DOCKER=true SECCOMP=no ./rootfs.sh "${distro}"
+	install -o root -g root -m 0550 -T ../../agent/kata-agent ${ROOTFS_DIR}/sbin/init
+	popd
+
+	pushd "${GOPATH}/src/${osbuilder_repo}"
+	sudo -E USE_DOCKER=1 DISTRO="${distro}" AGENT_INIT=yes ./image-builder/image_builder.sh ${ROOTFS_DIR}
+
+	image_path="/usr/share/kata-containers/"
+	sudo mkdir -p "${image_path}"
+	sudo install -D "${GOPATH}/src/${osbuilder_repo}/kata-containers.img" "${image_path}"
+	popd
+}
+
+main() {
+	build_image
+}
+
+main


.ci/install_kata_image_nosystemd.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
```
grahamwhaley, line 3: You may need to edit/update the copyright line

jongwu: you mean only upgrade 2019 to 2020? @grahamwhaley

grahamwhaley: yes, at minimum update the date.
Also, you can add your own company/name if you have made substantial edits - if you want to :-)

jongwu: thanks @grahamwhaley

GabyCT:
@jongwu thanks, just one question is this script going to be executed? or who is going to trigger this?

jongwu:
@GabyCT I raise this PR as I want to offer another choose for kata user. I think systemd is a little heave for kata, sometime not necessary for user (but I can't tell which group). one hundred to a few hundred millisecond will be wasted in every startup of kata.  if kata can work fine without systemd, it's better to hold it. But I'm not sure we are prepared to substitute systemd thoroughly, so maybe we can let this helper as an experimental one.

jongwu:
hello @grahamwhaley @GabyCT @devimc any comments?

.ci/install_kata_image_nosystemd.sh
```diff
@@ -0,0 +1,42 @@
...
+
+build_image() {
+	go get -d "${osbuilder_repo}" || true
+	pushd "${GOPATH}/src/${osbuilder_repo}/rootfs-builder"
+	export ROOTFS_DIR="${GOPATH}/src/${osbuilder_repo}/rootfs-builder/rootfs"
+
+	#Currently, only alpine support no systemd"
+	distro="alpine"
+	sudo -E GOPATH="${GOPATH}" USE_DOCKER=true SECCOMP=no ./rootfs.sh "${distro}"
+	install -o root -g root -m 0550 -T ../../agent/kata-agent ${ROOTFS_DIR}/sbin/init
```
devimc, line 26: not sure but I think this won't work, alpine is built around musl libc, did `./rootfs.sh` build this binary using musl?

jongwu:
hello @amshinde , I'm not sure that the error in jenkins ci is related with this PR or a random one, can you help trigger it again or do you have some idea about it?

chavafg:
@jongwu can you re-push here to trigger the DCO and WIP checks?

jongwu:
thanks @chavafg

jodh-intel:
@jongwu - can you answer the outstanding question on this PR please?

.ci/install_kata_image_nosystemd.sh
jongwu: sorry to have missed this comments so long.
I have tried this script both on x86 and arm64 and it works well. when I inspect the kata-agent binary using file, it shows:
kata-agent: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, not stripped
Does that mean this kata-agent build using musl?

jongwu:
thanks @jodh-intel

jodh-intel (approved):
Thanks @jongwu

.ci/install_kata_image_nosystemd.sh
devimc: thanks for confirming, I  guess I was wrong

pmores:
Hi, I'm not sure what the status of the PR is (it seems to have enough approvals but doesn't seem to be merged) so please ignore this if not relevant but I think existing tooling (`kata_install_image.sh` and the osbuilder machinery it relies on) should be able to handle this so I'm not sure if we want/need to have a special-purpose script for this.

If I'm wrong and `kata_install_image.sh` cannot currently handle this wouldn't it be better to try to fix it first?  I mean I can understand the advantages of a bunch of simple special-purpose scripts over a unified but possibly harder-to-understand build pipeline.  I think both can be valid approaches but I'm kind of dubious about mixing them.  Seeing as the unified pipeline seems to be the approach currently taken I'd probably stick with it, at least until we decide to rearchitect and refactor the whole thing...

jongwu:
Yeah @pmores -, fair enough, I will close this PR.

Issue #2549, opened 2020-05-15
metrics: report: grab: only lookup versions file when needed


grahamwhaley:
**Which feature do you think can be improved?**

The [metrics report grabdata](https://github.com/kata-containers/tests/blob/master/metrics/report/grabdata.sh) extracts the version for the Elastic image right at the start of the file, even though it is only used when doing the density tests.

```
# Get working elasticsearch version
VERSIONS_FILE="${SCRIPT_DIR}/../../versions.yaml"
ELASTICSEARCH_VERSION=$("${GOPATH}/bin/yq" read "$VERSIONS_FILE" "docker_images.elasticsearch.version")
ELASTICSEARCH_IMAGE="elasticsearch:$ELASTICSEARCH_VERSION"
```

This was causing an issue where a user had a different version of `yq` installed, so it grok'd, even though they were not running the density tests.

Move the version pull code down into the density function just before its use, so if we are going to fail, we only fail when necessary, and late.

PR #2550, opened 2020-05-15
metrics: report: evaluate elastic version later

grahamwhaley:
Evaluate the version of Elastic we use for the density tests later.
Do it in the function that uses it, not at initialisation, as we do
not always perform the density tests, and we'd rather only fail
when necessary.

Fixes: #2549


grahamwhaley:
@eadamsintel - this is the one you bumped into earlier this week - this is a minor improvement (it still won't catch the wrong version of `yq` installed), but at least now it won't fail on the density code if you are not running the density check.
@GabyCT - fyi, just a small tweak.

PR #2552, opened 2020-05-15
ci: Copy config file instead of moving

### LINKED ISSUE #2551: ci: original config file not available after install_runtime.sh
In [install_runtime.sh](https://github.com/kata-containers/tests/blob/master/.ci/install_runtime.sh#L57), function `enable_hypervisor_config` moves a configuration file instead of copying, removing the original file. The [vfio ci test](https://github.com/kata-containers/tests/blob/master/functional/vfio/run.sh) should use the original `configuration-qemu.toml` file, but instead falls back to `configuration.toml` because the original is not available.

cmaf:
Copy config file instead of moving so that the original is still
available.

Fixes #2551

Signed-off-by: Chelsea Mafrica <chelsea.e.mafrica@intel.com>

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -54,7 +54,7 @@ fi
 
 enable_hypervisor_config(){
 	local path=$1
-	sudo mv "$path" "${PKGDEFAULTSDIR}/configuration.toml"
+	sudo ln -f "$path" "${PKGDEFAULTSDIR}/configuration.toml"
 
 }
 
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -194,6 +194,9 @@ if [ -n "${METRICS_CI}" ]; then
 	echo "Running the metrics tests:"
 	"${tests_repo_dir}/.ci/run_metrics_PR_ci.sh"
 elif [ -n "${VFIO_CI}" ]; then
+	echo "Installing initrd image:"
+	TEST_INITRD=yes "${ci_dir_name}/install_kata_image.sh"
+
 	echo "Running VFIO tests:"
 	"${ci_dir_name}/run.sh"
 else
--- a/Makefile
+++ b/Makefile
@@ -301,10 +301,14 @@ vcpus:
 	bash -f integration/vcpus/default_vcpus_test.sh
 
 vfio:
-	bash -f functional/vfio/run.sh -s false -p qemu -m pc
-	bash -f functional/vfio/run.sh -s true -p qemu -m pc
-	bash -f functional/vfio/run.sh -s false -p qemu -m q35
-	bash -f functional/vfio/run.sh -s true -p qemu -m q35
+	bash -f functional/vfio/run.sh -s false -p qemu -m pc -i image
+	bash -f functional/vfio/run.sh -s true -p qemu -m pc -i image
+	bash -f functional/vfio/run.sh -s false -p qemu -m q35 -i image
+	bash -f functional/vfio/run.sh -s true -p qemu -m q35 -i image
+	bash -f functional/vfio/run.sh -s false -p qemu -m pc -i initrd
+	bash -f functional/vfio/run.sh -s true -p qemu -m pc -i initrd
+	bash -f functional/vfio/run.sh -s false -p qemu -m q35 -i initrd
+	bash -f functional/vfio/run.sh -s true -p qemu -m q35 -i initrd
 
 ipv6:
 	bash -f integration/ipv6/ipv6.sh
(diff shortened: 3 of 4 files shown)


.ci/install_runtime.sh
```diff
@@ -54,7 +54,7 @@ fi
 
 enable_hypervisor_config(){
 	local path=$1
-	sudo mv "$path" "${PKGDEFAULTSDIR}/configuration.toml"
+	sudo cp "$path" "${PKGDEFAULTSDIR}/configuration.toml"
```
devimc, line 57: `cp -a`  :wink:

cmaf: @devimc updated

devimc:
@cmaf - I feel (peter-tingle) vfio CI will fail
/test-vfio

cmaf:
@devimc it should fail, the vfio test doesn't check that only initrd or image is specified in the config yet

cmaf:
@devimc yes, I'll let you know when it is updated

.ci/install_runtime.sh
```diff
@@ -54,7 +54,7 @@ fi
 
 enable_hypervisor_config(){
 	local path=$1
-	sudo mv "$path" "${PKGDEFAULTSDIR}/configuration.toml"
+	sudo ln -f "$path" "${PKGDEFAULTSDIR}/configuration.toml"
```
amshinde, line 57: You mentioned in the commit copy, but seeing a link here. Do you want that? Any changes to one, will be reflected on the other.

Makefile
```diff
@@ -301,10 +301,14 @@ vcpus:
 	bash -f integration/vcpus/default_vcpus_test.sh
 
 vfio:
-	bash -f functional/vfio/run.sh -s false -p qemu -m pc
-	bash -f functional/vfio/run.sh -s true -p qemu -m pc
-	bash -f functional/vfio/run.sh -s false -p qemu -m q35
-	bash -f functional/vfio/run.sh -s true -p qemu -m q35
+	bash -f functional/vfio/run.sh -s false -p qemu -m pc -i image
```
amshinde, line 304: What is the true/false flag here, trying to see what combinations we are testing here.
Stepping back, do we want to test so many combinations?

devimc: sadly, yes, there are not enough resources to have a VFIO CI per feature (sandbox_cgroup_only, q35 or pc, image or initrd)

.ci/install_runtime.sh
cmaf: @amshinde thank you, I updated the commit to match the change

Makefile
cmaf: @devimc so we want to keep these 8 tests for now?

devimc: yes

grahamwhaley:
/test-ubuntu
@amshinde please re-review

.ci/jenkins_job_build.sh
```diff
@@ -194,6 +194,9 @@ if [ -n "${METRICS_CI}" ]; then
 	echo "Running the metrics tests:"
 	"${tests_repo_dir}/.ci/run_metrics_PR_ci.sh"
 elif [ -n "${VFIO_CI}" ]; then
+	echo "Installing initrd image:"
+	TEST_INITRD=yes "${ci_dir_name}/install_kata_image.sh"
```
wainersm, line 198: Since you passing the `TEST_INITRD=yes` to the installation script, perhaps you should export it to the environment because other scripts eventually called later can need that information. So I suggest this:

```
export TEST_INITRD=yes
"${ci_dir_name}/install_kata_image.sh"
```

cmaf: Thanks @wainersm, (is this the dismissed review? I can't tell)

I don't think other scripts use this variable right now, and we don't necessarily want all of them to have `TEST_INITRD` set to `yes`. However, I made these edits on @devimc's suggestion, so perhaps he can comment.

wainersm: > Thanks @wainersm, (is this the dismissed review? I can't tell)

Yes, I wanted to make this single comment but I pressed the wrong button :)
> 
> I don't think other scripts use this variable right now, and we don't necessarily want all of them to have `TEST_INITRD` set to `yes`. However, I made these edits on @devimc's suggestion, so perhaps he can comment.

Ok, I don't know those scripts in deep and this is the only thing that caught on my eyes.

devimc: @cmaf  you're right, however I don't see any inconvenience if you apply the changes suggested by @wainersm

cmaf: @devimc @wainersm I'll leave this as it is for now, thanks again for the feedback.

PR #2557, opened 2020-05-18
metrics: report: clean up report generation

grahamwhaley:
Fix up two issues with the report generator:
 - it was not very 'pleasant' at handling missing data, giving out errors and warnings. Make it smarter, and have it report 'no data found' for such cases.
- It looks like over at https://github.com/kata-containers/runtime/commit/a4b3c65c16f63f1690966f68426a703cd3af7ab5 the kata-env info changed 'shape', which gets output in the JSON data that the report generator parses. It was broken. Fix it up to handle the new Semver fields. (/cc @Pennyzct just fyi, as I think you added that Semver change - sorry if I didn't spot and make the link...)

### DIFF
--- a/metrics/report/report_dockerfile/dut-details.R
+++ b/metrics/report/report_dockerfile/dut-details.R
@@ -13,123 +13,130 @@ library(gridExtra)					# together.
 suppressMessages(suppressWarnings(library(ggpubr)))	# for ggtexttable.
 suppressMessages(library(jsonlite))			# to load the data.
 
-# A list of all the known results files we might find the information inside.
-resultsfiles=c(
-	"boot-times.json",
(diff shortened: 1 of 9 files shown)


grahamwhaley:
Hmm, seems I picked up 3 commits there when I only intended 2 ... let me have a look how/why that elastic one made it there - likely some local tree malarkey....

grahamwhaley:
OK, fixed the errant extra commit.
Also found and fixed up a few more rendering issues when we had partial or incomplete data. Tested with:
- one empty subdir (no results)
- one subdir with full results
- one subdir with full results and one empty subdir
- two subdirs with full results and one empty subdir

/test-ubuntu

Issue #2562, opened 2020-05-21
metrics: time: `date` now pre-pads ns with 0's


grahamwhaley:
# Description of problem

We specifically extract a ns resolution timestamp using `date` for the launch_times test, and use a formatter to request no pre-padding of the values.
Changes [here](https://lists.gnu.org/archive/html/bug-gnulib/2020-05/msg00208.html) and [here](https://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=commitdiff;h=6d0107a3) have changed the ns field padding, which has re-introduced the prepended 0's. They make the bash printf barf, as it sees them as octal numbers!

Fix by explicitly stripping off any leading 0's on the ns field.

PR #2563, opened 2020-05-21
metrics: time: explicitly strip leading 0's from date

grahamwhaley:
The `date` command `-N` format changed in version 8.32, and now
pre-pends 0's to pad the field. Explicitly strip those 0's, otherwise
it causes the later json prints to barf, as it sees them as octal.
This is also backwards compat with older date versions (if the 0's
are not there, the strip is benign).

Fixes: #2562


grahamwhaley:
As noticed over at https://github.com/kata-containers/runtime/issues/2692

Issue #2564, opened 2020-05-21
lib: is_a_kata_runtime should spot more kata runtimes


grahamwhaley:
**Which feature do you think can be improved?**

The common lib code has a utility function that [tries to identify when we are using a kata runtime](https://github.com/kata-containers/tests/blob/master/lib/common.bash#L55-L70) - but, it is a little hard wired. We can make it a touch more flexible. For instance, it currently does not recognise `kata-qemu-virtiofs`. We can probably just spot `kata*`, which feels will cover most cases.

```bash
is_a_kata_runtime(){
	case "$1" in
	"kata-runtime") ;&	# fallthrough
	"kata-qemu") ;&		# fallthrough
	"kata-fc")
		echo "1"
		return
		;;
```

PR #2565, opened 2020-05-21
common: match more kata runtimes in generic check func

grahamwhaley:
We have a generic 'is_a_kata_runtime' function that was a bit
hard wired to specific names. We have a growing collection of kata
runtime names now, so let's get a little more flexible in our spotting
by just going for any runtime named 'kata*'.
Specifically, this now matches kata-qemu-virtiofs for instance.

Fixes: #2564


grahamwhaley:
@eadamsintel @dagrh - this explains why we were not getting the kernel boot time breakdown for `kata-qemu-virtiofs`... fixed!

PR #2575, opened 2020-05-27
 k8s: e2e: Move focus list to yaml

jcvenegas:
For PRs the CI runs a subset of test to make sure e2e setup works.
Then it has a job that run once a day with full list of e2e tests.

Add list to a yaml like other test that are skipped.

Fixes: #2582

### DIFF
--- a/integration/kubernetes/e2e_conformance/run.sh
+++ b/integration/kubernetes/e2e_conformance/run.sh
@@ -28,6 +28,8 @@ KATA_HYPERVISOR="${KATA_HYPERVISOR:-}"
 # Overall Sonobuoy timeout in minutes.
 WAIT_TIME=${WAIT_TIME:-180}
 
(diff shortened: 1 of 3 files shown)


integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -52,42 +54,50 @@ get_sonobuoy() {
 
 }
 
+# Input
+#yaml list key
```
chavafg, line 58: missing space to have same alignment as other lines?

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -52,42 +54,50 @@ get_sonobuoy() {
...
+# yaml file
+# Output
+#string: | separated values of the list
```
chavafg, line 61: same as above

PR #2598, opened 2020-06-02
metrics: Support different hypervisors for footprint test

### LINKED ISSUE #2597: Modify footprint test to support different hypervisors

GabyCT:
This PR retrieves the correct hypervisor name instead of having qemu name
in all the cases. This will fix that if we are running with cloud hypervisor
or firecracker, we will see that in the results json and not qemu.

Fixes #2597

### DIFF
--- a/metrics/density/footprint_data.sh
+++ b/metrics/density/footprint_data.sh
@@ -11,7 +11,7 @@
 # - when do we terminate the test (cutoff points)
 #
 # There are a number of things we may wish to add to this script later:
-# - sanity check that the correct number of runtime components (qemu, shim etc.)
+# - sanity check that the correct number of runtime components (hypervisor, shim etc.)
 #  are running at all times
 # - some post-processing scripts to generate stats and graphs
 #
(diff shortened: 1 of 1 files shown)


metrics/density/footprint_data.sh
```diff
@@ -167,16 +167,16 @@ function get_proc_pss() {
...
 
-	total=$((proxy + shim + qemu + virtiofsd))
+	total=$((proxy + shim + hypervisor + virtiofsd))
 
 	local json="$(cat << EOF
 		"uss": {
 			"proxy": $proxy,
 			"shim": $shim,
-			"qemu": "$qemu",
+			"$KATA_HYPERVISOR": "$hypervisor",
```
wainersm, line 179: It uses `KATA_HYPERVISOR` variable on which a default value isn't declared in this script nor it seems to be in the sourced `common.bash`. I may be missing something (first time looking at this script), but wouldn't it need to define a default somewhere to be safe?

GabyCT: @wainersm it is declared here https://github.com/kata-containers/tests/blob/4fede8941f5ad6b820652eef1227681044429cbf/lib/common.bash#L28 and https://github.com/kata-containers/tests/blob/4fede8941f5ad6b820652eef1227681044429cbf/lib/common.bash#L145

GabyCT: you can see that the variable can have multiple definitions

GabyCT: if you are running a different hypervisor like `firecracker` or `clh` you need to setup the `KATA_HYPERVISOR` variable so it can perform the setup and install the correspondent hypervisor

wainersm: @GabyCT ah, ok, I realized I look at `metrics/lib/common.bash` only.

OK, this PR looks good to me.

Issue #2546, opened 2020-05-14
clh: Enable crio testing


jcvenegas:
**Which feature do you think can be improved?**

Enable crio testing for cloud hypervisor. Now that block [device is enabled ](https://github.com/kata-containers/runtime/pull/2681) we can use crio testing to exercise that code path

**How can it be improved?**
- [ ] Add a new job for crio and k8s
 

 \cc @likebreath

likebreath:
We decide to focus on enabling the cri-o in kata 2.0, where we will resume the work when the k8s+crio CI for qemu is stabilized in 2.0. Changing it medium-priority.

PR #2621, opened 2020-06-08
[DNM] clh: Add a ci job for crio tests

likebreath:
This patch extends the existing scripts to cover a new Jenkins job on
testing k8s with crio for cloud-hypervisor.

Fixes: #2546

Depends-on: github.com/kata-containers/runtime#2833

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -253,6 +253,14 @@ case "${CI_JOB}" in
 	export TEST_DOCKER="true"
 	export experimental_kernel="true"
 	;;
+"CLOUD-HYPERVISOR-K8S-CONTAINERD")
+	init_ci_flags
+	export CRI_CONTAINERD="yes"
+	export CRI_RUNTIME="containerd"
+	export KATA_HYPERVISOR="cloud-hypervisor"
+	export KUBERNETES="yes"
+	export experimental_kernel="true"
+	;;
 "CLOUD-HYPERVISOR-K8S-E2E-CRIO-MINIMAL")
 	init_ci_flags
 	export CRIO="yes"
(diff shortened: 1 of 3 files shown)


likebreath:
test-clh-crio

jcvenegas:
I see it is passing :D

likebreath:
By enabling the block device unplug (https://github.com/kata-containers/runtime/pull/2833), the cri-o integration tests are now passing quite many tests (at least the first 22 tests as shown [here](http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-crio/10/console)). 

The failing test is `test 23: privileged ctr device add`, which is skipped for firecracker w/ kata for now. Also, I re-ran the CI few times and did see few unstable tests, especially `test 18 ctr list label filtering` (`test 5 additional devices support` failed once). I will take a closer look into those failings later.

@jcvenegas @sboeuf @chavafg

sboeuf:
@likebreath 
Nice! I see you made some interesting progress on this issue. Basically, the CRI-O tests don't really match the standard usage of Cloud-Hypervisor with Kata Containers since most of the time virtio-fs is involved for adding and removing a container's rootfs.
So maybe the first question is to ask ourselves if we want to keep the CRI-O test using `devmapper` as the storage driver, or if instead, and specifically to Cloud-Hypervisor, we want to switch to overlay FS so that virtio-fs is being used?
That being said, I think this CRI-O testing using devmapper as the storage driver is interesting as it triggers some bugs that couldn't be caught through other tests, so I'd be in favor of keeping it the way it is.
/cc sameo

c3d:
Adding dnm label (as opposed to in-title)

jodh-intel:
@likebreath - is this PR still needed? If so, please can you rebase as it's conflicted.

Issue #2650, opened 2020-06-17
build_version in .ci/lib.sh fails for repeated builds


dgibson:
# Description of problem

On a clean system, in the runtime module
    # make
    # make check
    # rm $GOPATH/bin/golangci-lint
    # make check

# Expected result

Both make check runs complete and pass.

# Actual result

Second make check run fails:
```
INFO: Installing golangci-lint
INFO: Forcing golangci-lint version v1.18.0
~/go/src/github.com/golangci/golangci-lint ~/go/src/github.com/kata-containers/runtime
INFO: Using github.com/golangci/golangci-lint version v1.18.0
warning: refname 'v1.18.0' is ambiguous.
fatal: A branch named 'v1.18.0' already exists.
make: *** [Makefile:681: check-go-static] Error 128
```

# Further information

This problem occurs because the build_version shell function in the tests module .ci/lib.sh creates a new branch to check out the dependency (golangci-lint in this case) to the version we want.  When we attempt to build again, the checkout fails because the branch already exists (and it also makes the ref we're using ambiguous because there's now both a tag and a branch with the same name).

Removing the installed golanci-lint binary is necessary to reproduce, or we don't attempt to rebuild the tool at all.  The problem can also be triggered if there are unrelated failures building golangci-lint.

A similar problem can also occur with the mvdan.cc/xurls dependency.

PR #2651, opened 2020-06-17
Don't create branches when building dependencies at specific versions

dgibson:
For some dependencies we want to build a specific version, which is handled
by the build_version function in .ci/lib.sh.  It does this by creating a
new branch at the specific revision we want, then building.  However, this
new branch is left over, which means the check out will fail if we attempt
it again without manually removing the branch.

In fact, there's no real point to creating a branch: git allows us to just
checkout the commit we want directly.  That will leave the tree in
"detached HEAD" mode, which is awkward if you're developing in the tree,
but is just fine for building a dependency as we are here.

Fixes: #2650

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -80,7 +80,7 @@ function build_version() {
 
 	if [ "$version" != "HEAD" ]; then
 		info "Using ${github_project} version ${version}"
-		git checkout -b "${version}" "${version}"
+		git checkout "${version}"
 	fi
 
 	info "Building ${github_project}"


.ci/lib.sh
```diff
@@ -80,7 +80,7 @@ function build_version() {
 
 	if [ "$version" != "HEAD" ]; then
 		info "Using ${github_project} version ${version}"
-		git checkout -b "${version}" "${version}"
```
c3d, line 83: LGTM.

What about making the branche name something transient, e.g. with a date?  `test-$(date +%Y-%m-%d-%H:%M:%S)-${version}`?

dgibson: We could, but why bother?  I don't actually see any real benefit to creating the branch, and it pollutes the other tree with a bunch of stale branch tags.

c3d: Sorry, I was not awake enough, read the change backwards.

I assume `git checkout` should always succeed, possibly creating a local branch if you have some origin version here.

dgibson: I don't quite follow.  git checkout will only create a branch with -b, but it will still succeed without creating a branch (putting the tree in detached HEAD state).

c3d: If you give a sha1, then yes, `git checkout` will always succeed. If you pass a branch, it needs the branch to exist somewhere it can be tracked from. But I guess that is always the case here.

dgibson: Right, yes.  Of the examples I've seen, one uses a sha, one uses a tag.  A existing branch name would not previously have worked, since it would have tried to create a new branch with the same name as the existing one.  So, as a bonus the change should allow for versions to be set to existing branches.  That's probably not a wise choice, though, since the point here is to get a specific version to build.

jodh-intel: It's worth pointing out that this code isn't only run by the CI - if a developer runs https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh, that can call `build_version()` so I wonder if it would less surprising to those users if we changed this to use `-B` (aka "switch to branch and reset, but don't fail if branch already exists"):

```
git checkout -B "${version}" "${version}"
```

/cc @jcvenegas, @chavafg, @amshinde.

dgibson: Least surprising is a bit hard to judge.  But simply moving the other tree to a detached HEAD state seems less dangerous to me than adding or resetting a branch there.

That said, either one addresses the initial problem here.

jodh-intel: That's a good point - let's go with this.

PR #2653, opened 2020-06-17
ci: metrics: Move metrics CI config out out jenkins web inteface

jcvenegas:
Move what is used and installed out of jenkins web interface.
Now we can control what a metrics job for CLH install and
future changes can be done here not the job web interface.

If we want to modify jenkins_job_build.sh
in a PR may not be possible as it was the entry point.
Also, in CI jobs some variables that change what a job
does is hidden in the job.

This script should be minimal to avoid change it in the future.
So if jenkins_job_build.sh changes the changes will be taken
no matter the branch.

The script also tries to simply a Jenkins job internal script.


.ci/jenkins_job_build.sh
```diff
@@ -264,7 +293,7 @@ else
 	echo "WARN: Kata runtime is not installed"
 fi
 
-if [ -n "${METRICS_CI}" ] || [ -n "${METRICS_CI_CLH_BAREMETAL}" ]; then
+if [ -n "${METRICS_CI}" ]; then
```
GabyCT, line 296: @jcvenegas so we will rename the jobs for metrics to have export METRICS_CI_PROFILE="clh-baremetal"? the ones that run with CLH

jcvenegas: @GabyCT  we should export the CI_JOB to CLOUD-HYPERVISOR-METRICS-BAREMETAL , the profile is already set on this job id, https://github.com/kata-containers/tests/pull/2653/files#diff-25960c2e9b74d48d51e5a93d8df4455aR282

.ci/ci_entry_point.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Use workspace as gopath
+export WORKSPACE=${WORKSPACE:-$HOME}
+export GOPATH="${WORKSPACE}/go"
+
+# Repository where we store all tests
+test_repo="github.com/kata-containers/tests"
+
+# Clone as golang would do it with GOPATH
+test_repo_dir="${GOPATH}/src/${test_repo}"
```
chavafg, line 45: same here: `tests_repo_dir`?

.ci/ci_entry_point.sh
```diff
@@ -0,0 +1,66 @@
...
+mkdir -p "${test_repo_dir}"
+git clone "https://${test_repo}.git" "${test_repo_dir}"
+cd "${GOPATH}/src/${test_repo}"
```
chavafg, line 48: `cd $tests_repo_dir`?

.ci/jenkins_job_build.sh
```diff
@@ -7,8 +7,18 @@
 
 set -e
 
+export GOPATH=${WORKSPACE}/go
+# For all our jobs  will install golang from static binaries
```
chavafg, line 11: propose to rephrase: `# For all our jobs, golang will be installed using static binaries`

.ci/jenkins_job_build.sh
```diff
@@ -7,8 +7,18 @@
...
+export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:${PATH}
+export GOROOT="/usr/local/go"
+
+
 # List of all setup flags used by scripts
 init_ci_flags() {
+	# Make jobs work like in CI
+	# CI disable non-working tests
```
chavafg, line 19: nit: `#CI disables non...`

Issue #2673, opened 2020-06-24
creating test code base by merging instead of rebasing


bergwolf:
Right now we always rebase the testing branch on top of the targeting branch. It is not necessary because we only require the final code do not conflict with the targeting branch, but a rebase would require each commit to not conflict.

bergwolf:
closed by #2674

PR #2674, opened 2020-06-24
ci: resolve resulting code base by merging

bergwolf:
If we do rebase, it is required that every commit cannot conflict but a
merge only requires that the final code do not conflict.

Fixes: #2673

### DIFF
--- a/.ci/resolve-kata-dependencies.sh
+++ b/.ci/resolve-kata-dependencies.sh
@@ -56,8 +56,8 @@ apply_depends_on() {
 		dependency_branch="p${pr_id}"
 		git fetch origin "pull/${pr_id}/head:${dependency_branch}" && \
 			git checkout "${dependency_branch}" && \
-			git rebase "origin/${branch}"
-			# And show what we rebased on top of to aid debugging
+			git merge "origin/${branch}"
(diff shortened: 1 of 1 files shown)


.ci/resolve-kata-dependencies.sh
```diff
@@ -56,8 +56,8 @@ apply_depends_on() {
 		dependency_branch="p${pr_id}"
 		git fetch origin "pull/${pr_id}/head:${dependency_branch}" && \
 			git checkout "${dependency_branch}" && \
-			git rebase "origin/${branch}"
-			# And show what we rebased on top of to aid debugging
+			git merge "origin/${branch}"
```
jodh-intel, line 59: Could we make this `--ff-only`? Would that work for https://github.com/kata-containers/kata-containers/pull/330 ?

bergwolf: We don't need to require --ff-only either. Just simulate a github merge action is enough. And yes, I think this should work for kata-containers/kata-containers#330. waiting to verify.

bergwolf:
It makes sense to use `merge` instead of `rebase` so there is no dependency on kata-containers/kata-containers#330. This is ready to be reviewed/merged.

PR #2696, opened 2020-06-30
test: Check cgroups with sandbox

### LINKED ISSUE #2695: check cgroups with sandbox

GabyCT:
This test will check that certain cgroups are not present while running
a container when we have sandbox enabled.

Fixes #2695

### DIFF
--- a/integration/sandbox_cgroup/check_cgroups_sandbox.sh
+++ b/integration/sandbox_cgroup/check_cgroups_sandbox.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This will enable the sandbox_cgroup_only
+# to true and verify the cgroups
+
+dir_path=$(dirname "$0")
+source "${dir_path}/../../lib/common.bash"
+source "${dir_path}/../../.ci/lib.sh"
+tests_repo="${tests_repo:-github.com/kata-containers/tests}"
+TEST_SANDBOX_CGROUP_ONLY="${TEST_SANDBOX_CGROUP_ONLY:-}"
+image="centos"
+container_name="test_cgroups"
+payload="tail -f /dev/null"
+docker_cgroup_prefix="/sys/fs/cgroup/devices/docker"
+
+if [ -z "${TEST_SANDBOX_CGROUP_ONLY}" ]; then
+	info "Skip: TEST_SANDBOX_CGROUP_ONLY variable is not set"
+	exit 0
+fi
+
+if ! command -v docker >/dev/null; then
+	info "This test needs docker to run"
+	exit 1
+fi
+
+function setup() {
+	clean_env
+	check_processes
+}
+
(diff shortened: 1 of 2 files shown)


integration/sandbox_cgroup/check_cgroups_sandbox.sh
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function check_cgroups_sandbox() {
+	pushd "${GOPATH}/src/${tests_repo}
+	".ci/toggle_sandbox_cgroup_only.sh" true
+	docker run -d --name "${container_name}" --runtime "${RUNTIME}" "${image}" sh -c "${payload}"
+	get_id=$(docker inspect --format='{{.Id}}' "${container_name}")
+	# Verify first cgroup
+	first_cgroup="a *.* rwm"
+	check_first_cgroup=$(grep -Fx "${first_cgroup}" /sys/fs/cgroup/devices/docker/kata_"${get_id}"/devices.list)
```
jodh-intel, line 36: This looks like a docker-specific test? If so, could you:

- make that clear in the commit.
- maybe renaming the test to include "docker"?
- add a comment at the top of the file explaining that the caller should only run this test if docker is available.
- maybe check that docker is available on the system and just `exit 0` (or `exit 1`?) if docker not available.
- consider creating a global variable like this:

  ```sh
  docker_cgroup_prefix="/sys/fs/cgroup/devices/docker"
  ```

  ... and then referring to that in the test functions.

GabyCT:
@jodh-intel changes applied

integration/sandbox_cgroup/check_cgroups_sandbox.sh
```diff
@@ -0,0 +1,80 @@
...
+payload="tail -f /dev/null"
+docker_cgroup_prefix="/sys/fs/cgroup/devices/docker"
+
+if [ -z "${TEST_SANDBOX_CGROUP_ONLY}" ]; then
+	info "Skip: TEST_SANDBOX_CGROUP_ONLY variable is not set"
+	exit 0
+fi
+
+if ! command -v docker >/dev/null; then
+	info "This tests needs docker to run"
```
chavafg, line 26: typo: this test?

integration/sandbox_cgroup/check_cgroups_sandbox.sh
```diff
@@ -0,0 +1,80 @@
...
+
+	# Verify third cgroup
+	third_cgroup="b *:* rwm"
+	check_third_cgroup=$(grep -Fx "${third_cgroup}" "${docker_cgroup_prefix}"/kata_"${get_id}"/devices.list)
+	if [ ! -z "${check_third_cgroup}" ]; then
+		echo "${third_cgroup} should not be there"
+		exit 1
+	fi
+
+	".ci/toggle_sandbox_cgroup_only.sh" false
```
chavafg, line 64: I think this should be part of the teardown as if we exit earlier from this function, this will not be executed.

GabyCT:
merging this as the ARM CI is failing with something unrelated with this PR

PR #2703, opened 2020-07-02
tests: Fixes for k8s tests on kata 2.0

### LINKED ISSUE #2702: Check pod no information is left after running each k8s tests
We have random failures in the k8s tests where it is saying that pod information is left, however, we do not know which test is leaving that so maybe we should consider check the pod information after each k8s test is left.

GabyCT:
We should check that pod no information is left after running
each k8s test for kata 2.0. This PR also gives time to check
that the pods are running as we have random failures in the pod
creation and also skips tests where pod information is left
at virtcontainers directory.

Fixes #2702

### DIFF
--- a/integration/kubernetes/k8s-attach-handlers.bats
+++ b/integration/kubernetes/k8s-attach-handlers.bats
@@ -20,6 +20,9 @@ setup() {
 }
 
 @test "Running with postStart and preStop handlers" {
+	wait_time=20
+	sleep_time=2
+
 	# Create yaml
 	sed -e "s/\${nginx_version}/${nginx_image}/" \
 		"${pod_config_dir}/lifecycle-events.yaml" > "${pod_config_dir}/test-lifecycle-events.yaml"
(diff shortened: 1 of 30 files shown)


GabyCT:
5 tests have been included with skip due to leaving pod information while running

integration/kubernetes/k8s-attach-handlers.bats
```diff
@@ -28,7 +31,8 @@ setup() {
 	kubectl create -f "${pod_config_dir}/test-lifecycle-events.yaml"
 
 	# Check pod creation
-	kubectl wait --for=condition=Ready pod "$pod_name"
+	cmd="kubectl wait --for=condition=Ready pod $pod_name"
```
liubin, line 34: `kubectl wait` has a `timeout` argument with default value `30s`,  so how about to only add a `timeout` to `kubectl` command ?

```
      --timeout=30s: The length of time to wait before giving up.  Zero means check once and don't
wait, negative means wait for a week.
```

Issue #2765, opened 2020-08-05
Add new test case for `kubectl exec`


liubin:
`kubectl exec` has many options, should add test to ensure that all use cases are working.

PR #2766, opened 2020-08-05
ci: add test cases for kubectl exec command

liubin:
Add cases for test `kubectl exec`'s options, input, and return values.

Fixes: #2765

### DIFF
--- a/integration/kubernetes/runtimeclass_workloads/busybox-pod.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/busybox-pod.yaml
@@ -14,11 +14,17 @@ spec:
   containers:
   - name: first-test-container
     image: busybox
+    env:
+    - name: CONTAINER_NAME
+      value: "first-test-container"
     command:
         - sleep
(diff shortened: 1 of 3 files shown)


integration/kubernetes/runtimeclass_workloads/busybox-pod.yaml
```diff
@@ -14,13 +14,19 @@ spec:
   containers:
   - name: first-test-container
     image: busybox
+    env:
+    - name: CONTAINER_NAME
+      value: "first-test-container"
     command:
         - sleep
-        - "30"
+        - "60"
```
jodh-intel, line 22: Why do we need these changes?

liubin: I'm not sure, but be afraid of if the container will exit during the test are running. In that case kubectl exec maybe get "unable to upgrade connection: container not found" error

teawater (approved):

Issue #2783, opened 2020-08-13
Kubernetes case "initContainer with shared volume" maybe failed


liubin:
In case "initContainer with shared volume" (`k8s-shared-volume.bats`), current timestamp is used to check if initContainer is running before normal container.

But the timestamp's precision is second, so it will occur that the initContainer and normal container may be started at the same second.

PR #2785, opened 2020-08-13
ci: add sleep in init container

liubin:
In some case the `date +%s` may get the same value
in initContainer and normal container, use $EPOCHREALTIME
can ensure that the values are different.

Fixes: #2783

### DIFF
--- a/integration/kubernetes/runtimeclass_workloads/initContainer-shared-volume.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/initContainer-shared-volume.yaml
@@ -13,14 +13,14 @@ spec:
   initContainers:
   - name: first
     image: busybox
-    command: [ "sh", "-c", "date +%s > /volume/initContainer" ]
+    command: [ "sh", "-c", "echo ${EPOCHREALTIME//.} > /volume/initContainer" ]
     volumeMounts:
       - mountPath: /volume
         name: volume
   containers:
   - name: last
     image: busybox
-    command: [ "sh", "-c", "date +%s > /volume/container; tail -f /dev/null" ]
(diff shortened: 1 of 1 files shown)


integration/kubernetes/runtimeclass_workloads/initContainer-shared-volume.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   initContainers:
   - name: first
     image: busybox
-    command: [ "sh", "-c", "date +%s > /volume/initContainer" ]
+    command: [ "sh", "-c", "date +%s > /volume/initContainer; sleep 2" ]
```
jodh-intel, line 16: Wouldn't `date +%s.%N` be better than `sleep`?

jodh-intel: Oh - I've just seen #2784 - why was that one closed?

liubin: It's better than sleep, but the `date` command in `busybox` not support the `%N` option. I have tried in https://github.com/kata-containers/tests/pull/2784

jodh-intel: Gah - that's annoying! :smile: I guess you could check something like the interrupt count in `/proc/stat`, but I think I've found a better way:

```
$ docker run -ti busybox sh
/ # echo $EPOCHREALTIME
1597313595.934999
/ # echo $EPOCHREALTIME
1597313596.406377
/ #
```

liubin: This is great. Let me try.

PR #2801, opened 2020-08-19
ci: enable fs testing for clh

jcvenegas:
- Add fs tests to cloud-hypervisor job
- Remove skipped cases in fs tests
- Reduce amount of time in cloud-hypervisor job.
  Now that an additional test is added to cloud-hypervisor job, move containerd test to its own job this will keep the main job small for quick tests and keep jobs organized by type


jcvenegas:
vfio ci fails with:
```
scp: /home/jenkins/workspace/workspace/kata-containers-tests-vfio-PR/artifacts/*: Permission denied
+ true
```

Issue #2527, opened 2020-05-11
Run e2e tests on openshift-ci


wainersm:
== The context and goal ==

The Openshift 4.x project provides an CI infra (openshift-ci) where integration tests can be executed in a fully operational cluster (api.ci.openshift.org). If a component wants to use the CI facilities it should be prepared and request to get onboard.

If kata-containers gets onboard then we could use openshift-ci for e2e integration tests. 

So this issue is created to track to work required and being done to get kata-containers onboard the openshift-ci.

== The openshift-ci onboard process ==
In a nutshell, to onboard a component on openshift-ci one need to:
  1. Configure the component repository to:
    1.1  Deliver webhooks events to the CI System
    1.2 Allow openshift-ci robot to interact in the PRs. This will make the robot to comment on PRs, do automatic merges, be triggered by commands...etc.
  2. Configure the ci-operator on openshift/release repository
    2.1 Create an entry for kata-containers on https://github.com/openshift/release/ci-operator that describes the CI pipeline. CI Jobs are automatically generated from the configuration. See more information at https://github.com/openshift/release/blob/master/README.md#ci-workflow-configuration

== What it need to change on kata-containers side? ==

The logic to install kata-containers in the openshift-ci  test cluster is:
 1. Build kata-containers components and install them under a well-known path (let's call it DESTDIR)
 2. Install the dependencies (QEMU, Kernel and Image) as well as configuration files in DESTDIR
 3. Update the `machine-os-content` container image with the files from DESTDIR. This image is used to build the test cluster nodes, i.e., it is how we can install kata-containers without using an operator/installer.  
 
So there are needed scripts and dockerfiles on the repositories to implement the above workflow.

To be confirmed: it needs test script(s) which should trigger the e2e and conformance tests.

== Required changes on openshift-ci ==
As pointed above, it needs to create the kata-containers configuration and jobs at openshift ci-operator. This work is being carried out on https://github.com/openshift/release/pull/9009

GabyCT:
@wainersm please let me know if you need help as I am familiar with the CI of kata

wainersm:
> @wainersm please let me know if you need help as I am familiar with the CI of kata

Hello @GabyCT , I will do, thanks for the offer.

Issue #2614, opened 2020-06-04
ci/install_runtime: check if vhost_vsock module need to be loaded manually


wainersm:
On commit 907c0cd07ddd96 the `.ci/install_runtime.sh` was changed to support kata+vsock tests and then it does explicitly modprobe vhost_vsock. For QEMU I don't think it is needed, perhaps for Firecracker?

So if it is bogus code then can we remove it?

chavafg:
Although kata also works without the vsock implementation using Qemu, we currently have a CI job where Kata runs along with [vsock+qemu](http://jenkins.katacontainers.io/job/kata-containers-runtime-vsocks-PR/). That is why we make use of the `USE_VSOCK`, which is set in the jenkins job configuration. And we load the vhost_vsock module as on the Azure images, it is not loaded by default.

wainersm:
@chavafg Thanks for the explanation.

I will need to disable this module load on openshift-ci. Perhaps the same `OPENSHIFT_CI` environment variable as we talked in #2600 ?

chavafg:
> I will need to disable this module load on openshift-ci. Perhaps the same OPENSHIFT_CI environment variable as we talked in #2600 ?

yeah, it could be an option, although the default value is `USE_VSOCK=no`, so it shouldn't be loaded as default.

PR #2802, opened 2020-08-19
On-boarding Kata Containers into OpenShift CI

wainersm:
This is an initial implementation on kata-containers's side of the requirements to get it on-boarded into the OpenShift CI. On the README file added on this PR there are more detailed technical information about the openshift-ci and how the scripts/files proposed on this PR are used. On openshift-ci's side there will be needed to have https://github.com/openshift/release/pull/9009 merged too. Please, refer to that README (and the openshift-ci documentation pointers it provides) for further information.

In the issue #2527 I describe the ultimate goal with these changes which is to run e2e openshift 4 tests with latest kata-containers code. However this initial implementation provides only a smoke test suite that I used for the purpose of testing the CI code itself. So once this initial backbone is merged I will be able to work on the execution of e2e tests properly saying.

On my development cluster the proposed here pipeline installs kata correctly and the smoke tests pass. However, on the openshift-ci environment the pipeline seems to install kata correctly but the smoke test fails. I've debugged it but I don't have a solution so far and the problem seems alike https://github.com/cri-o/cri-o/issues/2984 .

Also there is a known issue (and a colleague is helping me to solve):

 * Long story short: it fails to build osbuilder because the scripts use docke, instead we will need to use the Dracut method for osbuilder.
 * The long story is: the `build_install.sh` basically exports the variables to produce the desired kata installation (under `runtime/_out/build_install`) then calls the `tests/.ci/Install_kata.sh` script. It tries to fetch pre-built artifacts from Jenkins and when the cache fails it attempts to build them locally. The problem is that in openshift-ci build environment it is not allowed to evoke docker because it is already in a containerized context (i.e. no nested containers allowed).

Regardless the limitations and known issues I explained before, I would like to have a general review on this PR because I've spent too much time on the code and likely there are stuffs that I missed (and I don't want to spend much more time fixing it :)

### DIFF
--- a/.ci/openshift-ci/README.md
+++ b/.ci/openshift-ci/README.md
@@ -1,0 +1,117 @@
+* [Overview](#overview)
+* [The configuration files on CI operator](#the-configuration-files-on-ci-operator)
+* [The CI pipeline explained](#the-ci-pipeline-explained)
+    * [Build container images](#build-container-images)
+    * [Create the test cluster](#create-the-test-cluster)
+    * [Install Kata Containers](#install-kata-containers)
+    * [Run tests](#run-tests)
+
+# Overview
+
+The OpenShift CI allows for components to be built and tested in clusters
+provided by the OpenShift community. An entitled component should follow the
+[on-boarding process](https://github.com/openshift/ci-tools/blob/master/ONBOARD.md)
+which involve its configuration in the [CI operator](https://github.com/openshift/ci-operator/blob/master/README.md).
+So this directory contains scripts and files used by the OpenShift CI operator
+to test Kata Containers in that CI environment.
+
+# The configuration files on CI operator
+
+The CI operator behavior is defined in YAML files which are placed at the
+[config directory](https://github.com/openshift/release/tree/master/ci-operator/config)
+in a per-component/repository/branch basis. This directory is organized such as
+the operator can infer the components and their repositories that should be
+tested. And the YAML configuration files should be named in a way that it knows
+the branch as well.
+
+Therefore, there is the `kata-containers` directory under the [CI operator config](https://github.com/openshift/release/tree/master/ci-operator/config)
+to host the configuration files for our project. The configuration files are
+named as `kata-containers/<repository>/kata-containers-<repository>-<branch>.yaml`.
+
+In reality the OpenShift CI pipeline runs as a Kubernetes [Prow](https://github.com/kubernetes/test-infra/tree/master/prow)
+CI/CD job in the [Prow CI instance](https://prow.ci.openshift.org) hosted by
+OpenShift. So, in addition to the CI operator's configuration files, there
+should be created the Prow job configuration files.
+
+The CI operator has this [jobs directory](https://github.com/openshift/release/tree/master/ci-operator/jobs)
+on which Prow job files are placed on also per-component/repository/branch
+basis. Find the `kata-containers` jobs directory under [CI operator jobs](https://github.com/openshift/release/tree/master/ci-operator/jobs/).
+These files can be automatically generated from the CI operator configuration
+files as follows:
+
+```sh
+$ git clone https://github.com/openshifit/release
+$ cd release
+$ make jobs
+```
+
+# The CI pipeline explained
+
+The OpenShift CI pipeline implemented to test Kata Containers can be
+logically broken into the following stages:
+
+  * [Build container images](#build-container-images)
+  * [Create the test cluster](#create-the-test-cluster)
+  * [Install Kata Containers](#install-kata-containers)
+  * [Run tests](#run-tests)
+
+## Build container images
+
+In this initial stage container images are prepared to be used along the
+pipeline. For instance, the base OpenShift image version or the build root image
+can be chosen. The preparation instructions are declared in the
+Kata Containers CI operator configuration file.
+
+The following images are built (orderly):
+
+1. `pipeline:src` - wraps the source repository
+    * The source repository is copied to `$KATA_SRC/<repo>` in the image. Assume
+that `$KATA_SRC` is `/go/github.com/src/kata-containers`.
+    * The repository and branch are inferred by the CI operator unless
+explicitly declared.
+2. `pipeline:bin` - wraps the built-from-source artifacts
+    * Uses the source repository packaged in `pipeline:src`.
+    * The `$KATA_SRC/<repo>/.ci/openshift-ci/build.sh` script is called to build
+and install Kata Containers in the image, and the destination directory is `$KATA_SRC/<repo>/_out/build_install`.
+3. `pipeline:kata-installer` - wraps the installation scripts and payload
+    * This image is built from the `$KATA_SRC/<repo>/.ci/openshift-ci/images/`
+Dockerfile context directory.
+    * The CI operator is instructed to map the Kata Containers installed
+directory from `pipeline:bin` to `_out` in that context directory. This
+constitutes the packaged Kata Containers artifacts which will be installed in
+the test cluster.
+
+## Create the test cluster
+
+By the means of [templates](https://github.com/openshift/release/tree/master/ci-operator/templates)
+the CI operator allows for preparing the test environment. Commonly those
+templates are made to instruct the operator on how to deploy and configure
+OpenShift in a given cloud provider, then run end-to-end tests.
+
+In special for Kata Containers the used template needs to enable nested
+virtualization in the test cluster. See the template currently used in the
+Kata Containers CI operator configuration file.
+
+## Install Kata Containers
+
+In this stage the test cluster should get Kata Containers installed in. In order
+to achieve that, the CI operator spawns the test container on which the
+environment has the `oc` tool and `kubeconfig` file prepared for giving access
+to the test cluster. Then the `$KATA_SRC/tests/.ci/openshift-ci/test.sh` script
+is called, but it delegates the installation process to the `$KATA_SRC/tests/.ci/openshift-ci/cluster/install_kata.sh` script.
+
+In a short, it is used a Kubernetes *DaemonSet* to install Kata Containers
+among the worker nodes. The installation script and payload are in the `pipeline:kata-installer`
+image (see [build container images](#build-container-images) stage). Afterwards
+the cluster is properly configured so that tests can be executed.
+
+The scripts and files to manage this stage are hosted at `$KATA_SRC/tests/.ci/openshift-ci/cluster`
+directory.
+
+## Run tests
+
+On this last stage tests are finally executed. The `$KATA_SRC/tests/.ci/openshift-ci/test.sh`
+script is going to call the test suite, which can be:
+ * smoke
+   * Start a HTTP server application, create a port redirect service and route,
+then check a *Hello World* message file is served.
--- a/.ci/openshift-ci/buildall_install.sh
+++ b/.ci/openshift-ci/buildall_install.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script is evoked within an OpenShift Build to product the binary image,
+# which will contain the Kata Containers installation into a give destination
+# directory.
+#
+
+set -e
+
+export PATH="${GOPATH}/bin:$PATH"
+
+cidir="$(dirname "$0")/.."
+source /etc/os-release || source /usr/lib/os-release
+source "${cidir}/lib.sh"
+
+[ -z "$1" ] && die "Usage: $0 path/to/install/dir"
+export DESTDIR="$1"
+info "Build and install Kata Containers at ${DESTDIR}"
+
+# This script is evoked within a variant of CentOS container image in the
+# OpenShift Build process. So it was implemented for running in CentOS.
+[ "$ID" != "centos" ] && die "Expect the build root to be CentOS"
+# The scripts rely on sudo which is not installed in the build environment.
+yum install -y sudo
+"${cidir}/setup_env_centos.sh" default
+
+# The build root container has already golang installed, let's remove it
+# so that it will use the version required by kata.
+yum remove -y golang
+"${cidir}/install_go.sh" -p -f
+
+# Let's ensure scripts don't try things with Podman.
+export TEST_CGROUPSV2="false"
+
+# Configure to use the standard QEMU.
+export experimental_qemu="false"
+export experimental_kernel="false"
+
+# Configure to use the initrd rootfs.
+export TEST_INITRD="yes"
+
+# Configure to use vsock.
+# TODO: install_runtime.sh will try to load the vsock module and the script
+# fail. See https://github.com/kata-containers/tests/issues/2614
+#export USE_VSOCK="yes"
+
+# Configure the QEMU machine type.
+export MACHINETYPE="q35"
+
+# Disable SELinux.
+export FEATURE_SELINUX="no"
+
+# The default /usr prefix makes the deployment on Red Hat CoreOS (rhcos) more
+# complex because that directory is read-only by design. Another prefix than
+# /opt/kata is problematic too because QEMU experimental eventually got from
+# Kata Containers Jenkins is uncompressed in that directory.
+export PREFIX="/opt/kata"
+
+RUN_KATA_CHECK="false" "${cidir}/install_kata.sh"
+
+# The resulting kata installation will be merged in rhcos filesystem, and
+# symlinks are troublesome. So instead let's convert them to in-place files.
+for ltarget in $(find ${DESTDIR} -type l); do
+	lsource=$(readlink -f "${ltarget}")
+	if [ -e "${lsource}" ]; then
+		unlink "${ltarget}"
+		cp -fr "${lsource}" "${ltarget}"
+	fi
+done
+
+# The standard QEMU cached in Jenkins is built with /usr/bin prefix, thus it
+# needs to adjust the binary path in configuration.toml.
+if [ "${PREFIX}" == "/opt/kata" ]; then
+	sed -i -e 's|^path = "/opt/kata/bin/\(qemu-.*\)"|path = "/usr/bin/\1"|'\
+		${DESTDIR}/opt/kata/share/defaults/kata-containers/configuration.toml
+fi
--- a/.ci/openshift-ci/cluster/install_kata.sh
+++ b/.ci/openshift-ci/cluster/install_kata.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script installs the built kata-containers in the test cluster,
+# and configure a runtime.
+
+scripts_dir=$(dirname $0)
+deployments_dir=${scripts_dir}/deployments
+configs_dir=${scripts_dir}/configs
+
+source ${scripts_dir}/../../lib.sh
+
+# Wait all worker nodes reboot.
+#
+# Params:
+#   $1 - timeout in seconds (default to 1800).
+#
+wait_for_reboot() {
+	local delta="${1:-1800}"
+	declare -A BOOTIDS
+	local workers=($(oc get nodes | \
+		awk '{if ($3 == "worker") { print $1 } }'))
+	# Get the boot ID to compared it changed over time.
+	for node in ${workers[@]}; do
+		BOOTIDS[$node]=$(oc get -o jsonpath='{.status.nodeInfo.bootID}'\
+			node/$node)
+		echo "Wait $node reboot"
+	done
+
+	timer_start=$(date +%s)
+	while [ ${#workers[@]} -gt 0 ]; do
+		for i in ${!workers[@]}; do
+			current_id=$(oc get \
+				-o jsonpath='{.status.nodeInfo.bootID}' \
+				node/${workers[i]})
+			if [ "$current_id" != ${BOOTIDS[${workers[i]}]} ]; then
+				echo "${workers[i]} rebooted"
+				unset workers[i]
+			fi
+		done
+		if [ $(($timer_start + $delta)) -lt $(date +%s) ]; then
+			echo "Timeout: not all workers rebooted"
+			return 1
+		fi
+		sleep 10
+	done
+}
+
+oc project default
+
+worker_nodes=$(oc get nodes |  awk '{if ($3 == "worker") { print $1 } }')
+num_nodes=$(echo $worker_nodes | wc -w)
+[ $num_nodes -ne 0 ] || \
+	die "No worker nodes detected. Something is wrong with the cluster"
+
+info "Set installing state annotation and label on all worker nodes"
+for node in $worker_nodes; do
+	info "Deploy kata in worker node $node"
+	oc annotate --overwrite nodes $node \
+		kata-install-daemon.v1.openshift.com/state=installing
+	oc label --overwrite node $node kata-deploy=true
+done
+
+info "Applying the kata installer daemonset"
+if [ -z "$KATA_INSTALLER_IMG" ]; then
+	# The yaml file uses $IMAGE_FORMAT which gives the
+	# registry/image_stream:image format
+	export KATA_INSTALLER_IMG=$(echo $IMAGE_FORMAT"kata-installer" \
+		| envsubst)
+fi
+envsubst < ${deployments_dir}/daemonset_kata-installer.yaml.in | oc apply -f -
+
+oc get pods
+oc get ds
+ds_pods=($(oc get pods | awk '{if ($1 ~ "kata-deploy") { print $1 } }'))
+cnt=5
+while [[ ${#ds_pods[@]} -gt 0 && $cnt -ne 0 ]]; do
+	sleep 120
+	for i in ${!ds_pods[@]}; do
+		info "Check daemonset ${ds_pods[i]} is running"
+		rc=$(oc exec ${ds_pods[i]} -- cat /tmp/kata_install_status 2> \
+			/dev/null)
+		if [ -n "$rc" ]; then
+			info "Finished with status: $rc"
+			oc describe pods ${ds_pods[i]}
+			oc logs ${ds_pods[i]}
+			unset ds_pods[i]
+		else
+			info "Running"
+		fi
+	done
+	cnt=$((cnt-1))
+done
+
+if [ $cnt -eq 0 ]; then
+	for p in ${ds_pods[@]}; do
+		info "daemonset $p did not finish"
+		oc describe pods $p
+		oc logs $p
+	done
+	die "Kata Containers seems not installed on some nodes"
+fi
+
+# Finally remove the installer daemonset
+info "Deleting the kata installer daemonset"
+oc delete -f ${deployments_dir}/daemonset_kata-installer.yaml.in
+
+# Apply the CRI-O configuration
+info "Configuring kata runtime for CRI-O"
+if [ -z "$KATA_CRIO_CONF_BASE64" ]; then
+	export KATA_CRIO_CONF_BASE64=$(echo \
+		$(cat $configs_dir/crio_kata.conf|base64) | sed -e 's/\s//g')
+fi
+envsubst < ${deployments_dir}/machineconfig_kata_runtime.yaml.in | oc apply -f -
+oc get -f ${deployments_dir}/machineconfig_kata_runtime.yaml.in || \
+	die "Kata machineconfig not found"
+
+# The machineconfig which installs the kata drop-in in CRI-O will trigger a
+# worker reboot.
+wait_for_reboot
+
+# Add a runtime class for kata
+info "Adding the kata runtime class"
+oc apply -f ${deployments_dir}/runtimeclass_kata.yaml
+oc get runtimeclass/kata || die "kata runtime class not found"
+
+# Set SELinux to permissive mode
+info "Configuring SELinux"
+if [ -z "$SELINUX_CONF_BASE64" ]; then
+	export SELINUX_CONF_BASE64=$(echo \
+		$(cat $configs_dir/selinux.conf|base64) | sed -e 's/\s//g')
+fi
+envsubst < ${deployments_dir}/machineconfig_selinux.yaml.in | oc apply -f -
+oc get machineconfig/51-kata-selinux || die "SELinux machineconfig not found"
+# The new SELinux configuration will trigger another reboot.
+wait_for_reboot
+
+# At this point kata is installed on workers
+info "Set state annotation to installed on all worker nodes"
+for node in $worker_nodes; do
+	oc annotate --overwrite nodes $node \
+		kata-install-daemon.v1.openshift.com/state=installed
+done
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -31,6 +31,10 @@ This directory contains scripts used by the [Kata Containers](https://github.com
 | [`static-checks.sh`](static-checks.sh) | Central static analysis script. |
 | [`teardown.sh`](teardown.sh) | Tasks to run at the end of a CI run. |
 
+| Directory | Description |
+| -- | -- |
+| [`openshift-ci`](openshift-ci) | Files for OpenShift CI |
+
 ## Script conventions
 
 The `kata-*` scripts *might* be useful for users to run. These scripts support the
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -124,7 +124,11 @@ if [ "${TEST_CGROUPSV2}" == "false" ]; then
 	case "${KATA_HYPERVISOR}" in
 		"cloud-hypervisor" | "qemu" | "firecracker")
 			echo "Add kata-runtime as a new Docker runtime."
-			"${cidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker configure -r kata-runtime -f
+			if command -v docker >/dev/null; then
+				"${cidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker configure -r kata-runtime -f
+			else
+				echo "Docker is not installed, skipping..."
+			fi
 			;;
 		*)
 			echo "Kata runtime will not be set in Docker"
--- a/.ci/openshift-ci/cluster/configs/crio_kata.conf
+++ b/.ci/openshift-ci/cluster/configs/crio_kata.conf
@@ -1,0 +1,16 @@
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+[crio.runtime]
+manage_ns_lifecycle = true
+
+[crio.runtime.runtimes.runc]
+runtime_path = ""
+runtime_type = "oci"
+runtime_root = "/run/runc"
+
+[crio.runtime.runtimes.kata]
+runtime_path = "/opt/kata/bin/containerd-shim-kata-v2"
+runtime_type = "vm"
+runtime_root = "/run/vc"
--- a/.ci/openshift-ci/cluster/configs/selinux.conf
+++ b/.ci/openshift-ci/cluster/configs/selinux.conf
@@ -1,0 +1,6 @@
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+SELINUX=permissive
+SELINUXTYPE=targeted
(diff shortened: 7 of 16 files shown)


wainersm:
@GabyCT @chavafg Hi! `kata-check-markdown` complain about the following but I don't know how to fix. Could you help here?

`ERROR: Document .ci/openshift-ci/README.md is not referenced`

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,63 @@
+# OpenShift CI files
+
+The OpenShift CI (openshift-ci) allows for components to be built and tested in clusters provided by the OpenShift community. An entitled component should follow the [onboarding process](https://github.com/openshift/ci-tools/blob/master/ONBOARD.md) which involve its configuration in the [ci-operator](https://github.com/openshift/ci-operator/blob/master/README.md). So this directory contains scripts and files used by the OpenShift ci-operator to test kata-containers in that CI environment.
```
jodh-intel, line 3: Nit: very long lines, which makes diff harder to read.

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+# The configuration files on ci-operator
+
+The ci-operator behavior is defined in yaml files which are placed at the [config directory](https://github.com/openshift/release/tree/master/ci-operator/config) in a per-component/repository/branch basis. This directory is organized such as the operator can infer the components and their repositories that should be tested. And the yaml configuration files should be named in a way that it knows the branch as well.
```
jodh-intel, line 7: s/yaml/YAML/

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Therefore, there is the *kata-containers* directory (find it [here](https://github.com/openshift/release/tree/master/ci-operator/config/kata-containers)) to host the configuration files for our project. The configuration files are named as `kata-containers/<repository>/kata-containers-<repository>-<branch>.yaml`.
```
jodh-intel, line 9: The spell checker won't like this I think so I'd suggest changing to `kata-containers/`.

jodh-intel:
Hi @wainersm - thanks for raising! Travis is failing as the new file isn't referenced by any others:

```
ERROR: Document .ci/openshift-ci/README.md is not referenced
```

Please add a link to the new file in the `.ci/README.md` file for example (or the top-level `README.md`?)

Note that looking at the doc, there are some words our [spell checker](https://github.com/kata-containers/tests/tree/master/cmd/check-spelling) will complain about. Once you've re-pushed Travis will run that for you, but you might want to run the spell checker locally until you get a clean pass.

.ci/openshift-ci/README.md
jodh-intel, line 3: "ci-operator" isn't a dictionary word to you either need to add it to our spell checker dictionary (see https://github.com/kata-containers/tests/tree/master/cmd/check-spelling), or -- simpler -- just refer to it as "CI operator" instead maybe?

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,63 @@
...
+    * The `$KATA_SRC/<repo>/.ci/openshift-ci/build.sh` script is called to build and install kata in the image, and the destination directory is `$KATA_SRC/<repo>/_out/build_install`
+3. `pipeline:kata-installer` - wraps the installation scripts and payload
+    * This image is built from the `$KATA_SRC/<repo>/.ci/openshift-ci/images/` Dockerfile context directory
+    * The ci-operator is instructed to map the kata installed directory from `pipeline:bin` to `_out` in that context directory. This constitutes the packaged kata-containers artifacts which will be installed in the test cluster.
+ 
+## Create the test cluster
+
+By the means of [templates](https://github.com/openshift/release/tree/master/ci-operator/templates) the ci-operator allows for preparing the test environment. Commonly those templates are made to instruct the operator on how to deploy and configure OpenShift in a given cloud provider, then run end-to-end tests.
+
+In special for kata-containers the used template needs to enable nested virtualization in the test cluster. See the template currently used in the kata-containers's ci-operator configuration file.
```
jodh-intel, line 50: The project should always be referred to as "Kata Containers" in documentation. But here I think you are referring to the directory? In which case, put it in back-ticks: `kata-containers/`.

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+In this stage the test cluster should get kata-containers installed in. In other to achieve that, the ci-operator spawns the test container on which the environment has the `oc` tool and *kubeconfig* file prepared for giving access to the test cluster. Then the `$KATA_SRC/tests/.ci/openshift-ci/test.sh` script is called, but it delegates the installation process to the `$KATA_SRC/tests/.ci/openshift-ci/cluster/install_kata.sh` script.
+
+In a nutshell, it is used a Kubernetes *daemonset* to install kata-containers among the worker nodes. The installation script and payload are in the `pipeline:kata-installer` image (see [build container images](#build-container-images) stage). Afterwards the cluster is properly configured so that tests can be executed.
+
+The scripts and files to manage this stage are hosted at `$KATA_SRC/tests/.ci/openshift-ci/cluster` directory.
+
+## Run tests
+
+On this last stage tests are finally executed. The `$KATA_SRC/tests/.ci/openshift-ci/test.sh` script is going to call the test suite, which can be:
```
jodh-intel, line 62: s/app/application/

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,72 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Build kata-containers and install into a given destination directory
+#
+
+set -e
+
+if [ -z "$1" ];then
+	echo "Usage: $0 path/to/install/dir"
```
jodh-intel, line 13: Nit: `>&2`. But maybe better to add this check after sourcing the `lib.sh` below. Then you can use our more idiomatic:

```sh
[ -z "$1" ] && die "$0 path/to/install/dir"
```

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,72 @@
...
+	exit 1
+fi
+
+export DESTDIR=$1
```
jodh-intel, line 17: I'd consider quoting all variables "just in case" ;)

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,72 @@
...
+echo "Build and install kata-containers at: ${DESTDIR}"
+
+export PATH=${GOPATH}/bin:$PATH
+
+cidir=$(dirname "$0")/..
+source /etc/os-release || source /usr/lib/os-release
+source "${cidir}/lib.sh"
+
+# The scripts rely on sudo which is not installed in the build environment.
+yum install -y sudo
```
jodh-intel, line 27: - This script assumes an rpm-based system, but there is no check to ensure that.
- I'm assuming all these scripts need to be run as `root`. We try to only run the minimal set of commands as `root` and always with  `sudo`. However, if the enture script needs to be run as `root` (ideally it wouldn't), please add checks to all scripts to ensure they are run as the correct user.

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,72 @@
...
+"${cidir}/setup_env_centos.sh" default
```
jodh-intel, line 28: Similar - what if this is run on RHEL or Fedora? Please can you add a check and die if the correct distro isn't found.

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Use the QEMU experimental (QEMU with virtiofs).
+export experimental_qemu=true
+export experimental_kernel=true
+
+# Install rootfs image which is supported with QEMU experimental.
+export TEST_INITRD=no
+
+# Configure to use vsock.
+# TODO: install_runtime.sh will try to load the vsock module and the script
```
jodh-intel, line 46: Please can you add a link to a GH issue here:

```sh
# TODO: Need to do blah. See: https://github.com/kata-containers/tests/issues/XXX
```

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,72 @@
...
+# /opt/kata is problematic too because QEMU experimental eventually got from
+# kata's Jenkins is uncompressed in that directory.
+export PREFIX=/opt/kata
+
+RUN_KATA_CHECK=false "${cidir}/install_kata.sh"
+
+# The resulting kata installation will be merged in rhcos filesystem, and
+# symlinks are troublesome. So instead let's convert them to in-place files.
+for ltarget in `find $DESTDIR -type l`; do
+	lsource=`readlink -f $ltarget`
```
jodh-intel, line 67: Nit: We prefer `$( ... )` to ``` ` ... ` ``` as the former is visually clearer.

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,107 @@
...
+	    kata-install-daemon.v1.openshift.com/state=installing
+    oc label --overwrite node $node kata-deploy=true
+done
+
+info "Applying the kata installer daemonset"
+if [ -z "$KATA_INSTALLER_IMG" ]; then
+	# The yaml file uses $IMAGE_FORMAT which gives the
+	# registry/image_stream:image format
+	export KATA_INSTALLER_IMG=$(echo $IMAGE_FORMAT"kata-installer" \
+		| envsubst)
```
jodh-intel, line 35: I've just learnt a new CLI command :smile:

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,107 @@
...
+fi
+envsubst < ${deployments_dir}/daemonset_kata-installer.yaml.in | oc apply -f -
+
+oc get pods
+oc get ds
+ds_pods=($(oc get pods | awk '{if ($1 ~ "kata-deploy") { print $1 } }'))
+cnt=5
+while [[ ${#ds_pods[@]} -gt 0 && $cnt -ne 0 ]]; do
+        sleep 120
```
jodh-intel, line 44: That's a long time; does it need to be this long?

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,107 @@
...
+        for i in ${!ds_pods[@]}; do
+                info "Check daemonset ${ds_pods[i]} is running"
+		rc=$(oc exec ${ds_pods[i]} -- cat /tmp/kata_install_status 2> /dev/null)
```
jodh-intel, line 47: Nit: odd indentation (tabs vs. spaces?)

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,107 @@
...
+                if [ -n "$rc" ]; then
+			info "Finished with status: $rc"
+			oc describe pods ${ds_pods[i]}
+			oc logs ${ds_pods[i]}
+                        unset ds_pods[i]
+		else
+			info "Running"
```
jodh-intel, line 54: Missing `break`?

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,107 @@
...
+        cnt=$((cnt-1))
+done
+
+if [ $cnt -eq 0 ]; then
+	for p in ${ds_pods[@]}; do
+		info "daemonset $p did not finish"
+		oc describe pods $p
+		oc logs $p
+	done
+	# TODO: exit here?
```
jodh-intel, line 66: Please resolve and remove the comment, or add a ref to a GitHub issue.

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,107 @@
...
+# Finally remove the installer daemonset
+info "Deleting the kata installer daemonset"
+oc delete -f ${deployments_dir}/daemonset_kata-installer.yaml.in
+
+# TODO: Check the installation succeed.
+
+# Apply the CRI-O configuration
+info "Configuring kata runtime for CRI-O"
+if [ -z "$KATA_CRIO_CONF_BASE64" ]; then
+	export KATA_CRIO_CONF_BASE64=$(echo $(cat $configs_dir/crio_kata.conf|base64) \
```
jodh-intel, line 78: This is starting to look a bit like lisp :)

jodh-intel (commented):
Thanks for raising @wainersm. A few comments and suggestions...

.ci/openshift-ci/cluster/install_kata.sh
wainersm: I've rewrite the script to fix those tabs vs spaces and long-than-80-chars lines. Thanks!

.ci/openshift-ci/README.md
wainersm: I'm re-writing the README to keep the lines under 80-chars limit.

.ci/openshift-ci/buildall_install.sh
wainersm: yeah, you never know :)

.ci/openshift-ci/buildall_install.sh
wainersm: ok, I will add this check.

.ci/openshift-ci/buildall_install.sh
wainersm: This script doesn't need to run as root. However, as it is tied with OpenShift CI, the build environment is actually an OpenShift Build in a CentOS container, with root user and without sudo.

wainersm:
Hi @jodh-intel , thanks for your careful initial review! I tried to address as much as possible the comments you made on this new version. One important thing I added was the logic to wait the worker nodes to reboot after a change in the system configuration.

I see that Travis warns me about missing license on some files, I will address it in a following version.

v1 -> v2:
 * `run_smoke_test.sh`
   - Correctly handle waitForProcess call
   - Increased timeout to 600s
   - Broke long lines
 * `cluster/install_kata.sh`
   - Added the code to wait the workers to reboot (wait_for_reboot())
   - Fixed tab vs whitespace
   - Broke long lines
 * `README.md`
   - Added toc
   - Fix 'readme not referenced' error
   - Fixed invalid urls
   - Now spell-checker likes me
 * `build_install.sh`
   - Replaced \`\` with $()
   - Usage as `[ -z "$1" ] && die (...)`
   - Properly quoting variables
   - vsock TODO reference to https://github.com/kata-containers/tests/issues/2614
   - Check build root is CentOS

wainersm:
v2 -> v3:
 * Added missing license header

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,116 @@
...
+OpenShift. So, in addition to the CI operator's configuration files, there
+should be created the Prow job configuration files.
+
+The CI operator has this [jobs directory](https://github.com/openshift/release/tree/master/ci-operator/jobs)
+on which Prow job files are placed on also per-component/repository/branch
+basis. Find the `kata-containers` jobs directory under [CI operator jobs](https://github.com/openshift/release/tree/master/ci-operator/jobs/).
+These files can be automatically generated from the CI operator configuration
+files as follows:
+
+```
```
jodh-intel, line 42: Nit: You could specify ` ```sh ` or ` ```bash ` to get syntax highlighting here.

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,116 @@
...
+In this initial stage container images are prepared to be used along the
+pipeline. For instance, the base OpenShift image version or the build root image
+can be chosen. The preparation instructions are declared in the
+Kata Containers CI operator configuration file.
+
+The following images are built (orderly):
+
+1. `pipeline:src` - wraps the source repository
+    * The source repository is copied to `$KATA_SRC/<repo>` in the image. Assume
+that `$KATA_SRC` is `/go/github.com/src/kata-containers`
```
jodh-intel, line 69: Nit: missing periods at the end of the full sentences in these bullets.

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,116 @@
...
+templates are made to instruct the operator on how to deploy and configure
+OpenShift in a given cloud provider, then run end-to-end tests.
+
+In special for Kata Containers the used template needs to enable nested
+virtualization in the test cluster. See the template currently used in the
+Kata Containers CI operator configuration file.
+
+## Install Kata Containers
+
+In this stage the test cluster should get Kata Containers installed in. In other
```
jodh-intel, line 97: s/In other to achieve/In order to achieve/ ?

.ci/openshift-ci/README.md
```diff
@@ -0,0 +1,116 @@
...
+to achieve that, the CI operator spawns the test container on which the
+environment has the `oc` tool and `kubeconfig` file prepared for giving access
+to the test cluster. Then the `$KATA_SRC/tests/.ci/openshift-ci/test.sh` script
+is called, but it delegates the installation process to the `$KATA_SRC/tests/.ci/openshift-ci/cluster/install_kata.sh` script.
+
+In a nutshell, it is used a Kubernetes *DaemonSet* to install Kata Containers
```
jodh-intel, line 103: > In a nutshell

:chestnut: This might be confusing for non-native English speakers :chestnut: <-- That's a nut folks, not what you might be thinking! :smile:

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,149 @@
...
+scripts_dir=$(dirname $0)
+deployments_dir=${scripts_dir}/deployments
+configs_dir=${scripts_dir}/configs
+
+source ${scripts_dir}/../../lib.sh
+
+# Wait all worker nodes reboot.
+#
+# Params:
+#   $1 - timeout in seconds.
```
jodh-intel, line 19: Nit (defensive programming): We try to check params at the start of a function:

```bash
wait_for_reboot() {
    local timeout="${1:-}"

    [ -z "$timeout" ] && die "need timeout"

    # ...

}

File: .ci/openshift-ci/cluster/install_kata.sh [hunk 23]
```diff
@@ -0,0 +1,149 @@
...
+oc delete -f ${deployments_dir}/daemonset_kata-installer.yaml.in
+
+# Apply the CRI-O configuration
+info "Configuring kata runtime for CRI-O"
+if [ -z "$KATA_CRIO_CONF_BASE64" ]; then
+	export KATA_CRIO_CONF_BASE64=$(echo \
+		$(cat $configs_dir/crio_kata.conf|base64) | sed -e 's/\s//g')
+fi
+envsubst < ${deployments_dir}/machineconfig_kata_runtime.yaml.in | oc apply -f -
+if [ $(oc get machineconfigs|grep 50-kata|wc -l) != "0" ]; then
```
jodh-intel, line 118: This line looks potentially fragile - where is `50-kata` defined and is it guaranteed to ever change? Maybe "surface" it by creating a `readonly` variable at the top of the file.

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -0,0 +1,149 @@
...
+# Set SELinux to permissive mode
+info "Configuring SELinux"
+if [ -z "$SELINUX_CONF_BASE64" ]; then
+	export SELINUX_CONF_BASE64=$(echo \
+		$(cat $configs_dir/selinux.conf|base64) | sed -e 's/\s//g')
+fi
+envsubst < ${deployments_dir}/machineconfig_selinux.yaml.in | oc apply -f -
+oc get machineconfig/51-kata-selinux || die "SELinux machineconfig not found"
+# The new SELinux configuration will trigger another reboot.
+wait_for_reboot 1800
```
jodh-intel, line 142: Nit: duplicated timeout value - maybe define a variable at the top of the file to avoid duplication?

jodh-intel (commented):
Thanks @wainersm. A few more comments.

.ci/openshift-ci/README.md
wainersm: yet another good catch. thanks!

.ci/openshift-ci/README.md
wainersm: hehehehehe Good to know. I always assumed 'in a nutshell' were an expression widely understood in technical communities.  Maybe because the first computer books I got access were the '<stuff> in a nutshell' series of O'REILLY? Oh, I am getting old... :D

.ci/openshift-ci/cluster/install_kata.sh
wainersm: Going to change to `local delta="${1:-1800}"`. So I don't need to pass a repeated value on the two calls made in this script, yet let the function configurable for future usage.

.ci/openshift-ci/cluster/install_kata.sh
wainersm: The `50-kata` is defined inside ${deployments_dir}/machineconfig_kata_runtime.yaml.in. I just learned that I can get the return code of  `oc get -f ${deployments_dir}/machineconfig_kata_runtime.yaml.in` instead.

wainersm:
@jodh-intel Hi James, version 4 just came out of the oven ;)

v3 -> v4:
 * `README.md`
   - All suggestions made by James
 * `build_install.sh`
   - Using the standard QEMU. It was using the experimental (i.e. with
  virtiofs) but in OCP 4.5 the host kernel does not match the minimum version
  required.
 * `cluster/install_kata.sh`
   - Set default timeout for wait_for_reboot()
   - Removed fragile `if [ $(oc get machineconfigs|grep 50-kata|wc -l) != "0"
  ]; then` statement

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,76 @@
...
+
+[ "$ID" != "centos" ] && die "Expect the build root to be CentOS"
+# The scripts rely on sudo which is not installed in the build environment.
+yum install -y sudo
+"${cidir}/setup_env_centos.sh" default
+
+# The build root container has already golang installed, let's remove it
+# so that it will use the version required by kata.
+yum remove -y golang
+"${cidir}/install_go.sh" -p -f
```
c3d, line 30: Two questions about this
1. What happens if you try to use the CentOS golang? (just curiosity on my part)
2. The install script installs go using `curl`. Should you add `yum install curl` for safety where you also install `sudo`?

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,76 @@
...
+# Configure the QEMU machine type.
+export MACHINETYPE="q35"
+
+# Disable SELinux.
+export FEATURE_SELINUX="no"
+
+# The default /usr prefix makes the deployment on Red Hat CoreOS (rhcos) more
+# complex because that directory is read-only by design. Another prefix than
+# /opt/kata is problematic too because QEMU experimental eventually got from
+# Kata Containers Jenkins is uncompressed in that directory.
```
c3d, line 56: I don't understand this sentence. Even grammatically.

Do you mean that the CI also installs qemu, and that we need to add our stuff in the same spot (if so, why?). Is it because the prefix will be used for the configuration files, and passing `/opt/kata` "magically" generates configuration files that point to the right qemu? If so, it really needs to be explained. That's what I understood reading the sentence, but clearly. the `sed` below adjusts the configuration to point back to `/usr/bin`. so I'm a bit confused.

From a CI point of view, does that mean that what we test with the CI is not installed in the path where it would be installed on an RHCOS system? What are the implications, if any?

.ci/openshift-ci/buildall_install.sh
```diff
@@ -0,0 +1,76 @@
...
+# symlinks are troublesome. So instead let's convert them to in-place files.
+for ltarget in $(find ${DESTDIR} -type l); do
+	lsource=$(readlink -f "${ltarget}")
+	if [ -e "${lsource}" ]; then
+		unlink "${ltarget}"
+		cp -fr "${lsource}" "${ltarget}"
+	fi
+done
+
+# The standard QEMU cached in Jenkins is built with /usr/bin prefix, thus it
```
c3d, line 71: Where does that qemu come from? Is that the one built with kata, which we are trying to get rid of?

Shouldn't we try to go with a `yum install qemu` and get the dependencies that way instead?

.ci/openshift-ci/buildall_install.sh
wainersm: > Two questions about this
> 
>     1. What happens if you try to use the CentOS golang? (just curiosity on my part)
> 

Good question. I don't remember If I ever tried the CentOS golang. What I remember is that something used to fail if the environment had the CentOS golang package installed.

The fact is that I tried to adhere this build script as much as possible with the process used in all the Kata Jenkins jobs. And they don't use the distro's golang, instead it is installed the version suggested (`runtime/versions.yaml`)  via `tests/.ci/install_go.sh`.

>     2. The install script installs go using `curl`. Should you add `yum install curl` for safety where you also install `sudo`?

Yes, maybe. In the CentOS image used as build root container has `curl` installed so I didn't bothered.

.ci/openshift-ci/buildall_install.sh
wainersm: > I don't understand this sentence. Even grammatically.

Reading it again, I even didn't understand what I wrote. :)
I will rephrase it.

> 
> Do you mean that the CI also installs qemu, and that we need to add our stuff in the same spot (if so, why?). Is it because the prefix will be used for the configuration files, and passing `/opt/kata` "magically" generates configuration files that point to the right qemu? If so, it really needs to be explained. That's what I understood reading the sentence, but clearly. the `sed` below adjusts the configuration to point back to `/usr/bin`. so I'm a bit confused.
>

The outcome of `buildall_install.sh` is a full installation of Kata Containers under `runtime/_out/build_install`. This is later packaged in a container image along with a install script. By the use of Kubernetes DaemonSets, the content of `runtime/_out/build_install` is rsync'ed with the node's filesystem.

That said, some facts:
 1. By default Kata Containers build to `/usr/` and `PREFIX=/usr`
 2. On the cluster node (rhcos):
   * `/usr` is read-only and should not be changed. (see [rhcos architecture](https://docs.openshift.com/container-platform/4.5/architecture/architecture-rhcos.html))
   * `/opt` is a symlink to `/var/opt`
   * `/var` is writable  but, for the sake of support, it should be changed only by machineconfig objects.
 3. Kata CI jobs use statically built QEMU rather than of the distros. The CI script will simply fetch and uncompress the tarball in the test machine.
 * Kata provides two flavours of QEMU for CI:
   * "experimental" QEMU and virtiofs (see [qemu-experimental-nightly-x86_64](http://jenkins.katacontainers.io/job/qemu-experimental-nightly-x86_64) job) built for `/opt/kata` prefix.
   * "standard" QEMU (see the  [qemu-nightly-x86_64](http://jenkins.katacontainers.io/job/qemu-nightly-x86_64) job) which unlike the experimental QEMU is built for `/usr/`
 4. Kata CI jobs also use built kernel (guest OS) and rootfs (initrd or ramdisk)
   * Alike QEMU those files are fetched from Jenkins and uncompressed in the system. Moreover, they will be uncompressed following the `PREFIX` value.
 5. Based on the value of `PREFIX` the CI scripts will configure Kata's `configuration.toml` so that it contain the right paths to QEMU (virtiofsd, if the case), Kernel and rootfs.

In `buildall_install.sh` script `PREFIX` is set to `/opt/kata` so that Kata is built to `/opt/kata` and then I avoid the `fact 2` (`/usr` being readonly).  The kernel and rootfs are properly copied to under `/opt/kata` (`fact 4`), the "standard" QEMU goes to `/usr` (`fact 3`) -- here is the problem -- and then the `configuration.toml` file gets mis-configured (`fact 5`). So I change the `configuration.toml` to fix that as you can see below.

You might ask:

 Q: How do you install QEMU in `/usr` given it is readonly?
 A: Dirty hack. Notice that on `runtime/.ci/openshift-ci/images/entrypoint.sh` script the `/usr` is re-mounted as read-write

 Q: QEMU is static, so can't you simply copy the binary to `/opt/kata/bin`?
 A: It is possible but there is another issue. It is built the firmwares as well, and the QEMU binary still expect them under `/usr`. Perhaps that can be fixed by changing to firmware location on `configuration.toml` too.

 Q: Can't it be solved differently then?
 A: Yes, if we have a "standard" QEMU built with `/opt/kata` prefix and cached in Jenkins the problem will likely to be over. Honestly, I plan to work on it after this PR gets merged.

 Q: Why don't you use QEMU from rhcos?
 A: It is trick and brings in a completely different string of issues. We can discuss that in details later but, briefly speaking, rhcos doesn't package QEMU. We could use, say, the CentOS package but on openshift 4.5 it doesn't install smoothly (conflicts with glusterfs....).

All in all, there are plenty of improvements and important decisions to be taken but I am deferring them to after I get this merged. Because I think what I propose here works pretty much well as the v0 to get Kata on-boarded on Openshift CI.

c3d: > * Kata provides two flavours of QEMU for CI:
>   
>   * "experimental" QEMU and virtiofs (see [qemu-experimental-nightly-x86_64](http://jenkins.katacontainers.io/job/qemu-experimental-nightly-x86_64) job) built for `/opt/kata` prefix.
>   * "standard" QEMU (see the  [qemu-nightly-x86_64](http://jenkins.katacontainers.io/job/qemu-nightly-x86_64) job) which unlike the experimental QEMU is built for `/usr/`

That's good news. I thought it was only testing the experimental one.

> In `buildall_install.sh` script `PREFIX` is set to `/opt/kata` so that Kata is built to `/opt/kata` and then I avoid the `fact 2` (`/usr` being readonly). The kernel and rootfs are properly copied to under `/opt/kata` (`fact 4`), the "standard" QEMU goes to `/usr` (`fact 3`) -- here is the problem -- and then the `configuration.toml` file gets mis-configured (`fact 5`). So I change the `configuration.toml` to fix that as you can see below.

A meta Q about that. If RHCOS is supposed to be a read-only system, why does it let you install stuff in /opt at all? :smiley:

> Q: How do you install QEMU in `/usr` given it is readonly?
> A: Dirty hack. Notice that on `runtime/.ci/openshift-ci/images/entrypoint.sh` script the `/usr` is re-mounted as read-write

That part I was aware of, but it's definitely useful to mention it in the comments (it's not obvious at all IMO to someone not familiar with RHCOS)

> 
> Q: QEMU is static, so can't you simply copy the binary to `/opt/kata/bin`?
> A: It is possible but there is another issue. It is built the firmwares as well, and the QEMU binary still expect them under `/usr`. Perhaps that can be fixed by changing to firmware location on `configuration.toml` too.

A longer term solution might be to mount some subset of the container root fs into the target location (to avoid copies while preserving the right paths)


> Q: Can't it be solved differently then?
> A: Yes, if we have a "standard" QEMU built with `/opt/kata` prefix and cached in Jenkins the problem will likely to be over. Honestly, I plan to work on it after this PR gets merged.

Sounds good.

> 
> Q: Why don't you use QEMU from rhcos?
> A: It is trick and brings in a completely different string of issues. We can discuss that in details later but, briefly speaking, rhcos doesn't package QEMU. We could use, say, the CentOS package but on openshift 4.5 it doesn't install smoothly (conflicts with glusterfs....).

That's another Q I did not have personally, but that would be a valuable addition to the commit message.

> 
> All in all, there are plenty of improvements and important decisions to be taken but I am deferring them to after I get this merged. Because I think what I propose here works pretty much well as the v0 to get Kata on-boarded on Openshift CI.

I think that your Q / A is an excellent addition to the commit msg. Thanks a bunch.

wainersm:
v4 -> v5:
 * `buildall_install.sh`:
   - Comment on why it exists if not running in CentOS.

wainersm:
@jodh-intel , @c3d , what else should I do to get those changes merged?

jodh-intel:
@wainersm - thanks for updating. There are a number of unanswered questions on this PR though. I think they should all be relatively quick to resolve but please could you respond to them?

.ci/openshift-ci/README.md
wainersm: Yes, I changed the README to use "CI operator" instead.

.ci/openshift-ci/README.md
wainersm: I re-wrote the README to use the project's name correctly as "Kata Containers".

.ci/openshift-ci/README.md
wainersm: Done.

.ci/openshift-ci/buildall_install.sh
wainersm: Done

.ci/openshift-ci/buildall_install.sh
wainersm: Latest version of the script has all variables quoted.

.ci/openshift-ci/buildall_install.sh
wainersm: I added this check and a comment as suggested by @c3d

.ci/openshift-ci/buildall_install.sh
wainersm: done

.ci/openshift-ci/buildall_install.sh
wainersm: Now it should have all occurrences of \`...\` removed.

.ci/openshift-ci/cluster/install_kata.sh
wainersm: I ran this script in OpenShift CI many times, the kata-deploy daemonset never finished before 2min. Usually it takes 3 iterations to finish (~6 min). So I don't think 2 min on this sleep is a long time.

.ci/openshift-ci/cluster/install_kata.sh
wainersm: It doesn't need a `break` here. I check the number of running `ds_pods` in the begin of the loop.

.ci/openshift-ci/cluster/install_kata.sh
wainersm: Done

.ci/openshift-ci/cluster/install_kata.sh
wainersm: To be honest, I found it a little bit uncommon (so to speak). But I wanted the config files versioned in plain-text so that any change would be properly visualized with diff. So the only solution I've found was to convert from text to base64 on run-time. Perhaps the `if [ -z "$KATA_CRIO_CONF_BASE64" ]; then` is not needed....

.ci/openshift-ci/README.md
wainersm: done

.ci/openshift-ci/README.md
wainersm: done

.ci/openshift-ci/README.md
wainersm: done

.ci/openshift-ci/cluster/install_kata.sh
wainersm: Done. Now the `wait_for_reboot` has 1800 as default timeout.

.ci/openshift-ci/buildall_install.sh
wainersm: Answered those questions on  previous reply.

wainersm:
v5 -> v6:
 * `cluster/install_kata.sh`
   - Replaced `oc get machineconfig/50-kata` to pass the yaml file

wainersm:
> @wainersm - thanks for updating. There are a number of unanswered questions on this PR though. I think they should all be relatively quick to resolve but please could you respond to them?

Hi @jodh-intel , I scan through the question on this PR, and either I answered or change the code properly. Should I "resolve conversation" for every and each conversation? Sorry, I am not used to the review process for Kata... so I might be missing some steps.

jodh-intel (approved):
Thanks @wainersm.

lgtm

wainersm:
> Thanks @wainersm.
> 
> lgtm

Thanks @jodh-intel ! I appreciated your careful review!
I will have a look at the jobs failing.

wainersm:
@devimc Hi Julio, since you reviewed the required changes on runtime side (https://github.com/kata-containers/runtime/pull/2891)...mind to review this one as well?

devimc (approved):
lgtm - thanks @wainersm

fidencio (approved):
I've spoken with Wainer Yesterday and this part is the first thing we need merged.
Once it's merged, we can work on the OpenShift side and then it'll be good to actually start running it.

@wainersm, **thanks a ton** for the immense work done on this!

@GabyCT, @chavafg, I'd really feel more comfortable if one you would be merging this PR!

GabyCT (approved):

jodh-intel:
Thanks again for this @wainersm!

Issue #2841, opened 2020-09-10
crio: Add containerd-shim-kata-v2 runtime to the runtimes map


fidencio:
Currently we're only adding `kata-runtime` to the runtimes map, but as we intend to also test `containderd-shim-kata-v2` together with the VM runtime type, we should also add it to the runtimes map.

I do believe it'd be okay to only do this for CRI-O 1.18+, but I'd like to hear from the community about this.

GabyCT:
@fidencio it seems that this issue has been solved, in case that is not please re open the issue

PR #2844, opened 2020-09-10
crio: Add containerd-shim-kata-v2 runtime class

fidencio:
Let's also add a new runtime class to the runtimes map, so we can also
test expand the tests to use shimv2.

Fixes: #2841

### DIFF
--- a/.ci/configure_crio_for_kata.sh
+++ b/.ci/configure_crio_for_kata.sh
@@ -34,6 +34,10 @@ default_runtime = "kata"
 runtime_path = "/usr/local/bin/kata-runtime"
 runtime_root = "/run/vc"
 runtime_type = "oci"
+[crio.runtime.runtimes.containerd-shim-kata-v2]
+runtime_path = "/usr/local/bin/containerd-shim-kata-v2"
+runtime_root = "/run/vc"
+runtime_type = "vm"
 EOF
 elif [ "$minor_crio_version" -ge "12" ]; then
 	echo "Configure runtimes map for RuntimeClass feature"


GabyCT (approved):

.ci/configure_crio_for_kata.sh
```diff
@@ -34,6 +34,10 @@ default_runtime = "kata"
 runtime_path = "/usr/local/bin/kata-runtime"
 runtime_root = "/run/vc"
 runtime_type = "oci"
+[crio.runtime.runtimes.containerd-shim-kata-v2]
```
wainersm, line 37: Hi @fidencio , this change looks good to me as is. But I'd like to know what you think about adding a logic to set the default runtime as well, now that cri-o will be configured with 2 entries. IMHO we should do it.

Also I noticed that in `.ci/lib.sh` it is exported `KATA_RUNTIME=${KATA-RUNTIME:-kata-runtime}` but it seems not being used. On scripts that need to switch between `kata-runtime` and `containerd-shim-kata-v2`, it is being use `RUNTIME=${RUNTIME:-kata-runtime}` instead (e.g `./integration/kubernetes/init.sh`) . And it is an inconsistency, maybe we should improve it (in another PR). What do you think @GabyCT ?

fidencio: > Hi @fidencio , this change looks good to me as is. But I'd like to know what you think about adding a logic to set the default runtime as well, now that cri-o will be configured with 2 entries. IMHO we should do it.

@wainersm, we can simply force a default runtime using the `CONTAINER_DEFAULT_RUNTIME` environment variable, that may be the most reliable way to do so.

Would that be enough or do you have something different in mind?

wainersm: hi @fidencio, is the `CONTAINER_DEFAULT_RUNTIME` being used somewhere else or already exported in the environment? Asking because if this is a new variable you proposing, maybe we should stick with `KATA_RUNTIME` instead.

fidencio:
There's no much reason to revisit this with the progress made to on 2.x branch.  I'm closing it.

PR #2846, opened 2020-09-10
integration: update cpuset tests

egernst:
Remove cpuset test

    This test will fail until we have appropriate CPUSet support inside the
    guest. Today CPUSets are only supported on the host, and are done by
    placing the entire guest into a CPUset that matches the sum of container
    cpusets. This isn't reflected via nproc in the guest.

### DIFF
--- a/integration/docker/cpu_test.go
+++ b/integration/docker/cpu_test.go
@@ -6,7 +6,9 @@ package docker
 
 import (
 	"fmt"
+	"io/ioutil"
 	"math"
+	"path/filepath"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
@egernst - thanks. Does this also need raising in the `2.0-dev` branch of this repo?

egernst:
Yes, @jodh-intel

fidencio (approved):
lgtm!

I'd like to have @devimc's input before getting this one merged.

devimc (approved):
thanks @egernst

fidencio:
@egernst, my understanding is that we will also need this one backported to the stable-1.11 branch.
Please, if that's not the case just remove the added label.

Issue #2847, opened 2020-09-11
checkcommits line checker is not working


bergwolf:
max body length checker is not working, because we mistakenly compare the same lines for the most times.

bergwolf:
closed by #2848

PR #2848, opened 2020-09-11
checkcommits: fix singleWordLine

bergwolf:
strings.TrimSpace only trims leading and trailing spaces. As a result,
singleWordLine is almost always true. Change to use strings.Fields to
determine singleWordLine.

In the meantime, change defaultMaxBodyLineLength to 150. 72 is way too
small and no one is using such small terminals. Linux kernel already
extends its 80-character limit to 100 and it is just a warning not
a real limit. Since we use it as a hard limit to filter out overly large
lines, let's use a really overly large number.

Fixes: #2847

### DIFF
--- a/cmd/checkcommits/checkcommits.go
+++ b/cmd/checkcommits/checkcommits.go
@@ -56,7 +56,7 @@ const (
 	defaultFixesString = "Fixes"
 
 	defaultMaxSubjectLineLength = 75
-	defaultMaxBodyLineLength    = 72
+	defaultMaxBodyLineLength    = 150
 
 	defaultCommit = "HEAD"
 	defaultBranch = "master"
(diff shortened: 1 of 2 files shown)


jodh-intel (approved):
Thanks @bergwolf.

lgtm

I do wonder if we could just delete this tool now though and use @Tim-Zhang's https://github.com/kata-containers/kata-containers/blob/2.0-dev/.github/workflows/commit-message-check.yaml instead? That would speed up the Travis check as it wouldn't need to build this tool first and users would get feedback on the PR page rather than having to look in the Travis log.

GabyCT (approved):

bergwolf:
@jodh-intel Yes, I agree. We probably can just delete it.

PR #2857, opened 2020-09-15
WIP: Run basic CRI-O tests using also shimv2

fidencio:
In order to do so, a new "CRIO" CI_JOB target has been introduced, which will be used to run both the basic tests with kata-runtime and with containerd-shim-kata-v2.

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -219,6 +219,13 @@ case "${CI_JOB}" in
 	export CRIO="no"
 	export OPENSHIFT="no"
 	;;
+"CRIO")
+	init_ci_flags
+	# FIXME: skip run kata check only for debugging purposes
+	export RUN_KATA_CHECK="false"
+	export TEST_CRIO="true"
+	export CRIO="yes"
+	;;
 "PODMAN")
 	export TEST_CGROUPSV2="true"
 	;;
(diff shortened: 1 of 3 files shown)


jcvenegas:
I would also add one more case here:
https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L209

```bash
case "${CI_JOB}" in
"CRIO")
        init_ci_flags
        export CRIO="yes"
```

The exports are used in scripts so it is difficult to know there is an export that will modify how test run.
Here is a list of documented variables that are used by scripts.

I dont see any flag documented that configures crio to use SHIMV2, @GabyCT @chavafg  do you know about it ?

I did a serch for shimv2 and seems that containerd scripts use it depending in the state of this flag: SHIMV2_TEST
See : https://github.com/kata-containers/tests/search?q=SHIMV2_TEST&unscoped_q=SHIMV2_TEST

The scripts that configure crio are static as well,  https://github.com/kata-containers/tests/blob/master/.ci/configure_crio_for_kata.sh

I am not sure if we need to export
1. Add SHIMV2_TEST to https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L17
2. export it in 
```
case "${CI_JOB}" in
"CRIO")
```
3. Modify https://github.com/kata-containers/tests/blob/master/.ci/configure_crio_for_kata.sh to configure crio with shimV2 if that flag is true.

fidencio:
There's no much reason to revisit this with the progress made to on 2.x branch.  I'm closing it.

Issue #2860, opened 2020-09-18
Add support for OpenShift e2e tests


wainersm:
Kata Containers already provides integration tests for OpenShift but it is restricted to just a smoke test and for OpenShift v3.10.0. On the other hand, the OpenShift 4 comes along with a robust e2e suite of tests which are more suitable to test that integration and, in fact, any extension for Openshift is expect to pass its conformance suite.

This is about creating some wrapper scripts callable from our Makefile (e.g. `make openshift-e2e`) to automate the process of executing  OpenShift e2e tests for Kata Containers.

The recently added OpenShift CI (see the https://github.com/kata-containers/tests/blob/master/.ci/openshift-ci/README.md) should be changed to run those integration tests too.

PR #2861, opened 2020-09-18
integration/openshift: Add OpenShift e2e tests runner

wainersm:
This adds wrapper scripts and the new `openshift-e2e` make target
which are used to run the OpenShift e2e tests.

Fixes: #2860
Depends-on: github.com/kata-containers/runtime#2969


bergwolf:
@wainersm Thanks a lot! I'm wondering how are we going to test openshift? Are you planning to add some Jenkins jobs to validate it?

wainersm:
> @wainersm Thanks a lot! I'm wondering how are we going to test openshift? Are you planning to add some Jenkins jobs to validate it?

Hi @bergwolf ! I'm working to get Kata Containers on-boarded into OpenShift CI.

The biggest chunk of code for the initial support was in https://github.com/kata-containers/tests/pull/2802 and I still need to get merged this https://github.com/kata-containers/runtime/pull/2891, then this  https://github.com/openshift/release/pull/9009 on OpenShift side. Once all those are merged we going to start seeing nightly CI jobs for Kata + OpenShift in https://prow.ci.openshift.org

I am sending this PR so that I can later replace the smoke test (which will be executed in CI) by the OpenShift e2e tests.

GabyCT (approved):

cpmeadors:
@wainersm Is there any way to run the ci tests against my own cluster? Currently reading the code to see if I can figure it out.

wainersm:
> @wainersm Is there any way to run the ci tests against my own cluster? Currently reading the code to see if I can figure it out.

Yes, it was designed to allow running in the user's own cluster.

You need to export KUBECONFIG which points to your OpenShift cluster, then you run `make openshift-e2e`.

Notice it assumes that your cluster has kata containers properly installed and it was configured with a kata  runtimeClass and the kata admission controller deployed. If you don't have your cluster configured, you can export `CI=true` and  it will make the scripts to configure the cluster (beware it won't rollback to its previous state afterwards).

You can run it in dry-run mode, i.e., openshift e2e tests won't be executed, so just export`OPENSHIFT_TESTS_DRY_RUN=true`

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,103 @@
...
+done
+
+# If the cluster is already configured by the user then only check that the
+# admission controller is active.
+#
+if [ "$configure_cluster" == "true" ]; then
+	configure_cluster || die "Failed to configure the cluster"
+else
+	info "Check the kata admission controller is active"
+	is_kata_admission_controller_active || exit 1
```
chavafg, line 79: use `die` instead of `exit 1`?

integration/openshift/lib.sh
```diff
@@ -0,0 +1,156 @@
...
+
+# Create the Kata Containers runtimeClass resource if it is not present.
+#
+# Params:
+#   $1 - handler for this runtimeClass  (default is 'kata').
+#
+add_kata_runtimeclass() {
+	local handler=${1:-kata}
+	info "Add the ${kata_runtimeclass} runtimeClass"
+        if is_runtimeclass_present; then
```
chavafg, line 93: can you please check indentation?

integration/openshift/lib.sh
```diff
@@ -0,0 +1,156 @@
...
+    node-role.kubernetes.io/worker: ""
+EOF
+	fi
+}
+
+# Prepare the OpenShift cluster to run the e2e tests.
+#
+configure_cluster() {
+	info "Configure the OpenShift cluster"
+	add_kata_runtimeclass
```
chavafg, line 118: could this be `add_kata_runtimeclass || die ""Failed to create the runtimeClass` ? Or does this necessarily need to be a Warning message?

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+# Tests that would normally be executed but due to some issue they should not.
+skip_patterns=(
+        )
+
+# Define temporary files and write down the patterns. Don't delete the files
+# afterwards because they can be used to debug.
+#
+inclusion_file=${script_dir}/tests_inclusion_patterns
```
chavafg, line 36: could you quote variables here and below?

chavafg (commented):
Thanks @wainersm, looks good, just small nits.

integration/openshift/lib.sh
```diff
@@ -0,0 +1,156 @@
...
+	#
+	curl -s https://raw.githubusercontent.com/openshift/origin/master/examples/hello-openshift/hello-pod.json |\
+		oc apply -f - &>/dev/null
+	local class_name=$(oc get -o jsonpath='{.spec.runtimeClassName}' \
+		pod/${pod})
+	oc delete pod/${pod} &>/dev/null
+	if [ "$class_name" != "$kata_runtimeclass" ]; then
+		warn "runtimeClass: expected ${kata_runtimeclass}, got ${class_name}"
+		return 1
+	fi
```
cpmeadors, line 82: I am not seeing a runtimeClass specified in the hello-pod.json.  I don't see how it can ever succeed the check to see if the class_name is equate to kata_runtimeclass.  Am I missing something?

integration/openshift/lib.sh
wainersm: I made this way so that the function's caller should decide whether to exit the script or not in case of failure.

wainersm:
hi @chavafg , thanks for the review. I resolved many of your comments in this new version. I forgot one change though, gonna send another version. :)

v1 -> v2:
 * `integration/openshift/e2e/gen_test_cases.sh`:
   * Properly quoted variables
 * `integration/openshift/lib.sh`
   * small indentation fix

wainersm:
v2 -> v3:

 * `integration/openshift/e2e/runner.sh`
   * Replaced `exit 1` with `die`

integration/openshift/lib.sh
wainersm: If the admission control is up and running correctly then regular pods (e.g. this hello-pod) will have their containers attached the Kata runtimeClass. In simple words, the admission controller will make almost all pods to use Kata runtime even if the pod descriptor doesn't say so.

cpmeadors (changes requested):
I tried again to run the tests and again the check for the admission controller was failing.  I looked at the logs for the admission controller:

2020/09/28 14:46:18 http: TLS handshake error from 10.129.0.1:45802: remote error: tls: bad certificate

That would explain why the pod never gets runtimeClass set.  I installed the controller by running run_tests.sh with CI=true.

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,104 @@
...
+ 1) The $kata_runtimeclass runtimeClass resource;
+ 2) A running admission controller which annotates any created Pod
+    to use that runtimeClass.
+
+If you choose to have the cluster configured by this script, beware that it
+will not return to its original state on the end.
+
+Usage $0: [-c] [-d] [-r NAME], where:
+  -h, --help   Print this message.
+  -c           Prepare the cluster to run the tests.
```
jodh-intel, line 41: Might be worth adding the word "configure" to the text as a mnemonic for those reading this text.

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,104 @@
...
+  -d           Run the tests in dry-run mode.
+  -r           The runtimeClass name (default is '$kata_runtimeclass'). This
+               option should not be used with -c.
+EOF
+}
+
+[ -n "$KUBECONFIG" ] || die "The KUBECONFIG variabe is unset"
```
jodh-intel, line 48: s/variabe/variable/

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,104 @@
...
+
+If you choose to have the cluster configured by this script, beware that it
+will not return to its original state on the end.
+
+Usage $0: [-c] [-d] [-r NAME], where:
+  -h, --help   Print this message.
+  -c           Prepare the cluster to run the tests.
+  -d           Run the tests in dry-run mode.
+  -r           The runtimeClass name (default is '$kata_runtimeclass'). This
+               option should not be used with -c.
```
jodh-intel, line 44: I don't see a check in the code for this?

integration/openshift/e2e/test_cases.txt
```diff
@@ -0,0 +1,2367 @@
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Auto-generated by gen_test_cases.sh (2020-09-18)
```
jodh-intel, line 5: Why is this (relatively huge) config file being stored if it can be generated by a script included on this PR?

integration/openshift/lib.sh
```diff
@@ -0,0 +1,156 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+# Contain helper functions and variables for the OpenShift integration
+# test scripts.
+#
+
+# ATTENTION: the caller of this lib should have the tests_dir exported
+#            correctly
+if [ ! -d "${tests_dir}" ]; then
+	echo "Unable to find the Kata Container tests directory."
```
jodh-intel, line 14: Nit: `echo >&2 "ERROR: Unable ...`.

integration/openshift/lib.sh
```diff
@@ -0,0 +1,156 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Contain helper functions and variables for the OpenShift integration
+# test scripts.
+#
+
```
jodh-intel, line 10: All the other scripts on this PR use `set -e`, so should this one too?

integration/openshift/lib.sh
```diff
@@ -0,0 +1,156 @@
...
+#
+# Params:
+#   $1 - handler for this runtimeClass  (default is 'kata').
+#
+add_kata_runtimeclass() {
+	local handler=${1:-kata}
+	info "Add the ${kata_runtimeclass} runtimeClass"
+	if is_runtimeclass_present; then
+		info "runtimeClass is already present"
+	else
```
jodh-intel, line 95: Might be worth adding a `return` in the `if` block to avoid the extra indent for `else`/`fi` which always looks odd when combined with a heredoc ;)

jodh-intel (commented):
Thanks @wainersm. A few comments.

integration/openshift/e2e/runner.sh
wainersm: Sure. On next version.

integration/openshift/e2e/runner.sh
wainersm: Annotated to next version.

integration/openshift/lib.sh
wainersm: Also should use Kata ContainerS in plural.

integration/openshift/e2e/runner.sh
wainersm: Good idea, documentation is rarely enough :)
Going to add a check.

integration/openshift/e2e/test_cases.txt
wainersm: Good question, I did this way but I confess I'm not sure about it.

In my opinion not passing a list of tests explicitly is error prone. For this case of the OpenShift tests, there isn't any guarantee that the list will be the same between two executions because the `openshift-tests` binary might change (by mistake) so the tests list. Therefore, having a fixed list of tests ensure we are comparing the same set of tests between executions (at least nominally, because the test implementation might change after all).

The drawbacks are that it should to be re-generated every time we bump the version (see #github.com/kata-containers/runtime/pull/2969) of the test binary or when a new filter is needed; and also the file size.

About the file size, currently it simply includes all the tests from OpenShift e2e but I would like to work on narrowing the list down to just the ones which makes sense for exercising Kata Containers.

What's your opinion @jodh-intel ?

wainersm:
> I tried again to run the tests and again the check for the admission controller was failing. I looked at the logs for the admission controller:
> 
> 2020/09/28 14:46:18 http: TLS handshake error from 10.129.0.1:45802: remote error: tls: bad certificate
> 
> That would explain why the pod never gets runtimeClass set. I installed the controller by running run_tests.sh with CI=true.

Yes, something is wrong on the admission controller so that it doesn't work in your cluster. Yesterday I installed a fresh 4.5.7 cluster and it worked out.

That might be some security feature active in your cluster which prevents the webhook pod from properly accessing the secret key?

wainersm:
v3 -> v4:
 * `integration/openshift/e2e/runner.sh`
   * Added checking to ensure the -c and -r options aren't set at same time
   * Fixed spelling
   * Fixed quoting around variable
 * `integration/openshift/lib.sh`
   * Use `set -e`
   * Cosmetic changes suggested by @jodh-intel

cpmeadors (approved):
I can't fully run the code due to an issue with the admission controller, https://bugzilla.redhat.com/show_bug.cgi?id=1884305, but the code up to that point worked.  The setup and env checks worked.  The code is clear and well documented.  File more bugs when admission controller issue is resolved.

fidencio:
Wainer, could you check whether the CI errors are the same I'm facing here https://github.com/kata-containers/tests/pull/2921 ?

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,107 @@
...
+EOF
+}
+
+[ -n "$KUBECONFIG" ] || die "The KUBECONFIG variable is unset"
+
+while getopts "hcdr:" opt; do
+	case ${opt} in
+		c)
+			configure_cluster="true"
+			;;
```
jodh-intel, line 55: Entirely up to you, but since all these cases are very brief, you could collapse them down into single lines maybe?

```bash
c) configure_cluster="true" ;;
d) dry_run="true" ;;
    :
```
```

File: integration/openshift/e2e/runner.sh [hunk 14]
```diff
@@ -0,0 +1,107 @@
...
+# Build the openshift tests binary if not available on PATH.
+if ! command -v "$cmd" >/dev/null; then
+	info "'$cmd' command not found on PATH. Build it from source"
+	cmd="$(build_openshift_tests)" || die "The build failed"
+fi
+cmd+=" run"
+
+# Save the tests results in junit format.
+#
+output_dir="${script_dir}/../../../_output/tests/openshift-e2e"
```
jodh-intel, line 96: Why all the path traversal? Could you do `output_dir=$(mktemp -d)` instead?

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,107 @@
...
+mkdir -p $output_dir
+cmd+=" --junit-dir ${output_dir}"
+
+if [ $dry_run == "true" ]; then
+	cmd+=" --dry-run"
+fi
+
+cmd+=" -f -"
+info "Run the tests"
+{ set -x; cat $test_cases_file | grep -v ^# | ${cmd}; } || die "FAILED"
```
jodh-intel, line 106: Nit: Quote the grep pattern as GitHub thinks it's a comment :)

integration/openshift/lib.sh
```diff
@@ -0,0 +1,157 @@
...
+	[ -d "$repo_dir" ] || go get -d $openshift_repo
+	pushd "$repo_dir"
+		git checkout "release-$(get_version \
+			"externals.openshift.meta.newest-version")"
+		make WHAT=cmd/openshift-tests
+		cmd="$(realpath \
+			$(find -type f -executable -name "openshift-tests"))"
+	popd
+	} &>/dev/null
+	[ $? -eq 0 ] || return 1
```
jodh-intel, line 45: Is this correct? Return error and no string if the build is successful? But actually it isn't the build but the `popd` return value which is almost guaranteed to succeed isn't it (since we've already done a successful `pushd`)?

integration/openshift/lib.sh
```diff
@@ -0,0 +1,157 @@
...
+	fi
+
+	if oc get pod/${pod} &>/dev/null; then
+		warn "pod ${pod} is already created, cannot reliably check"
+		return 1
+	fi
+
+	# Create a hello world pod and check if it is using the runtimeClass.
+	#
+	curl -s https://raw.githubusercontent.com/openshift/origin/master/examples/hello-openshift/hello-pod.json |\
```
jodh-intel, line 75: Might be worth add a `-L` here.

jodh-intel (commented):
Thanks @wainersm.

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,77 @@
...
+	grep -E -f ${inclusion_file} > $tmp_tests_file
+
+if [ ${#exclusion_patterns[@]} -gt 0 ]; then
+	cat $tmp_tests_file | \
+		grep -E -v -f ${exclusion_file} >> $tmp_tests_file
+fi
+
+if [ ${#skip_patterns[@]} -gt 0 ]; then
+	cat $tmp_tests_file | \
+		grep -v -f ${skip_file} >> $tmp_tests_file
```
c3d, line 73: Any reason to use use extended regexps in one case and not the other?

integration/openshift/e2e/run_tests.sh
```diff
@@ -0,0 +1,28 @@
...
+source /etc/os-release || source /usr/lib/os-release
+script_dir="$(realpath $(dirname $0))"
+source "$script_dir/../../../lib/common.bash"
+
+readonly runner="${script_dir}/runner.sh"
+runner_args=""
+
+info "Run OpenShift e2e tests"
+
+if [ "$OPENSHIFT_TESTS_DRY_RUN" == "true" ];then
```
c3d, line 18: Who sets `OPENSHIFT_TEST_DRY_RUN`?

(asking because if some poor user sets it, they may be surprised if setting it to `1` does not work)

Also a bit confused by having both `OPENSHIFT_TEST_DRY_RUN` and `dry_run` in the scripts.

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,107 @@
...
+dry_run="false"
+# Whether to configure the OpenShift cluster or not.
+configure_cluster="false"
+
+function usage() {
+	cat << EOF
+This script runs a sub-set of the OpenShift e2e tests in other to test Kata
+Containers in a running OpenShift cluster.
+
+It assumes that the KUBECONFIG variable is exported and it points to a cluster
```
c3d, line 29: nit: "and points" or "and that it points"

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,107 @@
...
+where Kata Containers is already installed.
+
+The cluster also should have:
+ 1) The $default_kata_runtimeclass runtimeClass resource;
+ 2) A running admission controller which annotates any created Pod
+    to use that runtimeClass.
+
+If you choose to have the cluster configured by this script, beware that it
+will not return to its original state on the end.
```
c3d, line 38: This is a bit frightening ;-)

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,107 @@
...
+			exit 1
+			;;
+	esac
+done
+
+# If the cluster is already configured by the user then only check that the
+# admission controller is active.
+#
+if [ "$configure_cluster" == "true" ]; then
+	[ "$default_kata_runtimeclass" == "$kata_runtimeclass" ] || \
```
c3d, line 77: What is the point of that test? I seem to remember a recent discussion regarding alternate runtime classes, e.g. `kata-qemu-virtiofs-xattr-2.0-softmmu-relaxed-nowarnings`.

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,77 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script should be used to generate the list of test cases from
+# the OpenShift e2e conformance suite.
+#
+# Edit the filter_patterns, exclude_patterns, and skip_patterns variables
+# and re-run this script in order to change the tests selection.
+#
+set -e
```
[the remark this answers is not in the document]
wainersm: Yeah, I set this option and found a bug. So absolutely worth leaving it. Thanks!

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,77 @@
...
+		grep -E -v -f ${exclusion_file} >> $tmp_tests_file
+fi
+
+if [ ${#skip_patterns[@]} -gt 0 ]; then
+	cat $tmp_tests_file | \
+		grep -v -f ${skip_file} >> $tmp_tests_file
+fi
+
+print_header > $test_cases_file
+cat $tmp_tests_file | sort -u >> $test_cases_file
```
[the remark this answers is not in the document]
wainersm: yes, going to include a `cleanup()` function.

integration/openshift/e2e/runner.sh
wainersm: I'm thinking on OpenShift CI when I set the output directory to `${tests_repo}/_output/tests/openshift-e2e`, because this way the generated junit file can be indexed properly. If I use a temp dir then the path is unpredictable to the caller CI scripts.

integration/openshift/e2e/runner.sh
wainersm: Even `vim` cannot highlight it properly :D

integration/openshift/lib.sh
wainersm: BTW, @fidencio told me the other day that `apply -f url` can handle this, so I will remove this `curl` call.

integration/openshift/e2e/runner.sh
jodh-intel: Hmm. Maybe the caller could pass an "output direction parameter" to avoid this script having to make assumptions and use paths like this (which look confusing and error prone potentially to my eyes ;)

integration/openshift/e2e/run_tests.sh
jodh-intel: Agreed. If we really need both, I'd do something like the following as early in the script as possible:

```bash
[ "$OPENSHIFT_TEST_DRY_RUN" = true ] && dry_run="true" and then only refer to `$dry_run` if possible.
```

integration/openshift/e2e/runner.sh
jodh-intel: Ouch. It does if you backslash the hash though ;)

integration/openshift/e2e/runner.sh
wainersm: I think you're right @jodh-intel, it deserves a parameter for the output in the `runner.sh`

Let me take that opportunity to ask one thing related... at begin I was attempted to use `KATA_TESTS_LOGDIR` (exported in `tests/lib/common.bash`) but I got the impression test scripts don't use it. Then the other day I saw in one of your PRs it is being used but only when `CI=true`. So I am curious about what is the orientation for the use of that variable (and others KATA_TESTS_*) in the scripts.

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,77 @@
...
+set -e
+
+GOPATH=${GOPATH:-$(go env GOPATH)}
+script_dir="$(realpath $(dirname $0))"
+source "$script_dir/../../../lib/common.bash"
+tests_dir="$script_dir/../../../"
+source "$script_dir/../lib.sh"
+
+# Tests that should be included. Leave it empty to pick them all.
+filter_patterns=(
```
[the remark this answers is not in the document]
wainersm: Hi @c3d , hey, thanks for reviewing too!

> I would not call them patterns.

How would you call them? I'm really expecting grep-like (extended) patterns in quotes as you mentioned.

> 
> The problem is that there are two expansions at play here, the first one being the shell, the second one from your `grep` below. So if you have `*.znort` here, this will really be expanded by the shell first.
> 

I didn't understand the second expansion you mentioned. Because It writes the patterns to a file, then use the `grep -f` to read them from the same file.
 
> Maybe having a commended-out example of the kind of pattern you expect here? I suspect it would be grep-like patterns in quotes, i.e. `'foo*'` would not match `fooa` (not glob) but would match `foooo` (because `grep`).

Yeah, it deserves a better text explaining that.

integration/openshift/e2e/gen_test_cases.sh
wainersm: Yes, the reason is: I did a mistake. :) Gonna fix that.

integration/openshift/e2e/runner.sh
wainersm: Cluster's life is hard my friend :)

integration/openshift/e2e/runner.sh
wainersm: I made this OpenShift e2e tests support thinking in two use cases:

1. Manual testing. The person should know what she/he is doing, is likely the cluster is already configured but, to be sure, the `runner.sh` is going to do some checking. The user might have configured a `runtimeClass` with, e.g., `my-kata-runtime` and this is going to break the checking. So It was introduced the `-r` option, allowing to overwrite the default `runtimeClass` name used on the check.
1. Automated testing. In this case "the thing" (likely an CI) doesn't have the cluster full configured, and here comes the `-c` option to rescue. It is used the http://github.com/kata-containers/tests/tree/master/kata-webhook code to configure the admission controller, which has hard-coded the `runtimeClass` name to `kata`.

Therefore, if someone passes `-c` and `-r` options together then the configuration step is going to use the default `runtimeClass` name (i.e. `kata`)  and the checking is going to use whatever was passed by the user, and will fail. That's the reason why I added that test.

Makes sense?

integration/openshift/e2e/run_tests.sh
wainersm: > Who sets `OPENSHIFT_TEST_DRY_RUN`?

Oh, I forgot to initialize the variable to `false` if not set. The caller of `make openshift-e2e` can export this variable to get the tests running on dry-run mode.

ps: The `CI` variable should be initialized likewise, and I will fix that.

> 
> (asking because if some poor user sets it, they may be surprised if setting it to `1` does not work)
> 
> Also a bit confused by having both `OPENSHIFT_TEST_DRY_RUN` and `dry_run` in the scripts.

Ok, the `runner.sh` is the "backend" script which just run the tests. While developing that script I found handy to have this `dry_run` variable set (which turn on the dry-run mode), because I didn't always needed to run all the tests for some unrelated changes.

I thought that users in general could be interested on this dry-run feature as well. But now that you asked, maybe I should simply remove it...if someone finds value on it then later I can put it back. What do you think @c3d @jodh-intel ?

integration/openshift/e2e/runner.sh
jodh-intel: That variable can be used by tests if needed. I think I added it for https://github.com/kata-containers/tests/blob/master/tracing/tracing-test.sh, but no other test really needs to use the feature since most just log to `stdout`/`stderr` and/or write to the journal.

Note that Jenkins collects up all these artifacts by calling the following at the end of a test run:

- https://github.com/kata-containers/tests/blob/master/.ci/teardown.sh

integration/openshift/e2e/gen_test_cases.sh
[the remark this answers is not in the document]
c3d: > Hi @c3d , hey, thanks for reviewing too!
> 
> > I would not call them patterns.
> 
> How would you call them? I'm really expecting grep-like (extended) patterns in quotes as you mentioned.

Sorry, should have been clear. I would call them "regexps", that way you know it's not glob patterns.

> 
> > The problem is that there are two expansions at play here, the first one being the shell, the second one from your `grep` below. So if you have `*.znort` here, this will really be expanded by the shell first.
> 
> I didn't understand the second expansion you mentioned. Because It writes the patterns to a file, then use the `grep -f` to read them from the same file.

If in your script someone writes 

```
filter_patterns=(
    *.txt
)
```

then the shell first interprets that as a glob pattern, so you end up with an array containing maybe something like `a.txt`, `b.txt`, and so on.


> 
> > Maybe having a commended-out example of the kind of pattern you expect here? I suspect it would be grep-like patterns in quotes, i.e. `'foo*'` would not match `fooa` (not glob) but would match `foooo` (because `grep`).
> 
> Yeah, it deserves a better text explaining that.

integration/openshift/e2e/gen_test_cases.sh
c3d: > Yes, the reason is: I did a mistake. :) Gonna fix that.

That sounds like a good reason ;-)

integration/openshift/e2e/run_tests.sh
c3d: No objection to removing it, no objection to leaving it either.

integration/openshift/e2e/runner.sh
c3d: It does. I would expand the error message then:

```
die "Configuring the cluster uses kata-webhook, which deploys only the default runtimeClass $default_kata_runtime_class, so unable to test with $runtime_class"
```

It's a bit verbose, but at least you know where to look.

wainersm: cool, I will improve the message. Thanks @c3d

integration/openshift/lib.sh
wainersm: The idea is to return the string or error if the build failed. But you are right, it is going to get the success of `popd`. I was almost sure I had tested this function...anyway, I will fix it by checking if `$cmd` is non-empty (or `return 1`).

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,104 @@
...
+	printf "%s\n" "${exclude_patterns[@]}" > "${exclusion_file}"
+	printf "%s\n" "${skip_patterns[@]}" > "${skip_file}"
+}
+
+function cleanup()
+{
+	rm -f "${inclusion_file}" "${exclusion_file}" "${skip_file}"
+}
+
+trap cleanup EXIT
```
jodh-intel, line 64: Nit: Move this call to `setup()`.

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,104 @@
...
+
+function print_header()
+{
+	cat <<EOF
```
jodh-intel, line 68: Nit: You could use `<<-EOF` and then tab indent the here doc lines to make the function look neater ;)

integration/openshift/e2e/gen_test_cases.sh
```diff
@@ -0,0 +1,104 @@
...
+		        die "Failed to build openshift-tests"
+
+	printf "%s\n" "${filter_patterns[@]}" > "${inclusion_file}"
+	printf "%s\n" "${exclude_patterns[@]}" > "${exclusion_file}"
+	printf "%s\n" "${skip_patterns[@]}" > "${skip_file}"
+}
+
+function cleanup()
+{
+	rm -f "${inclusion_file}" "${exclusion_file}" "${skip_file}"
```
jodh-intel, line 61: Nit: What about `$tmp_tests_file`? You could use a slightly different idiom to make cleanup more reliable:

```bash
files_to_delete=()

    :

files_to_delete+=("/tmp/foo")
files_to_delete+=("/tmp/bar")
    :

[ "${#files_to_delete[@]}" -gt 0 ] && rm -f "${files_to_delete[@]}"
```

integration/openshift/e2e/runner.sh
```diff
@@ -0,0 +1,119 @@
...
+function setup()
+{
+	mkdir -p $output_dir
+
+	# If the cluster is already configured by the user then only check that the
+	# admission controller is active.
+	#
+	if [ "$configure_cluster" == "true" ]; then
+		[ "$default_kata_runtimeclass" == "$kata_runtimeclass" ] || \
+			die "To configure the cluster it should use the default runtimeClass $default_kata_runtime_class, so do not use the -c and -r options together"
```
jodh-intel, line 59: Nit: super long line ;)

jodh-intel (approved):
Thanks @wainersm.

lgtm

integration/openshift/e2e/gen_test_cases.sh
wainersm: done.

integration/openshift/e2e/runner.sh
wainersm: I tried to improve, let me know what you think.

c3d:
Remaining issues have been addressed. Merging.

PR #2895, opened 2020-09-29
debug: use latest CRI-O

liubin:


.ci/configure_crio_for_kata.sh
```diff
@@ -16,9 +16,13 @@ crio_config_dir="/etc/crio/crio.conf.d"
 runc_flag="\/usr\/local\/bin\/crio-runc"
 kata_flag="\/usr\/local\/bin\/containerd-shim-kata-v2"
 
+echo "crio --version!!!!!!!!!!!!"
```
jodh-intel, line 19: I think you missed an exclamation mark there! :smile: But seriously, all the exclamation marks might cause you problems due to the special meaning of `!!`.

liubin: Sorry, this is only a test :-)

Issue #2865, opened 2020-09-22
enable crio k8s tests


bergwolf:
We can just enable `integration/kubernetes/k8s-block-volume.bats` and `integration/kubernetes/nginx.bats` as they both pass w/o issues on k8s+crio.

bergwolf:
Turns out `k8s-block-volume.bats` is still failing randomly. Let's keep this open until that is fixed. 

See:
http://jenkins.katacontainers.io/job/kata-containers-2.0-debian-PR/292/console
http://jenkins.katacontainers.io/job/kata-containers-2.0-centos-PR/292/console

```
# (in test file k8s-block-volume.bats, line 52)
#   `kubectl exec "$pod_name" -- sh -c "apk add --no-cache e2fsprogs"' failed
# Ignoring http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz: temporary error (try again later)
# unsatisfiable constraints:
# I0929 09:27:41.195732  120111 log.go:172] (0xc000a6a160) Data frame received for 3
# I0929 09:27:41.195753  120111 log.go:172] (0xc00070a000) (3) Data frame handling
# I0929 09:27:41.195767  120111 log.go:172] (0xc00070a000) (3) Data frame sent
#   e2fsprogs (missing):
#     required by: world[e2fsprogs]
```

Might be some network issue. Maybe we should consider using ubuntu container image instead of alpine to do the test, since alpine CDN is more likely to go out of service from time to time.

ariel-adam:
@bergwolf @fidencio is there still an issue with this?

fidencio:
This one can be closed as nowadays we have the same parity between the tests running against CRI-O and containerd.

PR #2896, opened 2020-09-29
k8s: use ubuntu instead of alpine

bergwolf:
It is much more stable and includes e2fsprogs by default, thus
we do not have to install e2fsprogs on the fly which is not the purpose
of the test and might hit network issues.

Fixes: #2865

### DIFF
--- a/integration/kubernetes/runtimeclass_workloads/pod-block-pv.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-block-pv.yaml
@@ -7,7 +7,7 @@ spec:
   runtimeClassName: kata
   containers:
     - name: my-container
-      image: alpine
+      image: ubuntu:18.04
       command: ["tail", "-f", "/dev/null"]
       volumeDevices:
         - devicePath: DEVICE_PATH
--- a/integration/kubernetes/k8s-block-volume.bats
+++ b/integration/kubernetes/k8s-block-volume.bats
@@ -48,8 +48,6 @@ setup() {
 	# Verify persistent volume claim is bound
 	kubectl get pvc | grep "Bound"
 
-	# Add the e2fsprogs package to get mkfs.ext4 installed
-	kubectl exec "$pod_name" -- sh -c "apk add --no-cache e2fsprogs"
 	# make fs, mount device and write on it
 	kubectl exec "$pod_name" -- sh -c "mkfs.ext4 $ctr_dev_path"
 	ctr_mount_path="/mnt"


fidencio (commented):
@bergwolf, size difference is huge!

While the alpine imahe has 5.85MB, the ubuntu one has 66.6MB.

bergwolf:
The PR removes the line from `k8s-block-volume.bats` but still fails on the exactly same line:
`kubectl exec "$pod_name" -- sh -c "apk add --no-cache e2fsprogs"' failed`

It seems that the CI is not running with the PR commits merged. @GabyCT @chavafg Could you take a look? According to @liubin, it starts to happen since yesterday.

bergwolf:
@fidencio we are downloading several different images already. How about using `quay.io/footloose/ubuntu18.04:latest`? It is already being used by other bats so we don't need to pull again.

fidencio:
> @fidencio we are downloading several different images already. How about using `quay.io/footloose/ubuntu18.04:latest`? It is already being used by other bats so we don't need to pull again.

Works for me!

fidencio (approved):
lgtm!

bergwolf:
w.r.t. CI branch issue, it should be fixed here https://github.com/kata-containers/kata-containers/pull/822

integration/kubernetes/runtimeclass_workloads/pod-block-pv.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   runtimeClassName: kata
   containers:
     - name: my-container
-      image: alpine
+      image: quay.io/footloose/ubuntu18.04:latest
```
devimc, line 10: why this image and not the image from docker.io?

fidencio: That's due to me mentioning the size difference is huge. Then @bergwolf go to a 18.04 image we already have.
But, hey, let me change my mind. Let's use the one from docker.io, everywhere.

@bergwolf, sorry for the noise. /o\

bergwolf: because this one is already used by other bats, so that we do not have to pull yet another image.

fidencio: @bergwolf, feel free to get back to using the default one from docker.io, my bad.

Later on, in the future, we can try to standardize those. As I said, sorry for the noise.

bergwolf: nvm, I've restored the original commit ;)

devimc (approved):
thanks @bergwolf

Issue #2904, opened 2020-09-30
Patch files do not  need checking license header 


liubin:
In https://github.com/kata-containers/kata-containers/pull/817 Travis failed:

```
ERROR: Required SPDX license header check ('SPDX-License-Identifier: Apache-2.0') failed for the following files:
tools/packaging/obs-packaging/qemu-aarch64/patches/0001-arm-enable-cold-plug-nvdimm-using-DT.patch
```

PR #2905, opened 2020-09-30
ci: skip checking license header for patch files

liubin:
Patch files from other projects are not created by Kata
containers, and should not check license header in CI.

Fixes: #2904


jodh-intel (approved):
Thanks @liubin.

lgtm

lifupan (approved):

Issue #2365, opened 2020-03-06
.ci/ci-fast-return.shCI not working properly


GabyCT:
From https://github.com/kata-containers/tests/pull/2335, we can see that the fastpast script did not run.

The error that we have is
```
12:15:38 INFO: Checking for any changed files that will prevent CI fastpath return
12:15:38 INFO: Skip patterns: [-e ^CODEOWNERS$ -e ^LICENSE$ -e .*\.md -e *\.travis\.yml -e .*\.txt -e ^VERSION$ ]
12:15:38 INFO: Check patterns: []
12:15:38 INFO: No force CI check patterns
12:15:38 INFO: No force check files found
12:15:38 INFO: No files to check in CI.
12:15:38 INFO: Fastpath short circuit returning from CI.
12:15:38 INFO: Files skipped are:
12:15:38 
12:15:38 cmd/check-spelling/README.md
12:15:38 
12:15:38 INFO: Checking labels for PR kata-containers/tests/2335
12:15:38 debconf: unable to initialize frontend: Dialog
12:15:38 debconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)
12:15:38 debconf: falling back to frontend: Readline
12:15:38 debconf: unable to initialize frontend: Readline
12:15:38 debconf: (This frontend requires a controlling tty.)
12:15:38 debconf: falling back to frontend: Teletype
12:15:38 dpkg-preconfigure: unable to re-open stdin: 
12:15:41 INFO: No forcing label found
12:15:41 /tmp/jenkins/workspace/kata-containers-tests-ubuntu-18-04-PR/go/src/github.com/kata-containers/tests/.ci/ci-fast-return.sh: line 447: exit: Reading: numeric argument required
```
See http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-18-04-PR/1393/consoleFull

/cc @grahamwhaley

grahamwhaley:
hmm, that looks like a dpkg problem, and not specifically to do with the fast path... looks like one of those cases where something is requiring a real tty, and we don't have those on the CI machines...
@chavafg - can you think of anything we landed recently that might have added such a dep?
I'll have a stare at the scripts around where the lable checks are performed to see if there is anything obvious...

grahamwhaley:
[...]

PR #2908, opened 2020-10-01
Fix fast return

jodh-intel:
Use the special file descriptor created for messages when installing `jq(1)` and `yq(1)` to avoid polluting the standard streams and causing the fast return script to fail.

Also drops the `-E` from the `sudo(8)` call for SUSE for consistency with the other distro logic.

Fixes: #2365.

### DIFF
--- a/.ci/ci-fast-return.sh
+++ b/.ci/ci-fast-return.sh
@@ -68,7 +68,7 @@ local_info() {
 #  to the empty string)
 #
 read_yaml() {
-	${cidir}/install_yq.sh >&2
+	${cidir}/install_yq.sh 1>&5 2>&1
 
 	res=$(yq read "$1" "$2")
 	[ "$res" == "null" ] && res=""
--- a/.ci/ci-fast-return.sh
+++ b/.ci/ci-fast-return.sh
@@ -81,16 +81,16 @@ install_jq() {
 	package="jq"
 	case "$ID" in
 		centos|rhel)
-			sudo yum -y install "${package}"
+			sudo yum -y install "${package}" 1>&5 2>&1
 			;;
 		debian|ubuntu)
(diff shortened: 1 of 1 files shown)


jodh-intel:
Current behaviour:

```bash
$ ghprbGhRepository="kata-containers/tests" ghprbPullId=111 ./ci-fast-return.sh
INFO: Checking for any changed files that will prevent CI fastpath return
INFO: No files found
INFO: Checking labels for PR kata-containers/tests/111
INFO: No forcing label found
./ci-fast-return.sh: line 447: exit: Reading: numeric argument required
$ echo $?
2
```

Behaviour with this PR applied:

```bash
$ ghprbGhRepository="kata-containers/tests" ghprbPullId=111 ./ci-fast-return.sh
INFO: Checking for any changed files that will prevent CI fastpath return
INFO: No files found
INFO: Checking labels for PR kata-containers/tests/111
Reading package lists... Done
Building dependency tree
Reading state information... Done
jq is already the newest version (1.6-1).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
INFO: No forcing label found
$ echo $?
0
```

fidencio (approved):
lgtm!

jodh-intel:
@GabyCT, @chavafg - I'm not clear how this script actually gets called though

Forward port PR: https://github.com/kata-containers/tests/pull/2909.

jodh-intel:
@chavafg, @GabyCT - do you think we need this backported to the stable branches? It's a bug fix, but I wonder if we need to as it's just a CI optimisation?

jodh-intel:
And here - 2 acks on this PR and we can speed up the CI folks!

devimc (approved):
thanks @jodh-intel

PR #2909, opened 2020-10-01
Fix fast return

jodh-intel:
Use the special file descriptor created for messages when installing `jq(1)` and `yq(1)` to avoid polluting the standard streams and causing the fast return script to fail.

Also drops the `-E` from the `sudo(8)` call for SUSE for consistency with the other distro logic.

Fixes: #2365.

### DIFF
--- a/.ci/ci-fast-return.sh
+++ b/.ci/ci-fast-return.sh
@@ -68,7 +68,7 @@ local_info() {
 #  to the empty string)
 #
 read_yaml() {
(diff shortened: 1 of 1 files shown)


jodh-intel:
2 acks here and we can speed up the CI folks!

devimc (approved):
thanks @jodh-intel

GabyCT (approved):

Issue #2914, opened 2020-10-02
CI: Add ability to skip CI if magic label set


jodh-intel:
The CI [fast return](https://github.com/kata-containers/tests/blob/2.0-dev/.ci/ci-fast-return.sh) feature currently allows you to force the CI to run by adding a magic `force-ci` label on a PR.

However, it doesn't allow you to force the CI to be skipped in exceptional circumstances.

Add an equivalent `force-skip-ci` feature so that the CI will be skipped if the magic `force-skip-ci` label is set for a PR. This would be useful, for example, for the set of PRs which remove the old GitHub issue templates (https://github.com/kata-containers/shim/pull/271, https://github.com/kata-containers/runtime/pull/2953, https://github.com/kata-containers/proxy/pull/285, etc). Those PRs still need to be approved but there is no value in grinding through multiple CI's (which cost real money to run!) for the removal of a file like this.

PR #2915, opened 2020-10-02
CI: Add ability to skip CI run in fast-return script

jodh-intel:
Allow the CI to be skipped if the magic `force-skip-ci` label is set for a PR.

Added test for new function.

Fixes: #2914.

### DIFF
--- a/.ci/ci-fast-return.sh
+++ b/.ci/ci-fast-return.sh
@@ -12,6 +12,7 @@
 #  - file name patterns from the YAML to see if we can skip all files in the PR
 #  - If potentially skipping, looks for a 'force' label on the PR to force the CI
 #    to run anyway
+#  - Looks for a force label on the PR to always skip the CI.
 
 set -e
 
@@ -41,6 +42,9 @@ unit_yamlfile_gitname="${script_dir_base}/${unit_yamlfile_rootname}"
 
 # Name of the label, that if set on a PR, will force the CI to be run anyway.
 force_label="force-ci"
+
+# Name of the label, that if set on a PR, will force the CI to not be run.
+skip_label="force-skip-ci"
 
 # The list of files to check is held in a global, to make writing the unit test
 # code easier - otherwise we would have to pass two multi-entry lists (the list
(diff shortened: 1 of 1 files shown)


jodh-intel:
fwiw, the `force-ci` label didn't exist in any of the repos. I've now added that label and `force-skip-ci` to enable both features.

jodh-intel:
Even if this passes all the CI checks and gets approved, let's not land this until Monday "just in case" ;-) Note the new tests, plus I have tested the functionality on a real repo, but... [Warsaw's Second Law ](https://barry.warsaw.us/software/laws.html) :smile:

jodh-intel:
Note the description of the `force-skip-ci` label y'all:

> Stop the CI running for a PR (remember: if you break it, you fix it!)

.ci/ci-fast-return.sh
```diff
@@ -256,29 +317,35 @@ check_force_label() {
...
-	local json=$(curl -sL https://api.github.com/repos/${ghprbGhRepository}/issues/${ghprbPullId}/labels)
+	local result=$(is_label_set "$repo" "$pr" "$label")
+	if [ "$result" -eq 1 ]; then
+		local_info "label '$label' found"
+		echo "1"
+		return 0
+	fi
 
-	install_jq
+	local_info "label '$label' not found"
```
wainersm, line 334: In this case the label *is* found, right?

wainersm (approved):
@jodh-intel I really tried to find any problem, but I didn't :)

GabyCT (approved):

.ci/ci-fast-return.sh
```diff
@@ -252,35 +321,62 @@ check_force_label() {
...
```
jodh-intel, line 334: @wainersm - is this the line you meant? This is correct because `is_label_set()` "returns" `1` if the label is found. I've quoted returns because all the functions actually return zero but echo a string to stdout as the actual return value. That was the original design to work within the confines of `set -e` so I've retained that to keep the diff minimal.

wainersm: hi @jodh-intel - I made a comment about that line but later and quickly I noticed I said something wrong. I deleted the comment, so never mind. You change looks good to me.

PR #2901, opened 2020-09-29
k8s: Enable the 'sysctls' test for clh

likebreath:
This patch enables the last skipping k8s integration tests for clh.

Fixes: #2324

Signed-off-by: Bo Chen <chen.bo@intel.com>

PR #2930, opened 2020-10-08
ci: Add flag "privileged_without_host_devices"

### LINKED ISSUE #2929: Addd flag `privileged_without_host_devices` for kata runtime while running tests
We should be using this flag in privileged mode. Since the sysctl tests are run in privileged mode, lets enable this flag in our tests to ensure correct behaviour.

amshinde:
This flag should be used while running privileged containers
with Kata. Lets add this flag as certain tests are run in privileged
containers.

Fixes #2929

### DIFF
--- a/.ci/configure_containerd_for_kata.sh
+++ b/.ci/configure_containerd_for_kata.sh
@@ -26,6 +26,7 @@ cat << EOT | sudo tee /etc/containerd/config.toml
              Root = ""
         [plugins.cri.containerd.runtimes.kata]
            runtime_type = "io.containerd.runc.v1"
+           privileged_without_host_devices = true
            pod_annotations = ["io.kata-containers.*"]
            [plugins.cri.containerd.runtimes.kata.options]
              BinaryName = "${kata_runtime_path}"
--- a/.ci/configure_crio_for_kata.sh
+++ b/.ci/configure_crio_for_kata.sh
@@ -32,6 +32,7 @@ if [ "$minor_crio_version" -ge "18" ]; then
 default_runtime = "kata"
 [crio.runtime.runtimes.kata]
 runtime_path = "/usr/local/bin/kata-runtime"
+privileged_without_host_devices = true
 runtime_root = "/run/vc"
 runtime_type = "oci"
 EOF


egernst (approved):

.ci/install_qemu.sh
```diff
@@ -19,6 +19,8 @@ PREFIX=${PREFIX:-/usr}
 KATA_DEV_MODE="${KATA_DEV_MODE:-}"
 
 CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
+echo CURRENT_QEMU_VERSION: $CURRENT_QEMU_VERSION
```
likebreath: Nit: not sure whether you intended to include it in the patch.

.ci/install_qemu.sh
```diff
@@ -19,6 +19,7 @@ PREFIX=${PREFIX:-/usr}
 KATA_DEV_MODE="${KATA_DEV_MODE:-}"
 
 CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
+
```
egernst, line 22: nit :)

likebreath (approved):
Thanks a lot, @amshinde.  New thing learnt everyday for myself. :)

.ci/install_qemu.sh
amshinde: You caught it before it repushed :) Spurous debug from previous work. Have gotten rid of it now.

amshinde: @likebreath Can you add retest your sysctl test PR by adding a depends-on on this PR?

likebreath:
The only unexpected failing CI is: `jenkins-ci-ubuntu-1804-cloud-hypervisor-k8s-e2e-crio-minimal `. I just restarted it to see whether it is persistent or not.

@jcvenegas Can you please take a look at the failure on the `e2e-crio-mini`? IIRC, since you added it, we don't see many failures from this CI job at all.

.ci/install_qemu.sh
likebreath: @amshinde I believe we can't do depends on for the same repo (sysctl test PR #2901). So we may have to land this one first.

GabyCT (approved):

Issue #2910, opened 2020-10-01
CentOS and Debian CI failing in kata 2.0


GabyCT:
I have seen that the CentOS and Debian CI are failing in kata 2.0 both of them are showing issues in the kubernetes tests (they are using CRIO).
In the case of CentOS http://jenkins.katacontainers.io/view/Kata%202.0/job/kata-containers-2.0-tests-centos-PR/, the `ngnix` test is failing
```
05:02:40 not ok 1 Verify nginx connectivity between pods
05:02:40 # (from function `waitForProcess' in file ../../.ci/lib.sh, line 186,
05:02:40 #  in test file nginx.bats, line 41)
05:02:40 #   `waitForProcess "$wait_time" "$sleep_time" "$cmd"' failed
05:02:40 # Image is up to date for docker.io/library/busybox@sha256:2ca5e69e244d2da7368f7088ea3ad0653c3ce7aaccd0b8823d11b0d5de956002
05:02:40 # Image is up to date for docker.io/library/nginx@sha256:881169baf03885268b54eb07c673bc27f394b263cb728dfd86ff2b65b3450932
05:02:40 # INFO: k8s configured to use runtimeclass
05:02:40 # deployment.apps/nginx-deployment created
05:02:40 # deployment.apps/nginx-deployment condition met
05:02:40 # service/nginx-deployment exposed
05:02:40 # pod/test-nginx created
05:02:40 # deployment.apps "nginx-deployment" deleted
05:02:40 # service "nginx-deployment" deleted
05:02:40 # pod "test-nginx" deleted
05:02:40 Failed at 84: bats "${K8S_TEST_ENTRY}"
05:02:40 [reset] Reading configuration from the cluster...
05:02:40 [reset] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -oyaml'
05:02:40 [preflight] Running pre-flight checks
05:02:40 [reset] Removing info for node "centos8-51c4c0" from the ConfigMap "kubeadm-config" in the "kube-system" Namespace
05:02:40 W1001 10:02:40.350968  129578 removeetcdmember.go:61] [reset] failed to remove etcd member: error syncing endpoints with etc: etcdclient: no available endpoints
05:02:40 .Please manually remove this etcd member using etcdctl
05:02:40 [reset] Stopping the kubelet service
05:02:40 [reset] Unmounting mounted directories in "/var/lib/kubelet"
```

[...]

PR #2934, opened 2020-10-09
versions: Update flannel version to v0.13.0-rc2

fidencio:
Recently we've seen CI issues on both Debian and CentOS because of
iptables > 1.8 has been causing a different set of connectivity issues.

The one CI is facing is due to a known issue with flannel*, which was
pointed out by Archana Shinde and updating flannel does solve the issue
without causing any other regression.

A huge thanks to Archana for nailing down the issue.

*: https://github.com/coreos/flannel/issues/1306

Fixes: #2910

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -40,7 +40,7 @@ externals:
 
   flannel:
     url: "https://github.com/coreos/flannel"
-    version: "862c448ef28fd890e2ac4e5fddc49e7fe9693b31"
+    version: "v0.13.0-rc2"
 
   xurls:
     description: |


fidencio:
ouch! ran the whole CI by mstake, because of the " " instead of '-". :-/

fidencio:
This is related to: https://github.com/kata-containers/kata-containers/issues/868

bergwolf (approved):

liubin (approved):

jodh-intel:
Hi @liubin - I'm not sure how, but this PR seems to have been merged with four WIP checks?!?. Also, the porting labels action isn't listed - tha's supposed to be required for all PRs.

/me is confused :|

fidencio:
@jodh-intel, I've removed the "do-not-merge" label at the moment I realized the PR actually does fix the issue. So, it may quite well be my fault.
About the porting label, I also set it to "needs-backport", but that should not be the only label checked IIRC.

What I'm asking myself here is ... do we have the same rules / restriction  of the master branch to the 2.0-dev? The question is not whether we should follow the same rules, but rather whether we have some mechanism in place to enforce the same rules.

jodh-intel:
@fidencio - I found out why the porting action didn't run - your clone wasn't current, so didn't include all the expected actions.

Here are the actions expected to run:

- https://github.com/kata-containers/tests/tree/master/.github/workflows

But your clone only has:

- https://github.com/fidencio/kata-tests/tree/wip/2.0-dev-update-flannel/.github/workflows

This isn't your fault - imho it demonstrates that the GitHub actions workflow concept is flawed:

- https://github.com/kata-containers/.github/issues/20
- https://github.com/kata-containers/kata-containers/pull/688#issuecomment-694089046

If GitHub supported actions in the magic [`.github`](https://github.com/kata-containers/.github) *repo*, the problem would disappear!

As to the question about which branch the rules apply to, afaik actions trigger if they are present in the "default branch" for a repo. Since 2.0-dev is currently the default, they work. When we merge 2.0-dev into master/main, that will then contain all the actions so that too will work.

fidencio:
@jodh-intel,
 
> As to the question about which branch the rules apply to, afaik actions trigger if they are present in the "default branch" for a repo. Since 2.0-dev is currently the default, they work. When we merge 2.0-dev into master/main, that will then contain all the actions so that too will work.

2.0-dev is not the default branch for the tests repo.  master is the default one, and master will become 1.12 at some point soon.
Considering that, is it possible that we will need to extend all the triggers / actions for 2.0-dev? Sorry if the question is too basic or not to clear, I'm really not familiar with the actions.

amshinde:
@fidencio Lets port this change to master branch of tests as well.

Issue #2950, opened 2020-10-13
ci should install virtiofsd when installing qemu


bergwolf:
We are using qemu 5.0 and 5.1 for CI now. It includes virtiofsd by default. We can just install them so that we can always test virtio-fs.

ariel-adam:
Manually closing this issue since a fix was merge

PR #2951, opened 2020-10-13
ci: install virtiofsd when installing qemu

bergwolf:
So that we can use it to test virtio-fs.

Depends-on: github.com/kata-containers/kata-containers#936
Fixes: #2950

### DIFF
--- a/.ci/install_kata.sh
+++ b/.ci/install_kata.sh
@@ -41,7 +41,6 @@ case "${KATA_HYPERVISOR}" in
 	"cloud-hypervisor")
 		"${cidir}/install_cloud_hypervisor.sh"
 		echo "Installing experimental_qemu to install virtiofsd"
-		export experimental_qemu=true
 		install_qemu
 		;;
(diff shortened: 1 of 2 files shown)


fidencio (approved):
Either that, or get kata pointing to the right directory when building under CentOS.
@wainersm, maybe this is something we can improve in the future.

In any case, for now, just linking is fine lgtm!

bergwolf:
```
11:03:40 INFO: The following package will be installed: crudini
11:03:41 Install os-tree
11:03:42 Job for docker.service failed because the control process exited with error code.
11:03:42 See "systemctl status docker.service" and "journalctl -xe" for details.
11:03:43 Build step 'Execute shell' marked build as failure
11:03:43 Performing Post build task...
11:03:43 Match found for :.* : True
11:03:43 Logical operation result is TRUE
11:03:43 Running script  : #!/bin/bash
```
ARM CI failed on installing `os-tree`. Disk is full?

PR #2987, opened 2020-10-19
Cleanup Repositories for SLES

### LINKED ISSUE #2986: Cleanup package repositories for SLES env
Most of the repositories that are added to the SLES nodes in `.ci/setup_env_sles.sh` can be removed nowadays as many of the additionally required packages are meanwhile part of the "PackageHub" repositories that are offered for SLES.

rhafer:
Most of the repositories that are added to the SLES nodes can be removed
as many of the additionally required packages are meanwhile part of the
"PackageHub" repositories that are offered for SLES.
This should also speed up things a bit as only a single "zypper refresh"
call is needed now.

Fixes #2986
Signed-off-by: Ralf Haferkamp <rhafer@suse.com>

### DIFF
--- a/cmd/container-manager/manage_ctr_mgr.sh
+++ b/cmd/container-manager/manage_ctr_mgr.sh
@@ -150,8 +150,17 @@ install_docker(){
 			docker_version_full=$(apt-cache madison $pkg_name | grep "$docker_version" | awk '{print $3}' | head -1)
 			sudo -E apt-get -y install "${pkg_name}=${docker_version_full}"
 		elif [[ "$ID" =~ ^opensuse.*$ ]] || [ "$ID" == "sles" ]; then
+			# We need docker 18.06 for now, which is only provided by an older Leap 15 release
+			# Instead of adding the repo to the system (via zypper addrepo) we just install
+			# the package and its dependencies directly here. Mainly to avoid other packages
+			# from that repo creeping in.
+			leap15_repo="https://download.opensuse.org/repositories/openSUSE:/Leap:/15.0:/Update/standard/rpms/x86_64"
 			sudo zypper removelock docker
-			sudo zypper -n  install 'docker<19.03'
+			sudo zypper -n --no-gpg-checks install \
+                            $leap15_repo/docker-18.06.1_ce-lp150.5.6.1.x86_64.rpm \
+                            $leap15_repo/containerd-1.1.2-lp150.4.3.1.x86_64.rpm \
+                            $leap15_repo/docker-runc-1.0.0rc5+gitr3562_69663f0bd4b6-lp150.5.3.1.x86_64.rpm \
+                            $leap15_repo/docker-libnetwork-0.7.0.1+gitr2664_3ac297bc7fd0-lp150.3.3.1.x86_64.rpm
 			sudo zypper addlock docker
 		fi
 	elif [ "$tag" == "swarm" ]; then
--- a/.ci/setup_env_sles.sh
+++ b/.ci/setup_env_sles.sh
@@ -12,35 +12,17 @@ source "/etc/os-release" || source "/usr/lib/os-release"
 source "${cidir}/lib.sh"
 arch=$("${cidir}"/kata-arch.sh -d)
 
-echo "Add repo for perl-IPC-Run"
-perl_repo="https://download.opensuse.org/repositories/devel:languages:perl/SLE_${VERSION//-/_}/devel:languages:perl.repo"
-sudo -E zypper addrepo --no-gpgcheck ${perl_repo}
-sudo -E zypper refresh
-
-echo "Add repo for myspell"
-leap_repo="http://download.opensuse.org/update/leap/15.0/oss/"
-leap_repo_name="leap-oss"
-sudo -E zypper addrepo --no-gpgcheck ${leap_repo} ${leap_repo_name}
-sudo -E zypper refresh  ${leap_repo_name}
-
-echo "Install perl-IPC-Run"
-sudo -E zypper -n install perl-IPC-Run
-
-echo "Add repo for filesystems"
-filesystems_repo="https://download.opensuse.org/repositories/filesystems/SLE_${VERSION//-/_}/filesystems.repo"
-sudo -E zypper refresh
-sudo -E zypper -n install xfsprogs
-
-echo "Add repo for hunspell and pandoc packages"
+echo "Add PackageHub repositories for additional dependencies which are not part of the main distro"
 sudo -E SUSEConnect -p PackageHub/${VERSION_ID}/${arch}
+sudo -E zypper refresh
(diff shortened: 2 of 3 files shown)


rhafer:
I haven't fully tested this yet but the packages mentioned in the comments ("perl-IPC-Run", "myspell/hunspell", "pandoc", "xfsprogs", "redis" and "crudini") are all available in the base SLES repos or via the PackageHub.

Will give it a more complete check tomorrow.

GabyCT:
@rhafer you can test it with `/test-sles`

cmd/container-manager/manage_ctr_mgr.sh
```diff
@@ -151,7 +151,7 @@ install_docker(){
 			sudo -E apt-get -y install "${pkg_name}=${docker_version_full}"
 		elif [[ "$ID" =~ ^opensuse.*$ ]] || [ "$ID" == "sles" ]; then
 			sudo zypper removelock docker
-			sudo zypper -n  install 'docker<19.03'
+			sudo zypper -n  install docker
```
fidencio, line 154: By the moment we do this, we lose control on what's the specific version of docker being installed. @GabyCT, IIRC, we're testing using a specific version for a reason, do you remember what's the reasoning?. Are we fine bumping it there?

@rhafer, out of curiosity, what's the reason 

All in all, I'm in favour of switching to newer versions of docker, it's a step forward.

rhafer: > @rhafer, out of curiosity, what's the reason

@fidencio The main reason is that recent SLES and openSUSE releases (that is SLES 15 SP2, and Leap 15.2) only ship with a newer docker version. If you think it's better I could pin it to `19.03`. However that would still be different to what's used on the other distros.

AFAIK one of the reason to stick with a old version of docker was that the Debian/Ubuntu packages had devicemapper support removed  with 18.09. However, it seems that that change has been reverted for 19.03. So it might indeed be possible try 19.03 for other distros as well. But I guess @GabyCT has some more insight into that.

GabyCT: yes we are on that version because of the devicemapper support

fidencio: @GabyCT, what are your feelings about moving to 19.03, considering the changes were reverted for 19.03?

GabyCT: for me it is ok to move on to 19.03 but I think the best is to do across all the other distros as we have documentation and so on where we are specifying a specific docker version
/cc @chavafg

GabyCT: https://github.com/kata-containers/documentation/blob/master/install/docker/sles-docker-install.md

chavafg: I'm fine with it, but as @GabyCT mentions, it would be good to have the same docker version across all distros. Would also be good to know if this (currently skipped) test can pass now with the update: https://github.com/kata-containers/tests/blob/master/integration/stability/hypervisor_stability_kill_test.sh#L28-L29.

rhafer: > Would also be good to know if this (currently skipped) test can pass now with the update: https://github.com/kata-containers/tests/blob/master/integration/stability/hypervisor_stability_kill_test.sh#L28-L29.

Hm, that is still failing with newer docker releases :(.  However I think I have a rough understanding why that is. I'll followup in https://github.com/kata-containers/runtime/issues/2611

rhafer:
While I was able to come up with a fix for at least one of the issues with newer docker versions (https://github.com/kata-containers/runtime/issues/2611). I am still seeing failures in "integration/stability/soak_parallel_rm.sh". Sometimes there's left over files in /run/vc/sbs/ (but very different to what is seen  in  kata-containers/runtime#2611). I am trying to open a new issue for that.

Looking for a workaround until that problem is addressed I was able to dig out a older docker packages for SLE/openSUSE. So I reworked this patch to install those.

I still left the adjustments for docker 19.03 intact in the oci_calls test. If you prefer I can also remove that again.

cmd/container-manager/manage_ctr_mgr.sh
```diff
@@ -150,8 +150,17 @@ install_docker(){
...
+			# the package and its dependencies directly here. Mainly to avoid other packages
+			# from that repo creeping in.
+			leap15_repo="https://download.opensuse.org/repositories/openSUSE:/Leap:/15.0:/Update/standard/rpms/x86_64"
 			sudo zypper removelock docker
-			sudo zypper -n  install 'docker<19.03'
+			sudo zypper -n --no-gpg-checks install \
+                            $leap15_repo/docker-18.06.1_ce-lp150.5.6.1.x86_64.rpm \
+                            $leap15_repo/containerd-1.1.2-lp150.4.3.1.x86_64.rpm \
+                            $leap15_repo/docker-runc-1.0.0rc5+gitr3562_69663f0bd4b6-lp150.5.3.1.x86_64.rpm \
+                            $leap15_repo/docker-libnetwork-0.7.0.1+gitr2664_3ac297bc7fd0-lp150.3.3.1.x86_64.rpm
```
jodh-intel, line 163: Hi @rhafer - Could you try to find a way to avoid hard-coding package versions and architectures here?

rhafer: @jodh-intel The only way I could think of  avoiding that would be to add the leap-15.0  repo to this system (via `zypper ar`), but that would mean that other packages from that outdated distro might get picked up by later `zypper in` calls. Which is what I'd like to avoid so I choose to install those packages via the full url.

jodh-intel: @rhafer - I'll wait for @GabyCT to comment, but I don't see how the current code isn't going to require constant maintenance tbh as versions get updated.

rhafer: > I don't see how the current code isn't going to require constant maintenance tbh as versions get updated.

I agree that this is not really nice. But will those versions really change? Currently kata 1.x is pretty much tied to the specific docker release installed by this patch (or is that going to change?). And for kata 2.0 I guess this patch is not really relevant because it doesn't even support docker or am I wrong about that?

fidencio:
I wonder whether this one still makes sense to be merged or not, for the final 1.13.0 release.  @jodh-intel, WDYT?

fidencio:
Okay, I'm closing this one as I we're cutting the **final** 1.13.0 release soon.

Sorry this PR got stasle, and sorry for not getting it merged.  kata-containers 1.x will reach its EOL by May 12th and we'll be only focusing on kata-containers 2.x from that time onwards.

Please, check us out at https://github.com/kata-containers/kata-containers.

PR #3014, opened 2020-10-28
ci: fix VFIO CI

### LINKED ISSUE #3013: ci: fix VFIO CI
vfio POD must be deleted after checking the net interfaces otherwise the following POD won't be created because there is POD already running with the same name and yaml.

devimc:
Delete vfio POD after checking the net interfaces otherwise the following PODs
won't be created since there is already a POD with the same name and yaml.
Due to bug #3013 I was unable to detect that VFIO hotplug is not working with
QEMU+initrd. Let's skip these tests until we get a fix.

fixes #3013

### DIFF
--- a/integration/kubernetes/vfio.sh
+++ b/integration/kubernetes/vfio.sh
@@ -174,20 +174,20 @@ main() {
 		sleep 5
 	done
 
-	# Skip clh + initrd:
+	# Skip clh/QEMU + initrd:
 	# https://github.com/kata-containers/kata-containers/issues/900
 	# run_test initrd "" cloud-hypervisor false
 	# run_test initrd "" cloud-hypervisor false
+	# run_test initrd "q35" qemu false
+	# run_test initrd "q35" qemu true
+	# run_test initrd "pc" qemu false
+	# run_test initrd "pc" qemu true
 	run_test image "" cloud-hypervisor false
 	run_test image "" cloud-hypervisor true
 	run_test image "q35" qemu false
 	run_test image "q35" qemu true
-	run_test initrd "q35" qemu false
-	run_test initrd "q35" qemu true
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):

integration/kubernetes/vfio.sh
```diff
@@ -127,6 +127,8 @@ run_test() {
 	else
 		info "Success: found 2 network interfaces"
 	fi
+
+	sudo -E kubectl delete pod "${pod_name}"
```
c3d, line 131: For symmetry, wouldn't it be nicer with a

```
kubectl delete -f "${SCRIPT_DIR}/runtimeclass_workloads/vfio.yaml"
```

?

Assuming maybe some day the manifest may need to add other resources.

.ci/configure_containerd_for_kata.sh
```diff
@@ -15,6 +15,8 @@ readonly runc_path=$(command -v runc)
 sudo mkdir -p /etc/containerd/
 
 cat << EOT | sudo tee /etc/containerd/config.toml
+[debug]
+  level = "debug"
```
c3d, line 19: This looks like a separate PR to me. It's unclear how this is related to the problem described in the PR title.

c3d (approved):
lgtm but with minor suggestions

.ci/configure_containerd_for_kata.sh
devimc: yeah - I agree, adding DNM and WIP label

likebreath (approved):
@devimc Great. Thanks for locating and fixing the issue. I wonder how it was passing before this fix..

GabyCT (approved):

PR #3032, opened 2020-11-06
k8s: Skip "k8s-copy-file" tests with CRI-O

fidencio:
k8s-copy-file.bats tests regressed after bumping CRI-O version to
v1.18.4.

This has been tracked on CRI-O side`[0]`, we already have a patch waiting
for reviews, and we should enouble this test back as soon as a new
release contains the fix.

`[0]:` https://github.com/cri-o/cri-o/issues/4353

Fixes: #3033 

### DIFF
--- a/integration/kubernetes/k8s-copy-file.bats
+++ b/integration/kubernetes/k8s-copy-file.bats
@@ -7,9 +7,10 @@
 
 load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
 load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
-issue="https://github.com/kata-containers/tests/issues/2574"
+issue="https://github.com/cri-o/cri-o/issues/4353"
 
 setup() {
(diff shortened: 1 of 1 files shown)


chavafg (approved):

jodh-intel (approved):
Thanks @fidencio.

lgtm

bergwolf:
We have rolled back to v1.18.3. Is it still necessary to skip the test?

bergwolf:
sorry for chiming in late...

fidencio:
> We have rolled back to v1.18.3. Is it still necessary to skip the test?

Oh, we shouldn't then. Would be possible to get back to v1.18.4?
I'd like to just disable the test that's breaking while we actually work on a solution for the problem, instead of fully rolling out the update.  By fully rolling out the update we may face even more issues when we move to v1.18.5.

fidencio:
I've opened https://github.com/kata-containers/kata-containers/pull/1093 to revert back to v1.18.4.

Issue #3034, opened 2020-11-06
k8s: e2e conformance: fix containerd skipped tests


chavafg:
We need to add a test that requires a two node cluster to the skipped tests. We currently use a one node cluster.

PR #3035, opened 2020-11-06
k8s: Fix e2e conformance skip list for containerd

chavafg:
Added to the list a test that requires a two node cluster.
Additionally, fixed a letter in another entry which requires
to be Uppercase.

Fixes: #3034.


jodh-intel (approved):

devimc (approved):
thanks @chavafg

Issue #3039, opened 2020-11-10
Enable kubelet debug log


liubin:
Through this https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md, the default log level is `2`, set it to `4` may get more logs.

From https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/ we can see that `ClusterConfiguration` does not support pass flags to kubelet.

PR #3040, opened 2020-11-10
ci: enable debug log for kubelet

liubin:
Default log level for kubelet is 2, changes it to 4 will get more detail logs.
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

Fixes: #3039


devimc (approved):
thanks @liubin

likebreath (approved):

Issue #3041, opened 2020-11-10
k8s: re-enable "k8s-copy-file" tests for CRI-O


fidencio:
k8s-copy-file has been skipped due to a regression on CRI-O (see: https://github.com/kata-containers/tests/issues/3033).

However, now the regression has been fixed and https://github.com/kata-containers/kata-containers/pull/1095 bumps the CRI-O version to the commit fixing the issue (till an official release is out).

PR #3042, opened 2020-11-10
k8s: Re-enable "k8s-copy-file" tests for CRI-O

fidencio:
This basically reverts commit ef5cda7463416997bc051b83fc87ac3100be93ae,
as the fix for this already reached CRI-O, and CRI-I version will be
bumped on the kata-containers side.

Fixes: #3041


devimc (approved):
thanks @fidencio

likebreath (approved):

Issue #3057, opened 2020-11-13
metrics: clh: Update midval for memory footprint


likebreath:
Thanks for @devimc 's work on disabling DAX of virtio-fs when using cloud-hypervisor in PR https://github.com/kata-containers/runtime/pull/3073, the memory footprint of clh is reduced substantially. We will need to update the expected `midval` for our reference footprint of the metrics CI.

PR #3058, opened 2020-11-13
metrics: clh: Update midval for memory footprint

likebreath:
As a result of disabling the virtio-fs DAX when using cloud-hypervisor,
the expected memory footprint of kata+clh is reduced substantially. This
PR updates the reference value of memory footprint used by our metrics
CI.

Depends-on: github.com/kata-containers/runtime#3073

Fixes: #3057


devimc (approved):
thanks @likebreath - lgtm

jcvenegas (approved):

PR #3063, opened 2020-11-17
Configure to use vsock on OpenShift CI

wainersm:
This is going to build and configure with support to vsock for use on the OpenShift CI job.

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -40,6 +40,9 @@ export SHAREDIR
 
 USE_VSOCK="${USE_VSOCK:-no}"
 
+# Whether running on OpenShift CI or not.
+OPENSHIFT_CI="${OPENSHIFT_CI:-false}"
+
 runtime_config_path="${SYSCONFDIR}/kata-containers/configuration.toml"
 
 PKGDEFAULTSDIR="${SHAREDIR}/defaults/kata-containers"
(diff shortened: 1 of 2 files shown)


devimc (approved):
thanks @wainersm

.ci/install_runtime.sh
```diff
@@ -110,13 +113,16 @@ if [ "$USE_VSOCK" == "yes" ]; then
...
-	if lsmod | grep -q "$vsock_module" &> /dev/null ; then
-		echo "Module ${vsock_module} is already loaded"
-	else
-		echo "Load ${vsock_module} module"
-		sudo modprobe "${vsock_module}"
+	# On OpenShift CI the vhost module should not be loaded on build time.
+	if [ "$OPENSHIFT_CI" == "false" ]; then
+		vsock_module="vhost_vsock"
+		echo "Check if ${vsock_module} is loaded"
+		if lsmod | grep -q "$vsock_module" &> /dev/null ; then
```
jodh-intel, line 120: It would be safer to anchor that string search with something like the following (you don't need the redirection I think?):

```bash
if lsmod | egrep -q "\<vhost_vsock\>" ; then
```

Also, there is a tacit assumption that a non-priv user can run `lsmod`, but I assume that's correct for this system?

jodh-intel (approved):

.ci/install_runtime.sh
wainersm: @jodh-intel updated this PR with a 3rd commit which address those comments. Is it now ready for merge?

wainersm:
Re-triggering the Travis build...

GabyCT (approved):

PR #3069, opened 2020-11-18
test: Add k8s test to verify init containers work

### LINKED ISSUE #2825: Add test to verify init containers work
Till the latest Kata release, init containers were broken with cloud hypervisor in k8s. We should have had a test in place that verifies this. Lets add this to our existing pod test yamls for qemu and clh.
Could be as simple as just running `date` to test init containers exit cleanly. Note, useful to test this with devicemapper to make sure block devices are unplugged correctly.
```
apiVersion: v1
kind: Pod
metadata:
  name: threecontainers
spec:
  runtimeClassName: kata-clh
  initContainers:
  - name: init-myservice
    image: busybox:1.28
    command: ['date']
```

cc @GabyCT @jcvenegas @likebreath

GabyCT:
This adds a k8s test to verify that init containers are not broken with k8s.

Fixes #2825

### DIFF
--- a/integration/kubernetes/k8s-init.bats
+++ b/integration/kubernetes/k8s-init.bats
@@ -1,0 +1,29 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2020 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
+
+setup() {
+	export KUBECONFIG="$HOME/.kube/config"
+	pod_name="test-init"
+	get_pod_config_dir
+}
+
+@test "Test init containers" {
+	# Create pod
+	kubectl create -f "${pod_config_dir}/pod-init.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready pod "$pod_name"
+	cmd="printenv"
+	kubectl exec $pod_name -- sh -c $cmd | grep "HOSTNAME=$pod_name"
+}
+
+teardown() {
+	kubectl delete pod "$pod_name"
+}
--- a/integration/kubernetes/runtimeclass_workloads/pod-init.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-init.yaml
@@ -1,0 +1,20 @@
+#
+# Copyright (c) 2020 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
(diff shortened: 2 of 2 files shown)


integration/kubernetes/k8s-init.bats
```diff
@@ -0,0 +1,29 @@
...
+	pod_name="test-init"
+	get_pod_config_dir
+}
+
+@test "Test init containers" {
+	# Create pod
+	kubectl create -f "${pod_config_dir}/pod-init.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready pod "$pod_name"
```
wainersm, line 22: In case there is really a bug this can wait forever, which seems the behaviour of `kubectl wait` if `timeout` is not passed. Or does the caller of this test sets a timeout?

integration/kubernetes/runtimeclass_workloads/pod-init.yaml
```diff
@@ -0,0 +1,19 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-init
+spec:
+  runtimeClassName: kata
+  initContainers:
```
wainersm, line 12: On the Pod's definition you may want to set `restartPolicy` to `never`, this way it won't try to restart the init container over and over in case there is a bug.

I found the information about `restartPolicy`of init containers here:
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

wainersm:
Hi @GabyCT , other than the comments I did in the commit I was thinking that maybe you could have 2 init containers defined. Could it unveil more bugs on kata containers? Or just one init container is enough to stress the code?

integration/kubernetes/k8s-init.bats
GabyCT: @wainersm it has a timeout

integration/kubernetes/runtimeclass_workloads/pod-init.yaml
GabyCT: @wainersm I added to the pods definition

integration/kubernetes/k8s-init.bats
wainersm: ah, ok, thanks @GabyCT

wainersm (approved):
LGTM

amshinde:
lgtm overall @GabyCT 
But I agree with @wainersm, lets have atleast two init containers defined instead of just one to help uncover any corner cases.

devimc (approved):
thanks @GabyCT

PR #3070, opened 2020-11-18
ci: Enable minimal kubernetes-e2e tests with CRI-O

fidencio:


fidencio:
Okay, this PR adds the very same minimal tests running with containerd.  The only exception is that we're skipping `[sig-storage] ConfigMap [It] should be consumable from pods in volume with mappings as non-root with FSGroup [LinuxOnly] [NodeFeature:FSGroup]`, and an issue was filed so we can investigate what causes this to fail on CRI-O only: https://github.com/kata-containers/tests/issues/3072.

I'm not sure, thought, about running those tests on Ubuntu.  For now it's like that was it's what is used when running the containerd tests, but I'd be in favour of either using CentOS or Fedora for this job (which I can just change accordingly).

@wainersm, @mrunalp, WDYT?

fidencio:
Interesting, https://github.com/kata-containers/tests/issues/3072 is not failing on Fedora.

fidencio:
The two fails we have, on **jenkins-ci-debian-10** and **jenkins-ci-centos-8**, are related to not space left on the devices to store the logs.  I'll re-run those as soon as we fix the issue.

The newly added CI is **jenkins-ci-fedora-32-crio-k8s-minimal**, and `/test-k8s-crio-minimal` is the way to trigger it (needs to be tweaked and documented).

fidencio:
@wainersm, @jcvenegas, mind to take a look?

GabyCT (approved):

fidencio:
This is a test comment: @kata-containers/redhat, @kata-containers/power

.ci/jenkins_job_build.sh
```diff
@@ -247,6 +247,13 @@ case "${CI_JOB}" in
 		export TEST_CRIO="true"
 	fi
 	;;
+"CRIO_K8S_MINIMAL")
```
jcvenegas, line 250: Nit: I dont know if could  help you but there is a function that you can call to set all env vars with the default values.
`init_ci_flags`. Just in case there are new in the future an you want to set vars explicitly.

jcvenegas (approved):
Just one nit, but I you feel would is ok not use `init_ci_flags`, we can merge.

.ci/jenkins_job_build.sh
fidencio: @jcvenegas, Do you mind if I don't change this now, but rather have a patch adding that as part of https://github.com/kata-containers/tests/pull/3071 ?

jcvenegas: sure

PR #3089, opened 2020-11-27
integration/openshift: Run NodeConformance e2e tests only

### LINKED ISSUE #3088: openshift-ci: restrict e2e tests to NodeConformance
Currently the CI job runs the entire conformance/parallel openshift e2e tests. Most of them aren't meant to test a runtime so the aren't useful to test kata. Besides the fact that some are failing because of the admission controller which is used to mutate all pods to use the crio's runtimeclass for kata.
A sub-set of test which is more adequate is the NodeConformance suite.

wainersm:
Instead of run the entire conformance/parallel openshift tests,
restrict to only the NodeConformance ones.

So the list of tests is re-generated on this commit.

Fixes #3088


snir911:
lgtm

fidencio (approved):
lgtm,

Thanks @wainersm!

devimc (approved):
thanks @wainersm

PR #3104, opened 2020-12-09
k8s/e2e: Make cleanup() fault tolerant

### LINKED ISSUE #3101: [ci-failure] sonobouy timing out
In jenkins-ci-ubuntu-1804-cloud-hypervisor-k8s-e2e-... for kata-containers/tests:

All the issues seem to be related to sonobouy timing out.

wainersm:
Sonobuoy may crash on start up, then cleanup() is called and
it will try to access resources that doesn't exist. So this fixes
some failures on that function:
 - e2e_result_dir "unbound variable"
 - Cannot print the log file if it wasn't created
 - sonobuoy status/delete can fail

This isn't going to fix #3101 , actually it is just an improvement for when the sonobuoy fail.

### DIFF
--- a/integration/kubernetes/e2e_conformance/run.sh
+++ b/integration/kubernetes/e2e_conformance/run.sh
@@ -152,13 +152,22 @@ run_sonobuoy() {
 }
 
 cleanup() {
-	info "Results directory "${e2e_result_dir}" will not be deleted"
-	info "View results"
-	cat ${e2e_result_dir}/plugins/e2e/results/global/e2e.log
-	info "View sonobuoy status"
-	sonobuoy status
-	# Remove sonobuoy execution pods
-	sonobuoy delete
+	if [ -d "${e2e_result_dir:-}" ]; then
+		info "Results directory "${e2e_result_dir}" will not be deleted"
+		log_file="${e2e_result_dir}/plugins/e2e/results/global/e2e.log"
+		if [ -f "${log_file}" ]; then
+			info "View results"
+			cat ${log_file}
+		else
+			warn "Tests results file ${log_file} not found"
+		fi
+	fi
+	{
+		info "View sonobuoy status"
+		sonobuoy status
+		# Remove sonobuoy execution pods
+		sonobuoy delete
+	} || true
 }
 
 trap "{ cleanup; }" EXIT


integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -109,7 +109,7 @@ run_sonobuoy() {
 		fi
 	fi
 	echo "running: ${cmd}"
-	eval "${cmd}"
+	eval "${cmd}" || die "Sonobuoy run failed"
```
wainersm, line 112: Here I am not sure the script should exit or if it should try to retrieve the results. It might have failed due to timeout, in case there are unfinished tests yet.

integration/kubernetes/e2e_conformance/run.sh
```diff
@@ -152,13 +152,22 @@ run_sonobuoy() {
...
 
 cleanup() {
-	info "Results directory "${e2e_result_dir}" will not be deleted"
-	info "View results"
-	cat ${e2e_result_dir}/plugins/e2e/results/global/e2e.log
-	info "View sonobuoy status"
-	sonobuoy status
-	# Remove sonobuoy execution pods
-	sonobuoy delete
+	if [ -d "${e2e_result_dir:-}" ]; then
```
devimc, line 155: are you are allowing overwrite here but not in the next line (`log_file="${e2e_result_dir}/plugins/e2e/results/global/e2e.log"`) ?

integration/kubernetes/e2e_conformance/run.sh
devimc: yeah - I think the same, you should let it run :smile:

devimc (commented):
thanks @wainersm - I've left some comments

integration/kubernetes/e2e_conformance/run.sh
wainersm: @devimc actually the idea is not to allow overwrite but instead avoid a "unbounded variable" error. If the `sonobuoy run` fails then `e2e_result_dir` doesn't get initialized. To make it more clear, I can initialize it along side the other variables (e.g. `e2e_result_dir=""`). What do you think?

devimc: @wainersm sounds good to me, thanks

fidencio (approved):
lgtm, thanks @wainersm

integration/kubernetes/e2e_conformance/run.sh
fidencio: Ah, I prefer returning earlier in case of failure.

chavafg: I think that in case there are some failed tests, this command will return non-zero and in that case we won't be able to retrieve the results for analysis, so I would not make it `die` here.

wainersm: @chavafg good point. I confess I could not run sonobuoy so I wasn't sure a test failure would end up on the process returning non-zero code. So I will remove this statement, rebase the code and hopefully be able to merge it :)

wainersm:
@devimc @chavafg @liubin Hi folks! one more approval to get it merged. :)

liubin (approved):

PR #3109, opened 2020-12-10
test: no need to install docker

egernst:
Kata 2.0 doesn't even support docker CLI. Do we need to install? Let's
see...

Fixes: #xxxx

Signed-off-by: Eric Ernst <eric.g.ernst@gmail.com>


egernst:
we need it for building rootfs, etc. Let's go ahead and close this PR, and we'll continue to debug what is going on w/ the docker pkg install failure.

Issue #3111, opened 2020-12-11
kata-webhook crashes when checking whether a pod is privileged or not, on a namespace different than `default`


fidencio:
While checking https://github.com/kata-containers/tests/issues/3082, a crash was found.

```
2020/12/11 14:16:45 http: panic serving 10.129.0.2:42398: runtime error: invalid memory address or nil pointer dereference
goroutine 8 [running]:
net/http.(*conn).serve.func1(0xc0001f8e60)
	/usr/local/go/src/net/http/server.go:1801 +0x147
panic(0xc53b00, 0x12a8d40)
	/usr/local/go/src/runtime/panic.go:975 +0x3e9
main.annotatePodMutator(0xe19600, 0xc0002f5bc0, 0xe27420, 0xc0002cd000, 0xa6d, 0xa80, 0x3d0)
	/go/src/kata-pod-annotate/main.go:58 +0x133
kata-pod-annotate/vendor/github.com/slok/kubewebhook/pkg/webhook/mutating.MutatorFunc.Mutate(0xd73118, 0xe19600, 0xc0002f5bc0, 0xe27420, 0xc0002cd000, 0x40a15f, 0xc00001e000, 0xcdfe80)
	/go/src/kata-pod-annotate/vendor/github.com/slok/kubewebhook/pkg/webhook/mutating/mutator.go:25 +0x4e
kata-pod-annotate/vendor/github.com/slok/kubewebhook/pkg/webhook/mutating.(*staticWebhook).mutatingAdmissionReview(0xc000079080, 0xe19600, 0xc0002f5bc0, 0xc000287b60, 0xe27420, 0xc0002ccc00, 0xe27420, 0xc0002cd000, 0xc0002ccc00)
	/go/src/kata-pod-annotate/vendor/github.com/slok/kubewebhook/pkg/webhook/mutating/webhook.go:131 +0x9d
kata-pod-annotate/vendor/github.com/slok/kubewebhook/pkg/webhook/mutating.(*staticWebhook).Review(0xc000079080, 0xe19600, 0xc0002f5bc0, 0xc000287b60, 0xcc9660)
	/go/src/kata-pod-annotate/vendor/github.com/slok/kubewebhook/pkg/webhook/mutating/webhook.go:123 +0x3c8
[...]

PR #3112, opened 2020-12-11
1.x | webhook: Check whether Privileged is nil before accessing its value

fidencio:
Otherwise we'll just crash.

Fixes: #3111

Signed-off-by: Adel Zaalouk <azaalouk@redhat.com>

### DIFF
--- a/kata-webhook/main.go
+++ b/kata-webhook/main.go
@@ -50,7 +50,7 @@ func annotatePodMutator(ctx context.Context, obj metav1.Object) (bool, error) {
 	}
 
 	for i := range pod.Spec.Containers {
-		if pod.Spec.Containers[i].SecurityContext != nil {
+		if pod.Spec.Containers[i].SecurityContext != nil && pod.Spec.Containers[i].SecurityContext.Privileged != nil {
 			if *pod.Spec.Containers[i].SecurityContext.Privileged {
 				fmt.Println("privileged container: ", pod.GetNamespace(), pod.GetName())
 				return false, nil


fidencio:
The image will have to be rebuilt after this PR is merged.

liubin (approved):

kata-webhook/main.go
```diff
@@ -50,7 +50,7 @@ func annotatePodMutator(ctx context.Context, obj metav1.Object) (bool, error) {
 	}
 
 	for i := range pod.Spec.Containers {
-		if pod.Spec.Containers[i].SecurityContext != nil {
+		if pod.Spec.Containers[i].SecurityContext != nil && pod.Spec.Containers[i].SecurityContext.Privileged != nil {
```
wainersm, line 53: hi @fidencio , this change looks good. But, in reason of my ignorance on this subject, allow me to ask one question.

If I understood correctly if this webhook is applied to a non-default namespace then the `Privileged` object *might be* nil, as a result it continues to process (eventually the pod is mutated). My doubt is if the lack of `Privileged` really means the container is not privileged or whether it should be checked another object for non-default namespaces.

fidencio: @wainersm,

I think it does and the reason is here: https://pkg.go.dev/k8s.io/api/core/v1#SecurityContext
There it mentions:
```
	// Run container in privileged mode.
	// Processes in privileged containers are essentially equivalent to root on the host.
	// Defaults to false.
	// +optional
	Privileged *bool `json:"privileged,omitempty" protobuf:"varint,2,opt,name=privileged"`
```
From that I understand that unless it's explicitly set, and set to `true`, it'll be false.

Does this answer your question?

wainersm: @fidencio it does answer! Thanks!

wainersm (approved):

amshinde (approved):

Issue #3103, opened 2020-12-09
2.0 tests: docker setup failing for ubuntu jobs


egernst:
I'm seeing failures on a select number of ubuntu jobs, with the following errors:

```
Job for docker.service failed because the control process exited with error code.
See "systemctl status docker.service" and "journalctl -xe" for details.
invoke-rc.d: initscript docker, action "start" failed.
● docker.service - Docker Application Container Engine
   Loaded: loaded (/lib/systemd/system/docker.service; enabled; vendor preset: enabled)
   Active: activating (auto-restart) (Result: exit-code) since Wed 2020-12-09 03:16:56 UTC; 6ms ago
     Docs: https://docs.docker.com
  Process: 31109 ExecStart=/usr/bin/dockerd -H fd:// (code=exited, status=1/FAILURE)
 Main PID: 31109 (code=exited, status=1/FAILURE)
dpkg: error processing package docker-ce (--configure):
 installed docker-ce package post-installation script subprocess returned error exit status 1
```

This is failing for: 
`jenkins-ci-ubuntu-18-04-initrd `
`jenkins-ci-ubuntu-1804-clh-containerd-k8s-e2e-minimal`
`jenkins-ci-ubuntu-1804-containerd-k8s-e2e-minimal `

egernst:
I see this succeeding for other ubuntu job:

```
Get:1 https://download.docker.com/linux/ubuntu bionic/stable amd64 docker-ce amd64 18.06.3~ce~3-0~ubuntu [40.2 MB]
Fetched 40.2 MB in 1s (38.0 MB/s)
Selecting previously unselected package docker-ce.
(Reading database ... 
(Reading database ... 5%
(Reading database ... 10%
(Reading database ... 15%
(Reading database ... 20%
(Reading database ... 25%
(Reading database ... 30%
(Reading database ... 35%
(Reading database ... 40%
(Reading database ... 45%
(Reading database ... 50%
(Reading database ... 55%
(Reading database ... 60%
(Reading database ... 65%
(Reading database ... 70%
(Reading database ... 75%
(Reading database ... 80%
(Reading database ... 85%
(Reading database ... 90%
(Reading database ... 95%
(Reading database ... 100%
(Reading database ... 103387 files and directories currently installed.)
Preparing to unpack .../docker-ce_18.06.3~ce~3-0~ubuntu_amd64.deb ...
Unpacking docker-ce (18.06.3~ce~3-0~ubuntu) ...
[...]

PR #3115, opened 2020-12-11
ci: stopping docker.socket before remove package docker-ce

amshinde:
If system has docker already installed, then remove it
by also making sure docker.socket is stopped as well.

Fixes: #3103


liubin (approved):

wainersm (approved):
I would expect the package manager to stop the service before the package removal. Anyway, this change looks good to me.

Issue #3120, opened 2020-12-14
Switch away from DockerHub for the k8s bats tests.


fidencio:
Docker Hub has introduced a new pull rate limit that's making testing on the users' (and I believe also on the project's) side extremely hard / frustrating.

In order to avoid this, let's move away from images stored on Docker Hub and use quay.io instead.

PR #3121, opened 2020-12-14
2.x | integration,k8s: Rely on quay.io images instead of dockerhub ones

fidencio:
The reason for such change is the new Docke Hub pull rate limit,
which is hitting us quite hard.

https://www.docker.com/increase-rate-limits

Fixes: #3120

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -33,7 +33,7 @@ docker_images:
   nginx:
     description: "Proxy server for HTTP, HTTPS, SMTP, POP3 and IMAP protocols"
     url: "https://hub.docker.com/_/nginx/"
-    version: "1.17.0-alpine"
+    version: "1.15-alpine"
 
 externals:
   description: "Third-party projects used specifically for testing"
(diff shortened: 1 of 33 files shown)


liubin (approved):

versions.yaml
```diff
@@ -33,7 +33,7 @@ docker_images:
   nginx:
     description: "Proxy server for HTTP, HTTPS, SMTP, POP3 and IMAP protocols"
     url: "https://hub.docker.com/_/nginx/"
-    version: "1.17.0-alpine"
+    version: "1.15-alpine"
```
GabyCT, line 36: why we are using an older version of alpine?

fidencio: Because that's what I could find in quay.io. :-)
Is there the need to test specifically against 1.17.0-alpine of would this be okay?

GabyCT: @fidencio not that I remembered, thanks

GabyCT (approved):

amshinde (approved):

PR #3122, opened 2020-12-15
integration/kubernetes: Strength the k8s-block-volume test

### LINKED ISSUE #3123: k8s-block-volume test seems to fail intermittently
Looking at the recent failed jobs on:

http://jenkins.katacontainers.io/job/kata-containers-2.0-fedora-PR/

I noticed the jobs 615, 612, 598 failed because of integration/kubernetes/k8s-block-volume.bats. On other jobs the test passed. It might indicate a intermittent failure.

wainersm:
Wait on the created volume to be "Available" before proceed with
the remaining of the test; Also refactor the code to use variables
that hold the name of the resources.

Fixes #3123

### DIFF
--- a/integration/kubernetes/k8s-block-volume.bats
+++ b/integration/kubernetes/k8s-block-volume.bats
@@ -13,6 +13,10 @@ setup() {
 	get_pod_config_dir
 
 	pod_name="pod-block-pv"
+	volume_name="block-loop-pv"
+	volume_claim="block-loop-pvc"
+	wait_time=10
+	sleep_time=2
 	ctr_dev_path="/dev/xda"
 	vol_capacity="500M"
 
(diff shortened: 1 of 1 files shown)


amshinde (approved):

amshinde:
Thanks @wainersm, lets hope this solves the block-volume test.

integration/kubernetes/k8s-block-volume.bats
```diff
@@ -33,6 +37,8 @@ setup() {
 	sed -i "s|HOSTNAME|$(hostname)|" "$tmp_pv_yaml"
 	sed -i "s|CAPACITY|${vol_capacity}|" "$tmp_pv_yaml"
 	kubectl create -f "$tmp_pv_yaml"
+	cmd="kubectl get pv/${volume_name} | grep Available"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
```
amshinde, line 41: I was hoping that we could make use of `kubectl wait` here, but looks like it does not work for persistent volume resources.

wainersm: @amshinde same here. I tried to use `kubectl wait` but found no way to check the status is available. :(

egernst: same concern. This is at least better than what we had...

egernst (approved):

PR #3125, opened 2020-12-16
tests: Exclude *pb_test.go from SPDX header checks

dgibson:
static_check_license_headers() correctly excludes *.pb.go files, which are
generated from .proto files.  However, it doesn't exclude *pb_test.go
which are also generated (from oci.proto and health.proto).  This can cause
bogus CI failures.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -460,6 +460,7 @@ static_check_license_headers()
 			--exclude="LICENSE" \
 			--exclude="*.md" \
 			--exclude="*.pb.go" \
+			--exclude="*pb_test.go" \
 			--exclude="*.png" \
 			--exclude="*.pub" \
 			--exclude="*.service" \


liubin (approved):

GabyCT (approved):

.ci/static-checks.sh
```diff
@@ -460,6 +460,7 @@ static_check_license_headers()
 			--exclude="LICENSE" \
 			--exclude="*.md" \
 			--exclude="*.pb.go" \
+			--exclude="*pb_test.go" \
```
wainersm, line 463: Double checking it doesn't need a `.` as in `*.pb_test.go` for a accurate match.

dgibson: It does not the generated files are `src/runtime/virtcontainers/pkg/agent/protocols/grpc/healthpb_test.go` and `src/runtime/virtcontainers/pkg/agent/protocols/grpc/ocipb_test.go`

PR #3138, opened 2021-01-05
.ci/setup_env_centos.sh: Fix "too many arguments" error

### LINKED ISSUE #3137: CentOS CI setup script can fail with "too many arguments" error
In `.ci/setup_env_centos.sh`: in case the environment has already the extra packages repository set, the script will fail with "too many arguments" message.

wainersm:
In case the environment has already the extra packages repository
set, the script will fail with "too many arguments" message. Fixed
it by double-quoting the command.

Fixes #3137

### DIFF
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -22,7 +22,7 @@ fi
 echo "skip_missing_names_on_install=0" | sudo tee -a /etc/yum.conf
 
 # Check EPEL repository is enabled on CentOS
-if [ -z $(yum repolist | grep "Extra Packages") ]; then
+if [ -z "$(yum repolist | grep 'Extra Packages')" ]; then
 	echo >&2 "ERROR: EPEL repository is not enabled on CentOS."
 	# Enable EPEL repository on CentOS
 	sudo -E yum install -y wget rpm


snir911 (approved):
lgtm

.ci/setup_env_centos.sh
```diff
@@ -22,7 +22,7 @@ fi
 echo "skip_missing_names_on_install=0" | sudo tee -a /etc/yum.conf
 
 # Check EPEL repository is enabled on CentOS
-if [ -z $(yum repolist | grep "Extra Packages") ]; then
+if [ -z "$(yum repolist | grep "Extra Packages")" ]; then
```
jodh-intel, line 25: Nit: I find double sets of double quotes confusing so since the grep expression doesn't contain variables, you could make it:

```suggestion
if [ -z "$(yum repolist | grep 'Extra Packages')" ]; then
```

jodh-intel (approved):
Thanks @wainersm.

.ci/setup_env_centos.sh
wainersm: @jodh-intel yeah, feels like double negative sometimes. I accepted your suggestion on the version I just pushed.

wainersm:
Rebased the commit and forced push just to get Travis triggered again (it failed before due an error to install xurls).

Issue #3159, opened 2021-01-14
 ci: Do not install musl Rust target for s390x


Jakob-Naucke:
s390x uses a glibc Rust target since musl is not available, therefore, a GNU target should be added on s390x when installing Rust through `.ci/install_rust.sh`.

PR #3160, opened 2021-01-14
ci: Do not install musl Rust target for s390x

Jakob-Naucke:
s390x uses a glibc Rust target since musl is not available, therefore, a GNU target should be added on s390x when installing Rust.

Fixes: #3159


jodh-intel (approved):
Thanks @Jakob-Naucke!

lgtm

cmaf (approved):

fidencio:
@Jakob-Naucke, I guess we need a rebase on those as well. :-/

PR #3170, opened 2021-01-19
ci/install_runtime: export PATH on sudo env properly

### LINKED ISSUE #3174: .ci/install_runtime.sh not always picking go from /usr/local/go/bin
The .ci/install_go.sh script is going to install go into /usr/local/go/bin. If I have (or some script) exported PATH as PATH=/usr/local/go/bin:$PATH and run .ci/install_runtime.sh it isn't guaranteed it will pick up that golang version to build the runtime. That happen because whether sudo -E will preserve PATH or not depend on how sudoers security policy was configured.

wainersm:
The .ci/install_go.sh script is going to install go into /usr/local/go/bin. If
I have (or some script) exported PATH as PATH=/usr/local/go/bin:$PATH and run
.ci/install_runtime.sh it isn't guaranteed it will pick up that golang version to
build the runtime.  That happen because whether `sudo -E` will preserve PATH or not
depend on how sudoers security policy was configured. Thus, the script was changed
to pass PATH explicitly in the sudo's call.

Fixes #3174 

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -57,7 +57,7 @@ build_install_shim_v2(){
 	fi
 	pushd "$runtime_src_path"
 	make
-	sudo -E make install
+	sudo -E PATH=$PATH make install
 	popd
 }
 


wainersm:
@fidencio if I am not mistaken you hit this issue as well. am I right?

fidencio:
> @fidencio if I am not mistaken you hit this issue as well. am I right?

Yes, I did.  And I didn't understand why `sudo -E` was not preserving the PATH on my ubuntu box.

jodh-intel (approved):
Thanks @wainersm.

wainersm:
Updated the commit to include the issue number.

fidencio:
@wainersm, the change itself looks good to me.  We'd need an additional patch to actually add `/usr/local/go/bin` to the PATH, as it seems to be done only by the variables passed by the jenkins.

devimc (approved):
thanks @wainersm

wainersm:
> @wainersm, the change itself looks good to me. We'd need an additional patch to actually add `/usr/local/go/bin` to the PATH, as it seems to be done only by the variables passed by the jenkins.

I thought a little about it and the problem is that developer will lost control over the go version being used. Afaik the build scripts doesn't check for the exact version which is installed by `.ci/install_go.sh` script, so developers can still use a version which is newer than the minimum required.

What if we open an issue to keep this discussion alive? Because I am about to merge this PR :)

fidencio:
> > @wainersm, the change itself looks good to me. We'd need an additional patch to actually add `/usr/local/go/bin` to the PATH, as it seems to be done only by the variables passed by the jenkins.
> 
> I thought a little about it and the problem is that developer will lost control over the go version being used. Afaik the build scripts doesn't check for the exact version which is installed by `.ci/install_go.sh` script, so developers can still use a version which is newer than the minimum required.
> 
> What if we open an issue to keep this discussion alive? Because I am about to merge this PR :)

Oh, I gave you the wrong impression.  I was not expecting that to be solved in this PR!  Please, go ahead and merge this one.

PR #3188, opened 2021-01-22
tests: Update agent stability test to use ctr

### LINKED ISSUE #3187: Update agent stability test to use ctr for kata 2.0

GabyCT:
This PR updates the agent stability test to use ctr instead of docker.

Fixes #3187

### DIFF
--- a/integration/stability/agent_stability_test.sh
+++ b/integration/stability/agent_stability_test.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2018 Intel Corporation
+# Copyright (c) 2018-2021 Intel Corporation
 #
 # SPDX-License-Identifier: Apache-2.0
 #
@@ -15,12 +15,10 @@ cidir=$(dirname "$0")
 source "${cidir}/../../metrics/lib/common.bash"
 
 # Environment variables
-IMAGE="${IMAGE:-busybox}"
+IMAGE="${IMAGE:-quay.io/prometheus/busybox:latest}"
 CONTAINER_NAME="${CONTAINER_NAME:-test}"
 PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
 
-# Set the runtime if not set already
-RUNTIME="${RUNTIME:-kata-runtime}"
 
 # Timeout is the duration of this test (seconds)
 # We want to stress the agent for a significant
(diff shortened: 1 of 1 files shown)


devimc (approved):
thanks @GabyCT

integration/stability/agent_stability_test.sh
```diff
@@ -15,12 +15,10 @@ cidir=$(dirname "$0")
 source "${cidir}/../../metrics/lib/common.bash"
 
 # Environment variables
-IMAGE="${IMAGE:-busybox}"
+IMAGE="${IMAGE:-docker.io/library/busybox:latest}"
```
snir911, line 18: nit, maybe worth using quay.io/prometheus/busybox:latest to avoid the pull limit

fidencio: That would be preferrable, indeed, thanks @snir911!

c3d: Shouldn't there be some shared file somewhere that points to the various repos, e.g. docker, quay, ... so that we don't reinvent the wheel in every single test file?

integration/stability/agent_stability_test.sh
```diff
@@ -30,24 +28,26 @@ start_time=$(date +%s)
...
-	docker exec $CONTAINER_NAME sh -c "exit 42" || echo "Fail expected"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "echo 'hello world' > file"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "rm -rf /file"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "touch /tmp/execWorks"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "ls /tmp | grep execWorks"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "rm -rf /tmp/execWorks"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "ls /etc/foo" || echo "Fail expected"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "cat /tmp/one" || echo "Fail expected"
+	sudo ctr t exec --exec-id 1 $CONTAINER_NAME sh -c "exit 42" || echo "Fail expected"
 }
```
c3d, line 47: There is a lot of repeated code here. Why not use a variable for the part of the command that is identical every line?

(Bonus: You could leave the variant of the command that uses docker, making it easier to switch).

c3d (commented):
Can we try to factor the code a little more to make it easier to maintain?

(See suggestions in comments)

GabyCT:
changes applied

snir911 (approved):
Thanks! LGTM, I agree with c3d that it would be great to have all tests pointing to some centralized image variables, but i suppose it should be done in different issue&PR

Issue #3194, opened 2021-01-26
ci: Document and init SHIMV2_TEST env var


jcvenegas:
SHIMV2_TEST env var is exported today in some CI jobs
but not explicitly set in our CI jobs. 

Lets add it as part of our documented variables.`

PR #3195, opened 2021-01-26
ci: Init and document SHIMV2_TEST envvar.

jcvenegas:
This variable is used to enable shimv2 for some tests,
shimv2 is default in kata 2.0, lets document it and
make it explicit at init.

Fixes: #3194

Signed-off-by: Carlos Venegas <jos.c.venegas.munoz@intel.com>


snir911 (approved):
lgtm, thanks!

GabyCT (approved):

PR #3201, opened 2021-01-28
tests: Don't bother using chronic in test scripts

dgibson:
A number of steps in the test/CI setup scripts use the "chronic" tool.
This is a simple utility that suppresses output from a command, unless it
returns a failure code.  That sounds useful but:
 * Only some of the many commands in the scripts use chronic, so the logs
   are still a huge swath of information that chronic doesn't do much to
   mitigate
 * When running in a CI, it's conceivable that information from these
   commands, even when successful, could be useful for debugging a later
   failure
 * It's another external dependency, albeit a pretty easy to supply one.

I think it's more trouble than it's worth, so simply remove it.

### DIFF
--- a/.ci/install_kubernetes.sh
+++ b/.ci/install_kubernetes.sh
@@ -25,10 +25,10 @@ if [ "$ID" == "ubuntu" ] || [ "$ID" == "debian" ]; then
 	deb http://apt.kubernetes.io/ kubernetes-xenial-unstable main
 EOF"
 
-	chronic sudo -E sed -i 's/^[ \t]*//' /etc/apt/sources.list.d/kubernetes.list
+	sudo -E sed -i 's/^[ \t]*//' /etc/apt/sources.list.d/kubernetes.list
 	curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
-	chronic sudo -E apt update
-	chronic sudo -E apt install --allow-downgrades -y kubelet="$kubernetes_version" kubeadm="$kubernetes_version" kubectl="$kubernetes_version"
+	sudo -E apt update
+	sudo -E apt install --allow-downgrades -y kubelet="$kubernetes_version" kubeadm="$kubernetes_version" kubectl="$kubernetes_version"
 elif [ "$ID" == "centos" ] || [ "$ID" == "fedora" ]; then
 	url="https://packages.cloud.google.com/yum/repos/kubernetes-el7-${ARCH}"
 	echo "Install ${url} for ${ARCH}"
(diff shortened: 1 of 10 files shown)


jodh-intel:
@egernst - you might want to weigh in on this. I'm all for making the CI easier to debug but I think we introduced `chronic` at your suggestion due to the CI producing too much output?


----

wrt CI output, the ideal would be something like:

- [Split the Jenkins CI into pipelines](https://github.com/kata-containers/ci/issues/15)

  Anyone have knowledge on this / want to help implement it? It's been a long-term wishlist item :)
- Restructure all output so that we show a "pre" and "post" log message for each major operation.

  Something like:

  > INFO: installing dependency X at version Y
  > INFO: installed dependency X at version Y

- Save all command `stdout` and `stderr` (maybe as artifacts?)

Then, if the CI failed, you'd see something like:

  > INFO: installing dependency X at version Y
  >
  > <<< stdout and stderr output from operation goes here >>>
  >
  > ERROR: Failed to install dependency X at version Y

But regardless of whether it passed or failed, you'd be able to view/download all output.

cmaf:
Hi @dgibson, this branch has conflicts. Could you please resolve?

dgibson:
@cmaf, right, I didn't look at this for a long time, since there didn't seem to be much response here.

The conflicts were pretty trivial so I've resolved them and repushed.

cmaf:
@dgibson thanks! @egernst, do you have any feedback on this based on @jodh-intel's comment?

ariel-adam:
@fidencio could you take a look?

dgibson:
> @dgibson - I wonder if you could add this to the AC meeting agenda for discussion?

Honestly, I'm not sure it's worth the much trouble to me.  I proposed this because I was trying to remove dependencies from the CI and this was an easy one to remove, and I wasn't convinced the output brevity was particularly worth it.  It's such a small difference either way though that a big discussion seems like extreme bikeshedding.

jodh-intel:
> I was trying to remove dependencies from the CI

ooi, was to allow you to run the CI locally maybe?

Maybe you could rework this PR like this:

```bash
KATA_CI_FULL_OUTPUT="${KATA_CI_FULL_OUTPUT:-}"

HIDE=chronic

[ -n "$KATA_CI_FULL_OUTPUT" ] && HIDE=
[ -n "$HIDE" ] && command -v "$HIDE" &>/dev/null || die "need $HIDE"

$HIDE sudo dnf blah

   :

$HIDE sudo zypper blah

   :

$HIDE even more commands to run

   :
```

That way, if you were running locally, you could dynamically remove the need to use `chronic`.

In fact, until we can restructure the CI to make it easier to debug failures we could add in extra [trigger phrase](https://github.com/kata-containers/community/wiki/Controlling-the-CI) to allow that variable to be specified via a GH comment. Something like:

```
/test-ubuntu-full-output
```

/cc @GabyCT, @chavafg, @egernst.

jodh-intel:
Alternatively, if you're not in favour of that, could you either rebase to resolve the conflicts or close?

Issue #3206, opened 2021-02-02
xurls URL has changed


jodh-intel:
The `xurls` tool we use in the CI's URL has changed:

```bash
$ curl -sIL https://mvdan.cc/xurls/cmd/xurls | head -1
HTTP/2 404
$ curl -sIL https://github.com/mvdan/xurls | head -1
HTTP/2 200
```

PR #3207, opened 2021-02-02
CI: Fix xurls url

jodh-intel:
The `xurls` tool used by the `static-checks.sh` script to look for
broken URLs has changed its URL (oh the irony!), so update the versions
database and switch to using the latest version.

Fixes: #3206.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -570,13 +570,12 @@ static_check_docs()
 
 		local version
 		local url
-		local dir
 
 		version=$(get_test_version "externals.xurls.version")
 		url=$(get_test_version "externals.xurls.url")
-		dir=$(echo "$url"|sed 's!https://!!g')
 
-		build_version "${dir}" "" "${version}"
+		# xurls is very fussy about how it's built.
+		GO111MODULE=on go get "${url}@${version}"
 	fi
 
 	info "Checking documentation"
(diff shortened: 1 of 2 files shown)


fidencio:
@jodh-intel, doesn't look like we're running a really small amount of CIs here?
I think something was done wrong by @fidencio, when dealing with `2.0-dev` -> `main` convertion.

fidencio:
> @jodh-intel, doesn't look like we're running a really small amount of CIs here?
> I think something was done wrong by @fidencio, when dealing with `2.0-dev` -> `main` convertion.

Interesting, on the jenkins side everything seems to be correct. I'll run a `/test` here.

fidencio:
Ah, now I see, it was just showing the required ones, okay, now it makes sense.

fidencio (approved):
lgtm, thanks @jodh-intel!

wainersm (approved):
Indeed an irony, @jodh-intel !

This change LGTM and I don't want to hold it since the checker has failed. But I wonder if we couldn't get xurls packaged from the distros. Idea that could be applied to other tools that currently we fetch & build (error prone). Worth an issue?

devimc (approved):

jodh-intel:
@wainersm - that's a good question. The problem with doing that I think is we may end up installing the distro version of golang. But that tends to be too old so would cause other issues).

GabyCT (approved):

PR #3211, opened 2021-02-03
tests: Add some kubectl describe information to teardown of many k8s …

### LINKED ISSUE #3210: CI failures on timeout waiting for k8s pods to be ready give no useful debugging
# Description of problem

A fairly likely failure mode for Kata in a Kubernetes context is for Kata pods to be created, but never become ready.  In this situation many of the k8s tests in CI will simply timeout in `kubectl wait` with no indication of why the pod might not be ready.
(replace this text with the list of steps you followed).  This can make CI failures extremely frustrating to debug.

dgibson:
…tests

When kubernetes tests in the CI fail, it can be difficult to work out why.
In particular, one fairly likely failure mode is for a pod to be created
but never become ready.  If this happens, the CI output will show a timeout
on the kubectl wait command, but no indication of why the pod didn't become
ready.

This adds a kubectl describe to the teardown() path for many of the tests,
which will give at least a little more indication of where to look for a
problem.  Technically the debug command will be executed for both passing
and failing tests, however bats will suppress the output for passing tests,
so it shouldn't clutter the output except in cases where it is likely to be
useful.

This is by no means attempting to add exhaustive debugging information to
all the tests, just a basic start for a good selection.  Further debug
instrumentation in the teardown paths seems likely for future improvements
to debugability.

fixes #3210

### DIFF
--- a/integration/kubernetes/k8s-attach-handlers.bats
+++ b/integration/kubernetes/k8s-attach-handlers.bats
@@ -36,6 +36,9 @@ setup() {
 }
 
 teardown(){
+	# Debugging information
+	kubectl describe "pod/$pod_name"
+
 	rm -f "${pod_config_dir}/test-lifecycle-events.yaml"
 	kubectl delete pod "$pod_name"
 }
--- a/integration/kubernetes/k8s-block-volume.bats
+++ b/integration/kubernetes/k8s-block-volume.bats
@@ -65,6 +65,9 @@ setup() {
 }
 
 teardown() {
+	# Debugging information
+	kubectl describe "pod/$pod_name"
+
 	# Delete k8s resources
 	kubectl delete pod "$pod_name"
 	kubectl delete pvc "$volume_claim"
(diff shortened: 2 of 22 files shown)


fidencio:
@GabyCT, @wainersm, mind to take a look?

fidencio (approved):
@dgibson, thanks for the contribution!

This is something that's quite useful for us, indeed.  It'll increase the amount of our logs, though, which is fine by me, but I've asked Gaby's and Wainer's input as well.

fidencio:
@dgibson, on a side note, about the ubuntu-ci failure, this is being tracked [here](https://github.com/kata-containers/tests/issues/3208) and @GabyCT is already debugging it.

Please, subscribe to the issue in order to get updates, then we'll be able to retrigger the job and have it passing here as well.

jodh-intel (approved):

GabyCT (approved):

Issue #2479, opened 2020-04-26
Update build instructions for kata-webhook


fidencio:
# Description of problem

There's no way to build kata-webhook following the instructions provided in https://github.com/kata-containers/tests/blob/master/kata-webhook/README.md

# Expected result

Instructions which allow the users to build kata-webhook without failures;

# Actual result

Following the instructions you'll face:
```
fidencio@laerte ~/go/src/github.com/kata-containers/tests/kata-webhook $ docker build -t katadocker/kata-webhook-example:latest .
Sending build context to Docker daemon  13.82kB
Step 1/7 : FROM golang:latest AS builder
 ---> 2421885b04da
Step 2/7 : WORKDIR /go/src/kata-pod-annotate
 ---> Using cache
 ---> 2fbde962f238
Step 3/7 : COPY . ./
 ---> Using cache
 ---> 20cf32a11a54
Step 4/7 : RUN CGO_ENABLED=0 go build -o /go/bin/kata-pod-annotate
 ---> Running in a90ab84e36bc
main.go:18:2: cannot find package "github.com/slok/kubewebhook/pkg/http" in any of:
	/usr/local/go/src/github.com/slok/kubewebhook/pkg/http (from $GOROOT)
	/go/src/github.com/slok/kubewebhook/pkg/http (from $GOPATH)
main.go:19:2: cannot find package "github.com/slok/kubewebhook/pkg/log" in any of:
	/usr/local/go/src/github.com/slok/kubewebhook/pkg/log (from $GOROOT)
	/go/src/github.com/slok/kubewebhook/pkg/log (from $GOPATH)
main.go:20:2: cannot find package "github.com/slok/kubewebhook/pkg/webhook/context" in any of:
	/usr/local/go/src/github.com/slok/kubewebhook/pkg/webhook/context (from $GOROOT)
	/go/src/github.com/slok/kubewebhook/pkg/webhook/context (from $GOPATH)
main.go:21:2: cannot find package "github.com/slok/kubewebhook/pkg/webhook/mutating" in any of:
	/usr/local/go/src/github.com/slok/kubewebhook/pkg/webhook/mutating (from $GOROOT)
	/go/src/github.com/slok/kubewebhook/pkg/webhook/mutating (from $GOPATH)
main.go:15:2: cannot find package "k8s.io/api/core/v1" in any of:
	/usr/local/go/src/k8s.io/api/core/v1 (from $GOROOT)
[...]

PR #3229, opened 2021-02-09
kata-webhook: simplify building kata-webhook container

jensfr:
Before one can build the container image the vendor directory from the
parent directory has to be copied to the kata-webhook directory.
Otherwise the build will fail because it can't find all necessary
packages. See issue https://github.com/kata-containers/tests/issues/2479

This is solved on the master branch for Kata 1.x by adding a manual step
to the README.md.

We can avoid this manual step  by copying in the vendor directory during
the docker build.

The vendor directory is ~28 megs and it takes less than a second to copy
it into the build env, so no noticeable slow down.

When this is accepted I can send a similar PR for the master branch.

Fixes: #3230
Signed-off-by: Jens Freimann <jfreimann@redhat.com>

### DIFF
--- a/kata-webhook/Dockerfile
+++ b/kata-webhook/Dockerfile
@@ -5,6 +5,7 @@ FROM golang:latest AS builder
 
 WORKDIR /go/src/kata-pod-annotate
 
+COPY ../vendor ./
 COPY . ./
 RUN CGO_ENABLED=0 go build -o /go/bin/kata-pod-annotate
 


fidencio (approved):
lgtm, thanks @jensfr!

fidencio:
Mind that you'll need to create a new issue and add a `Fixes: #NNN` to your patch.

jensfr:
@fidencio thanks! created issue #3230

GabyCT (approved):

Issue #3241, opened 2021-02-12
test: docker: Adjust the cpuset test


likebreath:
As we are fixing the `--cpuset-cpus` support in kata with docker (https://github.com/kata-containers/runtime/pull/3138), it exposes the [`cpuset-cpus` test](https://github.com/kata-containers/tests/blob/9b9e93c1caca28968742f9a38aa9e723f6c84e16/integration/docker/cpu_test.go#L193) is actually not testing the expected behavior of kata. By design, Kata does not constrain the guest vCPUs at all, because we don't track the mappings between guest vCPU and host CPUs. As a result, we are always allowing containers to use all onlined vCPUs in the guest. The current `cpuset-cpus` assumes the guest cgroups constrain the vCPU in the exactly same way as the CPUs are constrained on the host. That is wrong.

likebreath:
Fixed by #3242

PR #3242, opened 2021-02-12
test: docker: Adjust the `cpuset-cpus test`

likebreath:
As we are fixing the `--cpuset-cpus` support in kata with
docker (kata-containers/runtime#3138), it exposes the `cpuset-cpus test`
is actually not testing the expected behavior of kata. By design, Kata
does not constrain the guest vCPUs at all, because we don't track the
mappings between guest vCPU and host CPUs. As a result, we are always
allowing containers to use all onlined vCPUs in the guest. The current
cpuset-cpus assumes the guest cgroups constrain the vCPU in the exactly
same way as the CPUs are constrained on the host. That is wrong.

This patch adjusts the `cpuset-cpus test` to respect the actual behavior
of kata with `docker --cpuset-cpus`.

Depends-on: github.com/kata-containers/runtime#3138

Fixes: #3241

### DIFF
--- a/integration/docker/cpu_test.go
+++ b/integration/docker/cpu_test.go
@@ -152,7 +152,12 @@ var _ = Describe("CPU constraints", func() {
 		quota      int = 20000
 		period     int = 15000
 		cpusetCpus int = 0
-		cpusetMems int = 0
+		// We are now not constraining the guest cgroup/cpuset
+		// values at all, e.g. we are allowing all onlined vCPUs
+		// to be used by the container. In this test, we use the
+		// hard-coded value below assuming the default vCPU is '1'
+		cpusetCpusExpected string = "0-1"
+		cpusetMems         int    = 0
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


integration/docker/cpu_test.go
```diff
@@ -152,7 +152,12 @@ var _ = Describe("CPU constraints", func() {
 		quota      int = 20000
 		period     int = 15000
 		cpusetCpus int = 0
-		cpusetMems int = 0
+		// We are now not constraining the guest cgroup/cpuset
+		// values at all, e.g. we are allowing all onlined vCPUs
+		// to be used by the container. In this test, we use the
+		// hard-coded value below assuming the default vCPU is '1'
+		cpusetCpusExpected string = "0-1"
```
egernst, line 159: if we aren't ever setting, would it be easier to just not check the cpuset cpus, and not need expected values?

egernst (approved):

jcvenegas (approved):

cmaf (approved):
@likebreath lgtm

likebreath:
Except the failures on the `jenkins-ci-ubuntu-1804-firecracker` and `jenkins-ci-cri-containerd `, all other CI failures should not be related to the changes in the PR. 

The firecracker CI is failing as it is missing CPU hotplug functionality, and I will send a follow-up PR to disable the cpuset docker test for firecracker. Also, we will need another PR to fix the failure on the `jenkins-ci-cri-containerd` CI.

Merging this PR now.

Issue #3245, opened 2021-02-16
ci: fix scripts for centos 8


alicefr:
# Description of problem

The are several small errors to allow the `jenkins_job_build.sh` running on Centos 8. One of the main problems is that the wanted docker version is `18.04`, and this version is not available on Centos 8. The available versions are:
```bash
yum --showduplicate list docker-ce
Last metadata expiration check: 0:08:42 ago on Tue 16 Feb 2021 06:13:14 AM EST.
Available Packages
docker-ce.x86_64                                           3:19.03.13-3.el8                                            docker-ce-stable
docker-ce.x86_64                                           3:19.03.14-3.el8                                            docker-ce-stable
docker-ce.x86_64                                           3:19.03.15-3.el8                                            docker-ce-stable
docker-ce.x86_64                                           3:20.10.0-3.el8                                             docker-ce-stable
docker-ce.x86_64                                           3:20.10.1-3.el8                                             docker-ce-stable
docker-ce.x86_64                                           3:20.10.2-3.el8                                             docker-ce-stable
docker-ce.x86_64                                           3:20.10.3-3.el8                               
```

For opensuse/sles the docker version is set to `19.03` ([here](https://github.com/kata-containers/tests/blob/f4b8a8de9515843473266c41fe2753b2d0f663c5/cmd/container-manager/manage_ctr_mgr.sh#L151)).

# Expected result
Running without errors the Ci scripts on Centos 8

alicefr:
cc @fidencio

fidencio:
@GabyCT, what do you think?  Shall we consider a switch to `19.03` everywhere else?

GabyCT:
@fidencio for me it is better if we have one docker version across all the CI's in order to avoid random issues maybe caused for the different versions, wdyt @chavafg ? I believe that the only reason why we are using `18.04` is because it is supporting devicemapper so that is something that you need to take in consideration

alicefr:
@GabyCT is devicemapper still relevant? Docker has deprecated devicemapper support and podman/crio don't support it.

GabyCT:
@alicefr I believe it is, but maybe it is something that should be discussed on the AC meeting
/cc @chavafg

chavafg:
My understanding is that cri-o supports devicemapper, we also use it as part of the integration with Firecracker, and in some cases it brings better performance in Kata. I think this is still relevant, but as @GabyCT says, we can discuss in AC meeting if needed.

fidencio:
[...]

PR #3246, opened 2021-02-16
ci: fix scripts for centos 8

alicefr:
Fix various issues to be able to run `jenkins_job_build.sh` on centos 8. Right now, I hardcoded the docker version, until we decide which is the right one to pick. For more details check  #3245

cc @fidencio

### DIFF
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -75,6 +75,7 @@ declare -A packages=( \
 	[haveged]="haveged" \
 	[libsystemd]="systemd-devel" \
 	[redis]="redis" \
+	[make]="make" \
 )
 
 main()
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -50,7 +50,7 @@ echo "Install chronic"
 sudo -E yum -y install moreutils
 
 if [ "$centos_version" == "8" ]; then
-	chronic sudo -E yum install pkgconf-pkg-config
+	chronic sudo -E yum install -y pkgconf-pkg-config
 fi 
 
 declare -A minimal_packages=( \


chavafg:
@alicefr can you please address conflicts? Just merged #3272 from @fidencio to not use a specific old docker version, so that part may be fixed for centos.

alicefr:
Done! After @fidencio it  turned out to be one line fix :)

fidencio (approved):
thanks, @alicefr!

fidencio:
@chavafg, I re-enabled the CentOS CI via jenkins, but I still can't trigger it here.
If you happen to have some cycles later Today, let me know and I can learn what I've done wrong there.

c3d (approved):
/lgtm

chavafg:
let's try to run centos CI

/test-centos

@fidencio, looked at the configuration and it seems correct.

chavafg:
triggered for now, let's see how this runs. @fidencio later, you can try again and see if you are now able to trigger the job,

alicefr:
so the Centos job is running until the k8s tests and then fails: http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-centos-PR/301/console. I could reproduce that locally

GabyCT (approved):

GabyCT:
@fidencio, @alicefr it seems that CentOS CI is still failing in a k8s test, maybe you can skip the test and create an issue in order to avoid other PRs to be failing due to the CentOS CI

fidencio:
@GabyCT, the failures are really non deterministic.  It seems the machine is just slow and we're failing on different tests because of that.

GabyCT:
> @GabyCT, the failures are really non deterministic. It seems the machine is just slow and we're failing on different tests because of that.

so maybe we can merge this change and see how it behaves the CentOS CI (how stable it is) and if it continues to be failing then we should disable it in order to have the CI failing in each PR, wdyt @chavafg @fidencio  @alicefr

fidencio:
@GabyCT, I've increased the only timeout I was able to see as part of the configurations:
Captura de pantalla de 2021-03-17 21-33-30

Is that the right thing?  In any case, CI passed now.

GabyCT:
ok thanks @fidencio , let's merge this and see how it behaves with the modifications

Issue #3119, opened 2020-12-14
CI: cache qemu, kernel, image for 2.x


wainersm:
The ci install scripts for 1.x look for built and cached qemu, kernel and image files in http://jenkins.katacontainers.io. In case of cache miss those artifacts are built on job time. 

For 2.x it doesn't have the jobs on jenkins to build and cache those artifacts, thus they are built for every CI job. This issue is about enabling such as caching mechanism to 2.x as well.

wainersm:
@GabyCT @chavafg Fix if I am wrong but current cached artifacts can't be used with 2.x because their versions can diverge.

wainersm:
Reopening because I still need to address qemu.

wainersm:
I will leave it open until we don't have images (osbuilder) cached too.

GabyCT:
I think we can close this issue, the images is still an issue as it is not easy to solve

PR #3262, opened 2021-02-19
DNM: Don't build qemu from the CI scripts

dgibson:
Experiments in removing the time consuming and complex qemu build from the CI scripts.


fidencio:
@dgibson, this approach is not going to fly.  Kata Containers project is quite opinionated about the VMM, and image being used.
The right approach here is to ensure we have the latest QEMU, using our config, cached somewhere and we just pull it from "somehwere" rather than installing the distro version.
/cc @egernst @wainersm @alicefr

fidencio:
> @dgibson, this approach is not going to fly. Kata Containers project is quite opinionated about the VMM, and image being used.
> The right approach here is to ensure we have the latest QEMU, using our config, cached somewhere and we just pull it from "somehwere" rather than installing the distro version.
> /cc @egernst @wainersm @alicefr

Which, by the way, we already used to have for 1.x but was not carried over for 2.x.  That's something really needed, IMHO.

alicefr:
@fidencio do you plan to make the qemu binaries available also for other architectures then x86? 
I think it will be nice if there is a mechanism to tell the CI to build qemu (and the kernel) for certain PRs. For example, when you change something in the kernel config you need to bump the [version](https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kernel/kata_config_version). Maybe, with something similar, if the version has changed, you could trigger the qemu build, otherwise fetch the cache version. In this way, it allow the CI to test eventual changes in qemu

fidencio:
> @fidencio do you plan to make the qemu binaries available also for other architectures then x86?
> I think it will be nice if there is a mechanism to tell the CI to build qemu (and the kernel) for certain PRs. For example, when you change something in the kernel config you need to bump the [version](https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kernel/kata_config_version). Maybe, with something similar, if the version has changed, you could trigger the qemu build, otherwise fetch the cache version. In this way, this allows the CI to test eventual changes in qemu

That's the idea!

Theoretically we should only rebuild the VMMs & Kernel when there's a change on those, and QEMU & kernel are the most time consuming ones.  Agent (and consequently the initrd / image) could also be cached, but this one would be tricky and I sincerely don't think it's worth it, at least not as part of a first try.

@wainersm, do you remember whether we had an issue for this one? /cc @jcvenegas and @GabyCT as well.

dgibson:
So, I see a bunch of (ludicrously complicated) logic for downloading a cached qemu build, but it seems to be rarely if ever actually avoiding the build for CI runs.

I realize this won't work as it stands, which is why I'm experimenting here.  But I don't see any fundamental reason for Kata to be so opinionated about the VMM, my plan is to try to use the distro version and see how that works.  Or if it doesn't, at least that will let me understand *why* Kata is so opinionated and we can look to fixing that.

fidencio:
> So, I see a bunch of (ludicrously complicated) logic for downloading a cached qemu build, but it seems to be rarely if ever actually avoiding the build for CI runs.

As I mentioned before in https://github.com/kata-containers/tests/pull/3262#issuecomment-781874802, that's something we used to do for 1.x repos, but we stopped doing for 2.x but we should get back to it.

> I realize this won't work as it stands, which is why I'm experimenting here. But I don't see any fundamental reason for Kata to be so opinionated about the VMM, my plan is to try to use the distro version and see how that works. Or if it doesn't, at least that will let me understand _why_ Kata is so opinionated and we can look to fixing that.

In order to test something, IMHO, we need at least a few stable pieces.
Testing Kata Containers with Fedora QEMU, Ubuntu QEMU, CentOS QEMU, just will not provide the same results.  That's one of the reasons I see why we stick to one QEMU version, rather than dropping support to this or that distro because their policy for packages doesn't meet the bare minimum we test.

Also, please, mind that I've raised the idea of relying only on the distro's packages last year, that would simplify the work for the Kata Containers developers, but here I'm forwarding the message that I got, and that made me understand the "why".

Now, one thing is one thing and another thing is another thing,  if you want an option to skip QEMU installation as part of the scripts, as you're running them locally, all power to you and, please, provide the patch and I do see that as something I'd happily review and accept.  The same goes for the kernel, by the way.  However, in the CI environment, we'd be better not having those pieces moving according to the distro.

wainersm:
Building QEMU is really cumbersome but as @fidencio pointed out at least we have a stable piece to test against regardless the of distro.

One thing that's really needed is the caching working for 2.x, tracked in https://github.com/kata-containers/tests/issues/3119 . I already started the work (as you can see on the issue) but I was trying to automate the CI job completely. Wrong approach, maybe I could have created the job manually as a temporary solution.

Besides that, the QEMU build scripts should be improved. I am trying that as I work on bump to QEMU 2.0 (https://github.com/kata-containers/kata-containers/pull/1349). Just now I noticed that if `AGENT_INIT=true` then QEMU is built with instructions from `.ci/install_kata_qemu.sh`, instead of using the scripts from `kata-containers/tools/packaging`, therefore, need to refactor that too.

dgibson:
> > So, I see a bunch of (ludicrously complicated) logic for downloading a cached qemu build, but it seems to be rarely if ever actually avoiding the build for CI runs.
> 
> As I mentioned before in [#3262 (comment)](https://github.com/kata-containers/tests/pull/3262#issuecomment-781874802), that's something we used to do for 1.x repos, that we stopped doing for 2.x but we should get back to it.

Ok, I'm not really clear on what "stopped doing" means here.  There is certainly a bunch of caching logic in the scripts as used with Kata 2.x right now.  See `.ci/install_qemu.sh` in the tests repo, specifically.  I do rather suspect it's not working in most cases.  Certainly we seem to be rebuilding in all cases on the Ubuntu CI runs.  We might not be on the Debian run, at least I couldn't spot a qemu rebuild on one example of that I looked at recently.

dgibson:
> In order to test something, IMHO, we need at least a few stable pieces.

So, the problem I have with this argument is that there are several different things that "stable" can mean.

1) "Stable" meaning we use the same qemu version from one run to another of the same testset

Absolutely we should do that - we're looking for regressions in Kata, not regressions in qemu.

But... using the distro version still accomplishes that, since we're working with a fixed distro version.  Arguably better than the cached build logic, since that just grabs whatever the latest nightly build is, and only compares it against the "correct" version with a version number check, which is a pretty coarse instrument.

The fact that the cached build also has a fallback to a local build also means we have at least two different potential build paths, which need to be kept in sync.

2) "Stable" meaning we should use a generally reliable and bug-free qemu version

Again, no dispute there - but a distro tested version should again supply that at least as well as a built-from-source version.

3) "Stable" meaning we use the same qemu version across each different CI testset

This one I will dispute.  If Kata is fragile against fairly minor changes in qemu versions, that's a bad thing - it makes life harder for system integrators, if not for end users.  Therefore using the relevant distro version for each of our testsets *improves* the CI's coverage: if it fails on some distro due to how we invoke qemu, we should probably be avoiding doing that.

fidencio:
> Ok, I'm not really clear on what "stopped doing" means here.

So, when we think about caching something, the first question that comes to my mind is "where?".
Looking at the place where we could use to cache something (our jenkins), one can find: http://jenkins.katacontainers.io/job/qemu-nightly-x86_64/.  Going further, looking into what that job does, it does build qemu, every night, based on `kata-containers/runtime` repo.

Something like that was never worked out, never done, for 2.x repo.

Does this make things a little bit more clear?

My understanding is that we "stopped doing" the build not because it was a well thought decision to not do so, but rather a non desired side-effect of the change.

> There is certainly a bunch of caching logic in the scripts as used with Kata 2.x right now. See `.ci/install_qemu.sh` in the tests repo, specifically.

I guess there are a few things to cover here. :-).
1. https://github.com/kata-containers/tests/blob/a8d1ff390f22d94d02c08e40194be58c3bf1108b/.ci/install_qemu.sh#L144
```
			elif [ "$cached_qemu_version" == "$CURRENT_QEMU_VERSION" ]; then
				# If installing cached QEMU fails,
				# then build and install it from sources.
				install_cached_qemu || build_and_install_static_qemu
```
QEMU versions on 1.x and 2.x will differ and that's the reason we end up building QEMU every single time.


2. https://github.com/kata-containers/tests/blob/a8d1ff390f22d94d02c08e40194be58c3bf1108b/.ci/install_qemu.sh#L30
```
qemu_latest_build_url="${jenkins_url}/job/qemu-nightly-$(uname -m)/${cached_artifacts_path}"
```
That points exactly to the QEMU built from `kata-containers/runtime` repo.  So, 1.x.

3. A bunch of things need a good clean-up.
AFAIR, when 2.x started being developed, everything was just got "as they were" and moved there.  That was the decision made at that time, and regardless of what we think about it now, that was the decision made at that time.

Such decision led us to have a bunch of code that we don't use, but we still maintain.  One may argue it's easier to keep things "closer to each other", I don't agree.  One may argue the project had other priorities and those clean ups were left behind, and with that I agree, I also dislike the situation, but I agree with the statement.

Right now, I sincerely don't know if it's easier to remove all the references to the cached qemu and avoid folks getting confused with them, or just wait till @wainersm gets some free cycles and have things being cached for 2.x as well.  I sincerely don't know, but I'd suggest to leave that decision to @wainersm, or whoever will be doing this work.  Or, IOW, I have my opinion, but as I don't plan on working on this now, I simply trust whoever will work on this will now better what to do.

> I do rather suspect it's not working in most cases. Certainly we seem to be rebuilding in all cases on the Ubuntu CI runs. We might not be on the Debian run, at least I couldn't spot a qemu rebuild on one example of that I looked at recently.

If it works in any of the 2.x CI, it'd be a surprise for me, based on all said above. /o\.


And now I realise how extensive this reply is, @dgibson, and I hope you don't mind it. :-)
All in all, with all things above mentioned, that's the reason I looped @wainersm into this issue.

Are things clear for you?  If not, please, let us know!

fidencio:
@dgibson,

> > In order to test something, IMHO, we need at least a few stable pieces.
> 
> So, the problem I have with this argument is that there are several different things that "stable" can mean.
> 
>     1. "Stable" meaning we use the same qemu version from one run to another of the same testset
> 
> 
> Absolutely we should do that - we're looking for regressions in Kata, not regressions in qemu.

We're on the same page here.

> 
> But... using the distro version still accomplishes that, since we're working with a fixed distro version. Arguably better than the cached build logic, since that just grabs whatever the latest nightly build is, and only compares it against the "correct" version with a version number check, which is a pretty coarse instrument.
> 
> The fact that the cached build also has a fallback to a local build also means we have at least two different potential build paths, which need to be kept in sync.
> 
>     1. "Stable" meaning we should use a generally reliable and bug-free qemu version
> 
> 
> Again, no dispute there - but a distro tested version should again supply that at least as well as a built-from-source version.

This leads to a yet another problem.

If we go for distro QEMU, looking at our CI, we have:
* Ubuntu 18.04, which has QEMU **2.11** as part of their official repos: https://launchpad.net/ubuntu/bionic/+source/qemu
* Fedora 32, which has QEMU 4.2.1 as part of their official repos: https://koji.fedoraproject.org/koji/packageinfo?buildStart=50&packageID=3685&buildOrder=-completion_time&tagOrder=name&tagStart=1#buildlist
* Debian 10 (buster), which has QEMU 3.1 as part of their official repos: https://packages.debian.org/search?keywords=qemu
* CentOS, which is disabled by now, when using the advanced-virt repo would have new enough QEMU, but that's not exactly what comes with the distro.

As we're using virtiofs by default, it just means we can't test anymore against Ubuntu 18.04, and Debian 10. :-/

But this is just the tip of the iceberg, IMHO. 
* What does it mean be supported by "a distro"?
* Can we consider ourselves totally supported by Ubuntu if the user is having to our provided build of the VMM, and our provided build of the Kernel?
  We have to properly set up the expectation levels on those, and I don't think that's something we currently do. :-/


>     1. "Stable" meaning we use the same qemu version across each different CI testset
> 
> 
> This one I will dispute. If Kata is fragile against fairly minor changes in qemu versions, that's a bad thing - it makes life harder for system integrators, if not for end users. Therefore using the relevant distro version for each of our testsets _improves_ the CI's coverage: if it fails on some distro due to how we invoke qemu, we should probably be avoiding doing that.

Well, think about the QEMU versions on Ubuntu 18.04 (**2.11**), and the QEMU version I'm running on my host (**5.2.0**).  It's not a minor change in QEMU version.



There's also another thing we have to keep in mind, we must test what we ship and recommend users to use, which is what's used by `kata-deploy`.

I can totally see the argument of "we should **also** test what we ship as part of `kata-deploy`, but not limit our tests to that".  And I agree.  I agree, and I think we should add an option to the CI scripts to just not install VMMs, to just not generate an image  / initrd, and then it'd help the distros and users who'd prefer to use the system ones, allowing them to re-use our tests on the distro side.

That's the path I think we should follow here, sincerely.

Does this sound reasonable?

PR #3264, opened 2021-02-19
ci: Install ninja-build package on Ubuntu

### LINKED ISSUE #3263: setup: ninja-build package should be installed on Ubuntu for QEMU 5.2
The `.ci/install_qemu.sh`  script has a conditional where if
 1. `AGENT_INIT=false` then QEMU is built with statically linking. It uses the scripts in github.com/kata-containers/kata-containers/tools/packaging which are going to build it inside a container, so any build dependencies are already handled. 
 1. `AGENT_INIT=true` then QEMU is built with dynamically linking. Here it doesn't use the packaging scripts, and QEMU is built in the local host. Meaning that all build dependencies should be installed on the host.

The QEMU 5.2 requires the `ninja-build` package for Ubuntu. Thus, the CI setup script should be changed to handle the case 2) above.

Note: theoretically all setup scripts should be changed to get ninja-build installed. But in practice only one CI job test the `AGENT_INIT=true` and that is in Ubuntu.

wainersm:
The ninja-build package is needed on Ubuntu for building QEMU 5.2
onward.

Fixes #3263

### DIFF
--- a/.ci/setup_env_ubuntu.sh
+++ b/.ci/setup_env_ubuntu.sh
@@ -28,7 +28,7 @@ declare -A minimal_packages=( \
 
 declare -A packages=( \
 	[kata_containers_dependencies]="libtool automake autotools-dev autoconf bc libpixman-1-dev coreutils" \
-	[qemu_dependencies]="libcap-dev libattr1-dev libcap-ng-dev librbd-dev" \
+	[qemu_dependencies]="libcap-dev libattr1-dev libcap-ng-dev librbd-dev ninja-build" \
 	[kernel_dependencies]="libelf-dev flex" \
 	[crio_dependencies]="libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev thin-provisioning-tools" \
 	[bison_binary]="bison" \


GabyCT (approved):

fidencio (commented):
@wainersm, what's the ninja version needed?

I'm afraid the version present on ubuntu 18.04, that we run, may not be "new enough" for the QEMU needs.

fidencio (changes requested):
So, taking a look at QEMU 5.2.0 (https://github.com/qemu/qemu/blob/553032db17440f8de011390e5a1cfddd13751b0b/meson.build#L1) we can see that the meson version required there is `0.55.0`, while the version Ubuntu 18.04 provides is `0.45.1`.

The first version of Ubuntu that supports the minimum meson version required by QEMU is 20.10, supporting `0.55.3`.

Please, refer to https://packages.ubuntu.com/search?keywords=meson for more information.


With everything above said, wouldn't be better / easier to rely on `pip` to install the latest version of the meson (or at least a specific version of that)?

wainersm:
@fidencio thanks for bringing this up, I really didn't pay attention to the version.

Now I double-checked to ensure this change is correct. Where are the facts:

* Ubuntu 18.04 provides ninja-build 1.8.2
```
root@d2cce3ba5230:/# lsb_release -r    
Release:	18.04
root@d2cce3ba5230:/# ninja --version
1.8.2
```
* Meson 0.55 requires ninja-build >= 1.7 (https://github.com/mesonbuild/meson/blob/0.55/README.md)
* Meson is a git submodule of QEMU (https://github.com/qemu/qemu/blob/master/.gitmodules#L61). So it can be consumed from sources, and in fact, we are not installing the packaged meson.

fidencio (approved):
@wainersm, thanks for double-checking and I didn't know QEMU was bundling `meson`.
lgtm, thanks!

PR #3297, opened 2021-03-10
ci: Fixes and improvements to the vfio job

### LINKED ISSUE #3263: setup: ninja-build package should be installed on Ubuntu for QEMU 5.2

wainersm:
The main goal of this PR was to fix the failure on vfio job due to missing ninja-build package on the ubuntu host. Then I noticed that the script was using code from the master branch. I changed it but that caused more failures, and more fixes were needed. Along the way I implemented changes to the vfio test script to increase its debuggability, this comes as a bonus commit.

### DIFF
--- a/.ci/ci_entry_point.sh
+++ b/.ci/ci_entry_point.sh
@@ -16,6 +16,7 @@ set -o errtrace
 
 script_name=${0##*/}
 script_dir=$(dirname "$(readlink -f "$0")")
+CI_JOB="${CI_JOB:-}"
 
 handle_error() {
 	local exit_code="${?}"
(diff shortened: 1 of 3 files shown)


wainersm:
This PR fixes the problem of QEMU not being built, however the vfio test failed with:

```
15:12:42 + sudo -E kubectl wait --for=condition=Ready pod vfio
15:13:12 error: timed out waiting for the condition on pods/vfio
15:13:12 Failed at 123: sudo -E kubectl wait --for=condition=Ready pod "${pod_name}"
15:13:12 ++ handle_error 123
15:13:12 ++ local exit_code=1
15:13:12 ++ local line_number=123
15:13:12 ++ echo 'Failed at 123: sudo -E kubectl wait --for=condition=Ready pod "${pod_name}"'
```

chavafg (approved):

GabyCT (approved):

PR #3312, opened 2021-03-12
cri-o: Stop using devicemapper with CRI-O

fidencio:
CRI-O tests have been constantly failing on 1.x, and it seems to be due
to using devicemapper as the storage driver.

Fixes: #3330

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -89,14 +89,7 @@ fi
 pushd "${GOPATH}/src/${crio_repo}"
 echo "Installing CRI-O"
 make clean
-if [ "$ID" == "centos" ] || [ "$ID" == "fedora" ]; then
-	# This is necessary to avoid crashing `make` with `No package devmapper found`
-	# by disabling the devmapper driver when the library it requires is not installed
-	sed -i 's|$(shell hack/selinux_tag.sh)||' Makefile
-	make BUILDTAGS='exclude_graphdriver_devicemapper libdm_no_deferred_remove'
-else
-	make
-fi
+make BUILDTAGS='exclude_graphdriver_devicemapper libdm_no_deferred_remove'
 make test-binaries
 sudo -E PATH=$PATH sh -c "make install"
 sudo -E PATH=$PATH sh -c "make install.config"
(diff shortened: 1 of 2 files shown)


fidencio:
A few notes that I'd like to discuss with @chavafg, @GabyCT, and @jcvenegas:
* I'm not sure we cover the same tests that CRI-O does on their side, by running those CIs, but I don't think so.  With this in mind, I'll try to clone their VM, add a specific test (not to be enabled, but rather for helping us to debug), and ensure it passes here;
* By using overlayfs instead of device-mapper with CRI-O, we can see an increase in the footprint.  I wonder if the metrics CI shouldn't be using containerd (which does supper device-mapper, no issue there) for such tests;

.ci/install_crio.sh
```diff
@@ -93,10 +93,8 @@ if [ "$ID" == "centos" ] || [ "$ID" == "fedora" ]; then
 	# This is necessary to avoid crashing `make` with `No package devmapper found`
```
wainersm, line 93: Hi @fidencio, I am wondering if this block can also be removed.

fidencio: By this block, you mean disable the `selinux_tag` check? it was added a long time ago and, indeed, I think we can just drop it entirely.

wainersm: yes, drop the selinux_tag check entirely.

chavafg:
Hey @fidencio, 

> I'm not sure we cover the same tests that CRI-O does on their side, by running those CIs, but I don't think so. With this in mind, I'll try to clone their VM, add a specific test (not to be enabled, but rather for helping us to debug), and ensure it passes here;

You're right, we are only running the [ctr.bats](https://github.com/cri-o/cri-o/blob/master/test/ctr.bats). It would be good if we could increase the tests. 

> By using overlayfs instead of device-mapper with CRI-O, we can see an increase in the footprint. I wonder if the metrics CI shouldn't be using containerd (which does supper device-mapper, no issue there) for such tests

For metrics, in 1.x we are using docker, so shouldn't be an issue with this change.

Issue #3348, opened 2021-03-26
OOM test for jenkins-ci-ubuntu-18-04-initrd fails frequently


liubin:
OOM test in [jenkins-ci-ubuntu-18-04-initrd](http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-initrd-PR/) fails frequently.

Adjusting some parameters maybe help to improve the success.

PR #3356, opened 2021-03-29
kubernetes: change OOM test parameter to ensure it will be success

liubin:
Let container run more longer, and if fails, get the pod full YAML spec.

Fixes: #3348

Signed-off-by: bin liub bin@hyper.sh


lifupan (approved):

fidencio (approved):
lgtm, thanks @liubin!

PR #3361, opened 2021-03-29
k8s: Add empty dir when FSGroup is set test 

wainersm:
Added a test case to reproduce the error in k8s e2e test:

[sig-storage] EmptyDir volumes when FSGroup is specified [LinuxOnly] [NodeFeature:FSGroup] new files should be created with FSGroup ownership when container is non-root

### DIFF
--- a/integration/kubernetes/k8s-empty-dirs.bats
+++ b/integration/kubernetes/k8s-empty-dirs.bats
@@ -8,10 +8,22 @@
 load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
 load "${BATS_TEST_DIRNAME}/../../lib/common.bash"
 
+assert_equal() {
+	local expected=$1
+	local actual=$2
+	if [[ "$expected" != "$actual" ]]; then
+	echo "expected: $expected, got: $actual"
+	return 1
+	fi
+}
+
 setup() {
 	export KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
 	pod_name="sharevol-kata"
 	get_pod_config_dir
+	pod_logs_file=""
+	wait_time=20
+	sleep_time=2
 }
 
 @test "Empty dir volumes" {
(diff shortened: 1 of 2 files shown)


wainersm:
Failed `13:56:54 not ok 3 Liveness tcp probe` on http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-fedora-PR/605 which is unrelated with this PR. I will re-trigger the Fedora jobs, in case it be an intermittent failure.

GabyCT (approved):

integration/kubernetes/k8s-empty-dirs.bats
```diff
@@ -26,9 +38,27 @@ setup() {
...
+
+	cmd="kubectl get pods ${pod_name} | grep Completed"
+	waitForProcess "${wait_time}" "${sleep_time}" "${cmd}"
+
+	pod_logs_file="$(mktemp)"
+	kubectl logs "$pod_name" > "$pod_logs_file"
+	uid=$(cat $pod_logs_file | grep 'owner UID of' | sed 's/.*:\s//')
+	assert_equal "1001" "$uid" "owner UID of file"
+	gid=$(cat $pod_logs_file | grep 'owner GID of' | sed 's/.*:\s//')
+	assert_equal "123" "$gid" "owner GIG of file"
```
snir911, line 54: typo 
also i suppose you forgot to do something with $3 in assert_equal :)

snir911 (approved):
overall LGTM just see my in-line comment, thanks! (adding do-not-merge until you'd review it)

integration/kubernetes/k8s-empty-dirs.bats
wainersm: @snir911 even after you mention, it took me minutes to realize the typo was in "GIG". I call this "super-typo" :)

And regarding the "$3" parameter to assert_equal, the initial plan was to simple print it. But then I noticed the msg would appear on the stack track in case the assert fails. I agree it is weird and I will re-do it.

Thanks for the review!

snir911: well, is it actually a typo if you don't print it? dunno, it's like a tree falls in the forest and no one is around to hear it :)

wainersm: @snir911 on the updated version which I just pushed I removed the 3rd argument, so it should be fine now. Btw, I like the tree fall analogy :D

wainersm:
Fail on jenkins-ci-ubuntu-18-04-initrd can be ignored since it is not related with this PR.

PR #3402, opened 2021-04-07
openshift-ci: build kata containers on CentOS 8

### LINKED ISSUE #3389: openshit-ci: build kata containers on CentOS 8
Currently on OpenShift CI the build root environment is CentOS 7.

Recently the project started using Kernel 5.10 which requires gcc >= 4.9, but CentOS 7 comes with 4.8.5, thus the build fails. So it will be needed to use a container image with a newer gcc version, let's use CentOS 8, and it will require some changes on the scripts found at `.ci/openshift-ci`

wainersm:
This contain the changes required to switch the build root on OpenShift CI from CentOS 7 to 8. Thus, it fixes #3389.

### DIFF
--- a/.ci/openshift-ci/build_install.sh
+++ b/.ci/openshift-ci/build_install.sh
@@ -11,7 +11,8 @@
 
 set -e
 
-export PATH="${GOPATH}/bin:$PATH"
+export GOPATH="${GOPATH:-/go}"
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @wainersm!

GabyCT (approved):

PR #3417, opened 2021-04-14
CI: Remove CentOS 7 support

wainersm:
I don't think it makes sense to maintain the CI scripts for CentOS 7, so I propose its removal on this PR. While ago @fidencio sent a similar PR but at that time I was using CentOS 7 to build kata on openshift-ci; recently I changed it to use CentOS 8 which allow us now to proceed with the removal.

### DIFF
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -18,48 +18,44 @@ else
   centos_version=$(grep VERSION_ID /usr/lib/os-release | cut -d '"' -f2)
 fi
 
+[ "$centos_version" == 8 ] || die "This script is for CentOS 8 only"
+
 # Send error when a package is not available in the repositories
 echo "skip_missing_names_on_install=0" | sudo tee -a /etc/yum.conf
 
 # Check EPEL repository is enabled on CentOS
-if [ -z "$(yum repolist | grep 'Extra Packages')" ]; then
+if [ -z "$(dnf repolist | grep 'Extra Packages')" ]; then
 	echo >&2 "ERROR: EPEL repository is not enabled on CentOS."
 	# Enable EPEL repository on CentOS
-	sudo -E yum install -y wget rpm
+	sudo -E dnf install -y wget rpm
 	wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-${centos_version}.noarch.rpm
 	sudo -E rpm -ivh epel-release-latest-${centos_version}.noarch.rpm
 fi
 
 # Enable priority to CentOS Base repo in order to
 # avoid perl updating issues
-if [ "$centos_version" == "8" ]; then
-	repo_file=""
-	if [ -f /etc/yum.repos.d/CentOS-Base.repo ]; then
-		repo_file="/etc/yum.repos.d/CentOS-Base.repo"
-	elif [ -f /etc/yum.repos.d/CentOS-Linux-BaseOS.repo ]; then
-		repo_file="/etc/yum.repos.d/CentOS-Linux-BaseOS.repo"
-	else
-		die "Unable to find the CentOS base repository file"
-	fi
-	sudo echo "priority=1" | sudo tee -a "$repo_file"
-	sudo -E yum -y clean all
+repo_file=""
+if [ -f /etc/yum.repos.d/CentOS-Base.repo ]; then
+	repo_file="/etc/yum.repos.d/CentOS-Base.repo"
+elif [ -f /etc/yum.repos.d/CentOS-Linux-BaseOS.repo ]; then
+	repo_file="/etc/yum.repos.d/CentOS-Linux-BaseOS.repo"
+else
+	die "Unable to find the CentOS base repository file"
 fi
+echo "priority=1" | sudo tee -a "$repo_file"
+sudo -E dnf -y clean all
 
 echo "Update repositories"
-sudo -E yum -y update
+sudo -E dnf -y update
 
-if [ "$centos_version" == "8" ]; then
-	echo "Enable PowerTools repository"
-	sudo -E yum install -y yum-utils
-	sudo yum-config-manager --enable powertools
-fi
+echo "Enable PowerTools repository"
+sudo -E dnf install -y 'dnf-command(config-manager)'
+sudo dnf config-manager --enable powertools
 
 echo "Install chronic"
-sudo -E yum -y install moreutils
+sudo -E dnf -y install moreutils
 
-if [ "$centos_version" == "8" ]; then
-	chronic sudo -E yum install -y pkgconf-pkg-config
-fi 
+chronic sudo -E dnf install -y pkgconf-pkg-config
 
 declare -A minimal_packages=( \
 	[spell-check]="hunspell hunspell-en-GB hunspell-en-US pandoc" \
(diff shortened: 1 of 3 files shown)


egernst:
Just for my own curiosity -- do we have much handling for centos7 versus centos8 (ie, trying to understand the burden).

Thx @wainersm !

fidencio:
@egernst, as part of the group containing folks maintaining the CentOS 8 packages, I can tell you there's absolutely zero interest on adding / supporting kata-containers there.

We better be looking forward (CentOS 8, hopefully CentOS Stream). :-)

.ci/setup_env_centos.sh
```diff
@@ -18,48 +18,42 @@ else
...
+sudo -E yum install -y yum-utils
+sudo yum-config-manager --enable powertools
 
 echo "Install chronic"
 sudo -E yum -y install moreutils
 
-if [ "$centos_version" == "8" ]; then
-	chronic sudo -E yum install -y pkgconf-pkg-config
-fi 
+chronic sudo -E yum install -y pkgconf-pkg-config
```
fidencio, line 56: Can you also drop this chronic usage? :-)
It's not consistent across this file.

.ci/setup_env_centos.sh
```diff
@@ -106,15 +100,11 @@ main()
...
-		# exclusive. Let's pass --allowerasing so that coreutils-single
-		# is replaced.
-		yum_install_args+=" --allowerasing"
-	fi
-	chronic sudo -E yum -y install $yum_install_args $pkgs_to_install
+	# On centos:8 container image the installation of coreutils
+	# conflicts with coreutils-single because they mutually
+	# exclusive. Let's pass --allowerasing so that coreutils-single
+	# is replaced.
+	chronic sudo -E yum -y install --allowerasing $pkgs_to_install
```
fidencio, line 107: Same here. :-)

.ci/setup_env_centos.sh
```diff
@@ -18,48 +18,42 @@ else
...
 
 # Check EPEL repository is enabled on CentOS
 if [ -z "$(yum repolist | grep 'Extra Packages')" ]; then
-	echo >&2 "ERROR: EPEL repository is not enabled on CentOS."
+	info "EPEL repository is not enabled on CentOS. Going to enable it."
 	# Enable EPEL repository on CentOS
-	sudo -E yum install -y wget rpm
-	wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-${centos_version}.noarch.rpm
-	sudo -E rpm -ivh epel-release-latest-${centos_version}.noarch.rpm
+	sudo -E rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
fidencio, line 30: Can't we just `yum -y install epel-release`?

fidencio (commented):
In general looks good, I'd like to know your opinion on a few comments I left.

Thanks, @wainersm!

.ci/setup_env_centos.sh
```diff
@@ -18,48 +18,42 @@ else
...
-		repo_file="/etc/yum.repos.d/CentOS-Base.repo"
-	elif [ -f /etc/yum.repos.d/CentOS-Linux-BaseOS.repo ]; then
-		repo_file="/etc/yum.repos.d/CentOS-Linux-BaseOS.repo"
-	else
-		die "Unable to find the CentOS base repository file"
-	fi
-	sudo echo "priority=1" | sudo tee -a "$repo_file"
-	sudo -E yum -y clean all
+repo_file=""
+if [ -f /etc/yum.repos.d/CentOS-Base.repo ]; then
```
devimc, line 36: please define some variables/constants for these two paths

.ci/setup_env_centos.sh
```diff
@@ -18,48 +18,42 @@ else
...
+	repo_file="/etc/yum.repos.d/CentOS-Base.repo"
+elif [ -f /etc/yum.repos.d/CentOS-Linux-BaseOS.repo ]; then
+	repo_file="/etc/yum.repos.d/CentOS-Linux-BaseOS.repo"
+else
+	die "Unable to find the CentOS base repository file"
 fi
+sudo echo "priority=1" | sudo tee -a "$repo_file"
```
devimc, line 43: this first sudo is not needed

.ci/setup_env_centos.sh
```diff
@@ -18,48 +18,42 @@ else
...
+sudo -E yum -y clean all
```
devimc, line 44: uhmm why this? also this is centos 8, why don't use `dnf`?

.ci/setup_env_centos.sh
```diff
@@ -18,48 +18,42 @@ else
...
 echo "Update repositories"
 sudo -E yum -y update
 
-if [ "$centos_version" == "8" ]; then
-	echo "Enable PowerTools repository"
-	sudo -E yum install -y yum-utils
-	sudo yum-config-manager --enable powertools
-fi
+echo "Enable PowerTools repository"
+sudo -E yum install -y yum-utils
```
devimc, line 50: `s/yum/dnf/g` ?

devimc (commented):
thanks @wainersm - I've left some comments

wainersm:
> thanks @wainersm - I've left some comments

Hi @devimc, first, thanks for the careful review! I tried to not change much the script so basically on this PR I am deleting the "if centos 8 then do this" branches. But while I am here maybe I can improve the script entirely (replace yum with dnf...etc).

I am going to send an updated version soon. :)

.ci/setup_env_centos.sh
wainersm: I don't know the intention of the original author but I noticed that it is going to `21 files removed`.
I am changing the script to use dnf instead of yum.

.ci/setup_env_centos.sh
```diff
@@ -18,48 +18,43 @@ else
...
 # Check EPEL repository is enabled on CentOS
-if [ -z "$(yum repolist | grep 'Extra Packages')" ]; then
-	echo >&2 "ERROR: EPEL repository is not enabled on CentOS."
+if [ -z "$(dnf repolist | grep 'Extra Packages')" ]; then
+	info "EPEL repository is not enabled on CentOS. Going to enable it."
 	# Enable EPEL repository on CentOS
-	sudo -E yum install -y wget rpm
-	wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-${centos_version}.noarch.rpm
-	sudo -E rpm -ivh epel-release-latest-${centos_version}.noarch.rpm
+	sudo -E dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
fidencio, line 30: I'd simply drop the if statement and do a `sudo -E dnf -y install epel-release`.  If it's already installed, it won't fail.  And as we have the package as part of CentOS, I see no good reason to downloading it from somewhere else.

wainersm: Done.

.ci/setup_env_centos.sh
wainersm: @fidencio  good point.

I saw somewhere a discussion regarding whether to remove the chronic or not from kata scripts. So I think it's better to keep it until we don't have an agreement (or do we already have one?).

wainersm:
Hi @egernst ,

> Just for my own curiosity -- do we have much handling for centos7 versus centos8 (ie, trying to understand the burden).

Not really, but when I make changes to the script I need to at least test on both versions. And if I could avoid it... :)

fidencio (approved):
lgtm, thanks @wainersm! :-)

fidencio:
@devimc, @GabyCT, can we get a review and have this one merged as well? :-)

devimc (approved):
thanks @wainersm

Issue #3415, opened 2021-04-14
enhance kata-webhook to take into consideration other runtimeClasses' names


chavafg:
**Which feature do you think can be improved?**

Currently, the kata-webhook only works when k8s is configured to use `RuntimeClass: kata`, which is the RuntimeClass that we use in our CI, but if we want to use the webhook along with `kata-deploy`, it will not work as the runtimeclass name configured in k8s will differ.

kata-deploy configures k8s with other runtimeClass names, such as: `kata-clh` and `kata-qemu` . 

**How can it be improved?**

Modify the kata-webhook [`main.go` ](https://github.com/kata-containers/tests/blob/main/kata-webhook/main.go), so that it can be built using an environment variable that we can use to specify other values for the runtimeClass.
If environment variable is not set, then use `kata` as default.

PR #3420, opened 2021-04-15
kata-webhook: Use env var to change runtimeclass name

chavafg:
kata-webhook currently works with hardcoded RuntimeClass name kata.
If the k8s cluster is configured with other RuntimeClasses, such as
kata-clh or kata-qemu, the webhook will not work.

With this change, the webhook will be able to get the name of the
RuntimeClass from an environment variable so that you can specify
other values.

Updated documentation for this new change.

Fixes: #3415.

### DIFF
--- a/kata-webhook/README.md
+++ b/kata-webhook/README.md
@@ -5,13 +5,16 @@ Kata runtime class.
 
 ## How to build the admission controller
 
+> **Note:**
+> Only run this step if you are modifying the current webhook or don't
+> want to use the webhook available in docker hub.
+
 First build the admission controller image and the associated
 Kubernetes YAML files required to instantiate the admission
 controller.
 
 ```bash
 $ docker build -t katadocker/kata-webhook-example:latest .
-$ ./create_certs.sh
 ```
 
 > **Note:**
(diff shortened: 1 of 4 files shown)


kata-webhook/README.md
```diff
@@ -27,7 +30,15 @@ by default, except for workloads we know for sure will not work with Kata, use
 the [admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks)
 and sample admission controller we created by running
 
+> **Note:**
+> By default, the RuntimeClass name used in this webhook is `kata`. If your
+> cluster is configured with another RuntimeClass, you'll need to change the
+> `RUNTIME_CLASS` to the correct value in the [webhook.yaml](deploy/webhook.yaml#L27).
```
devimc, line 36: `RUNTIME_CLASS` .... variable ? dictionary ?

devimc: `deploy/webhook.yaml#L27` is too specific, line number can change at any moment

devimc (commented):
thanks @chavafg - I've left some comments

kata-webhook/deploy/webhook.yaml
```diff
@@ -20,8 +20,11 @@ spec:
     spec:
       containers:
         - name: pod-annotate-webhook
-          image: fuentess/kata-webhook:20200707
+          image: katadocker/kata-webhook-example:latest
```
eadamsintel, line 23: Is there any reason this shouldn't be used in a production environment or any reason why someone should need to create their own Kata webhook? If it can be used in production I would consider removing the "-example" from the name.

chavafg: I think it could be run in production, but the behavior of it should depend on the use case. Maybe someone would want to add additional changes to the spec or consider different scenarios where they do not want to use kata. Currently we have only added conditions that make sense for our CI.

kata-webhook/README.md
chavafg: Thanks @devimc, addressed both comments

devimc (approved):
thanks @chavafg

jcvenegas (approved):

Issue #3456, opened 2021-04-27
issues with btrfs when building containerd ubuntu 20.04


GabyCT:
I am trying to enable ubuntu 20.04 for the azure VMs that we have in our CI, however, I have seen issues like
```
17:27:55 + bin/containerd
17:27:57 # github.com/containerd/containerd/vendor/github.com/containerd/btrfs
17:27:57 vendor/github.com/containerd/btrfs/btrfs.go:38:10: fatal error: btrfs/ioctl.h: No such file or directory
17:27:57    38 | #include <btrfs/ioctl.h>
17:27:57       |          ^~~~~~~~~~~~~~~
17:27:57 compilation terminated.
17:28:10 make[1]: *** [Makefile:183: bin/containerd] Error 2
17:28:10 make[1]: Leaving directory '/tmp/cri-install-containerd.UccE/src/github.com/containerd/containerd'
17:28:10 make: *** [Makefile:132: _output/cri-containerd-f864905c.linux-amd64.tar.gz] Error 2
17:28:10 Failed at 48: make release
17:28:10 Failed at 53: ( cd "${GOPATH}/src/${cri_repository}" >> /dev/null; git fetch; git checkout "${cri_containerd_version}"; make release; local commit; commit=$(git rev-parse --short HEAD); tarball_name="cri-containerd-${commit}.${CONTAINERD_OS}-${CONTAIENRD_ARCH}.tar.gz"; sudo tar -xvf "./_output/${tarball_name}" -C / )
17:28:10 containerd not installed
```
It seems that is happening when building `containerd`, this issue has been solved for CRI-O using buildtags ...I am not sure if that can be also used to build `containerd`, any idea @fidencio ? thanks
http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-20.04-PR/3/consoleFull

GabyCT:
or should I just install the `btrfs` package?

fidencio:
@GabyCT, I **think** ubuntu supports and provides the needed packages for btrfs, so I'd consider adding the package when building containerd there.  I'll try to submit a patch in a few, so you can review whenever you get back online.

Thanks!

PR #3457, opened 2021-04-28
ubuntu: Install btrfs-progs as containerd dependency

fidencio:
To be fair, this is also a CRI-O dependency and later on we should
explore using buildtags based on installed packages (rather than
explicitly defining those).

Fixes: #3456


GabyCT (approved):

devimc (approved):
thanks @fidencio

PR #3458, opened 2021-04-28
k8s: Use the same agnhost image version

### LINKED ISSUE #3455: k8s: Use same agnhost image version across tests
Some of our kubernetes integration tests rely on the [agnhost image](https://github.com/kubernetes/kubernetes/tree/master/test/images/agnhost) but currently each use a different version.

They should be using the same version and ideally updated with Kubernetes release used for testing.

wainersm:
This is a refactor on some kubernetes integration tests to ensure they
rely on the same version of the agnhost image. It was added a new
entry on versions.yaml to record the exact image's name and version.

Fixes #3455

### DIFF
--- a/integration/kubernetes/k8s-empty-dirs.bats
+++ b/integration/kubernetes/k8s-empty-dirs.bats
@@ -41,12 +41,14 @@ setup() {
 @test "Empty dir volume when FSGroup is specified with non-root container" {
 	# This is a reproducer of k8s e2e "[sig-storage] EmptyDir volumes when FSGroup is specified [LinuxOnly] [NodeFeature:FSGroup] new files should be created with FSGroup ownership when container is non-root" test
 	pod_file="${pod_config_dir}/pod-empty-dir-fsgroup.yaml"
-	image=$(grep 'image:' ${pod_file} | head -1 | sed 's/.*\s//')
+	agnhost_name=$(get_test_version "container_images.agnhost.name")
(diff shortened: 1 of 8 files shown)


fidencio (approved):
thanks, @wainersm.
I like the idea and I wonder whether we should o the same with the other images we use more than once.   The ones I have in mind are the busybox and nginx ones.

Also, while trying to get the whole CI green on my side, I've noticed that a good practice we should add is having the full image name (including the registry, as you did here).

This is not done for some of the images, IIRC.

wainersm:
/retest-debian
As it seems to face intermittent networking issue.

wainersm:
/retest-centos
As an unrelated test failed with timeout.

GabyCT (approved):

wainersm:
> thanks, @wainersm.

And thank you (@fidencio) for the review and merge!

> I like the idea and I wonder whether we should o the same with the other images we use more than once. The ones I have in mind are the busybox and nginx ones.

I actually took that idea of test which use nginx and currently have similar approach. The busybox ones are always point to latest image.

> 
> Also, while trying to get the whole CI green on my side, I've noticed that a good practice we should add is having the full image name (including the registry, as you did here).
> 

Getting @egernst on this conversation...IIUC on his environment, for osbuilder, he uses a proxy to pull the images, so the full name (i.e. with registry) is a problem. For k8s testing I don't know if there is such as constraint.

PR #3481, opened 2021-05-03
ci: remove existing k8s+crio config before seting up new config

### LINKED ISSUE #3480: Running CRI_CONTAINERD_K8S job after running CRIO_K8S job on the same systems fails.
# Description of problem

When running CRI_CONTAINERD_K8S job after having run CRIO_K8S job on the same system, the initialization of the Kubernetes cluster fails, thus preventing the test to run properly.

# Expected result

CRI_CONTAINERD_K8S runs independently of previous job runs.

# Actual result

Kubernetes failed to configure, displaying an error message:
`[kubelet-check] It seems like the kubelet isn't running or healthy.`

# Further information

Journal shows that kubelet is still trying to use CRI-O:
`Apr 29 06:46:34 kata5 kubelet[966242]: W0429 06:46:34.899249  966242 clientconn.go:1208] grpc: addrConn.createTransport failed to connect to {/var/run/crio/crio.sock  <nil> 0 <nil>}. Err :connection error: desc = "transport: Error while dialing dial unix /var/run/crio/crio.sock: connect: no such file or directory". Reconnecting...
`

MatthieuSarter:
Removes existing k8s+crio config during setup to ensure k8s will not use crio when running a containerd test.

Fixes #3480

Signed-off-by: MatthieuSarter matthieu.sarter@ibm.com

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -93,6 +93,9 @@ install_extra_tools() {
 	echo "Install CNI plugins"
 	bash -f "${cidir}/install_cni_plugins.sh"
 
+	# Remove K8s + CRIO conf that may remain from a previous run
+	rm -f /etc/systemd/system/kubelet.service.d/0-crio.conf
+
 	[ "${CRIO}" = "yes" ] &&
 		echo "Install CRI-O" &&
 		bash -f "${cidir}/install_crio.sh" &&


.ci/setup.sh
```diff
@@ -93,6 +93,9 @@ install_extra_tools() {
 	echo "Install CNI plugins"
 	bash -f "${cidir}/install_cni_plugins.sh"
 
+	# Remove K8s + CRIO conf that may remain from a previous run
+	rm -f /etc/systemd/kubelet.service.d/0-crio.conf
```
devimc: missing `/system/` ?  `0-crio.conf` looks very specific, what about removing whole `kubelet.service.d` directory?

MatthieuSarter: Yep, system was missing, committed the wrong file.

Not sure about deleting the whole directory, on my system (Ubuntu 20.04 ppc64) there is a 10-kubeadm.conf file in this directory, that came with the kubeadm package.

fidencio (approved):
The joys of running the CI on baremetal. :-).
lgtm, thanks @MatthieuSarter!

fidencio:
@MatthieuSarter, I'd like to ask you for the following things before we get this one merged:
* Rebase this atop of current `main`;
* Ensure the `Signed-off-by: ` tag is present;
* Force push the changes into this branch;

Those should make the CI happy.

fidencio:
@MatthieuSarter, last thing, seems you manually added the `Signed-off-by: ` tag?  You should do it using `git commit -s` instead, and that seems to be the reason why the CI is complaining.

MatthieuSarter:
> 
> 
> @MatthieuSarter, last thing, seems you manually added the `Signed-off-by: ` tag? You should do it using `git commit -s` instead, and that seems to be the reason why the CI is complaining.

Yes, I dit the commit with an old IDE that does not support signing off, but for some reason it removed the arrows arround the e-mail adress... Will now use the git CLI.

snir911 (approved):

PR #3489, opened 2021-05-04
ci: Use cached kernels

wainersm:
Today I created two jobs on CI to nightly build either vanilla and experimental kernels:

  http://jenkins.katacontainers.io/job/kata-containers-2.0-kernel-experimental-x86_64-nightly
  http://jenkins.katacontainers.io/job/kata-containers-2.0-kernel-vanilla-x86_64-nightly

It is using the same mechanism and scripts as used with kata 1.x. So basically on this PR I am only adjusting urls.

### DIFF
--- a/.ci/install_kata_kernel.sh
+++ b/.ci/install_kata_kernel.sh
@@ -17,7 +17,7 @@ source "${cidir}/lib.sh"
 source "/etc/os-release" || source "/usr/lib/os-release"
 
 latest_build_url="${jenkins_url}/job/kata-containers-2.0-kernel-vanilla-$(uname -m)-nightly/${cached_artifacts_path}"
-experimental_latest_build_url="${jenkins_url}/job/kernel-experimental-nightly-$(uname -m)/${cached_artifacts_path}"
+experimental_latest_build_url="${jenkins_url}/job/kata-containers-2.0-kernel-experimental-$(uname -m)-nightly/${cached_artifacts_path}"
 PREFIX="${PREFIX:-/usr}"
 kernel_dir="${DESTDIR:-}${PREFIX}/share/kata-containers"
 
@@ -39,11 +39,10 @@ trap exit_handler EXIT
 get_current_kernel_version() {
 	if [ "$experimental_kernel" == "true" ]; then
 		kernel_version=$(get_version "assets.kernel-experimental.tag")
-		echo "${kernel_version}"
 	else
 		kernel_version=$(get_version "assets.kernel.version")
-		echo "${kernel_version/v/}"
 	fi
+	echo "${kernel_version/v/}"
 }
 
 get_kata_config_version() {
@@ -130,6 +129,7 @@ main() {
 	if [ "${experimental_kernel}" == "false" ]; then
 		cached_kernel_version=$(curl -sfL "${latest_build_url}/latest") || cached_kernel_version="none"
 	else
+		current_kernel_version+="-experimental"
 		cached_kernel_version=$(curl -sfL "${experimental_latest_build_url}/latest") || cached_kernel_version="none"
 	fi
 	info "current kernel : ${current_kernel_version}"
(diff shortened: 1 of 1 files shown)


fidencio:
@wainersm, lovely.
AFAIU we'd still need to implement a a log to verify whether we're patching the kernel or QEMU files, right?  I know, having this one will do, but I wonder whether we could have a post merge hook that would update the kernel / initrd files if changed, rather than having a nightly build for those.

wainersm:
It is working:

```
17:24:36 + sudo -E curl -fL --progress-bar http://jenkins.katacontainers.io/job/kata-containers-2.0-kernel-vanilla-x86_64-nightly/lastSuccessfulBuild/artifact/artifacts/vmlinux-5.10.25-84 -o /usr/share/kata-containers/vmlinux-5.10.25-84
17:24:36 
                                                                           0.1%
                                                                           0.5%
###                                                                        4.6%
################                                                          22.7%
###############################                                           43.5%
#############################################                             63.7%
############################################################              83.9%
######################################################################## 100.0%
17:24:37 + kernel_symlink=/usr/share/kata-containers/vmlinux.container
17:24:37 + info 'Installing /usr/share/kata-containers/vmlinux-5.10.25-84 and symlink /usr/share/kata-containers/vmlinux.container'
17:24:37 + echo -e 'INFO: Installing /usr/share/kata-containers/vmlinux-5.10.25-84 and symlink /usr/share/kata-containers/vmlinux.container'
17:24:37 INFO: Installing /usr/share/kata-containers/vmlinux-5.10.25-84 and symlink /usr/share/kata-containers/vmlinux.container
17:24:37 + sudo -E ln -sf /usr/share/kata-containers/vmlinux-5.10.25-84 /usr/share/kata-containers/vmlinux.container
```

wainersm:
> @wainersm, lovely.
> AFAIU we'd still need to implement a a log to verify whether we're patching the kernel or QEMU files, right? I know, having this one will do, but I wonder whether we could have a post merge hook that would update the kernel / initrd files if changed, rather than having a nightly build for those.

I did not understand that statement about patching qemu and kernel.

About the post-merge hook to trigger the kernel build once its version/configuration is updated, I think it is a good idea. But I prefer to get it done later. I still would like to work on caching the other assets: qemu, osbuilder, etc.

fidencio:
> > @wainersm, lovely.
> > AFAIU we'd still need to implement a a log to verify whether we're patching the kernel or QEMU files, right? I know, having this one will do, but I wonder whether we could have a post merge hook that would update the kernel / initrd files if changed, rather than having a nightly build for those.
> 
> I did not understand that statement about patching qemu and kernel.

Let's say, I send a patch touching `tools/packaging/kernel`.  In this case, I'd like the kernel to be built rather than using a cached one.
Do we already handle this case?

> 
> About the post-merge hook to trigger the kernel build once its version/configuration is updated, I think it is a good idea. But I prefer to get it done later. I still would like to work on caching the other assets: qemu, osbuilder, etc.

Oh, sorry, I didn't mean to implement this as part of this PR, but rather as an improvement in the future.
By the way I'd focus mostly on caching kernel & qemu for now.

jcvenegas:
@fidencio I think the `84` in `5.10.25-84-experimental` (from the `latest` file in the Jenkins job) should be mark as flag to know if the build changed, this is because new patches or even the build scripts changed. But I agree that we could do better, what I would like to try in the future is have a self-contained build setup.  This that given have all the files that are required to build an binary in one single directory, then just do a hash using all the files in the directory, if something changes in the directory, the hash changes and the cached asset is not valid so would require to build again.

jcvenegas (approved):
using an experimental job to try this makes sense to me, let see how it goes  to keep adding even not experimental builds

fidencio (approved):
lgtm, thanks a whole lot @wainersm!

GabyCT (approved):

wainersm:
> @fidencio I think the `84` in `5.10.25-84-experimental` (from the `latest` file in the Jenkins job) is used as a flag to know if the build changed, this is because new patches or even the build scripts changed. But I agree that we could do better, what I would like to try in the future is have a self-contained build setup. This that have all the files that are required to build a binary in one single directory, then just do a hash using all the files in the directory, if something changes in the directory, the hash changes and the cached asset is not valid so would require to build again.

IIRC the static-check will fail if you touched any of the kernel files and not bumped the configuration version

Yeah, the approach using hashes would be way to go.

wainersm:
> lgtm, thanks a whole lot @wainersm!

I didn't do much actually :) Kudos to those who implemented this whole (nice) cache mechanism.

Issue #3478, opened 2021-04-30
ARM CI failures in kata 2.x


GabyCT:
I have seen a lot of failures in ARM CI with the following
```
15:53:00 INFO: Containerd checks
15:53:00 bash integration/containerd/cri/integration-tests.sh
15:53:00 INFO: Stop crio service
15:53:00 INFO: Stop containerd service
15:53:01 Install cni config
15:53:02 Configure CNI
15:53:02 ~/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/containerd/cri ~/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/kata-containers/tests
15:53:02 make[1]: Entering directory '/home/jenkins/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/containerd/cri'
15:53:02 🇩 clean
15:53:02 make[1]: Leaving directory '/home/jenkins/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/containerd/cri'
15:53:02 INFO: containerd(cri): Check daemon works with runc
15:53:02 [plugins]
15:53:02   [plugins.cri]
15:53:02     [plugins.cri.containerd]
15:53:02 	default_runtime_name = "runc"
15:53:02       [plugins.cri.containerd.runtimes.runc]
15:53:02         runtime_type = "io.containerd.runtime.v1.linux"
15:53:02         [plugins.cri.containerd.runtimes.runc.options]
15:53:02           Runtime = "/usr/local/sbin/runc"
15:53:02 [plugins.linux]
15:53:02        shim = "/usr/local/bin/containerd-shim"
15:53:02 make[1]: Entering directory '/home/jenkins/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/containerd/cri'
15:53:02 🇩 _output/integration.test
15:53:02 go test -c github.com/containerd/cri/integration -o _output/integration.test
15:53:02 🇩 _output/containerd
15:53:17 go build -o _output/containerd \
15:53:17 	-tags 'seccomp apparmor' \
15:53:17 	-ldflags '-X github.com/containerd/cri/vendor/github.com/containerd/containerd/version.Version=f864905c-TEST' \
15:53:17 	-gcflags '' \
15:53:17 	github.com/containerd/cri/cmd/containerd
15:53:17 🇩 binaries
[...]

PR #3493, opened 2021-05-05
ci: workaround to fix arm ci failure on behalf of containerd and k8s test

jongwu:
Currently, before containerd integration test, containerd service is
stopped., which will disable docker service. Then, the TestImageLoad
will fail for it depends on docker service.

Here, we restart docker before containerd test to avoid that failure.

But these tests have been working well for years. I can't tell why it
is.

Fixes: #3478
Signed-off-by: Jianyong.wu <jianyong.wu@arm.com>

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -70,7 +70,7 @@ esac
 check_processes
 
 # Remove existing CNI configurations:
-cni_config_dir="/etc/cni/net.d"
+cni_config_dir="/etc/cni"
 cni_interface="cni0"
 sudo rm -rf /var/lib/cni/networks/*
 sudo rm -rf "${cni_config_dir}"/*
@@ -109,6 +109,13 @@ trap 'sudo -E sh -c "rm -r "${kubeadm_config_file}""' EXIT
 if [ "${BAREMETAL}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
 	sudo swapoff -a || true
 fi
+
+#reinstall kubelet to do deep cleanup
+if [ "$(command -v kubelet)" != "" ]; then
+	info "reinstall kubeadm, kubelet before initialize k8s"
+	bash -f "${SCRIPT_PATH}/../../.ci/install_kubernetes.sh"
+fi
+
 sudo -E kubeadm init --config "${kubeadm_config_file}"
 
 mkdir -p "$HOME/.kube"
--- a/.ci/install_kubernetes.sh
+++ b/.ci/install_kubernetes.sh
@@ -21,6 +21,9 @@ kubernetes_version=$(get_version "externals.kubernetes.version")
 ARCH=$("${cidir}"/kata-arch.sh -d)
 
 if [ "$ID" == "ubuntu" ] || [ "$ID" == "debian" ]; then
+	if [ "$(command -v kubelet)" != "" ]; then
+		apt purge kubelet -y
+	fi
 	sudo bash -c "cat <<EOF > /etc/apt/sources.list.d/kubernetes.list
 	deb http://apt.kubernetes.io/ kubernetes-xenial-unstable main
 EOF"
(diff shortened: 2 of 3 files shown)


integration/kubernetes/init.sh
```diff
@@ -109,6 +111,14 @@ trap 'sudo -E sh -c "rm -r "${kubeadm_config_file}""' EXIT
 if [ "${BAREMETAL}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
 	sudo swapoff -a || true
 fi
+
+#deep cleanup k8s in arm64 environment
+if [ "${arch}" == "arm64" ]; then
+	info "reinstall kubeadm kubelet before initialize k8s"
+	sudo apt purge kubelet -y
+	sudo apt install kubelet kubeadm -y
```
fidencio, line 119: Oi!

This is quite error prone.  We shouldn't be uninstalling kubelet (and all its dependencies) here and then installing it again, and if we do, we should be ensuring the version to be installed, as done in: https://github.com/kata-containers/tests/blob/main/.ci/install_kubernetes.sh

I wonder why the cleanup functions (https://github.com/kata-containers/tests/blob/main/.ci/aarch64  /clean_up_aarch64.sh) are not working properly and I wonder whether patching it there would be a better solution.

fidencio (changes requested):
I've made one point related to purging and reinstalling kubelet & its dependencies.  We shouldn't be doing that, at leasst not in the way proposed in this patch, as we may end up running on issues related to the version installed and whatnot.

integration/kubernetes/init.sh
jongwu: I try to cleanup the host hard, but it doesn't work. I have no other way than reinstall the kubelet. can I reinstall kubelet of the right version here?

fidencio: What about:
1. Check whether we have something installed and, if we do, we purge it as part of https://github.com/kata-containers/tests/blob/main/.ci/install_kubernetes.sh ... that would be the very first step before actually installing something;
2. We just call that script here (as that script already takes care of using the correct versions and whatnot);

IMHO, we could even do it independently of the architecture.

Does this sound reasonable?

jongwu: Nice! thanks @fidencio, let me try it.

fidencio:
Right now we got 2 green runs in a row.  I'd suggest to run `/retest-arm` a few more times Tomorrow to ensure things won't break.

fidencio:
Hmm. Interesting:
```
09:10:08 ctr: failed to launch qemu: exit status 1, error messages from qemu log: WARNING: Image format was not specified for '/usr/share/kata-containers/kata-flash0.img' and probing guessed raw.
09:10:09          Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
09:10:09          Specify the 'raw' format explicitly to remove the restrictions.
09:10:09 WARNING: Image format was not specified for '/usr/share/kata-containers/kata-flash1.img' and probing guessed raw.
09:10:09          Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
09:10:09          Specify the 'raw' format explicitly to remove the restrictions.
09:10:09 qemu-system-aarch64: warning: gic-version=host not relevant with kernel-irqchip=off as only userspace GICv2 is supported. Using v2 ...
09:10:09 qemu-system-aarch64: Number of SMP CPUs requested (46) exceeds max CPUs supported by machine 'mach-virt' (8)
09:10:09 : unknown
09:10:09 Ending hypervisor stability test
09:10:09 ctr: container id must be provided
09:10:09 Makefile:135: recipe for target 'stability' failed
09:10:09 make: *** [stability] Error 1
09:10:09 Failed at 22: sudo -E PATH="$PATH" CRI_RUNTIME="containerd" bash -c "make stability"
09:10:09 Build step 'Execute shell' marked build as failure
```

There was a whole bunch of those, and those are not related to the issue @jongwu was primarily solving.

jongwu:
Hello @fidencio -, I add a new jenkins node today, but the kernel version is too low that leads to that issue. Thus I mark it offline then there is only one jenkins node which can work properly for now.

fidencio:
@jongwu, right!
Let's keep calling `/retest-arm` here and we can keep posting the results ... and by the end of Today / beginning of the next week we can have this PR merged.

fidencio:
@GabyCT, @wainersm, would you mind to take a look at this one?

One future improvement that I would like to have in the future, is specifically having a BAREMETAL target, so we do some assumptions / take some paths only if the BAREMETAL target is taken.  This would benefit all the baremetal CIs equally, rather than having things done specifically by arch.

fidencio:
@jongwu, while here, mind to also add the `Signed-off-by: ...` tag to your commits? :-)

jongwu:
yeah, thanks @fidencio

fidencio (approved):
lgtm, and seems stable enough!
Thanks @jongwu!

@GabyCT, mind to take a look if this approach is acceptable?

jongwu:
Thanks @fidencio -, it seems jenkins-ci-ubuntu-20-04 pending for so long a time, what's wrong with it?

c3d (approved):
Looks good to me.

> But these tests have been working well for years. I can't tell why it is.

Is it possible that running non-containerd tests impacts the configuration?

PR #3498, opened 2021-05-05
CI: Use cached QEMU

### LINKED ISSUE #3119: CI: cache qemu, kernel, image for 2.x

wainersm:
his changed the installer script to fetch and install a built QEMU
if a suitable version is found on the project's Jenkins.

To detect changes that would flag a cache mismatch it is employed
a simple algorithm to generate a SHA-256 from the files and scripts
used to build QEMU. The cached and current sha-256 is compared to
determine whether or not use the cached asset.

Fixes #3119

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -417,3 +417,52 @@ get_pr_changed_file_details()
 {
 	get_pr_changed_file_details_full | grep -v "vendor/"
 }
+
+
+# Gets a list of files and/or directories to build a SHA-256 from their contents.
+# Returns the SHA-256 hash if succeeded, otherwise an empty string.
+sha256sum_from_files() {
+	local files_in=${1:-}
+	local files=""
+	local shasum=""
+
+	# Process the input files:
+	#  - discard the files/directories that don't exist.
+	#  - find the files if it is a directory
+	for f in $files_in; do
+		if [ -d "$f" ]; then
+			files+=" $(find $f -type f)"
+		elif [ -f "$f" ]; then
+			files+=" $f"
+		fi
+	done
+	# Return in case there is none input files.
+	[ -n "$files" ] || return 0
+
+	# Alphabetically sorting the files.
+	files="$(echo $files | tr ' ' '\n' | LC_ALL=C sort -u)"
+	# Concate the files and calculate a hash.
+	shasum="$(cat $files | sha256sum -b)" || true
+	if [ -n "$shasum" ];then
+		# Return only the SHA field.
+		echo $(awk '{ print $1 }' <<< $shasum)
+	fi
+}
+
+# Calculate a SHA-256 from all the files used to build QEMU.
+# Use this function to detect changes on build scripts which should force a
+# local build of QEMU.
+#
+# Note: bear in mind it is not used a comprehensive list of files but only
+# those that seems sufficient to detect changes. For example, this script is
+# sourced by many others but it is not considered.
+calc_qemu_files_sha256sum() {
+	local pkg_dir="${kata_repo_dir}/tools/packaging"
+	local files="${pkg_dir}/qemu \
+		${pkg_dir}/static-build/qemu \
+		${pkg_dir}/static-build/qemu.blacklist \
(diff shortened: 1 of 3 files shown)


wainersm:
@GabyCT @jcvenegas @chavafg hello!

This is just a draft but I would like to hear from you. Unlike kernel, we don't have a configuration number to check if the current and cached asset version match. As @jcvenegas suggested on #3489 , here I "implement" the idea of hashing all the files which are used to build qemu to compound a 'version sha' string which could be used for comparison.

.ci/lib.sh
```diff
@@ -417,3 +417,24 @@ get_pr_changed_file_details()
...
+}
+
+gen_qemu_files_sha256sum() {
+	local pkg_dir="${kata_repo_dir}/tools/packaging"
+	# Files and dirs
+	files="${pkg_dir}/qemu \
+		${pkg_dir}/static-build/qemu \
+		${pkg_dir}/static-build/qemu/qemu.blacklist \
+		${pkg_dir}/static-build/scripts \
+		${pkg_dir}/scripts"
```
fidencio, line 435: I'd also add the QEMU version, coming from the version file, to the dance.

Right now we have patches which would accurately get any new change, but if we ever happen to use vanilla QEMU, not having the version could mislead us.

wainersm: Good catch @fidencio ! I will add that file too.

.ci/lib.sh
```diff
@@ -417,3 +417,24 @@ get_pr_changed_file_details()
...
+sha256sum_from_files() {
+	# TODO: implement me!
+	echo "zzzzzzzzzzzzzzzzz"
+
+}
+
+gen_qemu_files_sha256sum() {
+	local pkg_dir="${kata_repo_dir}/tools/packaging"
+	# Files and dirs
+	files="${pkg_dir}/qemu \
```
jcvenegas, line 431: I think some of the scripts we use do `source` to other  scripts , even if we can not control what is sourced, it is still  a better approach than a number (that we need to modify manually) and I like the simplicity of the initial idea.

wainersm: Thanks for the feedback @jcvenegas . Still today I plan to send a working version of this PR.

wainersm:
Updated this PR with a version which is good for reviews.

I will trigger some jobs that should build QEMU locally given that a suitable cached assert does not exist yet. Then later I will create the QEMU cahe job to check it can properly pull from there.

GabyCT (approved):

wainersm:
@fidencio @GabyCT @jcvenegas , hi folks,

I created the job on Jenkins (http://jenkins.katacontainers.io/job/kata-containers-2.0-qemu-x86_64) which is already caching QEMU. The job's script pulls this PR, so it will needed to be adjusted afterwards.

I still have some improvements to be done on the algorithm that calculate the sha-256 to make it more reliable. But I will send it in another PR later.

It seems the mechanism is working fine. We can merge it, unless you find bugs or want me to implement more changes.

GabyCT:
> @fidencio @GabyCT @jcvenegas , hi folks,
> 
> I created the job on Jenkins (http://jenkins.katacontainers.io/job/kata-containers-2.0-qemu-x86_64) which is already caching QEMU. The job's script pulls this PR, so it will needed to be adjusted afterwards.
> 
> I still have some improvements to be done on the algorithm that calculate the sha-256 to make it more reliable. But I will send it in another PR later.
> 
> It seems the mechanism is working fine. We can merge it, unless you find bugs or want me to implement more changes.

thank you @wainersm

devimc (approved):
thanks @wainersm

fidencio (approved):
lgtm, thanks @wainersm!

Issue #3505, opened 2021-05-07
We don't run CRI-O unit tests against the `main`repo


fidencio:
As we still didn't get CRI-O unit tests to run against shimv2, we removed those from our tree as part 816aee46241808c7261107aa5491fb529e43eef8.  However, we left some documentation behind and that will break the static check.

PR #3506, opened 2021-05-07
readme: Remove reference to cri-o unit tests

fidencio:
As we still didn't get CRI-O unit tests to run against shimv2, we
removed those from our tree as part 816aee4. However, we left some
documentation behind and that will break the static check.

Fixes: #3505


wainersm (approved):

c3d (approved):
Thanks @fidencio

PR #3513, opened 2021-05-11
stable-2.1 | ci: workaround to fix arm ci failure.

fidencio:
Currently, before containerd integration test, containerd service is
stopped., which will disable docker service. Then, the TestImageLoad
will fail for it depends on docker service.

Here, we restart docker before containerd test to avoid that failure.

But these tests have been working well for years. I can't tell why it
is.

Also do cleanup before k8s initialization to avoid lots of bizarre
issues.
for example:
1. [ERROR Port-10250]: Port 10250 is in use
2. Unfortunately, an error has occurred:
		timed out waiting for the condition

Fixes: #3478
(cherry picked from commit d7ca6409275b88ef12fad30ec7c8c188e8766f47)

### DIFF
--- a/.ci/install_kubernetes.sh
+++ b/.ci/install_kubernetes.sh
@@ -21,6 +21,9 @@ kubernetes_version=$(get_version "externals.kubernetes.version")
 ARCH=$("${cidir}"/kata-arch.sh -d)
 
 if [ "$ID" == "ubuntu" ] || [ "$ID" == "debian" ]; then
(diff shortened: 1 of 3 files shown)


fidencio:
@jongwu :-)

jongwu:
Thanks @fidencio!

GabyCT (approved):

devimc (approved):

Issue #3516, opened 2021-05-11
Only uninstall kubernetes / restart docker when running a baremetal CI.


fidencio:
https://github.com/kata-containers/tests/pull/3493 fixed the issues we had with the ARM CI, but the changes done were done for any environment.

Let's restrict those changes a little bit and only run those on baremetal CIs.

c3d:
@fidencio I don't fully understand the rationale. Why would things be different or better in a VM? The changes in #3493 restart the docker daemon. Presumably, a same VM can be used for multiple tests and we'd see the same problem, no?

PR #3517, opened 2021-05-11
Restrict baremetal specific changes to baremetal CIs

fidencio:
#3493 fixed the issues we had with the ARM CI, but the changes done were done for any environment.

Let's restrict those changes a little bit and only run those on baremetal CIs.

Fixes: #3516

### DIFF
--- a/integration/containerd/cri/integration-tests.sh
+++ b/integration/containerd/cri/integration-tests.sh
@@ -168,19 +168,19 @@ restart_docker() {
 	#avoid the "Start request repeated too quickly" error
         if [ -f "/lib/systemd/system/docker.service" ]; then
 		cp /lib/systemd/system/docker.service ${back_file}
-                sed -i 's/StartLimitBurst.*$/StartLimitBurst=0/g' /lib/systemd/system/docker.service
+                sudo sed -i 's/StartLimitBurst.*$/StartLimitBurst=0/g' /lib/systemd/system/docker.service
         elif [ -f "/usr/lib/systemd/system/docker.service" ]; then
 		cp /usr/lib/systemd/system/docker.service ${back_file}
-                sed -i 's/StartLimitBurst.*$/StartLimitBurst=0/g' /usr/lib/systemd/system/docker.service
+                sudo sed -i 's/StartLimitBurst.*$/StartLimitBurst=0/g' /usr/lib/systemd/system/docker.service
         fi
         sudo systemctl daemon-reload
         sudo systemctl restart docker
 
 	#recover docker service file
 	if [ -f "/lib/systemd/system/docker.service" ]; then
-		mv ${back_file} /lib/systemd/system/docker.service
+		sudo mv ${back_file} /lib/systemd/system/docker.service
 	elif [ -f "/usr/lib/systemd/system/docker.service" ]; then
-		mv ${back_file} /usr/lib/systemd/system/docker.service
+		sudo mv ${back_file} /usr/lib/systemd/system/docker.service
 	fi
         sudo systemctl daemon-reload
 }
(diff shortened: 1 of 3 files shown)


integration/containerd/cri/integration-tests.sh
```diff
@@ -190,7 +190,7 @@ check_daemon_setup() {
 
 	if [ "${BAREMETAL}" == true ]; then
 		#restart docker service as TestImageLoad depends on it
-		restart_docker
+		sudo -E restart_docker
 	fi
```
jongwu, line 194: Why do we need "sudo" here? And it lead to the failure in arm ci.
`10:49:34 sudo: restart_docker: command not found`

jongwu:
@fidencio - arm ci turns green.

integration/kubernetes/init.sh
```diff
@@ -108,12 +108,12 @@ trap 'sudo -E sh -c "rm -r "${kubeadm_config_file}""' EXIT
 
 if [ "${BAREMETAL}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
```
jongwu, line 109: We need let the  "BAREMETAL" check separate from the swap check. So what about:
"
if [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
         sudo swapoff -a || true
fi

if [ "${BAREMETAL}" == true &&  "$(command -v kubelet)" != "" ]; then
       #reinstall kubelet to do deep cleanup
      	info "reinstall kubeadm, kubelet before initialize k8s"
	bash -f "${SCRIPT_PATH}/../../.ci/install_kubernetes.sh"	
fi
"

jongwu: The Arm Ci fails as the "wc -l /proc/swaps | awk '{print $1}'" is "1" in the Arm server which results in the "reinstall k8s" ops didn't start.

fidencio: I'm not sure about this, @jongwu.

The baremetal check to disable swap has always been there and shouldn't cause an issue (it wasn't already).
I'll split the check, and let it run a few more times.

fidencio: Ah, now I see, sorry, my bad!

integration/kubernetes/init.sh
```diff
@@ -111,7 +111,7 @@ if [ "${BAREMETAL}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt
 fi
 
 #reinstall kubelet to do deep cleanup
-if [ "$(command -v kubelet)" != "" ]; then
+if [ "{BAREMETAL}" == true && "$(command -v kubelet)" != "" ]; then
```
jongwu, line 114: Hi @fidencio -, Try `if [ "${BAREMETAL}" == true -a "$(command -v kubelet)" != "" ]; then`

integration/kubernetes/init.sh
```diff
@@ -111,7 +111,7 @@ if [ "${BAREMETAL}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt
 fi
 
 #reinstall kubelet to do deep cleanup
-if [ "$(command -v kubelet)" != "" ]; then
+if [ "{BAREMETAL}" == true -a "$(command -v kubelet)" != "" ]; then
```
jongwu, line 114: miss "$" before "{BAREMETAL}"

fidencio: Lack of attention is  killing me. :-)

fidencio:
Okay, it passes, now let's give it a try on a second run.

fidencio:
@jongwu, now it's good to go! :-)

jcvenegas (approved):

fidencio:
@snir911, may I bother you for an ack?

snir911 (approved):
LGTM!

PR #3520, opened 2021-05-12
k8s: enable cri runtime service rather than start it

### LINKED ISSUE #3519: Integration test for kubernetes: Enable cri runtime rather than start it
While using the init.sh script to setup a local dev environment, I've found that the cri runtime service was not enabled, and I had to re-start it on reboots.
The init script is starting the service, and doesn't need to enable it for its purpose, but doing so can simplify things for other use cases.

This is not a big issue, and doesn't affect CI, but it can help setup local environments.

littlejawa:
While using the init.sh script to setup a local dev environment, I've found
that the cri runtime service was not enabled, and I had to re-start it on
reboots.
Enabling it makes sure it is always available.

Fixes #3519

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -80,7 +80,7 @@ if ip a show "$cni_interface"; then
 fi
 
 echo "Start ${cri_runtime} service"
-sudo systemctl start ${cri_runtime}
+sudo systemctl enable --now ${cri_runtime}
 max_cri_socket_check=5
 wait_time_cri_socket_check=5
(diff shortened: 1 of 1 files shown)


fidencio (approved):
@littlejawa, lgtm, thanks!

Do we need to do the same for kubelet service?

littlejawa:
> Do we need to do the same for kubelet service?

no, kubelet was running in my VM, only crio needed to be restarted.

GabyCT (approved):

Issue #3515, opened 2021-05-11
kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd is randomly failing


fidencio:
I've noticed this one started happening after merging https://github.com/kata-containers/tests/pull/3489 and it makes me wonder whether we're skipping the installation of rustup for some reason or another.
```
12:44:07 I am ubuntu or debian
12:44:07 INFO: Configure chrony file /tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/kata-containers/tools/osbuilder/ubuntu_rootfs/etc/chrony/chrony.conf
12:44:07 /tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/kata-containers/tools/osbuilder/rootfs-builder/rootfs.sh: line 576: rustup: command not found
12:44:07 /tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/tests /tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/kata-containers/tools/osbuilder
12:44:07 /tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/kata-containers/tools/osbuilder
12:44:07 INFO: Build agent
12:44:07 /tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/kata-containers/src/agent /tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/kata-containers/tools/osbuilder
12:44:07 INFO: checkout failed!
12:44:07 make[1]: Entering directory '/tmp/jenkins/workspace/kata-containers-tests-ubuntu-PR-cloud-hypeprvisor-k8s-containerd/go/src/github.com/kata-containers/kata-containers/src/agent'
12:44:07 /bin/sh: 1: cargo: not found
12:44:07 make[1]: cargo: Command not found
[...]

PR #3528, opened 2021-05-14
ci: Ensure we install rust whenever we're installing kata

fidencio:
Let's ensure we will **always** install rust whenever kata-containers
will be installed.

I know, this is theoretically done on the osbuilder side, but we, as
part of the CI, should ensure this environment is ready from our side.

Fixes: #3515


egernst:
Is this fixing the issue it is associated with?

fidencio:
I wish, but that one happens inside the container, AFAIU.

I'll close this one as @pmores mentioned he'd like to investigate it.

Issue #3531, opened 2021-05-14
KSM unstable with virtio-fs


Jakob-Naucke:
KSM (Kernel Samepage Merging) test was disabled for ARM in #3441. It turns out that when using virtio-fs, due to having file-backed memory, this doesn't work to its full extent, works only sometimes (ARM, also x86 in my tests), or not at all (s390x). virtio-9p should be used for this test instead. PR is prepared.

PR #3532, opened 2021-05-14
 Disable virtio-fs for KSM

Jakob-Naucke:
Due to virtio-fs using file-backed memory, KSM (Kernel Samepage Merging) is unstable or, depending on architecture, might not work at all.

- Generalize `.ci/toggle_sandbox_cgroup_only.sh` to `.ci/set_kata_config.sh` so it can be used for other keys.
- Disable virtio-fs for KSM test. Keep virtio-9p if it was enabled.
- Give KSM a little time to kick in (just because there are no shared pages *immediately* after starting container doesn't mean it isn't working).
- Re-enable KSM test on ARM. **I'm assuming this will fix ARM, but I don't have the means to test it. Test before merging.**

Fixes: #3531

### DIFF
--- a/.ci/set_kata_config.sh
+++ b/.ci/set_kata_config.sh
@@ -21,9 +21,9 @@ declare -A sections
 declare -A valid_requests
 declare -A json_keys
 
-readonly sections=([sandbox_cgroup_only]=runtime)
-readonly valid_requests=([sandbox_cgroup_only]="true false")
-readonly json_keys=([sandbox_cgroup_only]=.Runtime.SandboxCgroupOnly)
+readonly sections=([sandbox_cgroup_only]=runtime [shared_fs]=hypervisor.qemu)
+readonly valid_requests=([sandbox_cgroup_only]="true false" [shared_fs]="virtio-fs virtio-9p")
+readonly json_keys=([sandbox_cgroup_only]=.Runtime.SandboxCgroupOnly [shared_fs]=.Hypervisor.SharedFS)
 
 option=${1:-}
 request=${2:-}
--- a/.ci/set_kata_config.sh
+++ b/.ci/set_kata_config.sh
@@ -17,36 +17,61 @@ source "${script_dir}/../lib/common.bash"
 # shellcheck source=./lib.sh
 source "${script_dir}/lib.sh"
 
-option="sandbox_cgroup_only"
-request=${1:-}
+declare -A sections
+declare -A valid_requests
+declare -A json_keys
+
+readonly sections=([sandbox_cgroup_only]=runtime)
+readonly valid_requests=([sandbox_cgroup_only]="true false")
+readonly json_keys=([sandbox_cgroup_only]=.Runtime.SandboxCgroupOnly)
+
+option=${1:-}
+request=${2:-}
 
 usage(){
 	cat <<EOT
 Usage:
-${script_name} <true|false|clean>"
+${script_name} <option> <request>
 
-false: Disable ${option}
-true:  Enable ${option}
+Valid options and requests:
+$(for opt in "${!valid_requests[@]}"
+do
+	echo "${opt}": "${valid_requests["$opt"]}"
+done)
 
 The configuration changes are applied in kata user config:
 ${KATA_ETC_CONFIG_PATH}
 
 Remove it if you want to use the stateless options.
 EOT
+	exit 1
 }
 
-case ${request} in
-	true)
-		;;
-	false)
-		;;
-	*)
-		usage
-		exit 1
-		;;
-esac
+# need explicit handling of both cases because all errors are trapped
+valid=$([[ -n "${option}" ]] && echo $? || echo $?)
+# trailing space to match whole words
+if [[ "${valid}" == 0 && ! "${!valid_requests[*]} " =~ "${option} " ]]
+then
+	echo >&2 "ERROR: unknown option: '${option}'"
+	valid=1
+fi
+[[ "${valid}" == 0 ]] || usage
 
-current_value=$(kata-runtime kata-env --json | jq ".Runtime.SandboxCgroupOnly")
+valid=$([[ -n "${request}" ]] && echo $? || echo $?)
+if [[ "${valid}" == 0 && ! "${valid_requests["${option}"]} " =~ "${request} " ]]
+then
+	echo >&2 "ERROR: unknown request: '${request}'"
+	valid=1
+fi
+[[ "${valid}" == 0 ]] || usage
+
+# quote non-boolean request
+[[ "$request" =~ ^(true|false)$ ]] || request="\"${request}\""
+
+readonly section="${sections["${option}"]}"
+readonly json_key="${json_keys["${option}"]}"
+
+current_value=$(kata-runtime kata-env --json | jq "${json_key}")
 if [ "$current_value" == "${request}" ]; then
 	info "already ${request}"
 	exit 0
@@ -73,10 +98,10 @@ if [ "${KATA_ETC_CONFIG_PATH}" != "${kata_config_path}" ]; then
 fi
 
 info "modifying config file : ${KATA_ETC_CONFIG_PATH}"
-sudo crudini --set "${KATA_ETC_CONFIG_PATH}" "runtime" "${option}" "${request}"
+sudo crudini --set "${KATA_ETC_CONFIG_PATH}" "${section}" "${option}" "${request}"
 
 info "Validate option is ${request}"
-current_value=$(kata-runtime kata-env --json | jq ".Runtime.SandboxCgroupOnly")
+current_value=$(kata-runtime kata-env --json | jq "${json_key}")
 
 [ "$current_value" == "${request}" ] || die "The option was not updated"
 
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -286,6 +286,13 @@ wait_ksm_settle(){
 
 	oldscan=$(cat /sys/kernel/mm/ksm/full_scans)
 
+	# Wait some time for KSM to kick in to avoid early dismissal
+	for ((t=0; t<5; t++)); do
+		pages=$(cat "${KSM_PAGES_SHARED}")
+		[[ "$pages" -ne 0 ]] && echo "Discovered KSM activity" && break
+		sleep 1
+	done
+
 	# Go around the loop until either we see a small % change
 	# between two full_scans, or we timeout
 	for ((t=0; t<$1; t++)); do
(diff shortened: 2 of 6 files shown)


fidencio:
@Jakob-Naucke, why you don't have how to test it before it gets merged?

Jakob-Naucke:
@fidencio sorry, what do you mean? Can't we `/test`/`/test-arm` as usual?

fidencio:
> @fidencio sorry, what do you mean? Can't we `/test`/`/test-arm` as usual?

Sorry, I got confused, my bad.  I read your "I'm assuming this will fix ARM, but I don't have the means to test it. Test before merging." as "we don't have means to test it before merging", and I asked myself about "Cant't we `/test-arm` this?" :-)

Anyways, my bad, sorry for the confusion.

GabyCT (approved):

Jakob-Naucke:
On CI failures:

- CLH tests appear to be failing at an earlier stage (Credentials using secrets), so I think it's unrelated?
- ARM failure uncovered a bug: The "do not restore virtio-fs if it was 9p before" functionality is working, but returning 1. Fixed and updated scripts to also print output from the 9p set.
- Metrics failure was because virtio-9p was kept after disabling KSM. Put restore function into `disable_ksm` as well.

Also fixed a subtle logic bug if `restore_virtio_fs` was called without having called `set_ksm_aggressive` and `virtio-9p` was enabled, `virtio-fs` would be set.
Re-requesting review.

GabyCT:
Metrics CI is failing with
```
10:08:15 time="2021-05-18T15:08:15Z" level=debug msg="Compare check for [memory-footprint-ksm]"
10:08:15 time="2021-05-18T15:08:15Z" level=debug msg="Checking value [mean]"
10:08:15 time="2021-05-18T15:08:15Z" level=debug msg=" Check minval (102313.860000 > 47667.200000)"
10:08:15 time="2021-05-18T15:08:15Z" level=warning msg="Failed Minval (102313.860000 > 47667.200000) for [memory-footprint-ksm]"
10:08:15 time="2021-05-18T15:08:15Z" level=debug msg=" Check maxval (113083.740000 < 47667.200000)"
10:08:15 time="2021-05-18T15:08:15Z" level=debug msg=Passed
10:08:15 time="2021-05-18T15:08:15Z" level=warning msg="Check for [memory-footprint-ksm] failed [Failed]"
```

Jakob-Naucke:
Hmm. Now the metrics failure is a bit surprising to me. I thought it was because it was running on 9p (which it was before) and that was going to be slower, but going by the logs now, some lines later:

```
17:08:15 restoring KSM settings
17:08:15 INFO: already "virtio-fs"
```

Also the static checks were cancelled (?)

GabyCT:
@Jakob-Naucke this is the error of static checks
```
GitHub Actions has encountered an internal error when running your job.
```

GabyCT:
I open an issue related with github actions https://github.com/actions/virtual-environments/issues/3412

Jakob-Naucke:
Digging a bit deeper, yes, the `memory-footprint-ksm` test is supposed to run with KSM (duh), and I think it should run with virtio-9p, even with the performance hit that this implies. @gabyct Do you think the targets could be lowered?

GabyCT:
> Digging a bit deeper, yes, the `memory-footprint-ksm` test is supposed to run with KSM (duh), and I think it should run with virtio-9p, even with the performance hit that this implies. @GabyCT Do you think the targets could be lowered?

Yes for me it works fine if we lowered the targets, on the other hand is this PR going to have an impact in https://github.com/kata-containers/tests/blob/main/integration/ksm/ksm_test.sh?

Jakob-Naucke:
> is this PR going to have an impact in https://github.com/kata-containers/tests/blob/main/integration/ksm/ksm_test.sh?

@GabyCT well, `ksm_test.sh` will run on ARM again, and virtio-9p will be used for that test...if that answers your question?

.ci/set_kata_config.sh
```diff
@@ -35,18 +46,30 @@ Remove it if you want to use the stateless options.
...
-	*)
-		usage
-		exit 1
-		;;
-esac
-
-current_value=$(kata-runtime kata-env --json | jq ".Runtime.SandboxCgroupOnly")
+if [[ -z "${option}" || ! ${!valid_requests[*]} =~ ${option} ]]
+then
+	# unknown option
```
jodh-intel, line 51: Why not make this self-documenting?

```bash
echo >&2 "ERROR: unknown option: '$option'"
usage
exit 1
```

Same comment for unknown request below.

.ci/set_kata_config.sh
```diff
@@ -35,18 +46,30 @@ Remove it if you want to use the stateless options.
...
+
+if [[ -z "${request}" || ! ${valid_requests["${option}"]} =~ ${request} ]]
+then
+	# unknown request
+	usage
+	exit 1
+fi
+
+# quote non-boolean request
+if [[ "${request}" != "true" && "${request}" != "false" ]]
```
jodh-intel, line 64: You could simplify this to something like:

```bash
[[ "$request" =~ ^(true|false)$ ]] || request="\"${request}\""
```

metrics/lib/common.bash
```diff
@@ -227,6 +227,17 @@ set_ksm_aggressive(){
 	sudo bash -c "echo ${KSM_AGGRESIVE_PAGES} > ${KSM_PAGES_FILE}"
 	sudo bash -c "echo ${KSM_AGGRESIVE_SLEEP} > ${KSM_SLEEP_FILE}"
 	sudo bash -c "echo 1 > ${KSM_ENABLE_FILE}"
+
+	# Disable virtio-fs and save whether it was enabled previously
+	set_virtio_out=$(sudo -E "${LIB_DIR}/../../.ci/set_kata_config.sh" shared_fs virtio-9p)
+	echo "${set_virtio_out}"
+	grep "already" <<< "${set_virtio_out}" > /dev/null || was_virtio_fs=true;
```
jodh-intel, line 234: Slightly shorter to use `grep -q` rather than redirection.

metrics/lib/common.bash
```diff
@@ -227,6 +227,17 @@ set_ksm_aggressive(){
 	sudo bash -c "echo ${KSM_AGGRESIVE_PAGES} > ${KSM_PAGES_FILE}"
 	sudo bash -c "echo ${KSM_AGGRESIVE_SLEEP} > ${KSM_SLEEP_FILE}"
 	sudo bash -c "echo 1 > ${KSM_ENABLE_FILE}"
+
+	# Disable virtio-fs and save whether it was enabled previously
+	set_virtio_out=$(sudo -E "${LIB_DIR}/../../.ci/set_kata_config.sh" shared_fs virtio-9p)
+	echo "${set_virtio_out}"
```
jodh-intel, line 233: Is this line needed?

metrics/lib/common.bash
```diff
@@ -273,9 +286,13 @@ wait_ksm_settle(){
 
 	oldscan=$(cat /sys/kernel/mm/ksm/full_scans)
 
+	# Wait some time for KSM to kick in to avoid early dismissal
+	local initial_wait=5
+	sleep ${initial_wait}
+
```
jodh-intel, line 292: I'm not sure I understand this. We're giving KSM 5 extra seconds to start but then waiting for 5 less seconds for it to settle? Can't we avoid the `sleep` and just increase the timeout in the loop below (or use a timed wait here for KSM to trigger before entering the loop?)

jodh-intel (commented):
Thanks @Jakob-Naucke. A few comments.

jodh-intel:
Hi @Jakob-Naucke - any update on this?

.ci/set_kata_config.sh
Jakob-Naucke: @jodh-intel PTAL again :) maybe this can be simplified further

.ci/set_kata_config.sh
Jakob-Naucke: Updated.

metrics/lib/common.bash
Jakob-Naucke: This tells you something like `INFO: already "virtio-fs"` if applicable, I think that should be visible.

metrics/lib/common.bash
Jakob-Naucke: Updated!

metrics/lib/common.bash
Jakob-Naucke: So without the proposed change, this is run inside the loop:
```sh
newpages=$(cat "${KSM_PAGES_SHARED}")    
[[ "$newpages" -eq 0 ]] && echo "No need to wait for KSM to settle" && return
```

and if KSM hasn't started its work right there and then, this will terminate.
> Can't we [...] use a timed wait here for KSM to trigger before entering the loop?

Isn't that what I'm doing with the `sleep`?

Jakob-Naucke:
Whoops! That was a rebase without the suggested changes.

metrics/lib/common.bash
jodh-intel: Unless I'm missing something, ... not really, no ;) You're just sleeping in the "hope" that KSM will start in 5 seconds right? If it takes 6 seconds, the loop below will fail. If it starts in 1 second, the test has to wait a further 4 seconds before it can move to the for loop below.

What I'm saying is can't you use an addional for loop like the one below to poll for some (reasonably large amount of) time for KSM to "start"? Once we know it's started, we can then fall into the "wait for KSM to settle" loop below. By having two loops:

- you can make the logic clearer
- you don't penalise systems where KSM starts quickly as you can jump out of the first loop and hit the second where you wait for KSM to settle.

Jakob-Naucke: Ah, now I get what you meant :) Let me get back.

Jakob-Naucke: Updated.

fidencio (approved):
@Jakob-Naucke, the series look good and clean!

I've noticed the static checks were failing, but the failures didn't seen related at all, thus I just restarted the checks.

Thanks for the work on this.

@jodh-intel, do you want to take a look at the final version? :-)

fidencio:
Note to self: We should **only** be used `sandbox_cgroup_only=true`, everywhere. But that's a subject for a different discussion.

fidencio:
GitHub action seems to be busted right now.
```
GO111MODULE=on go mod edit -replace=github.com/kata-containers/tests/metrics/exec=../exec
GO111MODULE=on go mod tidy
GO111MODULE=on go mod edit -replace=github.com/kata-containers/tests/metrics/exec=../exec
GO111MODULE=on go mod tidy
go mod tidy: go.mod file indicates go 1.16, but maximum supported version is 1.15
```

PR #3540, opened 2021-05-17
tests: Remove popular docker tests

### LINKED ISSUE #3539: remove popular docker tests

GabyCT:
Currently in kata 2.x we do not support docker and this test can't be
modify to use ctr to docker pull limitations, we should remove it.

Fixes #3539


devimc (approved):

fidencio:
This is sad and I wonder if there would be a way to work this issue around.
I wonder how much a docker hub subscription costs and whether we should actually pay for that, and take advantage of that as part of our CI.

/cc @kata-containers/architecture-committee

fidencio (approved):
anyways, we can revert this one in case we figure out a way to work around the dockerhub issue.

PR #3542, opened 2021-05-18
ci: Remove podman installation

### LINKED ISSUE #3541: remove podman installation
we do not support podman in kata 2.x, we should remove the installation scripts as they are not supported and no longer used

GabyCT:
Podman is not being supported in kata 2.x, we should remove installation
scripts as they are not longer used.

Fixes #3541


.ci/setup_env_fedora.sh
```diff
@@ -10,17 +10,9 @@ set -e
 cidir=$(dirname "$0")
 source /etc/os-release || source /usr/lib/os-release
 source "${cidir}/lib.sh"
-TEST_CGROUPSV2="${TEST_CGROUPSV2:-false}"
```
fidencio, line 13: What's the `TEST_CGROUPSV2` about, @GabyCT?

.ci/setup_env_fedora.sh
```diff
@@ -10,17 +10,9 @@ set -e
...
-
 echo "Install chronic"
 sudo -E dnf -y install moreutils
 
-if [ "${TEST_CGROUPSV2}" == "true" ]; then
-	echo "Install podman"
-	version=$(get_test_version "externals.podman.version")
-	sudo -E dnf -y install podman-"${version}"
```
fidencio, line 21: I think we should still install podman, but do it in the same way we do for docker and do not rely on a specific version anymore.

We still can use podman together with osbuilder and this should be, somehow, supported and tested.

.ci/setup_env_fedora.sh
GabyCT: @fidencio it was a flag used for the CI of podman so we can deleted it

.ci/setup_env_fedora.sh
GabyCT: @fidencio why we want to install podman? what is the benefit? what are we going to test?

GabyCT: @fidencio we are installing docker for other purposes but I do not see the benefit of using podman when we have nothing to test with kata 2.x

.ci/setup_env_fedora.sh
GabyCT: Currently in kata 2.x, in no one of the CIs this flag has been active, this flag belonged to the podman CI in kata 1.x

.ci/setup_env_fedora.sh
GabyCT: currently in kata 2.x, all these scripts have not been executed at any time, mainly because the CI flag that was used to run all these steps was used in kata 1.x but never used in kata 2.x

fidencio: Hmmm.  Don't we have the option of using osbuilder with either podman or docker?  I thought we had some test covering both.

If no, then we can remove it now and, maybe, later on, add a test to ensure we actually can use osbuilder with docker.

fidencio (approved):
lgtm, thanks for the patch and for the explanations, @GabyCT!

.ci/setup_env_fedora.sh
GabyCT: for podman in kata 1.x, we needed to create a custom image and it was tagged with that flag

devimc (approved):

Jakob-Naucke:
Sorry for bump after merge, but I'm running into this just now.
On Fedora-likes, ppc64le & s390x cannot use Docker because it's not packaged (neither by the distro nor by https://download.docker.com like for x86_64 & aarch64).
I see two ways moving forward for them:

1. Build guest rootfs on the host -- requires some changes here in `tests`.
2. Build guest rootfs with Podman -- requires re-enabling of Podman installation (plus some changes to use it if applicable).

Jakob-Naucke:
@fidencio :+1: will do

PR #3568, opened 2021-05-25
[stable-2.1] Restrict baremetal specific changes to baremetal CIs

fidencio:
#3493 fixed the issues we had with the ARM CI, but the changes done were done for any environment.

Let's restrict those changes a little bit and only run those on baremetal CIs.

Fixes: #3516


snir911 (approved):
LGTM! I'd also mention it's porting of #3517

jcvenegas (approved):

Issue #3592, opened 2021-06-07
integration: add CRI-O integration tests supporting kata 2.x


fgiudici:
We removed old CRI-O integration tests  (https://github.com/kata-containers/tests/issues/3411): the main reason is that they were not 2.x ready and the CRI-O ansible playbook for used for running integration tests also supports kata now.
The way tests are run in Kata is different though, so would be good to have back the Kata integration tests with CRI-O here, to be run easily (make crio) but updated to run with kata 2.x.
Since those tests only require kata installed and configured (cri-o would be downloaded and compiled but not installed and tests will be performed just using crictl without kubernetes) they could be handy to allow any developer to quickly (and cleanly) test locally that CRI-O integration works.

PR #3593, opened 2021-06-07
integration/cri-o: add integration tests kata 2.x ready

fgiudici:
Add back integration tests with CRI-O but 2.x ready as per issue https://github.com/kata-containers/tests/issues/3592.
Once kata is installed on the system, "make crio" should allow the tests to be run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,10 @@ else
 	bash sanity/check_sanity.sh
 endif
 
+crio:
+	bash .ci/install_bats.sh
+	RUNTIME=${RUNTIME} ./integration/cri-o/cri-o.sh
+
 docker-stability:
 	systemctl is-active --quiet docker || sudo systemctl start docker
 	cd integration/stability && \
@@ -178,6 +182,7 @@ help:
 .PHONY: \
 	check \
 	checkcommits \
+	crio \
 	docker \
 	docker-stability \
 	filesystem \
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ endif
 
 crio:
 	bash .ci/install_bats.sh
-	RUNTIME=${RUNTIME} ./integration/cri-o/cri-o.sh
+	./integration/cri-o/cri-o.sh
 
 docker-stability:
 	systemctl is-active --quiet docker || sudo systemctl start docker
--- a/integration/cri-o/cri-o.sh
+++ b/integration/cri-o/cri-o.sh
@@ -14,7 +14,10 @@ source "${SCRIPT_PATH}/../../metrics/lib/common.bash"
 source /etc/os-release || source /usr/lib/os-release
 
 export JOBS="${JOBS:-$(nproc)}"
-export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-$RUNTIME}"
+export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-containerd-shim-kata-v2}"
+export CONTAINER_DEFAULT_RUNTIME="${CONTAINER_DEFAULT_RUNTIME:-$CONTAINER_RUNTIME}"
+export RUNTIME_ROOT="${RUNTIME_ROOT:-/run/vc}"
+export RUNTIME_TYPE="${RUNTIME_TYPE:-vm}"
 
 # Skip the cri-o tests if TEST_CRIO is not true
 # and we are on a CI job.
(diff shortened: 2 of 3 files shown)


fgiudici:
CC: @fidencio , @littlejawa

Makefile
```diff
@@ -99,6 +99,10 @@ else
 	bash sanity/check_sanity.sh
 endif
 
+crio:
+	bash .ci/install_bats.sh
+	RUNTIME=${RUNTIME} ./integration/cri-o/cri-o.sh
```
fidencio, line 104: Which versions of CRI-O we care about testing?
RUNTIME doesn't work with newer versions, those use CONTAINER_DEFAULT_RUNTIME, IIRC.
CONTAINER_DEFAULT_RUNTIME doesn't work with older versions, though.

So, the best way to go for now is exporting both.

integration/cri-o/cri-o.sh
```diff
@@ -0,0 +1,92 @@
...
+# Skip CRI-O tests that currently are not working
+pushd "${crio_repository_path}/test/"
+for i in "${skipCRIOTests[@]}"
+do
+	sed -i '/'${i}'/a skip \"This is not working\"' "$GOPATH/src/${crio_repository}/test/ctr.bats"
+done
+
+IFS=$OLD_IFS
+
+# On other distros or on ZUUL, use overlay.
```
fidencio, line 74: ZUUL is no more a thing we use. :-)

Anyways, we should always use overlay with CRI-O, as devmapper is not exactly maintained there.

integration/cri-o/cri-o.sh
```diff
@@ -0,0 +1,92 @@
...
+# This will allow us to run tests with at least 2 different
+# storage drivers.
+export STORAGE_OPTIONS="$OVERLAY_STORAGE_OPTIONS"
```
fidencio, line 77: I see the point of doing this, but if unless we get it to fully work and being fully maintained on CRI-O side, I wouldn't bother testing with device-mapper there.

fidencio (commented):
@fgiudici, thanks for working on this, I've added a few comments as a review.

In general, solid work.

fidencio:
And the static checks are failing due to:
```
 ERROR: Commit 4f783e3bed9a5fcd80bdae2da69e55512635a22c: Failed to find subsystem in subject: "Revert \"integration: Remove old CRI-O tests\""
ERROR: checkcommits failed. See the document below for help on formatting
commits for the project.

https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format
```

Makefile
fgiudici: Uh, missed it!  I moved the env vars directly in the cri-o.sh script, RUNTIME here was ignored, so removing it!
The vars have shimv2 compatible defaults that will be picked up for CRI-O configuration by the CRI-O test script. The user may still export its own values if needed (exporting CONTAINER_DEFAULT_RUNTIME, RUNTIME_ROOT and RUNTIME_TYPE) but I would leave them out of the Makefile... I don't expect a use case where the user needs to rediefine the vars too. Could this option be viable?


> And the static checks are failing due to:
> 
> ```
> ERROR: Commit 4f783e3bed9a5fcd80bdae2da69e55512635a22c: Failed to find subsystem in subject: "Revert \"integration: Remove old CRI-O tests\""
> ERROR: checkcommits failed. See the document below for help on formatting
> commits for the project.
> 
> https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format
> ```

integration/cri-o/cri-o.sh
fgiudici: Oh, sure, removed :-)

integration/cri-o/cri-o.sh
fgiudici: Ok, I left the variable exported with the driver explicitly set to "overlay". This should be the default in CRI-O, so probably not needed...  but leaving it explicit doesn't harm, moved near the exported vars for CRI-O test script.

fgiudici:
@fidencio , many thanks for the review (and the hint for resolving the static check issue)!
Enforced also build of CRI-O binaries: we may download the repo or just change the branch, so let's ensure we also (re)build the binaries to match the expected CRI-O version.

jodh-intel (approved):
Thanks @fgiudici.

lgtm

fidencio (approved):
lgtm, thanks @fgiudici!

Issue #3606, opened 2021-06-08
footloose image not built for s390x


Jakob-Naucke:
The pod-block-pv k8s test uses an Ubuntu image from footloose, which isn't built for s390x. I suggest to use another provider. PR is prepared.

PR #1566, opened 2019-05-13
test: Add kata+footloose kubernetes test

GabyCT:
This adds kata+footloose kubernetes test where
using footloose we are able to connect to pod.

Fixes #1535

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

PR #3607, opened 2021-06-08
k8s: Use libpod instead of footloose

Jakob-Naucke:
In pod-block-pv k8s test, use libpod latest (has no 18.04 tag) instead
of footloose 18.04 for quay.io registry organization, because footloose
has no s390x image.

Fixes: #3606

### DIFF
--- a/integration/kubernetes/runtimeclass_workloads/pod-block-pv.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-block-pv.yaml
@@ -7,7 +7,7 @@ spec:
   runtimeClassName: kata
   containers:
     - name: my-container
-      image: quay.io/footloose/ubuntu18.04
+      image: quay.io/libpod/ubuntu:latest
       command: ["tail", "-f", "/dev/null"]
       volumeDevices:
         - devicePath: DEVICE_PATH


jodh-intel (approved):
lgtm

fidencio (approved):
@Jakob-Naucke, lgtm if you promise me that we'll use the same ubuntu source everywhere in the code.

Right now we're getting ubuntu pods from a few different places and would be good to standardise those sooner than later.

Jakob-Naucke:
So we use a bunch of images besides `quay.io/{prometheus/busybox,sjenning/nginx,libpod/{redis,ubuntu}}` (the latter being proposed), `registry.fedoraproject.org/fedora` and `k8s.gcr.io/e2e-test-images/agnhost`, all of which have very broad architecture support.
- The `quay.io/footloose/ubuntu18.04` image that I'm trying to remove here is also used for a test that actually centers around `footloose`. However, it has been set to skip since it was merged (?!) in #1566.
- `docker.io/containerstack/alpine-stress` is used for the k8s memory limit and OOM tests. It is also not built for s390x, but I'm currently not yet complaining about that :stuck_out_tongue_winking_eye: because I have to disable them anyways due to https://github.com/kata-containers/kata-containers/issues/1412 (any comment there @fidencio :innocent:?)

While we're at it: Use of `docker.io` images isn't entirely removed either, which can bite you when you're rate limited:
- The metrics tests implicitly use it a bunch, _mostly_ for Ubuntu, but also Fedora and `rocker/tidyverse`.
- The k8s nginx tests use it implicitly for nginx and busybox.
- The `soak_parallel_rm` test uses it for nginx.
- The containerd-CRI tests (external!) also still use it in some places, despite some efforts to remove them.

@fidencio what would you suggest to change, if anything?

Issue #3611, opened 2021-06-09
Fix packages & build arguments for RHEL & s390x


Jakob-Naucke:
- CI environment setups are incomplete or dependent on x86_64.
- not optimised for RHEL 8, assumes btrfs-progs, does not install k8s

PR is prepared (see for more info).

PR #3612, opened 2021-06-09
Fix packages & build arguments for RHEL and (mostly) s390x

Jakob-Naucke:
- ci: Fix packages in Fedora env setup
  - `ninja-build` (already enabled for Ubuntu)
  - `libpmem` only if available (no aarch64 & s390x)
  - `protobuf-compiler` (ppc64le & s390x)
  - `openssl`
  - `which`
- ci: Fix packages in RHEL env setup
  - dto. Fedora
  - `make`
- ci: Fix packages in Ubuntu env setup
  - `libssl`
  - `protobuf-compiler` (ppc64le & s390x) 
  - `musl` only if Rust toolchain available (no ppc64le & s390x)
- ci: Gear up for RHEL 8 
  - `/proc/sys/fs/may_detach_mounts` seems only enabled for RHEL 7 (not available on my RHEL 8 x86_64 and s390x machines), so restrict it
  - use EPEL according to RHEL version 
  - Install `python3`
- ci: Export CC=gcc on s390x Fedora-likes -- see also https://github.com/kata-containers/kata-containers/pull/1974, is also required for CRI-containerd and CRI-O
- ci: Fix btrfs for CentOS/RHEL
  - is not packaged, should not be installed on RHEL (like on CentOS)
  - should not be used for building containerd on either
  - do not build with seccomp on RHEL (like on CentOS)
- ci: Install k8s on RHEL
  - Include as supported distro in k8s installation
  - Requires `yum update --nobest` due to version mismatches which is a bit dirty. Any suggestions?
- ci: Install distro Docker on Ubuntu s390x -- not packaged by docker.com

Fixes: #3611
@jongwu && (@Amulyam24 || @bpradipt || @clnperez) PTAL again :)


.ci/setup.sh
```diff
@@ -146,7 +146,7 @@ main() {
 	sync
 	sudo -E PATH=$PATH bash -c "echo 3 > /proc/sys/vm/drop_caches"
 
-	if [ "$ID" == rhel ]; then
+	if [ "$ID" == rhel ] && [ "$PLATFORM_ID" == "platform:el7" ] ; then
```
jodh-intel, line 149: v7 only? If so, please add a comment explaining why. Else could you instead use something like this?:

```bash
[[ "$PLATFORM_ID" =~ ^platform:el[0-9][0-9]*$ ]]
```

cmd/container-manager/manage_ctr_mgr.sh
```diff
@@ -179,6 +182,17 @@ install_docker(){
 	sudo chmod g+rw /var/run/docker.sock
 }
 
+install_docker_s390x(){
+	log_message "Installing docker"
+	case "$ID" in
+		"ubuntu")
+			sudo apt-get install -y docker.io
+			;;
```
jodh-intel, line 190: You could compact this for brevity (and not the missing `;;` I added):

```bash
ubuntu) sudo apt-get install -y docker.io ;;
*) die "Unsupported distribution: $ID" ;;
```

fidencio:
@Jakob-Naucke, let me start saying I'm **loving** to see these cleanups, and in general the whole series look good.

One thing that I would like to ask is, do you care about RHEL-7?
Here, upstream, we don't test on RHEL (but we should take advantage of the free offerings to hook up a RHEL VM in our tests), and we don't test on CentOS 7.

I wonder if we could just get rid of anything older than 8 and be happy about that.
WDYT?

jodh-intel (commented):
Thanks @Jakob-Naucke. A couple of comments/queries.

bpradipt:
Note to self: the CI action seems to work only for `kata-containers` repo and not for `tests` repo .

fidencio:
> Note to self: the CI action seems to work only for `kata-containers` repo and not for `tests` repo .

@bpradipt, Without stepping on your toes, let me just share some info with you.

When I go to http://jenkins.katacontainers.io/, I can see one Power job (kata-containers-2.0-Power9-ubuntu-20-04-PR), but that's tied to the `kata-containers` repo.  What you have to do to get this **also** working for the tests is actually add a new job, cloning one of the jobs that have `-tests-` on their name, and configure it accordingly.

This is the drill for those two different repos and, thankfully, we only have two repos nowadays. :-)

bpradipt:
> > Note to self: the CI action seems to work only for `kata-containers` repo and not for `tests` repo .
> 
> @bpradipt, Without stepping on your toes, let me just share some info with you.
> 
> When I go to http://jenkins.katacontainers.io/, I can see one Power job (kata-containers-2.0-Power9-ubuntu-20-04-PR), but that's tied to the `kata-containers` repo. What you have to do to get this **also** working for the tests is actually add a new job, cloning one of the jobs that have `-tests-` on their name, and configure it accordingly.
> 
> This is the drill for those two different repos and, thankfully, we only have two repos nowadays. :-)

Thanks @fidencio

.ci/setup.sh
Jakob-Naucke: So I cannot really find a good source on where and when `/proc/sys/fs/may_detach_mounts` is supported.
There's this [Red Hat bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1441737) that claims it was introduced in 7 (RHEL only, no CentOS), but it's not on my x86_64 laptop or my s390x dev instance (RHEL 8.4 and 8.3, respectively). It's also not in the [kernel docs](https://www.kernel.org/doc/html/latest/admin-guide/sysctl/fs.html). I don't really have a RHEL 7 instance handy to test with.

cmd/container-manager/manage_ctr_mgr.sh
Jakob-Naucke: Updated, thanks!

Jakob-Naucke:
> One thing that I would like to ask is, do you care about RHEL-7?

@fidencio no, not really, until someone told me they cared :)
 
> I wonder if we could just get rid of anything older than 8 and be happy about that.
> WDYT?

What would that be?

- There's the limitation of `/proc/sys/fs/may_detach_mounts` to RHEL 7 (see also discussion in review), but I'd prefer a more definite source for its removal than "it doesn't work on my machine" anyways :)
- Then there's the parse of `/etc/os-release` for getting the EPEL URL, but I think this is more flexible for the future anyways than hard-coding 8.

fidencio:
> > One thing that I would like to ask is, do you care about RHEL-7?
> 
> @fidencio no, not really, until someone told me they cared :)

I'd say just drop anything older than 8, sincerely, for the following reasons:
* We don't run tests against CentOS / RHEL 7, at all.
* CentOS / RHEL 7, as it's, doesn't provide you a Kernel / VMM that's capable of running what we need for kata-containers.
* Having RHEL 7 modified to run a different Kernel / VMM than the one provided would lead to unsupported land.

I sincerely would like to see us moving forward and testing more on newer distros (CentOS Stream 8, 9?) rather than putting effort on 7.

@jodh-intel, do we strongly disagree? Why? What's your take on this? :-)

@Jakob-Naucke, do you need or run tests on RHEL 7 on your side?

Jakob-Naucke:
> @Jakob-Naucke, do you need or run tests on RHEL 7 on your side?

@fidencio no. s390x CI (soon™) will use 8 on the RHEL side.

fidencio:
@Jakob-Naucke,
```
12:01:40 E: Unable to locate package ninja-buildlibpmem-dev
```

.ci/setup_env_ubuntu.sh
```diff
@@ -49,7 +49,11 @@ if [ "${NAME}" == "Ubuntu" ] && [ "$(echo "${VERSION_ID} >= 20.04" | bc -q)" ==
 fi
 
 if [ "$(uname -m)" == "x86_64" ] && [ "${NAME}" == "Ubuntu" ] && [ "$(echo "${VERSION_ID} >= 18.04" | bc -q)" == "1" ]; then
-	packages[qemu_dependencies]+=" libpmem-dev"
+	packages[qemu_dependencies]+="libpmem-dev"
```
fidencio, line 52: We need a space before the `libpmem-dev`

.ci/setup_env_ubuntu.sh
```diff
@@ -49,7 +49,11 @@ if [ "${NAME}" == "Ubuntu" ] && [ "$(echo "${VERSION_ID} >= 20.04" | bc -q)" ==
...
+fi
+
+if [ "$(uname -m)" == "s390x" ]; then
+	packages[kernel_depencencies]+="libssl-dev"
```
fidencio, line 56: We need a space before the `libssl-dev`

.ci/setup_env_ubuntu.sh
Jakob-Naucke: Updated, sorry!

.ci/setup_env_ubuntu.sh
Jakob-Naucke: Updated.

Jakob-Naucke:
- Rebase for static check fix
- Export `$CC` at the top level
- Fix `$BUILDTAGS` for CRI-containerd (this is only relevant when _not_ exporting `$CC`)

Jakob-Naucke:
I just checked -- `/proc/sys/fs/may_detach_mounts` seems to be in RHEL/CentOS 7, and essentially nowhere else.
@fidencio remove the code bit that touches it and that's it for 7?

fidencio:
> I just checked -- `/proc/sys/fs/may_detach_mounts` seems to be in RHEL/CentOS 7, and essentially nowhere else.
> @fidencio remove the code bit that touches it and that's it for 7?

Yes, please! :-)

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

Issue #3622, opened 2021-06-15
integration/crio: do not modify crio bats file


littlejawa:
Running the integration test with crio multiple times, multiple comment lines are added to the ctr.bats file in cri-o sources.
Also if you edit the crio_skip_test.sh file to remove some entry (to troubleshoot some failing tests), the change is not taken into account if the test has been run before.

The root cause is that the cri-o.sh script is editing ctr.bats directly, on each run.
I would like to modify that by making a copy of ctr.bats, editing that copy, and running the tests from there.

PR #3623, opened 2021-06-15
integration/cri-o: do not edit ctr.bats

littlejawa:
Make a copy of ctr.bats before editing it.
This prevents multiple lines from being added to the file in cri-o sources.
It also makes it possible to change the list of skipped tests and take the
change into account.

Fixes: #3622

### DIFF
--- a/integration/cri-o/cri-o.sh
+++ b/integration/cri-o/cri-o.sh
@@ -68,10 +68,10 @@ IFS=''
 
 # Skip CRI-O tests that currently are not working
 pushd "${crio_repository_path}/test/"
-cp $GOPATH/src/${crio_repository}/test/ctr.bats $GOPATH/src/${crio_repository}/test/ctr_kata_integration_tests.bats
+cp ctr.bats ctr_kata_integration_tests.bats
 for i in "${skipCRIOTests[@]}"
 do
-	sed -i '/'${i}'/a skip \"This is not working\"' "$GOPATH/src/${crio_repository}/test/ctr_kata_integration_tests.bats"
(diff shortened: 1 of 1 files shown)


littlejawa:
CC: @fgiudici

fgiudici:
lgtm!

jodh-intel (approved):
Thanks @littlejawa.

lgtm

fidencio:
@littlejawa, for completeness, shouldn't we also remove the file we created after the run?

littlejawa:
@fidencio sure, it doesn't hurt. I'll add that.

fidencio (approved):
lgtm, thanks @littlejawa!

integration/cri-o/cri-o.sh
```diff
@@ -68,9 +68,10 @@ IFS=''
 
 # Skip CRI-O tests that currently are not working
 pushd "${crio_repository_path}/test/"
+cp $GOPATH/src/${crio_repository}/test/ctr.bats $GOPATH/src/${crio_repository}/test/ctr_kata_integration_tests.bats
```
uril, line 71: minor comment (it's good as is):
crio_repository_path is $GOPATH/src/${crio_repository} and the script just pushd into it
so you can simply cp ctr.bats ctr_kata_integration_tests.bats (no need for prefix)

littlejawa: Right. The "sed" line below was using the prefix, so I just copied it. But it's not needed.

snir911:
some of the jobs seems stuck
/retest

Issue #3631, opened 2021-06-18
Off-by-one KSM comparison


Jakob-Naucke:
on jenkins-ci-ubuntu-20-04:
```
07:31:38 Pages merged 226
...
07:31:42 Pages merged 226
07:31:42 Comparing merged pages between containers
07:31:42 ERROR: The merged pages on the second container is less than the first container
```

> ```[ "$second_pages_merged" -gt "$first_pages_merged" ] || die "The merged pages on the second container is less than the first container"```

I think that should be `-ge`

_Originally posted by @Jakob-Naucke in https://github.com/kata-containers/tests/issues/3627#issuecomment-862333022_

PR #3632, opened 2021-06-18
integration: Fix off-by-one in KSM comparison

Jakob-Naucke:
Fails when pages merged on second container is equal to first, which
should still be a pass

Fixes: #3631


fidencio (approved):
Nice catch, @Jakob-Naucke!

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

Issue #3661, opened 2021-06-30
CI: Add ability to skip CI run in fast-return script


jodh-intel:
We used to have this in the `master` branch (from PR https://github.com/kata-containers/tests/pull/2915), so let's copy it across to `main`.

PR #3662, opened 2021-06-30
CI: Add ability to skip CI run in fast-return script

jodh-intel:
Allow the CI to be skipped if the magic `force-skip-ci` label is set for a PR.

Cherry-picked from #2915.

Added new documentation for the fast return script.

Fixes: #3661

### DIFF
--- a/.ci/ci-fast-return.sh
+++ b/.ci/ci-fast-return.sh
@@ -12,6 +12,7 @@
 #  - file name patterns from the YAML to see if we can skip all files in the PR
 #  - If potentially skipping, looks for a 'force' label on the PR to force the CI
 #    to run anyway
+#  - Looks for a force label on the PR to always skip the CI.
 
 set -e
 
@@ -41,6 +42,9 @@ unit_yamlfile_gitname="${script_dir_base}/${unit_yamlfile_rootname}"
 
 # Name of the label, that if set on a PR, will force the CI to be run anyway.
 force_label="force-ci"
+
+# Name of the label, that if set on a PR, will force the CI to not be run.
+skip_label="force-skip-ci"
 
 # The list of files to check is held in a global, to make writing the unit test
 # code easier - otherwise we would have to pass two multi-entry lists (the list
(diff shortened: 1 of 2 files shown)


.ci/ci-fast-return.sh
```diff
@@ -374,6 +470,121 @@ doc.md" "$matches"
...
+	assertEquals "0" "$result"
+
+	result=$(is_label_set "repo" "pr" "")
+	assertEquals "0" "$result"
+
+	result=$(is_label_set "" "pr" "label")
+	assertEquals "0" "$result"
+
+	result=$(is_label_set "repo" "" "label")
+	assertEquals "0" "$result"
```
fidencio, line 489: It wouldn't hurt to have a negative test that would ensure 1 is returned when expected.

.ci/ci-fast-return.sh
```diff
@@ -113,6 +121,63 @@ check_not_matches() {
...
+	local labels=$(jq '.[].name' <<< $json)
+
+	# Check if we have the forcing label set
+	for x in $labels; do
+		# Strip off any surrounding '"'s
+		y=$(sed 's/"//g' <<< $x)
+
+		if [ "$y" == "$label" ]; then
+			echo "1"
+			return 0
```
fidencio, line 172: I must admit that looking at and checking whatever is printed to stdout rather than whatever is returned is confusing.

fidencio (approved):
lgtm, thanks @jodh-intel!

I've left one comment and one suggestion, feel free to ignore those if you think they don't make sense.

.ci/ci-fast-return.sh
jodh-intel: Agreed. I've added some logic to `is_label_set()` (see `testing_mode=`) to handle this scenario by faking the return value from that function (after having checked the params).

.ci/ci-fast-return.sh
jodh-intel: Also agreed. I think this is partly due to the constraints of shunit2 / `set -e`.

fidencio:
@jodh-intel, thanks for addressing the comments, my lgtm is still valid! :-)

fidencio:
Howdy!

I'd like to ask you to rebase your PR atop of current main, as #3668 has been merged and it has a fix that unblocks the CI.

Please, do the rebase and feel free to `@fidencio` in case you do not have the powers to run `/test`.

Sorry for the inconvenience and happy hacking!

jodh-intel:
One more ack and we can land this...

.ci/ci-fast-return.sh
```diff
@@ -374,6 +480,124 @@ doc.md" "$matches"
 
 }
 
+testIsLabelSet() {
```
Jakob-Naucke, line 483: Non-blocking question: Where are those tests picked up?

Jakob-Naucke (approved):
LGTM, although we should be vigilant to not overuse this.

.ci/ci-fast-return.sh
jodh-intel: They are for updating the scirpt: they aren't run by the CI (since if they fail, the CI is dead ;)

chavafg (approved):

Issue #3678, opened 2021-07-05
docker.io/containerstack/alpine-stress is not built for !x86_64


Jakob-Naucke:
and it's required for the `k8s-{memory,oom}` tests. Failing to find a suitable substitute, I suggest we make it ourselves. PR is prepared.

PR #3668, opened 2021-06-30
metrics: Update metrics ksm value

GabyCT:
This PR updates the metrics value for the memory ksm test. It seems
that the change of server modified the value so this PR fixes it.

Fixes #3667

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

PR #3636, opened 2021-06-18
ci: Use Podman on Fedora-likes

Jakob-Naucke:
since it's more aligned with CRI-O, e.g. for importing images.

- A comment in the CRI tests claims that their TestImageLoad test
  depends on it, but my experiments don't confirm this.
- Likewise, kubeadm depends on Docker according to a comment, but I
  can't reproduce that.

Fixes: #3635
Depended-on: github.com/kata-containers/kata-containers#2068
I-once-thought-it-depended-on: github.com/kata-containers/tests#3658
But-it-really-Depends-on: github.com/kata-containers/kata-containers#2150

Do you think that any of these are a blocker?

- ~Podman is incompatible with osbuilder Ubuntu in that it can't be debootstrapped, apt errors about setting up packages.~ fixed with an Ubuntu 20.04 guest
- The metrics tests use Docker a lot, but those are configured to CRI-containerd and we don't run them on Fedora-likes anyways.
- The packaging scripts also use Docker.

~Starting out as draft~ adding do-not-merge
because there are some open questions.

PR #3679, opened 2021-07-05
integration/k8s: Build custom stress image

Jakob-Naucke:
containerstack/alpine-stress is only built for x86_64 and I cannot find
a substitute. Let's build the image ourselves to have a `stress`
container available on all architectures.

Fixes: #3678
Depends-on: github.com/kata-containers/tests#3636

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -69,6 +69,20 @@ esac
 # Check no there are no kata processes from previous tests.
 check_processes
 
+echo "Build custom stress image"
+registry_image="docker.io/library/registry"
+arch=$("${SCRIPT_PATH}/../../.ci/kata-arch.sh")
+if [[ "${arch}" == "ppc64le" || "${arch}" == "s390x" ]]; then
+	# that image is not built for these architectures
+	registry_image="docker.io/ibmcom/registry:2.6.2.5"
+fi
+
+runtimeclass_files_path="${SCRIPT_PATH}/runtimeclass_workloads"
+pushd "${runtimeclass_files_path}/stress"
+[ "${container_engine}" == "docker" ] && sudo -E systemctl restart docker
+sudo -E "${container_engine}" build . -t "${stress_image}"
+popd
+
 # Remove existing CNI configurations:
 cni_config_dir="/etc/cni"
 cni_interface="cni0"
@@ -98,6 +112,12 @@ for i in $(seq ${max_cri_socket_check}); do
 done
 
 sudo systemctl status "${cri_runtime}" --no-pager
+
+if [ "${stress_image_pull_policy}" == "Always" ]; then
+	echo "Store custom stress image in registry"
+	sudo -E "${container_engine}" run -d -p ${registry_port}:5000 --name "${registry_name}" "${registry_image}"
+	sudo -E "${container_engine}" push "${stress_image}"
+fi
 
 echo "Init cluster using ${cri_runtime_socket}"
 kubeadm_config_template="${SCRIPT_PATH}/kubeadm/config.yaml"
@@ -139,7 +159,6 @@ kubectl get pods
 flannel_version="$(get_test_version "externals.flannel.version")"
 flannel_url="https://raw.githubusercontent.com/coreos/flannel/${flannel_version}/Documentation/kube-flannel.yml"
 
-arch=$("${SCRIPT_PATH}/../../.ci/kata-arch.sh")
 #Load arch-specific configure file
 if [ -f "${SCRIPT_PATH}/../../.ci/${arch}/kubernetes/init.sh" ]; then
         source "${SCRIPT_PATH}/../../.ci/${arch}/kubernetes/init.sh"
@@ -152,7 +171,6 @@ kubectl apply -f "$network_plugin_config"
 # we need to ensure a few specific pods ready and running
 wait_pods_ready
 
-runtimeclass_files_path="${SCRIPT_PATH}/runtimeclass_workloads"
 echo "Create kata RuntimeClass resource"
 kubectl create -f "${runtimeclass_files_path}/kata-runtimeclass.yaml"
 
--- a/integration/kubernetes/runtimeclass_workloads/stress/Dockerfile
+++ b/integration/kubernetes/runtimeclass_workloads/stress/Dockerfile
@@ -1,0 +1,9 @@
+#
+# Copyright (c) 2021 IBM Corp.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM quay.io/libpod/ubuntu
+RUN apt-get -y update && \
+    apt-get -y upgrade && \
+    apt-get -y install stress
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -61,6 +61,20 @@ jenkins_url="http://jenkins.katacontainers.io"
 # Path where cached artifacts are found.
 cached_artifacts_path="lastSuccessfulBuild/artifact/artifacts"
 
+# Info for running local registry
+registry_port="${REGISTRY_PORT:-5000}"
+registry_name="kata-registry"
+container_engine="${USE_PODMAN:+podman}"
+container_engine="${container_engine:-docker}"
+# We cannot use Podman's and Kubernetes' CNI at the same time, but Podman uses k8s' namespace
+if [ "${container_engine}" == "podman" ]; then
+	stress_image="localhost/stress-kata:latest"
+	stress_image_pull_policy="Never"
+else
+	stress_image="localhost:${registry_port}/stress-kata:latest"
+	stress_image_pull_policy="Always"
+fi
+
 # Clone repo only if $kata_repo_dir is empty
 # Otherwise, we assume $kata_repo is cloned and in correct branch, e.g. a PR or local change
 clone_kata_repo() {
--- a/integration/kubernetes/cleanup_env.sh
+++ b/integration/kubernetes/cleanup_env.sh
@@ -7,6 +7,7 @@
 # This script is used to reset the kubernetes cluster
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../../.ci/lib.sh"
 source "${SCRIPT_PATH}/../../lib/common.bash"
 
 cri_runtime="${CRI_RUNTIME:-crio}"
@@ -38,6 +39,8 @@ BAREMETAL="${BAREMETAL:-false}"
 if [ "${BAREMETAL}" == true ] && [ -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh" ]; then
 	bash -f "${SCRIPT_PATH}/cleanup_bare_metal_env.sh"
 fi
+
+sudo -E "${container_engine}" rm -f "${registry_name}" || true
 
 # Check no kata processes are left behind after reseting kubernetes
 check_processes
--- a/integration/kubernetes/k8s-memory.bats
+++ b/integration/kubernetes/k8s-memory.bats
@@ -14,14 +14,21 @@ setup() {
 	get_pod_config_dir
 }
 
+setup_yaml() {
+	sed \
+		-e "s/\${memory_size}/${memory_limit_size}/" \
+		-e "s/\${memory_allocated}/${allocated_size}/" \
+		-e "s|\${stress_image}|${stress_image}|" \
+		-e "s/\${stress_image_pull_policy}/${stress_image_pull_policy}/" \
+		"${pod_config_dir}/pod-memory-limit.yaml"
+}
+
+
 @test "Exceeding memory constraints" {
 	memory_limit_size="50Mi"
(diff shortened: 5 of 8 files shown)


integration/kubernetes/runtimeclass_workloads/stress/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM quay.io/libpod/ubuntu:latest
+RUN apt-get -y update && \
+    apt-get -y upgrade && \
+    apt-get -y install stress
```
jodh-intel, line 4: Looking at the original `Dockerfile` (https://github.com/containerstack/alpine-stress/blob/master/Dockerfile) shows that `stress` originates from:

- https://people.seas.harvard.edu/~apw/stress.

That's confirmed by the debian package page too:

- https://packages.debian.org/en/buster/stress.

However, the homepage no longer exists (which presumably explains why sid doesn't list it - https://packages.debian.org/en/sid/stress).

I wonder if we want to use `stress-ng` instead?:

- https://packages.ubuntu.com/hirsute/stress-ng

integration/kubernetes/init.sh
```diff
@@ -158,3 +158,17 @@ kubectl create -f "${runtimeclass_files_path}/kata-runtimeclass.yaml"
 
 # Enable the master node to be able to schedule pods.
 kubectl taint nodes "$(hostname)" node-role.kubernetes.io/master:NoSchedule-
+
+container_engine="${USE_PODMAN:+podman}"
+container_engine="${container_engine:-docker}"
+stress_image="stress-kata"
+pushd "${runtimeclass_files_path}/stress"
+sudo "${container_engine}" build . -t "${stress_image}"
+# Import into containerd
```
MatthieuSarter, line 167: This script may be used with containerd or CRI-O, so the CRI-O case should also be handled. It might require to have podman installed, as crictl does not allow to import images.

integration/kubernetes/runtimeclass_workloads/stress/Dockerfile
Jakob-Naucke: Hmmm, I did notice that, and I knew that stress-ng existed. There's also quay.io/kinvolk/stress-ng, which is at least also built for arm64, but not for the IBM architectures, so we'd probably still need the custom build. Let me TAL at how much we would have to change.

integration/kubernetes/init.sh
Jakob-Naucke: Yes. This is why this PR depends on #3636, which installs Podman and uses it for osbuilder on Fedora-likes. I.e.
- Fedora-likes will have Podman+CRI-O, and the image will be available to Kubernetes right away after line 166.
- Everything else will have Docker+containerd (lines below).
- I don't think Podman+containerd+k8s is used anywhere, but I just figured I might actually add support for it because it's quite simple.
- I think Docker+CRI-O+k8s is not possible without an external registry, so yes, that is one combination we would drop. But that's why the dependency on #3636 exists.

integration/kubernetes/runtimeclass_workloads/stress/Dockerfile
jodh-intel: Thanks @Jakob-Naucke. Hopefully, it should just be a case of changing the `apt` package name as `stress-ng` is tested on all arches fwics.

/cc @liubin and @GabyCT on the idea of switching to a different "stressor" package though.

Jakob-Naucke: Yes, but I think the flags in the tests we use it in (`integration/kubernetes/k8s-{memory,oom}.bats`) will change.

integration/kubernetes/init.sh
Jakob-Naucke: Updated as mentioned.

integration/kubernetes/runtimeclass_workloads/stress/Dockerfile
Jakob-Naucke: updated... it seems `stress-ng` is actually (and probably intentionally) forwards-comptaible.

jodh-intel: @Jakob-Naucke - nice! And thanks @ColinIanKing! :smile:

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

fidencio:
Howdy!

I'd like to ask you to rebase your PR atop of current main, as #3668 has been merged and it has a fix that unblocks the CI.

Please, do the rebase and feel free to `@fidencio` in case you do not have the powers to run `/test`.

Sorry for the inconvenience and happy hacking!

Jakob-Naucke:
rebase & fix depends tag
/test

fidencio:
I'm adding the `do-not-merge` label till we get `2.2.0-alpha1` release out.
Once we cut the release I'll come back removing the label from the PRs.

Jakob-Naucke:
added transitive PR dependency
/test

fidencio:

Jakob-Naucke:
So we are using Docker+CRI-O+k8s e.g. on the Debian CI (so it definitely works on non-Fedora-likes as well) and that's a combination I'd rather want to support. Taking a look at running a local registry.

GabyCT:
@Jakob-Naucke could you please do a rebase of this PR ...some CI fixes were done that I believe is impacting the CI thanks

Jakob-Naucke:
So the previously anticipated workflow of importing the built image with `ctr --namespace=k8s.io` for containerd or Podman with CRI-O is not going to fly, since Docker+CRI-O is both very much possible and even used in our CI (Debian).
I then looked into telling Docker to use the Kubernetes namespace, which would have looked a little like
```sh
export DOCKER_TLS_VERIFY=1
export DOCKER_HOST="$(yq read "$KUBECONFIG" "clusters[0].cluster.server") | sed "s/https/tcp/")"
yq read "$KUBECONFIG" "clusters[0].cluster.certificate-authority-data" | base64 -d > "$HOME/.kube/ca.pem"
export DOCKER_CERT_PATH="$HOME/.kube"
```

and while I could swear that this worked at one point in a VM where I failed to restart the cluster again, maybe it never worked and it merely worked with the minikube that I used for figuring out these variables.
Anyway, this leads to the error:
```
Error response from daemon: forbidden: User "system:anonymous" cannot get path "/v1.24/build"
```

My fight against resolving this has been unsuccessful and unless someone has an idea, since we recently got [quay.io/kata-containers](https://quay.io/organization/kata-containers), I wonder whether we should use that instead.
/cc @fidencio because I would need a bit of guidance there :)

Jakob-Naucke:
Re-requesting review because a lot has changed. This works with Docker and Podman, and is CRI runtime indepedent.
`docker.io/registry/registry` is not perfect due to rate limiting, ideas for alternatives? There is also `quay.io/ibm/registry`, but unlike `docker.io/ibmcom/registry` (which the PR currently uses for ppc64le & s390x), it only builds for s390x.
/test

Jakob-Naucke:
forgot second substitution in exceed memory test
/test

Jakob-Naucke:
So there is _something_ that `stress-ng` does differently from `stress`. They seem to behave similarly in some quick tests, but while the `k8s-memory` test is working, the `k8s-oom` test clearly works with `stress`, but not with `stress-ng` -- the OOMKill does not seem to happen.
(For reference, it works with these changes:)
```diff
diff --git a/integration/kubernetes/runtimeclass_workloads/pod-oom.yaml b/integration/kubernetes/runtimeclass_workloads/pod-oom.yaml
index 3dfcbd88..7b9d414a 100644
--- a/integration/kubernetes/runtimeclass_workloads/pod-oom.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-oom.yaml
@@ -20 +20 @@ spec:
-      args: ["-c", "sleep 2; stress-ng --vm 2 --vm-bytes 400M --timeout 30s"]
+      args: ["-c", "sleep 2; stress --vm 2 --vm-bytes 400M --timeout 30s"]
diff --git a/integration/kubernetes/runtimeclass_workloads/stress-ng/Dockerfile b/integration/kubernetes/runtimeclass_workloads/stress-ng/Dockerfile
index b25b541c..6546f030 100644
--- a/integration/kubernetes/runtimeclass_workloads/stress-ng/Dockerfile
+++ b/integration/kubernetes/runtimeclass_workloads/stress-ng/Dockerfile
@@ -9 +9 @@ RUN apt-get -y update && \
-    apt-get -y install stress-ng
+    apt-get -y install stress
```

Now I'm debating whether I should track the root cause of this (because this is probably a bug in our testing), or just keep using `stress`.

/cc @jodh-intel

Jakob-Naucke:
Okay, this was because `--vm` uses `malloc` in `stress`, but `mmap` in `stress-ng`, and to use `malloc` in `stress-ng`, you need `malloc`, so this seems to be a classical "failing to RTFM", although I couldn't pinpoint why it wouldn't work with `mmap`.
/test

ColinIanKing:
Yep, at the end of the stress-ng manual it states:

"Note  that the stress-ng cpu, io, vm and hdd tests are different imple‐
 mentations of the original stress tests and hence may produce different
 stress  characteristics."

It's hard to do a clean-room reimplementation of the original stress w/o knowing what it really is doing. :-/

Jakob-Naucke:
I can reproduce this: it does not work under CRI-O. Investigating...

Jakob-Naucke:
I have a hard time putting my finger on the difference between containerd and CRI-O here. But @ColinIanKing could you TAL at this (this is not in a container):
```
$ free -h
              total        used        free      shared  buff/cache   available
Mem:          7.6Gi       819Mi       5.9Gi       382Mi       974Mi       6.2Gi
Swap:            0B          0B          0B
$ stress --vm 1 --vm-bytes 16G --timeout 60s
stress: info: [145711] dispatching hogs: 0 cpu, 0 io, 1 vm, 0 hdd
stress: FAIL: [145711] (415) <-- worker 145712 got signal 9
stress: WARN: [145711] (417) now reaping child worker processes
stress: FAIL: [145711] (451) failed run completed in 3s
$ stress-ng --malloc 1 --malloc-bytes 16G --oomable --timeout 60s
stress-ng: info:  [146265] dispatching hogs: 1 malloc
stress-ng: info:  [146265] successful run completed in 60.14s (1 min, 0.14 secs)
```

shouldn't `stress-ng` be OOMed there?

ColinIanKing:
stress-ng will respawn OOM'd stressor processes since it is designed to keep the stress going until the timeout or bogo-ops limits are reached.  The --oomable option will override this default and allow stressors to be OOM'd without a restart.

Jakob-Naucke:
@ColinIanKing I used `--oomable`, check the line

ColinIanKing:
Ah, I overlooked that --oomable.  So the malloc stressor will allocate memory but Linux uses lazy memory allocation hence pages that are not touched won't cause a page to be physically mapped. The stressor is intended to thrash the libc memory allocator rather than eat loads of memory,  but I could add an option to make this stressor touch each page to force them to be physically mapped in.

Jakob-Naucke:
> So the malloc stressor will allocate memory but Linux uses lazy memory allocation hence pages that are not touched won't cause a page to be physically mapped. The stressor is intended to thrash the libc memory allocator rather than eat loads of memory, but I could add an option to make this stressor touch each page to force them to be physically mapped in.

Ah, that makes sense. I think I will switch back to `stress` for the time being. Such an option would be useful, but don't feel `stress`ed to do it :)

Jakob-Naucke:
I didn't actually push towards my fork, sorry for the noise -- /test

ColinIanKing:
--malloc-touch option now added to stress-ng to force all pages to be physically resident on a malloc/realloc.

Jakob-Naucke:
on ARM/s390x:
```
15:39:54 Err:1 http://ports.ubuntu.com/ubuntu-ports focal InRelease
15:40:14   Temporary failure resolving 'ports.ubuntu.com'
```

Let me check whether there's something wrong with the `libpod` image.
/test

Jakob-Naucke:
that should be fixed, /test

Jakob-Naucke:
all look like one-offs -- /retest-arm /retest-s390x /retest-clh

GabyCT (approved):

fidencio:
@Jakob-Naucke, out of curiosity, considering that now we have a quay.io account and we have some freedom on managing that, would be worth it to maintain the image there?

Jakob-Naucke:
@jodh-intel I am going to dismiss your review because the PR was very different at the time, it would be dishonest to the processes to merge right now.

> @Jakob-Naucke, out of curiosity, considering that now we have a quay.io account and we have some freedom on managing that, would be worth it to maintain the image there?

@fidencio just now I learnt about [quay.io being able to build based on Dockerfiles](https://docs.quay.io/guides/building.html), but we have yet to figure out the available hardware. We can follow the approach from this PR for the time being.

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

Issue #3691, opened 2021-07-09
Fix wrong runtime name


liubin:
Some code in tests is using `kata-runtime` as the runtime (process) name, which leads to some tests have been skipped.

PR #3692, opened 2021-07-09
tests: fix wrong runtime name

liubin:
Some code in tests is using kata-runtime as the
runtime (process) name, which leads to some tests
have been skipped.

And netmon test should be deleted too.

Fixes: #3691

### DIFF
--- a/.ci/configure_containerd_for_kata.sh
+++ b/.ci/configure_containerd_for_kata.sh
@@ -9,7 +9,6 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly kata_runtime_path=$(command -v kata-runtime)
 readonly runc_path=$(command -v runc)
 
 sudo mkdir -p /etc/containerd/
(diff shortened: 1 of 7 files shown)


fidencio (approved):
@liubin, this looks reasonable to me.
One thing though, is that we should get rid of netmon (and its references) everywhere.

It's currently present as part of configuration files.  It deserves an issue to track the total removal.

liubin:
> @liubin, this looks reasonable to me.
> One thing though, is that we should get rid of netmon (and its references) everywhere.
> 
> It's currently present as part of configuration files. It deserves an issue to track the total removal.

Yes, the netmon is no used anymore, all related things should be deleted.

amshinde:
> One thing though, is that we should get rid of netmon (and its references) everywhere.

@fidencio Just wanted to ask if we got rid of netmon altogether and if so was this functionality added to the shimv2 process?

jodh-intel (approved):
Ouch!! Thanks @liubin!

lgtm

PR #3693, opened 2021-07-09
ci: Temporarily disable govet checks

fidencio:
Let's hammer this issue down by temporarily disabling govet checks, as
it does affect a whole lot of code that's already part of our codebase
and, for some reason, it just started being opinionated now.

The proposal here is, disable it ... and do a somewhat taskforce to
re-enable this, skipping the checks wherever it makes sense to be
skipped (our tests, for instance).

Fixes: https://github.com/kata-containers/tests/issues/3694

Note:
The kind of issues govet will expose are all related to field alignment
in structs, which we could do better indeed.

Here's one example of the exposed issues:
```
INFO: Running golangci-lint on /home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/runtime/cli
kata-env.go:74:18: fieldalignment: struct with 128 pointer bytes could be 112 (govet)
type RuntimeInfo struct {
^
kata-env.go:86:18: fieldalignment: struct with 48 pointer bytes could be 24 (govet)
type VersionInfo struct {
^
kata-env.go:117:16: fieldalignment: struct with 32 pointer bytes could be 24 (govet)
type AgentInfo struct {
^
```

### DIFF
--- a/.ci/.golangci.yml
+++ b/.ci/.golangci.yml
@@ -18,7 +18,6 @@ linters:
     - gocyclo
     - gofmt
     - gosimple
-    - govet
     - ineffassign
     - misspell
     - staticcheck
@@ -33,6 +32,3 @@ linters-settings:
     min_complexity: 15
   unused:
     check-exported: true
-  govet:
-    enable:
-      - fieldalignment


Jakob-Naucke (approved):

c3d:
@fidencio  Could you add the kind of failures we now see to the commit message? Just so that we know what we are fixing?

c3d (approved):
With the caveat of better documenting what we fix, lgtm.

fidencio:
> @fidencio Could you add the kind of failures we now see to the commit message? Just so that we know what we are fixing?

It'll delay the fix for at least one hour. :-/

fidencio:
And the commit message already points to https://github.com/kata-containers/tests/issues/3694, which has the description of the problem.

c3d:
OK, understand the point about not restarting the CI.

(as for the issue showing the problem, as the owner of one project where the underlying issue database went poof, I can tell you this is not necessarily solid in the long run).

fidencio:
> OK, understand the point about not restarting the CI.
> 
> (as for the issue showing the problem, as the owner of one project where the underlying issue database went poof, I can tell you this is not necessarily solid in the long run).

Can you re-review to ensure it matches your expectation?

wainersm (approved):

c3d:
New text looks fine to me. Sorry our messages crossed and you'll have to wait one more round of CI.

.ci/.golangci.yml
```diff
@@ -33,6 +32,3 @@ linters-settings:
     min_complexity: 15
   unused:
     check-exported: true
-  govet:
-    enable:
-      - fieldalignment
```
jodh-intel, line 38: Rust ftw! :smile:

jodh-intel (approved):
Thanks @fidencio.

lgtm

.ci/.golangci.yml
fidencio: Don't get me to start on this. :-)

jodh-intel: :laughing: :laughing: :laughing:

PR #3696, opened 2021-07-09
ci/setup_env_centos: enable CentOS Stream support

### LINKED ISSUE #3695: CI setup script returns error on CentOS 8 Stream
# Description of problem

Cannot execute successfully  the .ci/setup.sh script on CentOS Stream 

# Expected result

The script will complete successfully 

# Actual result

The scripts ends with the error:
_ERROR: Unable to find the CentOS base repository file_

# Further information
The file name of the base repository for CentOS Stream should be added to the setup_env_centos.sh script

fgiudici:
The CentOS Stream base repository file name is different from the one
shipped with CentOS. This causes the script to fail.
Just add it. Nothing more needed to be CentOS Stream ready.

Fixes #3695

Signed-off-by: Francesco Giudici <fgiudici@redhat.com>

### DIFF
--- a/.ci/setup_env_centos.sh
+++ b/.ci/setup_env_centos.sh
@@ -29,7 +29,8 @@ sudo -E dnf -y install epel-release
 # Enable priority to CentOS Base repo in order to
 # avoid perl updating issues
 for repo_file_path in /etc/yum.repos.d/CentOS-Base.repo \
-	/etc/yum.repos.d/CentOS-Linux-BaseOS.repo; do
+	/etc/yum.repos.d/CentOS-Linux-BaseOS.repo \
+	/etc/yum.repos.d/CentOS-Stream-BaseOS.repo; do
 	if [ -f "$repo_file_path" ]; then
 		repo_file="$repo_file_path"
 		break


.ci/setup_env_centos.sh
```diff
@@ -29,7 +29,8 @@ sudo -E dnf -y install epel-release
 # Enable priority to CentOS Base repo in order to
 # avoid perl updating issues
 for repo_file_path in /etc/yum.repos.d/CentOS-Base.repo \
-	/etc/yum.repos.d/CentOS-Linux-BaseOS.repo; do
+	/etc/yum.repos.d/CentOS-Linux-BaseOS.repo \
+	/etc/yum.repos.d/CentOS-Stream-BaseOS.repo; do
```
fidencio, line 33: Hmm. I don't think this is the way to go, @fgiudici.

This will automatically "transform" the CentOS CI into a CentOS Stream CI, and there's absolutely no guarantee it'll keep working or that no conflcts will happen.

The way to enable a CentOS Stream CI would be:
1. Ensure we have a CentOS Stream image to be used from Azure (I don't think we have);
2. Add a .ci/setup_env_centos_stream.sh script;
3. Use it as a different entity than CentOS (as, in fact, it's a different entity).

Sorry for being a party pooper, but it's a nack from me.

fidencio (changes requested):

.ci/setup_env_centos.sh
fidencio: Another option, but this may complicate things a little bit ... would be to automatically detect whether you're running on CentOS or CentOS Stream and install the correct repos based on that.

That's something I'd be way more in favour to accept. :-)

fgiudici: To my understanding this change just enables the script to be run on CentOS Stream (if and when we will have such distro used in our CI) and has no effect on current CentOS CI. The intention was not to add CentOS Stream to our ci, but to make the setup script able to deal with it.
Here it just add a check to see if the CentOS-Stream-BaseOS.repo file exists and is already there: if so,  I expect we are in a CentOS Stream distro, otherwise the file shouldn't be there (but we will probably have CentOS-Linux-BaseOS.repo or CentOS-Base.repo instead). It shouldn't change anything.
I just stumbled into this while using the .ci/setup.sh script locally (yep, on CentOS Stream), and looked to me this could make it future proof in a safe way. If not, no problem to drop this PR.

fidencio: @fgiudici, I get that it will only check for the file and use it if found, and that finding the Stream file is the last option and that the file will not be present on a normal CentOS installation.

I get that it'll allow local users to run CI on CentOS Stream, but I'd rather be more explicit about supporting CentOS stream than silently making it work, because with this we create the expectation that it'll work for CentOS Stream, although we don't test / use it.

So, I'm in a personal quest to limit, more and more, what we actually support and focusing on what we actually claim to support, then we have people responsible for those bits in case something goes awry.

Does this make sense?  And sorry for being kinda strict here, but adding "just one more thing" is something that has been worrying me a lot lately. :-)

fgiudici: Ok, no problem, closing this PR then.

Issue #3705, opened 2021-07-12
containerd, not containerd.io should be removed


Jakob-Naucke:
Clean up scripts uninstall `containerd.io` on Ubuntu when the package is
called `containerd`. This can lead to Docker failing to start because
containerd was not fully removed and systemd still remembers the 
custom-installed unit file which points to `/usr/local`, as seen e.g. in
http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-PR/83
(although I can also reproduce this on x86?).

PR #3706, opened 2021-07-12
ci: s/containerd.io/containerd/ in cleanup

Jakob-Naucke:
Clean up scripts uninstall `containerd.io` on Ubuntu when the package is
called `containerd`. This can lead to Docker failing to start because
containerd was not fully removed and systemd still remembers the
custom-installed unit file which points to `/usr/local`, as seen e.g. in
http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-PR/83
(although I can also reproduce this on x86?).

Fixes: #3705

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -314,8 +314,8 @@ gen_clean_arch() {
 	info "Remove installed kubernetes packages and configuration"
 	if [ "$ID" == ubuntu ]; then
 		sudo rm -rf /etc/systemd/system/kubelet.service.d
-		sudo apt-get remove kubeadm kubelet kubectl docker.io docker-ce containerd.io docker-ce-cli -y
-		sudo apt autoremove -y
+		sudo apt-get autoremove -y kubeadm kubelet kubectl \
+			$(dpkg -l | awk '{print $2}' | grep -E '^(containerd(.\io)?|docker(\.io|-ce(-cli)?))$')
 	fi
 	# Remove existing CNI configurations and binaries.
 	sudo sh -c 'rm -rf /opt/cni/bin/*'


.ci/lib.sh
```diff
@@ -314,7 +314,7 @@ gen_clean_arch() {
 	info "Remove installed kubernetes packages and configuration"
 	if [ "$ID" == ubuntu ]; then
 		sudo rm -rf /etc/systemd/system/kubelet.service.d
-		sudo apt-get remove kubeadm kubelet kubectl docker.io docker-ce containerd.io docker-ce-cli -y
+		sudo apt-get remove kubeadm kubelet kubectl docker.io docker-ce containerd docker-ce-cli -y
```
jodh-intel, line 317: Couple of queries:

- Should we try to remove both `containerd` and `continerd.io`?
- If `containerd*` isn't installed, this command will fail. And since this script might be sourced with `set -e`, we probably need to add `|| true` at the end. Alternatively, the script could query the name of the containerd package before trying to remove it (which sounds safer to me).

/cc @GabyCT, @chavafg.

Jakob-Naucke: > * Should we try to remove both `containerd` and `continerd.io`?

@jodh-intel This is run for Ubuntu. It seems to me that `containerd.io` wasn't packaged on Ubuntu, ever. `containerd.io` was the package in RHEL/CentOS 7.
> * If `containerd*` isn't installed, this command will fail.

I just found out `apt-get remove` does not fail when a package is in index, but not installed. It does fail when it isn't in index. This will also fail on ppc64le & s390x since `docker-ce-cli` isn't referenced anywhere (`docker-ce` is referenced, but not installable, must use `docker.io`).
> And since this script might be sourced with `set -e`, we probably need to add `|| true` at the end. Alternatively, the script could query the name of the containerd package before trying to remove it (which sounds safer to me).

`"... docker* containerd*"` too broad?

jodh-intel: Hi @Jakob-Naucke.

> containerd.io wasn't packaged on Ubuntu, ever.

That implies this code never ran then, of if it did, it failed the job ;)

> `"... docker* containerd*"` too broad?

It might be. An `egrep` pattern of `^(containerd|containerd.io|docker|docker.io|...)$` might be safer if that looks reasonable (not sure about the `docker-ce` package names)?

/cc @GabyCT, @chavafg.

Jakob-Naucke: > That implies this code never ran then, of if it did, it failed the job ;)

This is only run on "bare metal" CIs, which isn't a lot of them. Also, _as is_, that code is not `set -e`'d, but your concern that it might be is right.
> > `"... docker* containerd*"` too broad?
> 
> It might be. An `egrep` pattern of `^(containerd|containerd.io|docker|docker.io|...)$` might be safer if that looks reasonable (not sure about the `docker-ce` package names)?

Sure. Let me get back to you.

Jakob-Naucke: @jodh-intel `download.docker.com` (which doesn't package for ppc64le/s390x) calls it `containerd.io`, that's where it's from. Excuse my s390x ignorance :)

jodh-intel: Yes, although this script isn't `set -e`, I think all users of this script are and `source` this file.

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm.

Jakob-Naucke:
The failures don't look related, also rebase
/test

Jakob-Naucke:
adding do not merge until release

fidencio:

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

fidencio:
@Jakob-Naucke, by the way, this also has to be backported to the stable-2.1 branch. Please, would you mind to provide the backport?

Jakob-Naucke:
> @Jakob-Naucke, by the way, this also has to be backported to the stable-2.1 branch. Please, would you mind to provide the backport?

@fidencio sure. s390x just passed, let me wait a couple more runs which will surely happen today to ensure it's stable.

Jakob-Naucke:
> @Jakob-Naucke, by the way, this also has to be backported to the stable-2.1 branch. Please, would you mind to provide the backport?

@fidencio Just now I noticed that kata-containers CI uses stable-2.1 tests and this fix couldn't possibly have been picked up. I'll go ahead and create the PR on stable-2.1 :smile:

Jakob-Naucke:
see #3741

Issue #3714, opened 2021-07-13
Do retry for wget in case nginx connectivity


liubin:
Though #3702 has been merged, the nginx connectivity still sometimes fails.

The get will run only once, adding retry will let it more robust.

PR #3715, opened 2021-07-13
tests: do retry for running wget command in pod

liubin:
Only run wget once may fail without retries.
The wget command in busybox image doesn't support
many options that used to control the retry, so here a
while-loop of shell is used.

And this commit also rename nginx.bats
to k8s-nginx-connectivity.bats to follow the naming rule
with other files under the same directory.

Fixes: #3714


fidencio (approved):
lgtm, thanks @liubin!

fidencio:
@wainersm, please, take a look at this one as well.

jodh-intel (approved):
Thanks @liubin.

lgtm

PR #3716, opened 2021-07-13
static-checks: Friendly messages when required commands aren't found

wainersm:
The `.ci/static-checks.sh` script relies on external tools provided by the distro (e.g. `chronic` and `pandoc`) as well as on in-house ones which are built and installed under `$GOPATH/bin`. For a first comer to ensure that those tools are installed and searchable may not be obvious. So on this PR I tried to make the script more friendly by pointing out where those tools live (what they eat? :) and how to install some of them.

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -164,6 +164,8 @@ Parameters:
 Notes:
 
 - If no options are specified, all non-skipped tests will be run.
+- Some required tools may be installed in \$GOPATH/bin, so you should ensure
+  that it is in your \$PATH.
 
 Examples:
 
@@ -239,6 +241,9 @@ static_check_commits()
 	# for main branch
 	(cd "${tests_repo_dir}" && make checkcommits && git remote set-branches origin 'main' && git fetch -v)
 
+	command -v checkcommits &>/dev/null || \
+		die 'checkcommits command not found. Ensure that "$GOPATH/bin" is in your $PATH.'
+
 	# Check the commits in the branch
 	{
 		checkcommits \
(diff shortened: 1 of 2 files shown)


devimc (approved):
thanks @wainersm

.ci/static-checks.sh
```diff
@@ -310,6 +325,8 @@ static_check_go_arch_specific()
 
 		info "Forcing ${linter} version ${linter_version}"
 		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin "${linter_version}"
+		command -v $linter &>/dev/null || \
+			die "$linter command not found. Ensure that \"\$GOPATH/bin\" is in your \$PATH."
```
liubin, line 329: Nit: Some usages of `die`  are using single quotes to quote the error message, which will not need backslashes to escape the dollar char. So should all use single quotes?

fidencio:
I'm adding the `do-not-merge` label till we get `2.2.0-alpha1` release out.
Once we cut the release I'll come back removing the label from the PRs.

.ci/static-checks.sh
wainersm: Hi @liubin , this line is the only exception because the tool's name is in a variable (`$linter`). The other error messages I had the tool's name hard-coded in the text so I could use single-quotes. Or am I missing something? Thanks for reviewing it!

amshinde (approved):

.ci/static-checks.sh
```diff
@@ -239,6 +251,9 @@ static_check_commits()
 	# for main branch
 	(cd "${tests_repo_dir}" && make checkcommits && git remote set-branches origin 'main' && git fetch -v)
 
+	command -v checkcommits &>/dev/null || \
+		die 'checkcommits command not found. Ensure that "$GOPATH/bin" is in your $PATH.'
```
jodh-intel, line 255: Nit: Alternatively, you could create a function that accepts the command name and calls die specifying `$cmd` if not found to avoid the duplication in all these calls.

.ci/static-checks.sh
```diff
@@ -230,6 +232,16 @@ pkg_to_path()
 	go list -f '{{.Dir}}' "$pkg"
 }
 
+# Check that chronic is installed, otherwise die.
+need_chronic() {
+	local first_word
+	[ -z "$chronic" ] && return
+	first_word="${chronic%% *}"
```
jodh-intel, line 239: What is this variable for?

jodh-intel (approved):

fidencio:

GabyCT (approved):

Issue #3728, opened 2021-07-19
Update SUSE support, especially for s390x


Jakob-Naucke:
quite the bunch, see PR

PR #3634, opened 2021-06-18
ci: Minor script fixes

Jakob-Naucke:
- CRI-containerd installation needs our $BUILDTAGS for its $BUILD_TAGS
  env -- not the other way around...
- Install ninja-build for QEMU on CentOS
- Install openssl non-`-devel` on s390x Fedora-likes for generating
  X.509 certificates during kernel build
- Install curl as general dependency on Ubuntu (used for checking
  versions)
- Fix non-fatal dependency typos in Ubuntu setup

Fixes: #3633

Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

PR #3629, opened 2021-06-17
containerd: fix the issue of getting cri version

lifupan:
Since 1.5.0, containerd had embedded cri and there's no
need to get the cri version.

Depends-on: github.com/kata-containers/kata-containers#2040

Fixes:#3630

Signed-off-by: fupan.lfp <fupan.lfp@antgroup.com>

PR #3729, opened 2021-07-19
Update SUSE support, especially for s390x

Jakob-Naucke:
- ci: Apply fixes from CentOS/RHEL to SLE
  - no seccomp & btrfs
  - CC=gcc on s390x
- ci: Re-apply $BUILD_TAGS fix
  - PR #3634 fixed use of the $BUILD_TAGS variable, which was later overwritten in #3629. Re-apply this fix.
- ci: Add a non-k8s CRI_CONTAINERD job
  - Kubernetes is really hard to set up on SUSE (I failed to set it up even on x86). The "blessed" way to run k8s on SUSE is Kubic, but that is e.g. not built for s390x. Add a job for cri-containerd without k8s.
- integration: Copy configuration.toml
  - Copy `/usr/share/defaults/kata-containers/configuration.toml` to `/etc/kata-containers` (previously only done on Ubuntu). Also fix trailing space.
- ci: Fix SLE package install
  - Skip all the individual repositories, just add openSUSE Leap OSS. On s390x, use 15.3 or newer, because that is the earliest openSUSE that is built.
  - Allow pandoc to not be installed (e.g. currently not in openSUSE Leap 15.3 OSS s390x)
  - Install libpmem on supported architectures only (cf. #2341)
  - Install ninja for QEMU (cf. #3264)
  - Install protobuf for required architectures (cf. #3612)
- ci: Fix openSUSE package install
  - Allow pandoc to not be installed (e.g. currently not in openSUSE Leap 15.3 OSS s390x)
  - Install libpmem on supported architectures only (cf. #2341)
  - Install ninja for QEMU (cf. #3264)
  - Install protobuf for required architectures (cf. #3612)
- ci: Support SUSE Docker install for s390x
  - s390x can't use download.docker.com, must use distro package. Support SUSE Linux Enterprise & openSUSE.

Fixes: #3728

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -37,6 +37,12 @@ else
 	export GOPATH="${GOPATH:-$HOME/go}"
 fi
 
+if [ "$(uname -m)" == "s390x" ] && grep -Eq "\<(fedora|suse)\>" /etc/os-release 2> /dev/null; then
+	# see https://github.com/kata-containers/osbuilder/issues/217
+	export CC=gcc
+fi
+
+
 tests_repo="${tests_repo:-github.com/kata-containers/tests}"
 lib_script="${GOPATH}/src/${tests_repo}/lib/common.bash"
 source "${lib_script}"
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -42,7 +42,7 @@ install_from_source() {
 		cd "${GOPATH}/src/${cri_containerd_repo}" >>/dev/null
 		git fetch
 		git checkout "${cri_containerd_tarball_version}"
-		make BUILDTAGS="${BUILD_TAGS:-}" cri-cni-release
+		make BUILD_TAGS="${BUILDTAGS:-}" cri-cni-release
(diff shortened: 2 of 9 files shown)


GabyCT (approved):

.ci/lib.sh
```diff
@@ -37,6 +37,12 @@ else
 	export GOPATH="${GOPATH:-$HOME/go}"
 fi
 
+if [ "$(uname -m)" == "s390x" ] && grep -Eq "\<(fedora|suse)\>" /etc/os-release 2> /dev/null; then
+	# see https://github.com/kata-containers/osbuilder/issues/217
+	export CC=gcc
```
fidencio, line 42: Not for now, but shouldn't we also expand the test to support RHEL?  I get that CentOS is out of question as CentOS doesn't provide a 390x build.

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

.ci/lib.sh
Jakob-Naucke: @fidencio RHEL puts `ID_LIKE="fedora"` in its `/etc/os-release`, this test was written with that in mind.

fidencio: Perfect then, thanks @Jakob-Naucke!

PR #3748, opened 2021-07-22
metrics/storage: Webtooling script improvements.

dborquez:
* This PR breaks the code in functions that are executed with timeout
* Cleans the environment beforethe workload starts and at the end

Fixes kata-containers#3747

Signed-off-by: desparza <david.esparza.borquez@intel.com>

### DIFF
--- a/metrics/storage/webtooling.sh
+++ b/metrics/storage/webtooling.sh
@@ -13,6 +13,7 @@ set -o pipefail
 # General env
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/../lib/common.bash"
+NUM_CONTAINERS="$1"
 
 TEST_NAME="web-tooling"
 IMAGE="docker.io/library/local-web-tooling:latest"
@@ -26,11 +27,12 @@ file_path="/web-tooling-benchmark"
 file_name="output"
 CMD="mkdir -p ${TESTDIR}; cd $file_path && node dist/cli.js > $file_name"
 PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
-NUM_CONTAINERS="$1"
+
 # This timeout is related with the amount of time that
 # webtool benchmark needs to run inside the container
-timeout=$((180 * "$NUM_CONTAINERS"))
-declare -a CONTAINERS_ID
+timeout=600
+INITIAL_NUM_PIDS=1
+
 cpu_period="100000"
 cpu_quota="200000"
 
@@ -47,8 +49,49 @@ help(){
 cat << EOF
 Usage: $0 <count>
    Description:
-        <count> : Number of containers to run.
+       <count> : Number of containers to run.
 EOF
+}
+
+verify_task_is_completed_on_all_containers() {
+	local containers=( $(sudo ctr c list -q) )
+	local sleep_secs=10
+	local max=$(bc <<<"$timeout / $sleep_secs")
+	local wip_list=()
+	local count=1
+	local sum=0
+	local i=""
+
+	while (( $sum < $NUM_CONTAINERS )); do
+
+	    for i in "${containers[@]}"; do
+		# Only check containers that have not completed the workload at this step
+		num_pids=$(sudo ctr t metrics "$i" | grep pids.current | grep pids.current | xargs | cut -d ' ' -f 2)
+
+	        if [ "$num_pids" -lt "$INITIAL_NUM_PIDS" ]; then
+                    ((sum++))
+                else
+                    wip_list+=("$i")
+                fi
+            done
+
+            # hold the list of containers that are still running the workload
+            containers=(${wip_list[*]})
+            wip_list=()
+
+            info "loop $count of $max: sleeping for $sleep_secs seconds"
+            sleep $sleep_secs
+            ((count++))
+	done
+}
+
+check_containers_are_up() {
+	info "Verify that the containers are running"
+	local containers_launched=0
+	while (( $containers_launched < $NUM_CONTAINERS )); do
+		containers_launched="$(sudo ctr t list | grep -c "RUNNING")"
+		sleep 1
+	done
 }
 
 save_config(){
@@ -74,55 +117,76 @@ function main() {
 		exit 1
 	fi
 
-	containers=()
-	# Check tools/commands dependencies
-	local cmds=()
-	cmds+=("docker")
+	local i=0
+	local containers=()
+	local cmds=("docker")
+	local not_started_count=$NUM_CONTAINERS
 
 	restart_containerd_service
+	# Check tools/commands dependencies
 	init_env
 	check_cmds "${cmds[@]}"
 	check_ctr_images "$IMAGE" "$DOCKERFILE"
-
 	metrics_json_init
-
 	save_config
+
+	info "Creating $NUM_CONTAINERS containers"
 
 	for ((i=1; i<= "$NUM_CONTAINERS"; i++)); do
 		containers+=($(random_name))
 		# Web tool benchmark needs 2 cpus to run completely in its cpu utilization
 		sudo -E ctr run -d --runtime "${CTR_RUNTIME}" --cpu-quota "${cpu_quota}" --cpu-period "${cpu_period}" "$IMAGE" "${containers[-1]}" sh -c "$PAYLOAD_ARGS"
+		((not_started_count--))
+		info "$not_started_count remaining containers"
 	done
 
-	# We verify that number of containers that we selected
-	# are running
-	for i in $(seq "$timeout") ; do
-		echo "Verify that the containers are running"
-		containers_launched=$(sudo ctr c list -q | wc -l)
-		[ "$containers_launched" -eq "$NUM_CONTAINERS" ] && break
-		sleep 1
-		[ "$i" == "$timeout" ] && return 1
-	done
+	# Check that the requested number of containers are running
+	local timeout_launch="10"
+	check_containers_are_up & pid=$!
+	(sleep "$timeout_launch" && kill -HUP $pid) 2>/dev/null & pid_tout=$! 
+
+	if wait $pid 2>/dev/null; then
+		pkill -HUP -P $pid_tout
+		wait $pid_tout
+	else
+		warn "Time out exceeded"
+		return 1
+	fi
+
+	# Get the initial number of pids in a single container before the workload starts
+        INITIAL_NUM_PIDS=$(sudo ctr t metrics "${containers[-1]}" | grep pids.current | grep pids.current | xargs | cut -d ' ' -f 2)
+	((INITIAL_NUM_PIDS++))
 
 	# Launch webtooling benchmark
-	CONTAINERS_ID=($(sudo ctr c list -q))
-	for i in "${CONTAINERS_ID[@]}"; do
-		sudo ctr t exec -d --exec-id "$(random_name)" "$i" sh -c "$CMD"
+	local pids=()
+	local j=0
+	for i in "${containers[@]}"; do
+		$(sudo ctr t exec -d --exec-id "$(random_name)" "$i" sh -c "$CMD") &
+		pids[${j}]=$!
+		((j++))
 	done
 
-	CONTAINERS_ID=($(sudo ctr c list -q))
-	for j in $(seq 1 "$timeout"); do
-		FILE_CMD="cat $file_path/$file_name | grep Geometric"
-		# Here we are using i to iterate with the proper container id and perform an exec
-		check_file_content=$(sudo ctr t exec --exec-id "$(random_name)" "$i" sh -c "$FILE_CMD")
-		[ ! -z "$check_file_content" ] && break
-		sleep 1
+	# wait for all pids
+	for pid in ${pids[*]}; do
+	    wait $pid
 	done
+
+	info "All containers are running the workload..."
+
+	# Verify that all containers have completed the assigned task
+	verify_task_is_completed_on_all_containers & pid=$!
+	(sleep "$timeout" && kill -HUP $pid) 2>/dev/null & pid_tout=$!
+	if wait $pid 2>/dev/null; then
+		pkill -HUP -P $pid_tout
+		wait $pid_tout
+	else
+		warn "Time out exceeded"
+		return 1
+	fi
 
 	RESULTS_CMD="cat $file_path/$file_name"
-	CONTAINERS_ID=($(sudo ctr c list -q))
-	for i in "${CONTAINERS_ID[@]}"; do
-		sudo ctr t exec --exec-id "$(random_name)" "$i" sh -c "$RESULTS_CMD" >> "$TMP_DIR"/results
+	for i in "${containers[@]}"; do
+		sudo ctr t exec --exec-id "$RANDOM" "$i" sh -c "$RESULTS_CMD" >> "$TMP_DIR/results"
 	done
 
 	# Save configuration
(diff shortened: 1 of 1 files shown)


GabyCT:
@dborquez thanks for the PR , I will review it and test it :)

metrics/storage/webtooling.sh
```diff
@@ -51,6 +52,44 @@ Usage: $0 <count>
 EOF
 }
 
+clean() {
+	sudo systemctl restart containerd && sleep 1
+	local containers=( $(sudo ctr c list -q) )
+	for i in "${containers[@]}"; do
```
jodh-intel, line 58: Nit: `local i`

metrics/storage/webtooling.sh
```diff
@@ -51,6 +52,44 @@ Usage: $0 <count>
 EOF
 }
 
+clean() {
+	sudo systemctl restart containerd && sleep 1
```
jodh-intel, line 56: Do we need this `sleep`?

metrics/storage/webtooling.sh
```diff
@@ -51,6 +52,44 @@ Usage: $0 <count>
...
+	        # Only check containers that have not completed the task at this step
+	        if [[ ! " ${completed[@]} " =~ " ${i} " ]]; then
+                    check_file_content=$(sudo ctr t exec --exec-id "$(random_name)" "$i" sh -c "$FILE_CMD")
+                    if [ ! -z "$check_file_content" ]; then
+                        ((count=count+1))
+                        completed+=("$i")
+                    fi
+                fi
+            done
+            sleep 10
```
jodh-intel, line 80: This is a long sleep. If it has to be this long, can you add a log call so it's clear what is happening. Also, do we need a timeout in this loop so we could log something like this each time through:

```bash
local sleep_secs=10

  :

echo "INFO: loop $count of $max: sleeping for $sleep_secs"
sleep $sleep_secs
```

jodh-intel (commented):
Thanks @dborquez. A couple of comments.

metrics/storage/webtooling.sh
dborquez: thank you

metrics/storage/webtooling.sh
marcel-apf: Yes, I saw this issue in my environment.
However choosing a random number of seconds might not be enough.
Querying the service status might be better (every second for X times) might be better but also it might not be enough since the service might be active but trying to access containerd might result in an error since is not "ready".
Maybe a specialized restart_containerd function in lib that ensures containerd API is ready?

The easiest way to reproduce is to restart containerd and pull some image.

metrics/storage/webtooling.sh
dborquez: Thank you, fixed.

Jakob-Naucke: @dborquez it seems you forgot to push the change?

dborquez: @Jakob-Naucke Hi, I was testing some fixes. But I'm done.

Jakob-Naucke: @dborquez after making your changes, you should amend them to your commit, rebase it, and force-push it, see https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#github-workflow

metrics/storage/webtooling.sh
dborquez: thank you @jodh-intel, sleep removed

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
+	clean
 	init_env
 	check_cmds "${cmds[@]}"
 	check_ctr_images "$IMAGE" "$DOCKERFILE"
-
 	metrics_json_init
-
 	save_config
 
+	echo "INFO: creating $NUM_CONTAINERS containers"
```
marcel-apf, line 148: Please use `info "Creating..."` instead of `echo INFO:...`

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
+
 	for ((i=1; i<= "$NUM_CONTAINERS"; i++)); do
-		containers+=($(random_name))
+		c_name="$(random_name)"
+		containers+=("$c_name")
 		# Web tool benchmark needs 2 cpus to run completely in its cpu utilization
-		sudo -E ctr run -d --runtime "${CTR_RUNTIME}" --cpu-quota "${cpu_quota}" --cpu-period "${cpu_period}" "$IMAGE" "${containers[-1]}" sh -c "$PAYLOAD_ARGS"
+		sudo -E ctr run -d --runtime "${CTR_RUNTIME}" --cpu-quota "${cpu_quota}" --cpu-period "${cpu_period}" "$IMAGE" "${c_name}" sh -c "$PAYLOAD_ARGS"
+		((not_started_count--))
+		echo "INFO: $not_started_count remaining containers"
```
marcel-apf, line 156: Same here

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
+	# Check that the requested number of containers are running
+	local timeout_launch="10"
+	check_containers_are_up & pid=$!
+	(sleep "$timeout_launch" && kill -HUP $pid) 2>/dev/null & pid_tout=$! 
+
+	if wait $pid 2>/dev/null; then
+		pkill -HUP -P $pid_tout
+		wait $pid_tout
+	else
+		echo "INFO: Time out exceeded"
```
marcel-apf, line 168: Please use `warn` or `error`  instead of info

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
-		# Here we are using i to iterate with the proper container id and perform an exec
-		check_file_content=$(sudo ctr t exec --exec-id "$(random_name)" "$i" sh -c "$FILE_CMD")
-		[ ! -z "$check_file_content" ] && break
-		sleep 1
+	# wait for all pids
+	for pid in ${pids[*]}; do
+	    wait $pid
 	done
 
+	echo -e "\nINFO:All containers are running the workload...\n"
```
marcel-apf, line 190: Use `info` please. I suppose the new line is on purpose?

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
+
+	# Verify that all containers have completed the assigned task
+	verify_task_is_completed_on_all_containers & pid=$!
+	(sleep "$timeout" && kill -HUP $pid) 2>/dev/null & pid_tout=$!
+	if wait $pid 2>/dev/null; then
+		pkill -HUP -P $pid_tout
+		wait $pid_tout
+	else
+		echo "INFO: Time out exceeded"
```
marcel-apf, line 199: warn/error

metrics/storage/webtooling.sh
```diff
@@ -47,10 +53,61 @@ help(){
...
+            wip_list=()
+
+            echo "INFO: loop $count of $max: sleeping for $sleep_secs seconds"
+            sleep $sleep_secs
+            ((count++))
+	done
+}
+
+check_containers_are_up() {
+	echo "Verify that the containers are running"
```
marcel-apf, line 103: Use `info` please

metrics/storage/webtooling.sh
```diff
@@ -47,10 +53,61 @@ help(){
...
-        <count> : Number of containers to run.
+       <count> : Number of containers to run.
 EOF
 }
 
+clean() {
+	local containers=( $(sudo ctr c list -q) )
+	local i=""
+	for i in "${containers[@]}"; do
+            	echo "INFO: loop $count of $max: sleeping for 1 second"
```
marcel-apf, line 64: Use `info` please

marcel-apf (commented):
I only have a few comments regarding the logging, thank you

metrics/storage/webtooling.sh
dborquez: @marcel-apf  thank you, changes applied

metrics/storage/webtooling.sh
dborquez: @marcel-apf  thank you, changes applied

metrics/storage/webtooling.sh
dborquez: @Jakob-Naucke done, thanks.

GabyCT:
@dborquez it seems that the containers are being killed before they finished running the workload which is generating empty results

metrics/storage/webtooling.sh
```diff
@@ -47,10 +53,61 @@ help(){
...
+	done
+}
+
+verify_task_is_completed_on_all_containers() {
+	local containers=( $(sudo ctr c list -q) )
+	local sleep_secs=10
+	local max=$(bc <<<"$timeout / $sleep_secs")
+	local wip_list=()
+	local count=1
+	local sum=1
```
dborquez, line 76: @GabyCT The issue you mention is here, the sum variable need to be initialized to 0.

metrics/storage/webtooling.sh
```diff
@@ -13,8 +13,13 @@ set -o pipefail
 # General env
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/../lib/common.bash"
+NUM_CONTAINERS="$1"
+
+# runtimes: io.containerd.runc.v2, io.containerd.run.kata.v2
+CTR_RUNTIME="${CTR_RUNTIME:-io.containerd.run.kata.v2}"
```
GabyCT, line 19: you do not need to specify again the value of `CTR_RUNTIME` as it is already defined at the common.bash

metrics/storage/webtooling.sh
```diff
@@ -47,10 +53,61 @@ help(){
 cat << EOF
 Usage: $0 <count>
    Description:
-        <count> : Number of containers to run.
+       <count> : Number of containers to run.
 EOF
 }
 
+clean() {
```
GabyCT, line 60: maybe this function can be implemented here https://github.com/kata-containers/tests/blob/09a62469fe1343d2a34d0ff933b37b8176ab6dd3/lib/common.bash#L168, as it is kind of similar of what you are doing which I believe will help other tests that are using this function

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
 	metrics_json_init
-
 	save_config
 
+	info "Creating $NUM_CONTAINERS containers"
+
 	for ((i=1; i<= "$NUM_CONTAINERS"; i++)); do
-		containers+=($(random_name))
+		c_name="$(random_name)"
+		containers+=("$c_name")
```
GabyCT, line 152: why have two variables ? why not just `c_name+="$(random_name)"` or just `containers+=($(random_name))`

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
+	((INITIAL_NUM_PIDS++))
 
 	# Launch webtooling benchmark
-	CONTAINERS_ID=($(sudo ctr c list -q))
-	for i in "${CONTAINERS_ID[@]}"; do
-		sudo ctr t exec -d --exec-id "$(random_name)" "$i" sh -c "$CMD"
+	local pids=()
+	local j=0
+	for i in "${containers[@]}"; do
+		$(sudo ctr t exec -d --exec-id "$i" "$i" sh -c "$CMD") &
```
GabyCT, line 180: this is kind of weird `sudo ctr t exec -d --exec-id "$i" "$i"` you are using twice `$i` the variable , I prefer as it was before as it is easier to read

metrics/storage/webtooling.sh
```diff
@@ -74,55 +131,78 @@ function main() {
...
+		warn "Time out exceeded"
+		return 1
+	fi
+
 	RESULTS_CMD="cat $file_path/$file_name"
-	CONTAINERS_ID=($(sudo ctr c list -q))
-	for i in "${CONTAINERS_ID[@]}"; do
-		sudo ctr t exec --exec-id "$(random_name)" "$i" sh -c "$RESULTS_CMD" >> "$TMP_DIR"/results
+	for i in "${containers[@]}"; do
+		sudo ctr t exec --exec-id "$RANDOM" "$i" sh -c "$RESULTS_CMD" >> "$TMP_DIR/results"
```
GabyCT, line 205: where `RANDOM` is defined? I could not see it

metrics/storage/webtooling.sh
GabyCT: use RANDOM instead of two $i

metrics/storage/webtooling.sh
dborquez: got it, thank you @GabyCT

metrics/storage/webtooling.sh
dborquez: @GabyCT got it, removed from the script

metrics/storage/webtooling.sh
dborquez: @GabyCT good recommendation to compact the code.

metrics/storage/webtooling.sh
dborquez: @GabyCT the purpose to use the same variable to name both the pid and the container, was because you can avoid an additional random generation call.

metrics/storage/webtooling.sh
dborquez: $RANDOM is an internal Bash function that returns an integer in the range 0 - 32767

metrics/storage/webtooling.sh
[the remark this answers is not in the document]
dborquez: @GabyCT function clean( ) removed, thanks

metrics/storage/webtooling.sh
```diff
@@ -51,6 +52,44 @@ Usage: $0 <count>
...
+verify_task_is_completed_on_all_containers() {
+	local FILE_CMD="cat $file_path/$file_name | grep Geometric"
+	local containers=( $(sudo ctr c list -q) )
+	local completed=()
+	local count=1
+	while [ $count -le $NUM_CONTAINERS ]; do
+	    for i in "${containers[@]}"; do
+	        # Only check containers that have not completed the task at this step
+	        if [[ ! " ${completed[@]} " =~ " ${i} " ]]; then
+                    check_file_content=$(sudo ctr t exec --exec-id "$(random_name)" "$i" sh -c "$FILE_CMD")
```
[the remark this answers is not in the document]
dborquez: @GabyCT thanks, solved: https://github.com/kata-containers/tests/blob/c150e69764e1d1ca009c0fb8dceec6d2b2715000/metrics/storage/webtooling.sh#L70

GabyCT (approved):

metrics/storage/webtooling.sh
```diff
@@ -26,11 +27,12 @@ file_path="/web-tooling-benchmark"
 file_name="output"
 CMD="mkdir -p ${TESTDIR}; cd $file_path && node dist/cli.js > $file_name"
 PAYLOAD_ARGS="${PAYLOAD_ARGS:-tail -f /dev/null}"
-NUM_CONTAINERS="$1"
+
 # This timeout is related with the amount of time that
 # webtool benchmark needs to run inside the container
-timeout=$((180 * "$NUM_CONTAINERS"))
-declare -a CONTAINERS_ID
+timeout=600
```
jcvenegas, line 33: Can we say that timeout is not dependent on the number of containers?

jcvenegas (approved):

PR #3547, opened 2021-05-19
Allow to reproduce CI jobs locally on VMs

wainersm:
This is a PoC aiming to allow to reproduce (Jenkins) CI jobs locally on VMs.

The implementation I am proposing here use vagrant (https://www.vagrantup.com) to orchestrate the VM creation, provision, connection and destroy. Under the hooks It uses Libvirt + QEMU but vagrant is flexible enough to allow other providers like VirtualBox. I do believe the libvirt and virtualbox can coexist.

Initially I was creating a VM from Ubuntu and it was needed to use virtualbox. As I had libvirt already installed on my host, switched to Fedora 32. However I plan to evolve the Vagrant file so that the user can choose the distro (ubuntu, centos, fedora...etc). I glanced at https://www.vagrantup.com/docs/multi-machine, it seems doable.

Other than that, maybe the guest and host files could share the repositories (instead of simply copy). Vagrant allows to use NFS or 9p for that purpose. I tried NFS but it failed so I switched to the more simple approach (rsync).

I updated the README.md so there you can find the rationale behind this feature, some details of implementation and how to use it.

Here the tests passed with CI_JOB="CRIO_K8S" and failed with CI_JOB="CRI_CONTAINERD_K8S". It might need to adjust some of the .ci scripts for the later job. I will give a try for the other jobs, although I think that VFIO and METRICS, for example, won't work out.

All in all, please share you opinion, thoughts, reviews and, if possible, help me by testing it in your environment.

PR #3751, opened 2021-07-23
Allow to reproduce CI jobs locally on VMs - part 2

wainersm:
This is a continuation of  https://github.com/kata-containers/tests/pull/3547 ((Allow to reproduce CI jobs locally on VMs).

The goal here is make even easier to execute the CI job whether on localhost or inside VM, by introducing makefile targets and some scripts (called *runners*) that abstract the environment where the job will be executed.

Although I didn't think in every and each details, the initial design is meant to allow us to add new runners (e.g. containers) easily. And for the vm_runner, back-ends other than Vagrant (e.g. pure libvirt scripts) can also be implemented.

There are other improvements already on my mind, for instance, allow the user to login the VM to debug a problem. But I will leave those for a part 3.

A typical usage session would be:
```
[wmoschet@wainer-laptop .ci]$ make help
Job related targets

  job-run-NAME             - Run the job NAME. Use KATA_TESTS_JOB_RUNNER to set the runner.
  job-list                 - List all available CI jobs
help: CI targets
  vm-help                  - Help for the VM runner targets.
Environment variables:
  KATA_TESTS_JOB_RUNNER:   - local or vm (default)
[wmoschet@wainer-laptop .ci]$ make vm-help
vm-help: Manage CI VMs with the following targets:
  vm-clean                 - Destroy all VMs
  vm-list                  - Print all supported VM names
Environment variables
  KATA_TESTS_VM_ENGINE: set the VM engine. Defaults to 'vagrant'.
                        Available engines: vagrant
  KATA_TESTS_VM_NAME:   the VM name.
[wmoschet@wainer-laptop .ci]$ make vm-list
The avaliable VMs:
fedora
ubuntu
[wmoschet@wainer-laptop .ci]$ make job-list
baremetal-pmem cri_containerd_k8s cri_containerd_k8s_initrd cri_containerd_k8s_complete cri_containerd_k8s_minimal crio_k8s crio_k8s_complete crio_k8s_minimal cloud-hypervisor-k8s-crio cloud-hypervisor-k8s-containerd cloud-hypervisor-k8s-containerd-minimal Ccloud-hypervisor-k8s-containerd-fullL firecracker vfio virtiofs_experimental metrics metrics_experimental
[wmoschet@wainer-laptop .ci]$ KATA_TESTS_VM_NAME=fedora make job-run-crio_k8s
[wmoschet@wainer-laptop .ci]$ make vm-clean
```

### DIFF
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -1,0 +1,25 @@
+# Copyright (c) 2021 Red Hat
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Include the VM runner specific targets.
+include job/vm/Makefile.include
+
+.PHONY: help
+help:
+	@echo "Job related targets"
+	@echo ""
+	@echo "  job-run-NAME             - Run the job NAME. Use KATA_TESTS_JOB_RUNNER to set the runner."
+	@echo "  job-list                 - List all available CI jobs"
+	@echo "help: CI targets"
+	@echo "  vm-help                  - Help for the VM runner targets."
+	@echo "Environment variables:"
+	@echo "  KATA_TESTS_JOB_RUNNER:   - local or vm (default)"
+
+.PHONY: job-list
+job-list:
+	@bash ./job/job_runner.sh -l
+
+job-run-%:
+	@bash ./job/job_runner.sh -j "$*"
--- a/.ci/job/job_runner.sh
+++ b/.ci/job/job_runner.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+cidir="$(dirname $0)/../"
+source "$cidir/lib.sh"
+
+#
+# Variables that the user can overwrite
+#
+KATA_TESTS_JOB_RUNNER="${KATA_TESTS_JOB_RUNNER:-vm}"
+
+#
+# Script variables
+#
+
+ALL_AVAILABLE_JOBS=(
+	"baremetal-pmem"
+	"cri_containerd_k8s"
+	"cri_containerd_k8s_initrd"
+	"cri_containerd_k8s_complete"
+	"cri_containerd_k8s_minimal"
+	"crio_k8s"
+	"crio_k8s_complete"
+	"crio_k8s_minimal"
+	"cloud-hypervisor-k8s-crio"
+	"cloud-hypervisor-k8s-containerd"
+	"cloud-hypervisor-k8s-containerd-minimal"
+	"Ccloud-hypervisor-k8s-containerd-fullL"
+	"firecracker"
+	"vfio"
+	"virtiofs_experimental"
+	"metrics"
+	"metrics_experimental")
+
+usage() {
+	cat <<-EOF
+	Run a CI job.
+	User can select the runner which will trigger the job within a given environment.
+
+	Usage: $0 [-h] [-l] [-j JOB_ID]
+
+	Options:
+	       -l: list all defined jobs
+	       -j JOB_ID: run the job of JOB_ID
+
+	Environment variables:
+	        KATA_TESTS_JOB_RUNNER: switch between 'local' and 'vm' (default) runners"
+	EOF
+}
+
+list_jobs_id() {
+	echo ${ALL_AVAILABLE_JOBS[@]}
+}
+
+#
+# Note:
+#
+# Any function named as _do_run_NAME is interpreted as the implementation of
+# the NAME runner. So if you want to add new runners then you just need to
+# create a function with that name pattern.
+#
+
+# Local runner implementation.
+_do_run_local() {
+	local job_id="$1"
+	local runner_script="${cidir}/run.sh"
+	local setup_script="${cidir}/setup.sh"
+	# Convert the ID to uppercase.
+	typeset -u CI_JOB="$job_id"
+	# Export it.
+	typeset -x CI_JOB
+	bash "$setup_script"
+	bash "$runner_script"
+}
+
+# VM runner implementation.
+_do_run_vm() {
+	local job_id="$1"
+	local runner_script="${cidir}/job/vm/vm_runner.sh"
+	# Convert the ID to uppercase.
+	typeset -u CI_JOB="$job_id"
+        # Export it.
+        typeset -x CI_JOB
+	bash "$runner_script" -r "$job_id"
+}
+
+# Check if the runner $1 exists.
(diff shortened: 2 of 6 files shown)


Vagrantfile
```diff
@@ -76,6 +76,7 @@ EOF
         # This switches back to cgroup v1. It requires a reboot.
         sudo dnf install -y grubby
         sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
+        sudo dnf install -y make
```
dgibson, line 79: Why not combine this with the install of grubby above?

.ci/Makefile
```diff
@@ -0,0 +1,25 @@
...
+help:
+	@echo "Job related targets"
+	@echo ""
+	@echo "  job-run-NAME             - Run the job NAME. Use KATA_TESTS_JOB_RUNNER to set the runner."
+	@echo "  job-list                 - List all available CI jobs"
+	@echo "help: CI targets"
+	@echo "  vm-help                  - Help for the VM runner targets."
+	@echo "Environment variables:"
+	@echo "  KATA_TESTS_JOB_RUNNER:   - local or vm (default)"
+
```
dgibson, line 19: The new interface does look a lot nicer.  A couple of overall concerns:
 - Can the docs be updated to describe this new preferred way to run tests
 - Adding a few more hundred lines of fairly complex bash to the thousands of existing lines of fragile bash code makes me kinda nervous.  But maybe it's a necessary interim step.

.ci/job/vm/Makefile.include
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2021 Red Hat
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Specific Make targets for the VM runner.
+#
+
+KATA_TESTS_VM_ENGINE ?= vagrant
+supported_vm_engines:=$(shell ./job/vm/vm_runner.sh -e)
+
```
dgibson, line 10: You have two layers of abstraction here: first local vs. vm, then vagrant vs. other possible VM engines.  Is it worth the complexity?  Could you just have "local" and "vagrant" as the top level options.

Vagrantfile
```diff
@@ -93,7 +93,6 @@ EOF
     export osbuilder_distro="$distro"
 
     cd ${kata_tests_repo_dir}
-    sudo -E PATH=$PATH -H -u #{guest_user} bash -c '.ci/setup.sh'
```
dgibson: This seems like it definitely needs a docs update, because if you create a Vagrant vm directly (which you might want to do for some more specific test not covered by the existing jobs) it will no longer be set up to do that as is.

.ci/job/job_runner.sh
```diff
@@ -0,0 +1,131 @@
...
+_do_run_local() {
+	local job_id="$1"
+	local runner_script="${cidir}/run.sh"
+	local setup_script="${cidir}/setup.sh"
+	# Convert the ID to uppercase.
+	typeset -u CI_JOB="$job_id"
+	# Export it.
+	typeset -x CI_JOB
+	bash "$setup_script"
+	bash "$runner_script"
```
dgibson, line 77: Enforcing running the setup and runner immediately after each other worries me a bit.  Yes, that's currently the safe way to run the CI jobs, but
 - The setup script takes rather a long time
 - It means the setup and job output will be combined together, perpetuating Kata's existing problem of having huge logs where it's hard to figure out where things went wrong.

dgibson (commented):
I'm unclear why the first few patches still show up in this PR, even though they're merged already.  Does this need a rebase?

.ci/job/job_runner.sh
```diff
@@ -0,0 +1,131 @@
...
+	"cri_containerd_k8s_initrd"
+	"cri_containerd_k8s_complete"
+	"cri_containerd_k8s_minimal"
+	"crio_k8s"
+	"crio_k8s_complete"
+	"crio_k8s_minimal"
+	"cloud-hypervisor-k8s-crio"
+	"cloud-hypervisor-k8s-containerd"
+	"cloud-hypervisor-k8s-containerd-minimal"
+	"Ccloud-hypervisor-k8s-containerd-fullL"
```
GabyCT, line 32: /s/Ccloud-hypervisor-k8s-containerd-fullL/cloud-hypervisor-k8s-containerd-full

.ci/job/job_runner.sh
```diff
@@ -0,0 +1,131 @@
...
+	"firecracker"
+	"vfio"
+	"virtiofs_experimental"
+	"metrics"
```
GabyCT, line 36: mmm I think that these jobs will be hard to reproduce as they run in baremetal and even if somebody runs the tests locally...the results will not match

.ci/job/vm/vm_runner.sh
```diff
@@ -0,0 +1,148 @@
...
+source "${cidir}/lib.sh"
+
+# The VM engine. Default to Vagrant.
+KATA_TESTS_VM_ENGINE=${KATA_TESTS_VM_ENGINE:-"vagrant"}
+# The default VM name.
+KATA_TESTS_VM_NAME=${KATA_TESTS_VM_NAME:-"fedora"}
+
+# Some jobs cannot run inside a VM. For instance, those which need bare-metal
+# machine. So here it keeps a list of unsupported jobs.
+UNSUPPORTED_JOBS=(
```
GabyCT, line 20: what about metrics? also I believe vfio needs baremetal

.ci/job/job_runner.sh
dgibson: As long as the host allows for nested virt, I think it should be workable.

.ci/job/vm/vm_runner.sh
dgibson: No, surprisingly vfio does not run on baremetal right now: it explicitly looks for a second virtio-net device in its "host" to then pass through into the Kata guests.

wainersm:
Updated this pull request just to fix conflicts with latest version of Vagrantfile.

wainersm:
@dgibson @GabyCT @jodh-intel I really appreciated your review on this PR however I am thinking in closing it.

I explain: David's https://github.com/kata-containers/tests/pull/3751#discussion_r676294881 made me thoughtful whether this is the right time for adding another layer of complexity to the CI scripts. I really think the propose changes on this PR adds a more friendly interface to allow users running CI locally, but at this point I don't know if the Vagrant stuff has been useful to others than me. Therefore, first I will promote the use of Vagrant, fix whatever problems come up, and if I feel this extra interface is worth having in then I will re-open this PR.

GabyCT:
ok thanks @wainersm

PR #3759, opened 2021-07-27
Cleanups and simplifications to .ci/install_kata_kernel.sh

dgibson:
A number of small simplifications and cleanups to this CI script.

@marcel-apf, FYI

### DIFF
--- a/.ci/install_kata_kernel.sh
+++ b/.ci/install_kata_kernel.sh
@@ -17,7 +17,6 @@ source "${cidir}/lib.sh"
 source "/etc/os-release" || source "/usr/lib/os-release"
 
 latest_build_url="${jenkins_url}/job/kata-containers-2.0-kernel-vanilla-$(uname -m)-nightly/${cached_artifacts_path}"
-experimental_latest_build_url="${jenkins_url}/job/kata-containers-2.0-kernel-experimental-$(uname -m)-nightly/${cached_artifacts_path}"
 PREFIX="${PREFIX:-/usr}"
 kernel_dir="${DESTDIR:-}${PREFIX}/share/kata-containers"
 
@@ -25,8 +24,6 @@ kernel_repo_dir="${kata_repo_dir}/tools/packaging"
 kernel_arch="$(arch)"
 readonly tmp_dir="$(mktemp -d -t install-kata-XXXXXXXXXXX)"
 packaged_kernel="kata-linux-container"
-#Experimental kernel support. Pull from virtio-fs GitLab instead of kernel.org
-experimental_kernel="${experimental_kernel:-false}"
 
 exit_handler() {
 	rm -rf "${tmp_dir}"
@@ -35,11 +32,7 @@ exit_handler() {
 trap exit_handler EXIT
 
 get_current_kernel_version() {
-	if [ "$experimental_kernel" == "true" ]; then
-		kernel_version=$(get_version "assets.kernel-experimental.tag")
-	else
-		kernel_version=$(get_version "assets.kernel.version")
-	fi
+	kernel_version=$(get_version "assets.kernel.version")
 	echo "${kernel_version/v/}"
 }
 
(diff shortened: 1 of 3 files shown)


.ci/install_kata_kernel.sh
```diff
@@ -14,133 +14,69 @@ set -o errtrace
...
 	done
 
 	pushd "${kernel_dir}" >/dev/null
-	info "Verify download checksum"
-	if [ "${experimental_kernel}" == "true" ]; then
-		sudo -E curl -fsOL "${experimental_latest_build_url}/sha256sum-kernel" || return 1
-	else
-		sudo -E curl -fsOL "${latest_build_url}/sha256sum-kernel" || return 1
-        fi
+	sudo -E curl -fsOL "${latest_build_url}/sha256sum-kernel" || return 1
```
jodh-intel, line 60: Does this need to use `sudo`? We try to avoid running commands that hit the network as `root` unless absolutely essential.

jodh-intel (commented):
Thanks @dgibson. Just one comment.

fidencio:
I'd prefer to **not** have this one merged till @wainersm is back from his PTO.  I know some of the parts touched is used on the OpenShift CI he's running, and I wouldn't like to face the chance of having that broken for the next 20 days or so.

@dgibson, would be okay for you to wait till @wainersm is back from PTO, so he can also review this PR?

amshinde (approved):

fidencio:
Adding the `do-not-merge` label just to avoid getting it mistakenly merged, before we hear back from David,

.ci/install_kata_kernel.sh
dgibson: Maybe not.. note that's not a change in my patchset, though.  I'll look at adding another patch to the set which removes some unnecessary sudos.

dgibson:
> I'd prefer to **not** have this one merged till @wainersm is back from his PTO. I know some of the parts touched are used on the OpenShift CI he's running, and I wouldn't like to face the chance of having that broken for the next 20 days or so.
> 
> @dgibson, would be okay for you to wait till @wainersm is back from PTO, so he can also review this PR?

Hmm.. 20 days is longer than I'd ideally like to wait, but if we have to so be it.

I'm not averse to splitting the patchset up, so if there's ones you think are low risk we could go ahead with those and postpone the rest until later.

.ci/install_kata_kernel.sh
dgibson: Looking closer we can't simply remove the `sudo` invocations: in each case the curl is writing to files in locations that the ordinary user won't have access to.  We could however `curl` without `sudo` to a temporary location then copy into the final location with `sudo`.  I'll look into implementing this.

fidencio:
> > I'd prefer to **not** have this one merged till @wainersm is back from his PTO. I know some of the parts touched are used on the OpenShift CI he's running, and I wouldn't like to face the chance of having that broken for the next 20 days or so.
> > @dgibson, would be okay for you to wait till @wainersm is back from PTO, so he can also review this PR?
> 
> Hmm.. 20 days is longer than I'd ideally like to wait, but if we have to so be it.

Thanks for understanding.

.ci/install_kata_kernel.sh
dgibson: Ok, I've added a patch removing the `sudo curl`s to the series.

.ci/install_kata_kernel.sh
```diff
@@ -14,133 +14,71 @@ set -o errtrace
...
 	sudo mkdir -p "${kernel_dir}"
 	local kernel_binary_name="${kernel_binary}-${cached_kernel_version}"
 	local kernel_binary_path="${kernel_dir}/${kernel_binary_name}"
-	if [ "${experimental_kernel}" == "true" ]; then
-		sudo -E curl -fL --progress-bar "${experimental_latest_build_url}/${kernel_binary_name}" -o "${kernel_binary_path}" || return 1
-	else
-		sudo -E curl -fL --progress-bar "${latest_build_url}/${kernel_binary_name}" -o "${kernel_binary_path}" || return 1
-	fi
+	curl -fL --progress-bar "${latest_build_url}/${kernel_binary_name}" -o "${tmp_dir}/${kernel_binary_name}" || return 1
+	sudo mv ${tmp_dir}/${kernel_binary_name} ${kernel_binary_path}
```
jodh-intel, line 50: I wonder if we should use `install` to ensure root user/group ownership?

jodh-intel (approved):
Thanks @dgibson.

lgtm

GabyCT (approved):

Issue #3764, opened 2021-07-27
Add environment variables to the job output


marcel-apf:
**Which feature do you think can be improved?**

Being able to reproduce a CI failure would help developers to better debug failed tests.
Currently there is no way to run the same CI job(test) on a local server since the workflow is based on environment variables we don't have access to.

**How can it be improved?**

Add the list of the environment variables to the output log. 

Then once a job fails we could:
```sed -e '1,/Environment vars/ d' procenv-root.log |  grep "CI\|KATA\|CRI\|CTR\|METRICS\|TEST\|SANDBOX\|K8S\|KUBERNETES\|VIRTIO\|experimental" | sort | xargs -n1 echo export```

And receive the exact env list that can be exported and used by the setup/run scripts.

```export CI_JOB=CRIO_K8S
export CI=true
export CRI_CONTAINERD=no
export CRIO=yes
export CRI_RUNTIME=crio
export CTR_RUNTIME=io.containerd.run.kata.v2
export DEFSANDBOXCGROUPONLY=false
export experimental_kernel=false
export experimental_qemu=false
export KATA_DOCKER_TIMEOUT=30
export KATA_ETC_CONFIG_PATH=/etc/kata-containers/configuration.toml
export KATA_HYPERVISOR=qemu
export KATA_KSM_THROTTLER=no
export KATA_NEMU_DESTDIR=/usr
export KATA_NET_TIMEOUT=30
export KATA_OBS_REPO_BASE=http://download.opensuse.org/repositories/home:/katacontainers:/releases:/x86_64:/master
export KATA_QEMU_DESTDIR=/usr
export KUBERNETES=yes
export METRICS_CI_CLOUD=
export METRICS_CI=false
export METRICS_CI_PROFILE=
export METRICS_JOB_BASELINE=
export MINIMAL_K8S_E2E=false
export RUN_KATA_CHECK=true
export SHIMV2_TEST=true
export TEST_CGROUPSV2=false
export TEST_CRIO=false
```

PR #3765, opened 2021-07-27
ci/setup: Print the environment variables

marcel-apf:
Being able to reproduce a CI failure would help developers to
better debug failed tests.
Currently there is no way to run the same CI job(test) on a
local server since the workflow is based on environment variables we don't have access to.

Add the list of the environment variables to the procenv-user-log .

Fixes: #3764

Signed-off-by: Marcel Apfelbaum <marcel@redhat.com>

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -459,3 +459,13 @@ calc_qemu_files_sha256sum() {
 
 	sha256sum_from_files $files
 }
+
+# Outputs to stdout the environment variables of the current script.
+print_environment() {
+	if [ -n "$(command -v env)" ]
+	then
+		echo environment variables for $(basename "$0"):
+		env
+		echo "#########################################"
+	fi
+}
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -152,6 +152,7 @@ main() {
 
 	[ "$setup_type" = "minimal" ] && info "finished minimal setup" && exit 0
 
+	print_environment
 	install_docker
 	enable_nested_virtualization
 	install_kata


fidencio (approved):
@marcel-apf, lgtm!

One question that I have is why doing this as part of the teardown and not (also) in the setup function?  Anyways, it's an improvement from the current situation.

marcel-apf:
> @marcel-apf, lgtm!
> 
> One question that I have is why doing this as part of the teardown and not (also) in the setup function? Anyways, it's an improvement from the current situation.

Because some scripts may add / change environment variables and I would want the env snapshot when the test failed / finished.

I can add the env also to the start of the run script if you think it worth it.

Thanks,
Marcel

fidencio:
> > @marcel-apf, lgtm!
> > One question that I have is why doing this as part of the teardown and not (also) in the setup function? Anyways, it's an improvement from the current situation.
> 
> Because some scripts may add / change environment variables and I would want the env snapshot when the test failed / finished.
> 
> I can add the env also to the start of the run script if you think it worth it.

Let me start emphasizing that I think what you proposed is quite reasonable improvement.

In an ideal world, just calling `make $target` (see the targers [here](https://github.com/kata-containers/tests/blob/main/Makefile)) would set all the env vars we need for us and that's it.  However, it doesn't seem to be the case, considering what you're facing, right?

With this in mind, I'd like to see, for instance, the env vars being printed at the beginning of each target so I know what I have to set to locally run a specific test or target, as some of the env-vars may just change between one test and the next, and we're just getting the last set when printing in the teardown.  Does this make sense?

marcel-apf:
> 
> In an ideal world, just calling `make $target` (see the targers [here](https://github.com/kata-containers/tests/blob/main/Makefile)) would set all the env vars we need for us and that's it. However, it doesn't seem to be the case, considering what you're facing, right?
> 
> With this in mind, I'd like to see, for instance, the env vars being printed at the beginning of each target so I know what I have to set to locally run a specific test or target, as some of the env-vars may just change between one test and the next, and we're just getting the last set when printing in the teardown. Does this make sense?

@fidencio agreed, I will give it a try as a separate PR.

marcel-apf:
@fidencio BTW, this PR is not good enough, I don't see the CI_JOB env var...
/do-not-merge

devimc:
uhmm.. hold on.. :thinking:  `procenv` already do this, isn't it ??   http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-20.04-PR/ (procenv-jenkins.log.gz)

cc @jodh-intel

marcel-apf:
@devimc you are right! thanks for catching it. I looked a few times and just missed them. The actual reason is I was looking for the CI_JOB env variable that wasn't there... 

I am going to replace the commit in this PR with another that outputs of the environment variables at the beginning of each script.
@fidencio suggested to use the make targets, but maybe at script level, at least the ones that run the tests may give us a better granularity.

marcel-apf:
Posted a second try.
At some point we will need to do what @fidencio said and print the environment variables for each make target / relevant script, but for now I want to solve a smaller issue. I want to be able to reproduce a jenkins job on a local setup by running the setup/run scripts on the same environments variables like jenkins. I want to see at least the exported CI_JOB.

Then we can discuss if printing the environment variables at other stages.

.ci/lib.sh
```diff
@@ -459,3 +459,16 @@ calc_qemu_files_sha256sum() {
 
 	sha256sum_from_files $files
 }
+
+# Outputs to stdout the environment variables of the current script.
+print_environment() {
+	local have_env="no"
+	[ -n "$(command -v env)" ] && have_env="yes"
+
+	if [ "${have_env}" = "yes" ]
```
devimc, line 468: can we simplify this?

`if [ -n "$(command -v env)" ]`

marcel-apf: Definitely.
BTW, How can we trigger the tests?

marcel-apf: @devimc repushed

devimc:
@marcel-apf thanks, CI is not happy

```
ERROR: Commit 6d9c8b5dde00cb19402e9d2428430f54dcdb4f23: Failed to find subsystem in subject: "[PATCH] ci/setup: Print the environment variables"
```

marcel-apf:
> @marcel-apf thanks, CI is not happy
> 
> ```
> ERROR: Commit 6d9c8b5dde00cb19402e9d2428430f54dcdb4f23: Failed to find subsystem in subject: "[PATCH] ci/setup: Print the environment variables"
> ```
Thanks, repushed

devimc (approved):
thanks @marcel-apf

PR #3767, opened 2021-07-27
metrics: Get kata-env when using kata with containerd

### LINKED ISSUE #3766: fix result json with the proper kata-runtime kata-env when running with containerd

GabyCT:
This PR fixes the issue of extracting the proper kata-runtime kata-env
when we are running with the runtime of io.containerd.run.kata.v2.

Fixes #3766

### DIFF
--- a/metrics/lib/json.bash
+++ b/metrics/lib/json.bash
@@ -82,7 +82,7 @@ EOF
 	# Now add a runtime specific environment section if we can
 	local iskata=$(is_a_kata_runtime "$RUNTIME")
 	if [ "$iskata" == "1" ]; then
-		local rpath="$(command -v $RUNTIME)"
+		local rpath="$(command -v kata-runtime)"
 		local json="$(cat << EOF
 	"kata-env" :
 	$($rpath kata-env --json)


metrics/lib/json.bash
```diff
@@ -82,7 +82,7 @@ EOF
 	# Now add a runtime specific environment section if we can
 	local iskata=$(is_a_kata_runtime "$RUNTIME")
 	if [ "$iskata" == "1" ]; then
-		local rpath="$(command -v $RUNTIME)"
+		local rpath="$(command -v kata-runtime)"
```
fidencio, line 85: Out of curiosity, what are we currently passing as "RUNTIME"?

GabyCT: @fidencio metrics scripts as they are only using containerd we have as a runtime io.containerd.run.kata.v2

fidencio: Just to completely clear, then "$RUNTIME" in this case is "containerd-shim-kata-v2", and we need to use `kata-runtime` instead.  Do I understand this correctly, @GabyCT?

GabyCT: @fidencio exactly we in order to get the kata-env (which give us hypervisor information, image information, etc) we need to use kata-runtime as we can't extract it from containerd-shim-kata-v2

fidencio (approved):
lgtm, thanks @GabyCT!

devimc (approved):

Issue #3772, opened 2021-07-29
 static-checks are not being triggered when a PR is updated.


fidencio:
For some reason it's been a week or two since static-checks are not being triggered when a PR is updated. We've came up with a few workarounds for this, such as closing & reopening a PR, or adding / removing labels.

This seems to be a regression introduced by 649a5a38c90a7eee351768d5c99c57a1dad636f4, where we explicitly enumerated the actions that would trigger the static-checks.

/cc @jodh-intel

PR #3773, opened 2021-07-29
ci: Run static checks when PRs are updated

fidencio:
Looking at the changes that could cause the static-checks not to run
when a PR is updated I think 649a5a38c90a7eee351768d5c99c57a1dad636f4
could be the one that introduced such a regression.

Let's (try to) fix this by enforcing the workflow to run also when the
PR has been "edited" and "synchronized".

Fixes: #3772

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,9 @@ on:
   pull_request:
     types:
       - opened
+      - edited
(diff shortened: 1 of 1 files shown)


amshinde (approved):

devimc (approved):
thanks @fidencio

PR #3775, opened 2021-07-29
Fix kata-webhook

dmaizel:
* Fixed api version of the `MutatingWebhookConfiguration` resource
* Added both `sideEffects` and  `admissionReviewVersions` fields because their default value is now removed and these field made required. (according to https://kubernetes.io/docs/reference/using-api/deprecation-guide/)
* Because the default value of `failurePolicy` was changed to `Fail` instead of `Ignore` I changed it back.
* Updated everything to use the new version of [kubewebhook](https://github.com/slok/kubewebhook)


kata-webhook/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM golang:latest AS builder
 
 WORKDIR /go/src/kata-pod-annotate
 
-COPY ../vendor ./
+# COPY ../vendor ./
```
snir911: Note that this reverts of the previous commit
I'd squash the 2 fixes and the "fix indentation" into single "webhook: minor fixes" commit of similar.
Also IIRC subject must include "area:" (e.g. **webhook:** minor fixes) and some text in the commit body to pass the checks (the text you added is in the PR not in the commit)
Thanks @dmaizel !

fidencio:
@dmaizel, seems like you did a rebase atop of `master`, while a rebase should've been done atop of `main`.

dmaizel:
> @dmaizel, seems like you did a rebase atop of `master`, while a rebase should've been done atop of `main`.

My bad, fixed...

snir911 (approved):
LGTM, Thanks @damaizel!

**NOTE that it requires to push a new quay.io/kata-containers/kata-webhook-example image!**

fidencio:
@snir911, @dmaizel, before I approve, merge, and push a new image, wouldn't be nice to give it a try and ensure it doesn't break our tests?

The way for doing this would be, for instance:
1. Create an image on your own repo, with this patch applied.
2. Update the yaml file under `kata-webhook/deploy/` to point to your image, rather than the canonical one.
3. Call `/test` with that, so we ensure our tests will pass using the new image.

Once we have all the tests green we can go ahead and merge this one without any doubt that it'll work as we expect.

Does this sound reasonable?

dmaizel:
> @snir911, @dmaizel, before I approve, merge, and push a new image, wouldn't be nice to give it a try and ensure it doesn't break our tests?
> 
> The way for doing this would be, for instance:
> 
> 1. Create an image on your own repo, with this patch applied.
> 2. Update the yaml file under `kata-webhook/deploy/` to point to your image, rather than the canonical one.
> 3. Call `/test` with that, so we ensure our tests will pass using the new image.
> 
> Once we have all the tests green we can go ahead and merge this one without any doubt that it'll work as we expect.
> 
> Does this sound reasonable?

Absolutely. I already did the 2 steps before opening the PR, but I wasn't aware of the third one.

fidencio:
Here we can find a PR opened that's taking advantage of this series: https://github.com/kata-containers/tests/pull/3809
As expected, all the tests there are passing. \o/

PR #3778, opened 2021-07-30
tests/integration: Add basic debug information to k8s-memory teardown

dgibson:
Since 591c556 "tests: Add some kubectl describe information to teardown of
many k8s tests", most of the Kubernetes bats tests had a kubectl describe
in the teardown path which helps to debug at least some CI failures when
pods fail to reach Ready state.

Unfortunately the k8s-memory.bats test was missed, probably because it
didn't have any teardown code to begin with.  Correct that oversight.


marcel-apf (approved):

fidencio (approved):
lgtm, thanks @dgibson!

PR #3791, opened 2021-08-03
integration-tests.sh: Add TestContainerSwap to test guest swap

teawater:
Add TestContainerSwap to test guest swap in integration-tests.sh.

Fixes: #3776

Signed-off-by: Hui Zhu <teawater@antfin.com>

### DIFF
--- a/integration/containerd/cri/integration-tests.sh
+++ b/integration/containerd/cri/integration-tests.sh
@@ -109,6 +109,9 @@ ci_cleanup() {
 
 create_containerd_config() {
 	local runtime="$1"
+	# kata_annotations is set to 1 if caller want containerd setup with
+	# kata annotations support.
+	local kata_annotations=${2-0}
 	[ -n "${runtime}" ] || die "need runtime to create config"
 
 	local runtime_type="${containerd_runtime_type}"
@@ -122,13 +125,17 @@ create_containerd_config() {
 	# Remove dots.  Dots are used by toml syntax as atribute separator
 	runtime="${runtime//./-}"
 
-	cat << EOT | sudo tee "${CONTAINERD_CONFIG_FILE}"
+cat << EOT | sudo tee "${CONTAINERD_CONFIG_FILE}"
 [plugins]
   [plugins.cri]
     [plugins.cri.containerd]
-	default_runtime_name = "$runtime"
+        default_runtime_name = "$runtime"
       [plugins.cri.containerd.runtimes.${runtime}]
         runtime_type = "${runtime_type}"
+        $( [ $kata_annotations -eq 1 ] && \
+        echo 'pod_annotations = ["io.katacontainers.*"]' && \
+        echo '        container_annotations = ["io.katacontainers.*"]'
+        )
         [plugins.cri.containerd.runtimes.${runtime}.options]
           Runtime = "${containerd_runtime}"
 [plugins.linux]
@@ -178,6 +185,10 @@ check_daemon_setup() {
 }
 
 testContainerStart() {
+	# no_container_yaml set to 1 will not create container_yaml
+	# because caller has created its own container_yaml.
+	no_container_yaml=${1-0}
+
 	local pod_yaml=${REPORT_DIR}/pod.yaml
 	local container_yaml=${REPORT_DIR}/container.yaml
 	local image="busybox:latest"
@@ -187,7 +198,9 @@ metadata:
   name: busybox-sandbox1
 EOF
 
-	cat << EOF > "${container_yaml}"
+	#TestContainerSwap has created its own container_yaml.
+	if [ $no_container_yaml -ne 1 ]; then
+		cat << EOF > "${container_yaml}"
 metadata:
   name: busybox-killed-vmm
 image:
@@ -195,6 +208,7 @@ image:
 command:
 - top
 EOF
+	fi
 
 	sudo cp "$default_containerd_config" "$default_containerd_config_backup"
 	sudo cp $CONTAINERD_CONFIG_FILE "$default_containerd_config"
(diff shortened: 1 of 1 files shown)


integration/containerd/cri/integration-tests.sh
```diff
@@ -129,18 +130,35 @@ create_containerd_config() {
 	# Remove dots.  Dots are used by toml syntax as atribute separator
 	runtime="${runtime//./-}"
 
-	cat << EOT | sudo tee "${CONTAINERD_CONFIG_FILE}"
+	if [ $kata_annotations -eq 1 ]; then
```
Jakob-Naucke, line 133: Does anything bad happen if we just always pass all annotations?

Jakob-Naucke (commented):
@teawater don't you want to actually run the tests? Also, when I run them, I get
```
cat: can't open '/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes': No such file or directory
```

across x86_64 and s390x

integration/containerd/cri/integration-tests.sh
teawater: > Does anything bad happen if we just always pass all annotations?

I think pass all annotations is OK for kata.

GabyCT (approved):

integration/containerd/cri/integration-tests.sh
```diff
@@ -207,6 +225,8 @@ check_daemon_setup() {
 }
 
 testContainerStart() {
+	no_container_yaml=$1
```
wainersm, line 228: Same rationale here, make "$1" optional.

integration/containerd/cri/integration-tests.sh
```diff
@@ -116,6 +116,7 @@ ci_cleanup() {
 
 create_containerd_config() {
 	local runtime="$1"
+	local kata_annotations="$2"
```
wainersm, line 119: A mere suggestion: make "$2" optional and defaulting to 0 so that you don't need to change the calls to `create_containerd_config`.

Documenting the parameters would be nice too :)

integration/containerd/cri/integration-tests.sh
```diff
@@ -129,18 +130,35 @@ create_containerd_config() {
 	# Remove dots.  Dots are used by toml syntax as atribute separator
 	runtime="${runtime//./-}"
 
-	cat << EOT | sudo tee "${CONTAINERD_CONFIG_FILE}"
+	if [ $kata_annotations -eq 1 ]; then
+		cat << EOT | sudo tee "${CONTAINERD_CONFIG_FILE}"
```
wainersm, line 134: The header of the file (until the annotation sections) and the footer (from the annotation on) are repeated in the `else` statement. Maybe you could do:
 1. Print the header
 2. Print the annotation sections if `$kata_annotations` equals to 1
 3. Print the footer
 
An alternative is to embedded the logic to print the annotation sections inside the cat's heredoc. For example:
```
cat << EOT | sudo tee "${CONTAINERD_CONFIG_FILE}"
[plugins]
  [plugins.cri]
    [plugins.cri.containerd]
        default_runtime_name = "$runtime"
      [plugins.cri.containerd.runtimes.${runtime}]
        runtime_type = "${runtime_type}"
        $( [ $kata_annotations -eq 1 ] && \
        echo 'pod_annotations = ["io.katacontainers.*"]' && \
        echo '        container_annotations = ["io.katacontainers.*"]'
        )
        [plugins.cri.containerd.runtimes.${runtime}.options]
          Runtime = "${containerd_runtime}"
[plugins.linux]
       shim = "${containerd_shim_path}"
EOT
```
The above command will produce a blank line if `$kata_annotations` equals to 0 though. Don't know if it will cause a problem. Also I confess that solution seems a bit ugly to me, maybe there should exist a best "bashism" way.

integration/containerd/cri/integration-tests.sh
```diff
@@ -319,6 +342,135 @@ TestContainerMemoryUpdate() {
...
+	if [ $swappiness -ne 60 ]; then
+		testContainerStop
+		die "The VM swappiness $swappiness without swap device is not right"
+	fi
+	swap_in_bytes=$(crictl exec $cid cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes)
+	if [ $swap_in_bytes -lt 1125899906842624 ]; then
+		testContainerStop
+		die "The VM swap_in_bytes $swap_in_bytes without swap device is not right"
+	fi
+	swap_size=$(($(crictl exec $cid cat /proc/meminfo | grep "SwapTotal:" | awk '{print $2}')*1024))
```
wainersm, line 370: On next statement you compare with zero, so I don't think you need to multiply by 1024.

integration/containerd/cri/integration-tests.sh
```diff
@@ -319,6 +342,135 @@ TestContainerMemoryUpdate() {
...
+		die "The VM swap size $swap_size without swap device is not right"
+	fi
+	testContainerStop
+
+	# Test with swap device
+	cat << EOF > "${container_yaml}"
+metadata:
+  name: busybox-killed-vmm
+annotations:
+  io.katacontainers.container.resource.swappiness: "100"
```
wainersm, line 382: Expected swappiness, swap_in_bytes and memory_limit_in_bytes values could be set in variables.

wainersm: Also, is there a rationale behind those values? or you arbitrarily chose them?

integration/containerd/cri/integration-tests.sh
```diff
@@ -319,6 +342,135 @@ TestContainerMemoryUpdate() {
...
+		testContainerStop
+		die "The VM swappiness $swappiness with swap device is not right"
+	fi
+	swap_in_bytes=$(crictl exec $cid cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes)
+	if [ $swap_in_bytes -ne 1610612736 ]; then
+		testContainerStop
+		die "The VM swap_in_bytes $swap_in_bytes with swap device is not right"
+	fi
+	swap_size=$(($(crictl exec $cid cat /proc/meminfo | grep "SwapTotal:" | awk '{print $2}')*1024))
+	if [ $swap_size -ne 536870912 ]; then
```
wainersm, line 404: 536870912 because `swap_size = swap_in_bytes - memory_limit_in_bytes`?

integration/containerd/cri/integration-tests.sh
```diff
@@ -319,6 +342,135 @@ TestContainerMemoryUpdate() {
...
+- top
+EOF
+	testContainerStart 1
+	swappiness=$(crictl exec $cid cat /sys/fs/cgroup/memory/memory.swappiness)
+	if [ $swappiness -ne 100 ]; then
+		testContainerStop
+		die "The VM swappiness $swappiness without swap_in_bytes is not right"
+	fi
+	swap_in_bytes=$(crictl exec $cid cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes)
+	if [ $swap_in_bytes -lt 1125899906842624 ]; then
```
wainersm, line 431: I didn't get that number. Shouldn't swap_in_bytes be `2 * memory_limit_in_bytes`?

integration/containerd/cri/integration-tests.sh
```diff
@@ -319,6 +342,135 @@ TestContainerMemoryUpdate() {
...
+	local container_yaml=${REPORT_DIR}/container.yaml
+
+	info "Test container with guest swap"
+
+	create_containerd_config "${containerd_runtime_test}" 1
+	sudo sed -i -e 's/^#enable_guest_swap.*$/enable_guest_swap = true/g' "${kata_config}"
+
+	# Test without swap device
+	testContainerStart 0
+	swappiness=$(crictl exec $cid cat /sys/fs/cgroup/memory/memory.swappiness)
```
jodh-intel, line 359: Isn't this assuming a cgroups v1 system? What about Fedora (cgroups v2)? Or `/proc/sys/vm/swappiness`.

integration/containerd/cri/integration-tests.sh
teawater: yes

integration/containerd/cri/integration-tests.sh
teawater: swap_in_bytes is not set, it should be a value that bigger than 1125899906842624.

integration/containerd/cri/integration-tests.sh
teawater: Updated.

integration/containerd/cri/integration-tests.sh
teawater: Updated.

integration/containerd/cri/integration-tests.sh
teawater: Updated.

integration/containerd/cri/integration-tests.sh
teawater: Updated.

integration/containerd/cri/integration-tests.sh
teawater: Moved code for get swap_size, swappiness and swap_in_bytes to a function getContainerSwapInfo.

integration/containerd/cri/integration-tests.sh
teawater: I would like to set swappiness.
Because default value is 60, so this part set it to 100.

liubin (approved):

integration/containerd/cri/integration-tests.sh
Bevisy: Sorry to bother you, but we can use "echo -n" to avoid to produce a blank line. fix in #3943  :smile:

Jakob-Naucke:
@teawater sorry for bump after merge, but
1. I don't think we are running the test, that requires something like
```diff
diff --git a/integration/containerd/cri/integration-tests.sh b/integration/containerd/cri/integration-tests.sh
index 4c065ff3..175422ee 100755
--- a/integration/containerd/cri/integration-tests.sh
+++ b/integration/containerd/cri/integration-tests.sh
@@ -517,6 +517,7 @@ main() {
                        make -e test-integration
        done
 
+       TestContainerSwap
        TestContainerMemoryUpdate 1
        TestContainerMemoryUpdate 0

```
2. Does the test work on your machine? I get
```
[integration-tests.sh:320] INFO: Test container with guest swap
[plugins]
  [plugins.cri]
    [plugins.cri.containerd]
        default_runtime_name = "io-containerd-kata-v2"
      [plugins.cri.containerd.runtimes.io-containerd-kata-v2]
        runtime_type = "io.containerd.kata.v2"
        pod_annotations = ["io.katacontainers.*"]
        container_annotations = ["io.katacontainers.*"]
        [plugins.cri.containerd.runtimes.io-containerd-kata-v2.options]
          Runtime = "io.containerd.kata.v2"
[plugins.linux]
       shim = "/usr/local/bin/containerd-shim"
[integration-tests.sh:203] INFO: restart containerd service
FATA[0000] pulling image: rpc error: code = Unknown desc = server is not initialized yet
```

on x86_64 Ubuntu 20.04.

PR #3816, opened 2021-08-11
test master -> 1.x-eol rename

fidencio:
I'm using a PR from @GabyCT as a way to ensure the tests are working.  The reason for using @GabyCT's PR is basically because all the CIs passed on a PR with this patch. :-)


marcel-apf (approved):

fidencio:
Everything is work as expected after the branch renaming!

Issue #3817, opened 2021-08-11
Deprecation of the dep tool


dmaizel:
# Unable to pass static_check_vendor after updating a package
After updating a package that I'm using in kata-webhook/main.go, I'm not able to pass the `static_check_vendor` static check because the Gopkg.lock file is out of sync. After googling this issue I found that the dep repo is deprecated, as you can see here: https://github.com/golang/dep

PR #3825, opened 2021-08-16
Switch to using go mod instead of `dep` tool

fidencio:
This work, as it's now, is a tentative to switch to using go mod.  The changes are quite a lot, but that's due to our vendor code being updated and I'm raising this PR basically to figure out what was broken during this transition.

We may need deeper changes, as the one proposed by Daniel Maizel to fix the kata-webhook, to fix other components; we may need deeper changes to get rid of other `go.mod` files that are scattered across our repo.  But, right now, let's try to see how big the breakage will be.

Fixes: #3817


.ci/static-checks.sh
```diff
@@ -918,55 +918,22 @@ static_check_files()
...
-		# PR changed files so check if it changed any vendored files
-		vendor_files=$(echo "$files" | grep -E "\<vendor/" || true)
-
-		if [ -n "$vendor_files" ]
-		then
-			result=$(echo "$files" | egrep "\<${vendor_ctl_file}\>" || true)
-			[ -n "$result" ] || die "PR changes vendor files, but does not update ${vendor_ctl_file}"
-		fi
+	STATUS=$(git status --porcelain)
+	if [[ -z $STATUS ]]; then
```
jodh-intel, line 928: Nit: Single square brackets work here.

jodh-intel (approved):
Thanks @fidencio.

lgtm

Issue #3828, opened 2021-08-17
stable-2.1: containerd integration test fails when using the release/1.3 branch


jongwu:
in stable-2.1 branch, containerd will switch to release/1.3 before integration test which does not accord with main branch.
I think it lead to the failure of containerd integration test.
see http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-ARM-PR/2105/console

PR #3829, opened 2021-08-17
stable-2.1: using master branch in containerd test

jongwu:
using master branch of containerd to keep consistent with kata main branch
in containerd integration test.

Fixes: #3828
Depends-on: github.com/kata-containers/kata-containers#2441

### DIFF
--- a/integration/containerd/cri/integration-tests.sh
+++ b/integration/containerd/cri/integration-tests.sh
@@ -336,7 +336,9 @@ main() {
 	pushd "${GOPATH}/src/${cri_repo}"
 
 	git reset HEAD
-	git checkout release/1.3
+	git checkout master
+	# switch to the default pause image set by containerd:1.5.x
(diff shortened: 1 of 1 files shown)


integration/containerd/cri/integration-tests.sh
```diff
@@ -336,7 +336,9 @@ main() {
 	pushd "${GOPATH}/src/${cri_repo}"
 
 	git reset HEAD
-	git checkout release/1.3
+	git checkout master
```
fidencio, line 339: I don't think we should do that, @jongwu.

There are a whole bunch of things that would need to be changed that would be way too intrusive for a backport.

jongwu: hello @fidencio -, I'm trying to fix the arm ci issue on stable-2.1, so I come up with that changing the branch for containerd. but it doesn't work. I will close this PR, but I want some idea from you about how to save the arm ci. see http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-ARM-PR/2110/consoleFull

PR #3842, opened 2021-08-18
metrics: Fixes exit non-zero exit returned by clean_env_ctr

### LINKED ISSUE #3836: metrics: exit non-zero exit returned by clean_env_ctr
tests/lib/common.bash:clean_env_ctr() function is used to stop and remove all current containerd instances, but ctr returns a non zero exit for each killed container.

dborquez:
Issues a SIGKILL to terminate the container.

Fixes #3836

Signed-off-by: David Esparza <david.esparza.borquez@intel.com>


lib/common.bash
```diff
@@ -167,12 +167,11 @@ clean_env()
...
 {
-	check_containers=$(sudo ctr c list -q | wc -l)
-	if ((${check_containers})); then
-		sudo ctr tasks kill $(sudo ctr task list -q)
-		sudo ctr tasks rm -f $(sudo ctr task list -q)
-		sudo ctr c rm $(sudo ctr c list -q)
-	fi
+	local containers=( $(sudo ctr c list -q) )
+	for i in "${containers[@]}"; do
+		sudo ctr task kill -s SIGKILL "$i"
```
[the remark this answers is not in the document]
dborquez: @devimc got catch thank you, fixed.

lib/common.bash
```diff
@@ -167,12 +167,11 @@ clean_env()
...
 {
-	check_containers=$(sudo ctr c list -q | wc -l)
-	if ((${check_containers})); then
-		sudo ctr tasks kill $(sudo ctr task list -q)
-		sudo ctr tasks rm -f $(sudo ctr task list -q)
-		sudo ctr c rm $(sudo ctr c list -q)
-	fi
+	local containers=( $(sudo ctr c list -q) )
+	for i in "${containers[@]}"; do
+		sudo ctr task kill -s SIGKILL $(sudo ctr task ls | grep "$i" | awk '{print $1}')
```
devimc, line 172: I wonder if `sudo ctr tasks rm ....` is still needed, why did you remove it?

dborquez: The command` ctr tasks rm` is not required to close and delete containers properly. Additionally that line is causing the non-zero exit.

devimc: ok, got it, thanks

devimc (approved):

GabyCT (approved):

egernst (commented):
Kind of a nit, but you can do the grep as part of the awl, so you don’t mix. 

Ie something like  awk ‘/$i/{print $1}’
(Warning this was written on phone so YMMV :))

egernst (changes requested):
you shouldn’t need to do a loop; you can just pass all the tasks to the task kill command. Same for container rm. 

Doing this will remove my awk/grep nit :)

dborquez (commented):
@egernst  thank you for your feedback, I have tested using the abbreviated method as you suggested, but containers can't close successfully. this is the output:

```
1)
sudo ctr tasks kill -s SIGKILL $(sudo ctr task list -q)
returns 0

2) 
sudo ctr c rm $(sudo ctr c list -q) 
returns:

ERRO[0000] failed to delete container "kata-sGdmoq"      error="cannot delete a non stopped container: {running 0 0001-01-01 00:00:00 +0000 UTC}"
ERRO[0000] failed to delete container "kata-snM0r0"      error="cannot delete a non stopped container: {running 0 0001-01-01 00:00:00 +0000 UTC}"
ERRO[0000] failed to delete container "kata-tCDw1x"      error="cannot delete a non stopped container: {running 0 0001-01-01 00:00:00 +0000 UTC}"
ERRO[0000] failed to delete container "kata-tRTcd2"      error="cannot delete a non stopped container: {running 0 0001-01-01 00:00:00 +0000 UTC}"
```

egernst:
Odd, I didn't expect that. Okay. WDYT about filtering in awk instead of grep :)

Thanks for the PR, and for giving it a try.

dborquez:
@egernst thank you, changes applied

egernst (approved):

GabyCT:
@dborquez it seems that there is an error related with this PR in one of the CI tests, please review the failure

GabyCT:
@dborquez please put the label if this PR does not need a backport

GabyCT:
@dborquez it seems that the vcpu integration test is failing with
```
13:48:19 ctr: cannot delete a non stopped container: {running 0 0001-01-01 00:00:00 +0000 UTC}
13:48:19 [default_vcpus_test.sh:173] ERROR: sudo ctr run --runtime=io.containerd.run.kata.v2 -d quay.io/prometheus/busybox:latest test sh -c nproc | grep 4
13:48:19 make: *** [Makefile:155: vcpus] Error 1
```
probably you need to stop the container and then delete it

Issue #3847, opened 2021-08-19
docker can't be used by jenkins user


jongwu:
in the current implementation, jenkins user is added to docker group and set the /run/docker.sock to rw for docker group. but it doesn't work until relogin the session. so the error will occur when jenkins user run docker:
docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/create": dial unix /var/run/docker.sock: connect: permission denied

PR #3848, opened 2021-08-19
docker: provides rw permission of docker.sock for all uses

jongwu:
Currently, the permission of docker.sock is set to be
srw-rw---- 1 root docker 0 xx y zz:tt /var/run/docker.sock
and add the jenkins user to docker group. But that configuration will
not take effect unitl relogin the session, which means the docker
can't be used by jenkins user in the whole CI journey.
It seems there is no way to log out and log back in the auto ci. So, I
have to set the permission of docker.sock to rw for all user, even
though it will open a secure hole.

Fixes: #3847

### DIFF
--- a/cmd/container-manager/manage_ctr_mgr.sh
+++ b/cmd/container-manager/manage_ctr_mgr.sh
@@ -181,7 +181,7 @@ install_docker(){
 	fi
 	restart_docker_service
 	sudo gpasswd -a ${USER} docker
-	sudo chmod g+rw /var/run/docker.sock
+	sudo chmod 666 /var/run/docker.sock
 }
 
 install_docker_s390x(){


jongwu:
@fidencio @jodh-intel , help look at this

egernst:
Did you try something like: 

exec sg <new group name> newgrp `id -gn`

jongwu:
Hello @egernst -, I have tried newgrp but it does work. Both sg and newgrp need interactivity thus are not proper to be used in the CI, I think.

jongwu:
Hello @fidencio -, any comments about this? I think it is a long standing bug.

GabyCT (approved):

GabyCT:
@jongwu thanks and only put if this needs or not a backport label

jongwu:
thanks @GabyCT

GabyCT:
> hello @chavafg -, do you mean run these changes on the ci machine instead of in the ci scripts? but the ci scripts will reinstall the docker then the changes will be invalid.

^^^ @chavafg

fidencio:
Hey!  First of all, sorry for coming late to the party, I was off for one week.

So, I tend to prefer the approach suggested by @chavafg, even if it means we manually create the docker group & add the jenkins user to it.  I know, this is something for CI only, but having docker socket as `666` feels like papering over the issue.

Out of curiosity, how are those machines set-up, @jongwu?  Is there a script that's used to set them up?  If so, is this part of the kata-containers repo?

GabyCT:
@jongwu could you please a take a look

jongwu:
OK @fidencio -, I will handle this issue in the ci machine. thus I will close this PR.

Issue #3858, opened 2021-08-20
openshift-ci: fix failling test "Managed cluster should set requests but not limits"


wainersm:
The following Kubernetes e2e conformance test has failed since the job on OpenShift CI was upgraded to use ocp 4.8:

     [sig-arch] Managed cluster should set requests but not limits [Suite:openshift/conformance/parallel]

PR #3861, opened 2021-08-20
kata-webhook: request cpu/memory to the pod and a few improvements

wainersm:
Primarily this should fix a test case which is failing on OpenShift CI (see #3858). While I am at it, I improved the `create-certs.sh` script a little bit.

### DIFF
--- a/kata-webhook/deploy/webhook.yaml
+++ b/kata-webhook/deploy/webhook.yaml
@@ -33,6 +33,10 @@ spec:
             - name: webhook-certs
               mountPath: /etc/webhook/certs
               readOnly: true
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "250Mi"
       volumes:
         - name: webhook-certs
           secret:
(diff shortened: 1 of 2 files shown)


kata-webhook/deploy/webhook.yaml
```diff
@@ -33,6 +33,10 @@ spec:
             - name: webhook-certs
               mountPath: /etc/webhook/certs
               readOnly: true
+          resources:
```
egernst, line 36: seems fine -- any particular reason for 1 CPU instead of 100m, or 100 CPU? 

Same for memory

snir911 (approved):
LGTM, thanks, also fine with lowering the requests

GabyCT (approved):

GabyCT:
@wainersm please put the label if this PR needs or not a backport thanks

kata-webhook/deploy/webhook.yaml
wainersm: @egernst for no particular reason. But maybe it should be better set 100m?

wainersm: Just updated this PR; setting "cpu==100m" and "memory==205Mi", which should be enough.

GabyCT:
@wainersm does this needs backport?

PR #3881, opened 2021-08-24
backport ci: Introduce the ci_job_flags.sh file

GabyCT:
In .ci/jenkins_job_build.sh it is exported a bunch of variables
as per job-type, they are used to control the setup of the system
and the execution of tests.

This creates the ci_job_flags.sh as the new home for those variables
so that external scripts (not just jenkins_job_build.sh) can source
the variables before running the same setup and run tests scripts.


devimc (approved):
thanks @GabyCT

chavafg (approved):

jcvenegas (approved):

PR #3886, opened 2021-08-26
k8s: enable memory hotplug test for arm64

jongwu:
enable memory hotplug test in k8s integration test

Fixes: #3885


fidencio (approved):
lgtm, thanks @jongwu!

Issue #3887, opened 2021-08-26
CCv0 | Update the CI for to work with the CCv0 requirements


stevenhorsman:
We now have separate branches and build/test jobs for CCv0 work, so we should start using that to make the CCv0 builds have awareness of this context and start adding customisations that we are using.

## Task list
- [x] Add an environment variable accessible by the different build jobs to reflect if we are running `CCv0` builds, or not
- [x] Add our build customisations into the scripts so we could potentially build a `CCv0` release from the branch, if the environment variable states we are running `CCv0`:
  - [x] Add skopeo, gnupg gpgme-devel as packages into the kata image. This might require changing the default image OS to fedora
  - [x] Add umoci into the rootfs before converting to an img as it is available as a package in the distro versions that Kata currently uses
  - [x] Updating the version of containerd that is pulled in to point to a containerd WIP PR as suggested by Mike Brown

stevenhorsman:
The associate PR #3888 has merged, closing this item

Issue #2577, opened 2020-05-27
clh: ci: enable metrics CI


jcvenegas:
# Description of problem
 We should enable  metrics CI
Items to consider:
- [ ] enable in test PR repository
- [ ] enable in runtime PR repository
- [ ] Save in artifacts summary of data
- [ ] Make sure scripts data is taken from cloud-hypervisor and not qemu.

grahamwhaley:
I'll sketch out a few hints of what/where might need modifying....

The metrics CI is invoked from the generic jenkins script [here](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L265-L267), which does:

```bash
if [ -n "${METRICS_CI}" ]; then
	echo "Running the metrics tests:"
	"${tests_repo_dir}/.ci/run_metrics_PR_ci.sh"
```

That obviously invokes [this](https://github.com/kata-containers/tests/blob/master/.ci/run_metrics_PR_ci.sh). In there you will see it calls three funcs - init/run/check.
- if you want to swap from testing qemu to testing clh (and no longer test qemu), I think you will be able to `export RUNTIME=` to set which runtime gets tested.
- if you want to test both qemu and clh, then I think you can put a simple loop around the init/run/check calls, with a `for runtime in "kata-qemu kata-clh"` style, and invoke the testing multiple times. You probably will need a little hackery to move the results files around to ensure the second run does not overwrite the first if you want to store all the artifacts back up on Jenkins.

None of it should be hard - the `RUNTIME` is not very hard wired into the metrics. But, you should first check that the metrics do run with `RUNTIME=kata-clh`. I don't think I've tried that (or possibly not for some time).

Oh, and I guess, if you are going to test multiple runtimes, it might worth considering setting them up as separate Jenkins jobs, as that will:
- avoid the first run interfering or perturbing the second run if they are on the same machine run one after the other
- allow them to run in parallel, speeding up the overall CI time (but, of course, at the cost of more parallel resource).

jcvenegas:
Updates on this:https://github.com/kata-containers/tests/pull/2598 got merged :)

PR #3888, opened 2021-08-26
CCv0 | ci: Add CCv0 features into CI script

stevenhorsman:
Make the build understand if they are running against a target branch
of CCv0 or main and update branches dynamically based on this

Add code to get containerd from our CCv0 PR rather than a released build

Add new CCv0 condition to the code that created the kata image.
CCv0 requires some extra packages for running and debug. It also needs
umoci, which is not yet availabe as a distro package, so wget it direct

Fixes: #3887
Depends-on: github.com/kata-containers/kata-containers#2577

### DIFF
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -23,12 +23,13 @@ source "${script_dir}/lib.sh"
 #Use cri contaienrd tarball format.
 #https://github.com/containerd/cri/blob/master/docs/installation.md#release-tarball
 CONTAINERD_OS=$(go env GOOS)
-CONTAIENRD_ARCH=$(go env GOARCH)
+CONTAINERD_ARCH=$(go env GOARCH)
 
 cri_containerd_tarball_version=$(get_version "externals.cri-containerd.version")
 cri_containerd_repo=$(get_version "externals.cri-containerd.url")
 
 cri_containerd_version=${cri_containerd_tarball_version#v}
+cri_containerd_pr="5911" # TODO - put in version.yaml and pull dynamically $(get_version "externals.cri-containerd.pr_id")
 
 echo "Set up environment"
 if [ "$ID" == centos ] || [ "$ID" == rhel ] || [ "$ID" == sles ]; then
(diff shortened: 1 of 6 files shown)


.ci/install_cri_containerd.sh
```diff
@@ -23,12 +23,13 @@ source "${script_dir}/lib.sh"
...
 #https://github.com/containerd/cri/blob/master/docs/installation.md#release-tarball
 CONTAINERD_OS=$(go env GOOS)
-CONTAIENRD_ARCH=$(go env GOARCH)
+CONTAINERD_ARCH=$(go env GOARCH)
 
 cri_containerd_tarball_version=$(get_version "externals.cri-containerd.version")
 cri_containerd_repo=$(get_version "externals.cri-containerd.url")
 
 cri_containerd_version=${cri_containerd_tarball_version#v}
+cri_containerd_pr="5911" # TODO - put in version.yaml and pull dynamically $(get_version "externals.cri-containerd.pr_id")
```
jodh-intel, line 32: This looks like a temporary tweak which can be dropped when it gets merged by the upstream. If so, it might be safer to add a `do-not-merge` label to this PR. That way, we can still test if the code works in the Kata CI, but we won't land it by mistake.

Alternatively, if you preceive it may take a while to get that PR landed, are you happy to update this script once the PR does land?

Either way, I think it would be useful to add a call like the following to make this stand out in the CI output:

```bash
warn "Using patched Confidential Computing containerd version: see https://github.com/containerd/containerd/pull/5911"
```

(Note that it's best to spell out what "CC" refers to as for some of the "old hands", that acronym can also stand for https://github.com/clearcontainers :smile:)

jodh-intel (commented):
Thanks @stevenhorsman. A few comments.

.ci/install_kata_image.sh
```diff
@@ -42,13 +43,28 @@ build_rust_image() {
...
+			if [ ${CCV0} == "yes"]; then
+				# CCv0 is using skopeo and gpg packages and we've added a few more for debugging. The default distro of ubuntu is really back level, so some are missing there, so use fedora
+				# We also need to split the rootfs create and image build so we can add umoci to the rootfs directly
+				distro=fedora
+				EXTRA_PKGS="skopeo gnupg gpgme-devel vim iputils net-tools iproute"
+				sudo -E USE_DOCKER="${use_docker:-}" DISTRO="${distro}" EXTRA_PKGS="${EXTRA_PKGS}" \
+					make -e "rootfs"
+				rootfs_target="$(shell pwd)/$(DISTRO)_rootfs"
+				# CCv0 is using umoci which isn't in a fedora package we can access yet, so grabbing from their webstire
+				sudo wget https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64 -O ${rootfs_target}/usr/local/bin/umoci
```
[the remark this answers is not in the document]
stevenhorsman: Thanks for the advice. I think I've updated this as required

.ci/install_cri_containerd.sh
stevenhorsman: Hey @jodh-intel, thanks very much for the review and suggestions. I've added the warn statement, but am a little unsure about the hard-coded version change. It's a bit of chicken and egg and I originally made the proper change in version.yaml, but this required an update in `kata-containers` repo that I couldn't test with this PR, so I was going to get this in, then update kata-containers, then come back to tests and fix, but that's not necessarily very efficient.
Alternatively I do have a PR open in that repo now (https://github.com/kata-containers/kata-containers/pull/2556), so I could put the version change in there, and make this change proper and depend on that?

jodh-intel (approved):
Thanks @stevenhorsman.

Either method is fine. We just need to ensure we don't forget to update this code and continue to rely on a patched version of containerd. You could prefix the warn message with "FIXME" possibly as that would help if we ever resurect the "FIXME checker" :smile:

.ci/install_cri_containerd.sh
stevenhorsman: As an update I decided to make the version change in https://github.com/kata-containers/kata-containers/pull/2556, so once that has been merged I'll update the code here to remove the TODO'd temporary code in this PR

stevenhorsman: As a further update, based on feedback from Samuel, I've separated the version change from the PR and created #2577 as a stand-alone PR with this change in now.

Issue #3901, opened 2021-08-31
Honor USE_DOCKER


Jakob-Naucke:
I revert the patch and it started to work; that made me think it was related with Podman. I wanted to double-check that, so I need a way of setting `USE_DOCKER=yes` and this really be honored even if in Fedora.

_Originally posted by @wainersm in https://github.com/kata-containers/tests/issues/3760#issuecomment-908708076_

PR #3902, opened 2021-08-31
ci: Honor USE_DOCKER

Jakob-Naucke:
so you can choose to use Docker rather than Podman on Fedora-likes.

Fixes: #3901


marcel-apf (approved):
Thanks for the fixes

snir911 (approved):

PR #3903, opened 2021-08-31
WIP: vfio-jenkins

marcel-apf:
Signed-off-by: Marcel Apfelbaum <marcel@redhat.com>


snir911 (approved):
LGTM!! if passes :) makes much more sense

marcel-apf:
> LGTM!! if passes :) makes much more sense

Passed once... let's see if we can repeat it

.ci/vfio_jenkins_job_build.sh
```diff
@@ -249,45 +249,20 @@ install_dependencies() {
...
 			;;
 		fedora|centos|rhel)
 			# cloud image dependencies
 			deps=(xorriso curl qemu-img openssh)
 
-			# QEMU dependencies
-			deps+=(libcap-devel libattr-devel libcap-ng-devel librbd-devel gcc glib2-devel pixman-devel psmisc ninja-build)
-
-			sudo dnf install -y ${deps[@]}
+			sudo dnf install -y ${deps[@]} qemu-kvm
```
devimc, line 259: should we change the path to `qemu-kvm` ? sorry I don't know how distro packages work

https://github.com/kata-containers/tests/blob/270df0df73438db91695365e9c26acc4fe5e9c0b/.ci/vfio_jenkins_job_build.sh#L233

marcel-apf: @devimc weirdly enough the qemu binary remains the same! (verified on both fedora and ubuntu)

devimc:
thanks @marcel-apf  - lgtm I'm just waiting for the VFIO CI

GabyCT:
@marcel-apf could you please add if this PR needs or not a backport thanks

marcel-apf:
> @marcel-apf could you please add if this PR needs or not a backport thanks

done

marcel-apf:
> thanks @marcel-apf - lgtm I'm just waiting for the VFIO CI

Appreciated!
It takes a loooong time for some reason. I hope is not "stuck".

devimc:
@marcel-apf VFIO CI is failing 

```
15:29:09 + /usr/bin/qemu-system-x86_64 -m 16384M -smp cpus=8 -cpu host,host-phys-bits -machine q35,accel=kvm,kernel_irqchip=split -device intel-iommu,intremap=on,caching-mode=on,device-iotlb=on -drive file=/home/jenkins/functional-vfio-test/image.img,if=virtio,aio=threads,format=raw -drive file=/home/jenkins/functional-vfio-test/config.iso,if=virtio,media=cdrom -daemonize -enable-kvm -device virtio-rng-pci -display none -vga none -netdev user,hostfwd=tcp:127.0.15.1:10022-:22,hostname=kata-metric4,id=net0 -device virtio-net-pci,netdev=net0,disable-legacy=on,disable-modern=off,iommu_platform=on,ats=on -netdev user,id=net1 -device virtio-net-pci,netdev=net1,disable-legacy=on,disable-modern=off,iommu_platform=on,ats=on
15:29:09 qemu-system-x86_64: Parameter 'type' expects a netdev backend type
15:29:09 ++ handle_error 233
```

.ci/vfio_jenkins_job_build.sh
fidencio: Hmm.  I'd double check with the downstream QEMU maintainers as the path on CentOS / RHEL / Fedora seems to be different between each other.  RHEL doesn't support QEMU being called directly, this its binary is under `/usr/libexec/`.

I wonder if we should be trying to run it in all systems ... or just lower the scope and support it only on, let's say, Ubuntu & Fedora.

snir911: I think in practice it used only in ubuntu, I'm not sure why fedora/centos/rhel is there

marcel-apf: @fidencio I tested the installation with a clean Fedora container and that was the path

marcel-apf:
> @marcel-apf VFIO CI is failing
> 
> ```
> 15:29:09 + /usr/bin/qemu-system-x86_64 -m 16384M -smp cpus=8 -cpu host,host-phys-bits -machine q35,accel=kvm,kernel_irqchip=split -device intel-iommu,intremap=on,caching-mode=on,device-iotlb=on -drive file=/home/jenkins/functional-vfio-test/image.img,if=virtio,aio=threads,format=raw -drive file=/home/jenkins/functional-vfio-test/config.iso,if=virtio,media=cdrom -daemonize -enable-kvm -device virtio-rng-pci -display none -vga none -netdev user,hostfwd=tcp:127.0.15.1:10022-:22,hostname=kata-metric4,id=net0 -device virtio-net-pci,netdev=net0,disable-legacy=on,disable-modern=off,iommu_platform=on,ats=on -netdev user,id=net1 -device virtio-net-pci,netdev=net1,disable-legacy=on,disable-modern=off,iommu_platform=on,ats=on
> 15:29:09 qemu-system-x86_64: Parameter 'type' expects a netdev backend type
> 15:29:09 ++ handle_error 233
> ```

That is simply strange, The error suggests the qemu is compiled without user level networking (slirp), however the qemu-kvm package includes slirp.
The only other explanation is that another version of qemu was installed on the "host" machine after the qemu-kvm installation.... I will have a look

.ci/vfio_jenkins_job_build.sh
fidencio: For Fedora, for sure.  For RHEL I don't think so.  That's the reason I'd double check with maintainers.
Even our spec file for kata-containers explicitly points at `/usr/libexec/` for the QEMU path on RHEL.

marcel-apf:
Wow:
```
23:29:04 + sudo apt-get install -y xorriso curl qemu-utils openssh-client qemu-kvm
23:29:04 Reading package lists...
23:29:04 Building dependency tree...
23:29:04 Reading state information...
23:29:05 xorriso is already the newest version (1.5.2-1).
23:29:05 curl is already the newest version (7.68.0-1ubuntu2.6).
23:29:05 openssh-client is already the newest version (1:8.2p1-4ubuntu0.3).
23:29:05 qemu-kvm is already the newest version (1:4.2-3ubuntu6.17).
23:29:05 qemu-utils is already the newest version (1:4.2-3ubuntu6.17).
```
That means the qemu-kvm installed the right binary on the host, come other CI had overridden with a kata specific version, and now the binary is different so it will not install it again... I'll force it

.ci/vfio_jenkins_job_build.sh
marcel-apf: yes, with RHEL won't work, on the other hand we don't use or test it....

marcel-apf:
@fidencio re-pushed; this time I used only the packages that install the binary and also dropped support for the rhel and centos which we don;t check anyway. We can easily add it back once we will add it to the tests.

.ci/vfio_jenkins_job_build.sh
devimc: > yes, with RHEL won't work, on the other hand we don't use or test it.... 

... yet - @wainersm any future plan to test VFIO on RHEL/centOS ?

marcel-apf:
vfio passed once, let's try it again with all the tests
/test

devimc (approved):
thanks @marcel-apf

.ci/vfio_jenkins_job_build.sh
dgibson: The `qemu-kvm` package is, I believe, just a wrapper which depends on the correct `qemu-system-*` package (at least in Fedora, and I think it's similar in Ubuntu).  It may also install some extra dependencies and possibly a wrapper `qemu-kvm` binary, but it doesn't fundamentally change the qemu binary.

That said some distros (e.g. RHEL) put the qemu binary in a different place.  For now, I think covering only Ubuntu and Fedora is plenty.

dgibson (approved):
I love it.

fidencio (approved):
lgtm, thanks @marcel-apf!

fidencio:
@marcel-apf, while the backport of this PR isn't exactly needed, wouldn't be better to have this backported to the `stable-2.2` branch?  The VFIO CI is one of the complicated ones, that every now and then gives us (well, Julio) some headache.  Making it simpler is a good improvement also for the stable branch, I think.

No, this is not mandatory, it's just a suggestion. :-)

fidencio:
As the issue is not related to this PR, I'm going ahead and merging it.

.ci/vfio_jenkins_job_build.sh
```diff
@@ -249,45 +249,22 @@ install_dependencies() {
...
+		fedora)
 			# cloud image dependencies
 			deps=(xorriso curl qemu-img openssh)
 
-			# QEMU dependencies
-			deps+=(libcap-devel libattr-devel libcap-ng-devel librbd-devel gcc glib2-devel pixman-devel psmisc ninja-build)
-
-			sudo dnf install -y ${deps[@]}
+			sudo dnf install -y ${deps[@]} qemu-system-x86-core
+			sudo dnf reinstall -y qemu-system-x86-core
```
jodh-intel, line 261: Late to the party, but we should rework this slightly in the next PR to remove the hard-coded architecture using the usual `arch=$(uname -m)` idiom. Or better, call https://github.com/kata-containers/tests/blob/main/.ci/kata-arch.sh as that makes life easier for some arches.

Issue #3889, opened 2021-08-30
ci: vfio: dependencies resolving is failing not on main


snir911:
# Description of problem

When you add a depends-on not on the main branch, vfio ci will fail to fetch the correct branch and apply the depended patch

# Expected result

all dependencies will be resolved correctly

# Actual result

depended are merged wrongly with main branch and messes the code

# Further information

For example, i received the following error when depends-on was applied on stable-2.1
`ci/jenkins_job_build.sh: line 165: .ci/ci_job_flags.sh: No such file or directory`

GabyCT:
@snir911 I believe that the failure on `vfio` CI was due to a lack of backporting as it was fixed

GabyCT:
This was fixed on https://github.com/kata-containers/tests/commit/e0e6dd94f39c8d60587530f2c96f2009725de6a2

PR #3890, opened 2021-08-30
vfio: checkout target branch

snir911:
It will fix deps resolving with vfio (on stable branches)

Fixes: #3889
Signed-off-by: Snir Sheriber <ssheribe@redhat.com>

PR #3913, opened 2021-08-31
ci: fix deps resolving

snir911:
IIUC this should fix the mess with the dependencies between branches
When PR for kata-containers is send and this PR **Depends-on** some PR in tests, this tests PR is not merged/applied until the resolve-kata-dependencies.sh is called, hence, sourcing lib.sh before resolve-kata-dependencies.sh will result in lib.sh from the plain tests target branch (without merging the tests PR we depends-on). if the tests PR changes touched lib.sh these changes are not used (although we "depends-on" them"), this may cause a mess

Thing is, i think i cannot really test it before it's merged @Jakob-Naucke @marcel-apf  WDYT?
(https://github.com/kata-containers/tests/pull/3890 or similar might still be needed to fix completely the issue on stable)
@fidencio 

Fixes: #3889

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -12,7 +12,6 @@ set -o errtrace
 
 cidir=$(dirname "$0")
 source "/etc/os-release" || source "/usr/lib/os-release"
-source "${cidir}/lib.sh"
 
 CI_JOB=${CI_JOB:-}
 ghprbPullId=${ghprbPullId:-}
@@ -165,6 +164,7 @@ fi
 # according to the job type.
 pushd "${GOPATH}/src/${tests_repo}"
 source ".ci/ci_job_flags.sh"
+source "${cidir}/lib.sh"
 popd
 
 "${ci_dir_name}/setup.sh"


Jakob-Naucke (approved):
> Thing is, i cannot really test it before it's merged

_adds "VFIO job dependency resolving" to wishlist_

I think we don't need `lib.sh` until this point, so it should be fine. OOI, what variable is this for? `kata_default_branch`?

GabyCT:
@snir911  could you please add if this PR needs or not a backport thanks

snir911:
@Jakob-Naucke it's the USE_PODMAN which is set wrongly , maybe others as well

wainersm (commented):
Good catch!

I'm afraid there are other code being called before resolve the deps which should not. For example, https://github.com/kata-containers/tests/blob/842e6fc14f703bc53629cea66df12fe4eb66da12/.ci/jenkins_job_build.sh#L118

It's fair if you don't want to fix that and other possible issues on this commit though.

One more thing: could you put a comment in the script explaining that lib.sh should be sourced after the resolve deps scripts?

snir911:
@wainersm yes, i guess it's the same if you depends on changes in install_go.sh, maybe we can make the resolving earlier, or at least having a comment explaining this edge case sensitivity, I'll do that tomorrow

GabyCT (approved):

Issue #3918, opened 2021-09-01
cri integration test  fall randomly on Arm


jongwu:
the cri integration test on arm falls randomly like:
`03:00:23 🇩 test-integration
03:00:23 === RUN   TestImageLoad
03:00:23     image_load_test.go:41: docker save image into tarball
03:00:23     image_load_test.go:53: make sure no such image in cri
03:00:25     image_load_test.go:60: load image in cri
03:00:25     image_load_test.go:67: make sure image is loaded
03:00:25     image_load_test.go:79: create a container with the loaded image
03:00:25 E0831 19:00:29.503033  187329 remote_runtime.go:130] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to setup network for sandbox "58e9b0c7b1643ebd82e5390c20b198e37b23c284573f1e1ae3cd86c2170a5548": failed to set bridge addr: "cni0" already has an IP address different from 10.88.0.1/16
03:00:29     image_load_test.go:82: 
03:00:29         	Error Trace:	image_load_test.go:82
03:00:29         	Error:      	Received unexpected error:
03:00:29         	            	rpc error: code = Unknown desc = failed to setup network for sandbox "58e9b0c7b1643ebd82e5390c20b198e37b23c284573f1e1ae3cd86c2170a5548": failed to set bridge addr: "cni0" already has an IP address different from 10.88.0.1/16
03:00:29         	Test:       	TestImageLoad
03:00:29 --- FAIL: TestImageLoad (5.71s)`

it results from that the baremental environment is not clean. should clean up cni flannel and k8s related files and interfaces.

PR #3919, opened 2021-09-01
lib: clean up k8s related evironment before real test

jongwu:
clean up stale network interfaces, files and process about cni, flannel
and k8s before tests.

Fixes: #3918

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -307,6 +307,20 @@ delete_containerd_cri_stale_resource() {
 	sudo rm -f /etc/systemd/system/containerd.service
 }
 
+cleanup_network_interface() {
+	local FLANNEL=$(ls /sys/class/net | grep flannel)
+	local CNI=$(ls /sys/class/net | grep cni)
+
+	[ "$FLANNEL" != "" ] && sudo ip link del $FLANNEL
+	[ "$CNI" != "" ] && sudo ip link del $CNI
+
+	FLANNEL=$(ls /sys/class/net | grep flannel)
+	CNI=$(ls /sys/class/net | grep cni)
+
+	[ "$FLANNEL" != "" ] && info "$FLANNEL doesn't clean up"
+	[ "$CNI" != "" ] && info "$CNI doesn't clean up"
+}
+
 gen_clean_arch() {
 	# Set up some vars
 	stale_process_union=( "docker-containerd-shim" )
@@ -322,6 +336,8 @@ gen_clean_arch() {
 		info "delete stale docker resource under ${stale_docker_dir_union[@]}"
 		delete_stale_docker_resource
 	fi
+	info "remove containers started by ctr"
+	clean_env_ctr
 	info "delete stale kata resource under ${stale_kata_dir_union[@]}"
 	delete_stale_kata_resource
 	info "Remove installed kata packages"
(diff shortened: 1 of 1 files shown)


.ci/lib.sh
```diff
@@ -322,6 +322,8 @@ gen_clean_arch() {
 		info "delete stale docker resource under ${stale_docker_dir_union[@]}"
 		delete_stale_docker_resource
 	fi
+	info "remove containers started by ctr"
+	sudo sh -c 'ctr c rm $(ctr c ls -q)'
```
marcel-apf, line 326: Please use the 'clean_env_ctr' function instead

.ci/lib.sh
```diff
@@ -331,15 +333,21 @@ gen_clean_arch() {
 	info "Remove installed containerd-cri related binaries and configuration"
 	delete_containerd_cri_stale_resource
 	info "Remove installed kubernetes packages and configuration"
+	[ "$(pgrep kubelet)" != "" ] && sudo sh -c 'kubeadm reset -f'
```
marcel-apf, line 336: I suppose we need this because it is suspected something else was broken before it. How do we know the system  behaves ok after this line?

.ci/lib.sh
```diff
@@ -331,15 +333,21 @@ gen_clean_arch() {
...
 	sudo sh -c 'rm -rf /opt/cni/bin/*'
-	sudo sh -c 'rm -rf /etc/cni'
-	sudo sh -c 'rm -rf /var/lib/cni'
+	sudo sh -c 'rm -rf /etc/cni /etc/kubernetes/'
+	sudo sh -c 'rm -rf /var/lib/cni /var/lib/etcd /var/lib/kubelet'
+	sudo sh -c 'rm -rf /run/flannel'
+
+	#cleanup stale network interface
+	sudo ip link del $(ls /sys/class/net|grep flannel)
+	sudo ip link del $(ls /sys/class/net|grep cni)
```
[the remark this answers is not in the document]
jongwu: what about testing if there are those network interfaces before delete them?

.ci/lib.sh
jongwu: sorry, I'm not catch your meaning. what's your meaning by "system ok". does the command "kubeadm reset -f" do something harmful to the system?
In my case, sometimes k8s is running before ci, so we must stop it. that's why I use "kubeadm reset -f". Do you have better  method to stop k8s service?

.ci/lib.sh
jongwu: it is just one line. Do you think we must wrap it?

marcel-apf: @jongwu, the clean_env_ctr function is already defined, just use it

.ci/lib.sh
marcel-apf: What am saying is that "kubeadm reset -f" is used in order to "fix" a previous issue.
Once you run this command you assume the problem is fixed.
Is there some way to ensure the problem is fixed?"

.ci/lib.sh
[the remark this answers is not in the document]
marcel-apf: That would be great, list them so we know what we had and what we delete.
Ensuring an empty interface list after deleting the interfaces will also help us to understand the operation succeeded.

.ci/lib.sh
jongwu: thanks @marcel-apf

.ci/lib.sh
jongwu: That's a good idea to test if the "k8s" is off. But I don't know what we can if it doesn't work. What about checking if kubelet is running and give a warning if it is?

.ci/lib.sh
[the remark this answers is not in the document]
jongwu: that sounds great!

.ci/lib.sh
marcel-apf: @jongwu if you don't have an easy way to ensure the kubeadm configuration is fine now, just disregard  my comment.

marcel-apf (approved):

marcel-apf:
@jongwu thanks for addressing my comments, I approved the PR.

marcel-apf:
Also the green CI color makes us look good!

jongwu:
@marcel-apf -， I make a change base on your comments. let's test it again.
/test

GabyCT:
@jongwu it seems that this has an impact into the metrics CI which is a baremetal at the boot time test, the CI is also installing k8s, maybe this can run only for ARM CI?

GabyCT:
@jongwu yes reviewing the data from metrics I think if you put the minpercent to 15 the issue can be solved properly, putting this value does not affect the performance in kata because it is saying that is taking less time to boot which is not a performance degradation

GabyCT (approved):

Issue #3871, opened 2021-08-24
ci: fails to remove running container in cleanup


Amulyam24:
In kubernetes cleanup, docker fails to remove running container

```
[reset] Deleting files: [/etc/kubernetes/admin.conf /etc/kubernetes/kubelet.conf /etc/kubernetes/bootstrap-kubelet.conf /e
tc/kubernetes/controller-manager.conf /etc/kubernetes/scheduler.conf]
[reset] Deleting contents of stateful directories: [/var/lib/etcd /var/lib/kubelet /var/lib/dockershim /var/run/kubernetes
 /var/lib/cni]

The reset process does not clean CNI configuration. To do so, you must remove /etc/cni/net.d

The reset process does not reset or clean up iptables rules or IPVS tables.
If you wish to reset iptables, you must do so manually by using the "iptables" command.

If your cluster was setup to utilize IPVS, run ipvsadm --clear (or similar)
to reset your system's IPVS tables.

The reset process does not clean your kubeconfig files and you must remove them manually.
Please, check the contents of the $HOME/.kube/config file.
Error response from daemon: Could not kill running container 06c6108602b8631c9311b39495542cf03adeb23ba9153ebda3fee0d4de8cc
17e, cannot remove - Cannot kill container 06c6108602b8631c9311b39495542cf03adeb23ba9153ebda3fee0d4de8cc17e: connection er
ror: desc = "transport: Error while dialing dial unix:///run/containerd/containerd.sock: timeout": unavailable
```

PR #3925, opened 2021-09-01
ci/lib: Introduce the registry_server_teardown()

wainersm:
This created the registry_server_teardown function which should
be used to remove the registry container after its use.

The original code is from integration/kubernetes/cleanup_env.sh but
changed to check before whether the container exists or not.

This is just an improvement to #3871.

Fixes: #3871

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -76,6 +76,23 @@ else
 	stress_image_pull_policy="Always"
 fi
 
+# Some tests will start a local registry server for container images.
+# Use this function to clean up the server container after its use.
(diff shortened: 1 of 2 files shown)


jodh-intel (approved):
Thanks @wainersm.

lgtm

Jakob-Naucke (approved):
thanks @wainersm, yeah, checking for running containers is better than blindly `|| true`ing the command like I did.

wainersm:
I don't know how those jobs failures are related with my PR. So I will just re-run the metrics job, it might be an intermittent issue.

Issue #3934, opened 2021-09-02
checkcommits: should not inspect the subject line of revert commit


wainersm:
**Which feature do you think can be improved?**

The checkcommits (cmd/checkcommits/checkcommits.go) program can be improved such as commits originated from `git revert` should be accepted without following the commit's subject rules.

For example, I just faced the following problem when I tried to revert a commit to the tests repo:
1. I did `git revert 3622e70b6fd160573b1127ba1fe3bc1fb2198ddc`
2. The subject line (auto-created by git) is *Revert "integration/kubernetes: Add cpu/memory constraints in resource quota"*
3. I edited the commit to include the *Fixes* and *Signed-off-by* tags, then git push
4. The *static-checks* job first complained about the lack of subsystem in the subject line
```
ERROR: Commit 1d80dcd: Failed to find subsystem in subject: "Revert "integration/kubernetes: Add cpu/memory constraints in resource quota""  
```
5. I changed the subject line to *k8s: Revert "integration/kubernetes: Add cpu/memory constraints in resource quota"*
6. git push. Now the static-checks complains about the subject being too long
7. I changed the subject line again, to *k8s: Revert "Add cpu/memory constraints in resource quota"*
8. git push. Now it is all fine.

**How can it be improved?**

IMHO the checkcommits should not inspect the line with pattern `^Revert ".*"`. In other words, it should be allowed commits with subject line auto-created by `git revert`.

wainersm:
@GabyCT what do you think?

fidencio:
fwiw, +1 for the proposal.

wainersm:
@fidencio btw, changes on the process such as that should pass through the board revision?

fidencio:
> @fidencio btw, changes on the process such as that should pass through the board revision?

/cc @kata-containers/architecture-committee for input

It should only go to the architecture committee meeting if there's some disagreement that only could be solved there.  Otherwise, business as usual. :-)

egernst:
This proposal seems fine to me.

GabyCT:
thank you @wainersm , I like the idea :)

jodh-intel:
Agreed - good idea! I'll create a PR...

PR #3938, opened 2021-09-03
checkcommits: Handle revert commits

jodh-intel:
Revert commits need to be treated differently since:

- They are formatted by `git(1)`, not the user.
- The summary includes a "`Revert `" prefix which may make the subject
  line "too long".
- Reverts imply urgency suggesting something important broke so we need
  to undo it quickly with the minimum of fuss!

Updated the `checkcommits` tool to ignore most checks if the
commit is a revert.

Fixes: #3934.


cmd/checkcommits/checkcommits.go
```diff
@@ -216,6 +232,11 @@ func checkCommitBody(config *CommitConfig, commit *Commit) error {
 		return err
 	}
 
+	if commit.revertCommit {
+		// Don't perform checks on revert commits.
+		return nil
```
fidencio, line 237: Will this also ignore the SoB check, @jodh-intel?  If that's the case (and it seems to be from a quick look in a code I'm not familiar with), I think we should do the SoB check before returning nil.

fidencio (commented):
thanks @jodh-intel, I've raised on question but series look good!

cmd/checkcommits/checkcommits.go
```diff
@@ -115,6 +120,17 @@ func checkCommitSubject(config *CommitConfig, commit *Commit) error {
 		return fmt.Errorf("Commit %v: pure whitespace subject", commit.hash)
 	}
 
+	lcSubject := strings.ToLower(subject)
+
+	if strings.HasPrefix(lcSubject, revertPrefix) {
```
devimc, line 125: with this change we are forcing  revert commits to have the `revert` prefix, right?  should this be documented?

devimc (commented):
thanks @jodh-intel -  I have a question

cmd/checkcommits/checkcommits.go
```diff
@@ -59,9 +62,11 @@ const (
 	defaultMaxBodyLineLength    = 150
 
 	defaultCommit = "HEAD"
-	defaultBranch = "master"
+	defaultBranch = "main"
 
 	versionSuffix = "for kata-containers"
+
+	revertPrefix = "revert"
```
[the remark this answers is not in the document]
jodh-intel: I personally prefer all lower-case, but yeah, let's keep it consistent with `git` - updated.

cmd/checkcommits/checkcommits.go
jodh-intel: Agreed - fixed.

cmd/checkcommits/checkcommits.go
jodh-intel: Agreed. I'll try to find a good location and raise a doc change PR for the main repo...

fidencio: `git revert` automatically adds that, @devimc.  Are you thinking about partial reverts?

devimc: just in case someone edit it and remove the revert prefix

cmd/checkcommits/checkcommits.go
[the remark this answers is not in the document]
fidencio: I think your approach was fine as it was, to be honest ... as you're lowering the subject a few lines below.
It was just confusing to read at first.  You mentioned you updated it ... are you then also lowering the revert prefix a few lines below?  (Now I think even more that I'd prefer your first approach, sorry).

cmd/checkcommits/checkcommits.go
```diff
@@ -115,6 +121,17 @@ func checkCommitSubject(config *CommitConfig, commit *Commit) error {
...
 
+	lcSubject := strings.ToLower(subject)
+
+	if strings.HasPrefix(lcSubject, revertPrefix) {
+		// A revert commit is formatted by git(1), not the user,
+		// so always allow it.
+		commit.revertCommit = true
+		commit.subsystem = revertPrefix
+
+		return nil
```
wainersm, line 132: Just wanted to confirm that even returning here it is still required the "Fixes" on the commit's body.
I got a little confused by the line below as it also checks for the "Fixes", but apparently on the subject line.

https://github.com/kata-containers/tests/blob/e74a844a2c295a06ef7acf91cc19d3976a5c63ff/cmd/checkcommits/checkcommits.go#L157

jodh-intel:
Branch updated folks. Note two significant changes:

- I've added some doc for revert commits.
- The tool now explicitly ignores the body of revert commits

GabyCT (approved):

Issue #3942, opened 2021-09-03
"vagrant up fedora" does not work on Ubuntu 20.04 host


wainersm:
The `vagrant up fedora` command fails on Ubuntu 20.04 with the following error:
```
Vagrant attempted to execute the capability 'reboot'
on the detect guest OS 'fedora', but the guest doesn't
support that capability. This capability is required for your
configuration of Vagrant. Please either reconfigure Vagrant to
avoid this capability or fix the issue by creating the capability.
```
@Jakob-Naucke  reported the issue here: https://github.com/kata-containers/tests/pull/3923#issuecomment-912672427

Currently it is required to configure the Fedora guest with cgroups v1 and that requires a reboot. Vagrant on Ubuntu host seems unable to reboot guests (whether Fedora or Ubuntu guests doesn't matter, it still fails).

`vagrant up ubuntu` on Ubuntu host works fine as it doesn't require reboot.

Possible solutions:

1. (mitigation actually) Mentions in https://github.com/kata-containers/tests/blob/main/README.md#using-vagrant-to-test-your-code-changes that `vagrant up fedora` on Ubuntu host is not supported.
2. Setup an account on [Vagrant cloud](https://app.vagrantup.com) to store a version of the Fedora's box with cgroups v1 enabled, therefore we just fetch the image and no reboot is needed.
3. Work to get the CI scripts running on Fedora with cgroups v2. That would be the best solution IMHO.

cpmeadors:
Removing need-reveiw and assigning to @wainersm

Issue #3757, opened 2021-07-27
No straightforward way to clean up after failed "vagrant up"


dgibson:
# Description of problem

Sometimes `vagrant up` fails, and when it does it's not at all clear how to clean up to try again.

# Expected result

Working instructions in `README.md` to remove a failed `vagrant up` install.

# Actual result

After a failed `vagrant up`, I tried:

 - Retrying `vagrant up` directly
```
$ vagrant up fedora
Bringing machine 'fedora' up with 'libvirt' provider...
==> fedora: Running action triggers before up ...
==> fedora: Running trigger...
==> fedora: Checking if box 'fedora/32-cloud-base' version '32.20200422.0' is up to date...
Name `tests_fedora` of domain about to create is already taken. Please try to run
`vagrant up` command again.
```
 - `vagrant destroy`
```
$ vagrant destroy fedora
Name `tests_fedora` of domain about to create is already taken. Please try to run
`vagrant up` command again.
```

# Further information

The only way I've yet found to return to a clean state is this complicated procedure (some steps may be unnecessary depending on exactly when/how the `vagrant up` failed: 
```
$ git clean -fdx
$ rm -rf ~/.vagrant.d
$ virsh destroy tests_fedora
$ virsh undefine tests_fedora
```

dgibson:
@marcel-apf for your attention.

PR #3965, opened 2021-09-14
Vagrant: boot q35 machine and other improvements

### LINKED ISSUE #3752: Vagrant testing VM uses ancient i440FX machine type
# Description of problem

The recently merged Vagrantfile for creating a VM with CI setup is a great improvement.  However, it currently runs the VM as qemu's i440FX based machine.  That means the "host" for the tests has no PCIe, and can't have an IOMMU which prevents reasonably testing a number of Kata features such as VFIO.

# Expected result

Vagrant VM runs with `q35` qemu machine which has PCIe and allows for an IOMMU.

# Actual result

Vagrant VM runs with i440FX based 'pc' machine, and attempting to add an IOMMU will result in a qemu error.

wainersm:
I started this string of commits by changing the Vagrantfile so that the VM boots with q35 machine type (fix #3752). As q35 is required to run the vfio tests (because it needs iommu and one extra virtio-net device) and the Vagrantfile had no support for such as test, I also changed it so that if the user exports `CI_JOB=VFIO` then extra arguments are passed to QEMU to allow testing Kata + VFIO.

Other changes came after to address the issues #3757 and #3942.

I tested those changes on my laptop x86_64 (Intel) with Fedora 33. I booted the fedora and ubuntu guests for VFIO and non-VFIO tests (`CI_JOB=CRIO_K8S`) normally (although sometimes the SSH connection fails after reboot; usually I destroy/up an 2nd time and it works). I don't have a ubuntu bare-metal to test the VFIO job but I will give a try with nested VMs.

Still about the VFIO arguments being passed to QEMU...now I realized that they are specific for Intel processors, I should be working on the AMD equivalents when/if time permits (if that is your case, i.e., you'd like to run VFIO tests on AMD, then please let me know).

**Warn**: if you have any vagrant boxes created from our old Vagrantfile then I recommend you to destroy them all *before* apply this pull request. The reason is because I changed the prefix name which is used to create the domain in libvirt; so it may leave orphan domains on your machine.

### DIFF
--- a/.ci/vagrant-cleaner.sh
+++ b/.ci/vagrant-cleaner.sh
@@ -1,0 +1,160 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+cidir="$(dirname $(readlink -f "$0"))"
+
+# Print message to stderr (add the 'ERROR:' prefix) and exit 1.
+die() {
+	echo -e "ERROR: $*" >&2
+	exit 1
+}
+
+usage() {
+	cat <<-EOF
+		This script helps you to clean up the VMs and Vagrant's control
+		directories.
+		By default it will attempt to destroy all VMs. For additionally
+		remove control directories use one or more of the flags shown
+		below.
+
+		Use: $0 [-g] [-h] [-l] [-n VM_NAME], where
+		  -g: also remove the vagrant's global directory (~/.vagrant.d)
+		      Implies -l.
+		  -h: print this help.
+		  -l: also remove the vagrant's local directory.
+		  -n: destroy the VM. Do not remove any vagrant's directories.
+
+		Caution: only use the -g flag if you know what you are doing.
+	EOF
+}
+
+# Return the VM names.
+get_vms() {
+	local vms=()
+	for vm in $(vagrant status --machine-readable | grep 'metadata,provider'); do
+		vms+=($(awk -F, '{ print $2}' <<< "$vm"))
+	done
+	echo ${vms[@]}
+}
+
+# Use vagrant to gentle destroy the VM.
+#
+# Parameters:
+#   $1 - the VM name
+vagrant_destroy() {
+	local vm="$1"
+	if vagrant destroy --force "$vm"; then
+		echo "VM '$vm' destroyed: OK"
+	else
+		echo "VM '$vm' destroyed: FAILED"
+		return 1
+	fi
+}
+
+# Remove the domain from libvirt.
+#
+# Parameters:
+#   $1 - the VM name
+libvirt_cleanup() {
+	local vm="$1"
+	# TODO: find a way to discovery the domain prefix.
+	local prefix="tests_"
+	local domain="${prefix}${vm}"
+	if ! virsh dominfo "$vm" &> /dev/null; then
+		echo "Domain '$domain' does not exist. Nothing to do."
+	elif virsh destroy "$domain" && virsh undefine "$domain"; then
+		echo "Domain '$domain' cleaned on libvirt: OK"
+	else
+		echo "Domain '$domain' cleaned on libvirt: FAILED"
+		return 1
+	fi
+}
+
+# Entrypoint function to remove the VM. In case it cannot be destroyed with
+# vagrant, it will forcibly remove the domain from libvirt.
+#
+# Parameters:
+#   $1 - the VM name
+vm_wipeout() {
+	local vm="$1"
+	if ! vagrant status "$vm" &>/dev/null; then
+		echo "VM '$vm' does not exist. Nothing to do."
+		return 0
+	fi
+
+	vagrant_destroy "$vm"
+	if [ $? -ne 0 ]; then
+		echo "WARN: Attempt to clean up the domain on libvirt"
+		libvirt_cleanup "$vm"
+	fi
+}
+
+# Remove vagrant's control directories, whether local or global.
+#
+# Parameters:
+#  $1 - set to 1 to remove the global directory as well. Defaults to 0.
+cfgs_wipeout() {
+	local global=${1:-0}
+	local local_cfg="${cidir}/../.vagrant"
+	local global_cfg="${HOME}/.vagrant.d"
+
+	echo "Remove vagrant's local directory: $local_cfg"
+	rm -rf "$local_cfg"
+	if [ "$global" -eq 1 ]; then
+		echo "Remove vagrant's global directory: $global_cfg"
+		rm -rf "$global_cfg"
+	fi
+}
+
+main() {
+	local vms=()
+	local ret=0
+	local single_vm=""
+	local local_cfg=0
+	local global_cfg=0
+	while getopts "aghln:" OPT; do
+		case $OPT in
+			g) global_cfg=1;;
+			h) usage; exit 0;;
+			l) local_cfg=1;;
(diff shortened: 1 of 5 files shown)


Vagrantfile
```diff
@@ -44,6 +44,25 @@ Vagrant.configure("2") do |config|
     if host_arch == "x86_64"
       lv.machine_type = "q35"
     end
+    # The VM needs one additional virtio-net device and iommu enabled
+    # for the vfio tests.
+    if job == "VFIO" and host_arch == "x86_64"
+      # Overwrite the default -machine already built by vagrant as
+      # additional arguments aren't allowed on `machine_type` property.
```
dgibson, line 53: I'm not sure what you mean by this.  It's definitely possible to have multiple `-machine` arguments (and frequently useful to give machine arguments separately from the machine type).

Vagrantfile
```diff
@@ -44,6 +44,25 @@ Vagrant.configure("2") do |config|
...
+      lv.qemuargs :value => "-machine"
+      lv.qemuargs :value => "q35,accel=kvm,kernel-irqchip=split,usb=off,dump-guest-core=off"
+      lv.qemuargs :value => "-device"
+      lv.qemuargs :value => "intel-iommu,intremap=on,caching-mode=on,device-iotlb=on"
+      # Currently the vfio test picks the last virtio-net device from lspci's
+      # output. Here we add the device in a PCIe root port with higher slot
+      # number on the hope it will be the last in the list.
+      lv.qemuargs :value => "-device"
+      lv.qemuargs :value => "pcie-root-port,port=0x16,chassis=7,id=pcie.7,multifunction=on,bus=pcie.0,addr=0x04"
```
dgibson, line 62: Any reason to use 0x04, rather than a really high slot number?  I think q35 has integrated devices in slot 0x1f and maybe 0x1e as well, but you can still go much higher than 4.

dgibson (commented):
Changes look good, to the extent that I understand them, with the exception of a couple of nits I've commented on.  Currently trying the changes out as well, I'll follow up when I have some results from that.

Vagrantfile
wainersm: No particular reason. I can change it to a higher slot so decrease the probability of the test not catching that device.

Vagrantfile
wainersm: I need to add `kernel-irqchip=split` to the machine type. Setting setting `lv.machine_type=q35,kernel-irqchip=split` won't work as it seems the vagrant-libvirt plug-in cannot parse correctly the comma-separated string.

I wasn't sure I could give `-machine` multiple times and QEMU would sum up the arguments (in my mind the last `-machine`'s value would overwrite all the previous)

So a command-line like `-machine pc-q35-5.1,accel=kvm,usb=off,dump-guest-core=off (...) -machine=q35,kernel-irqchip=split` is equivalent to `-machine pc-q35-5.1,accel=kvm,usb=off,dump-guest-core=off,kernel-irqchip=split`?

wainersm:
I just updated this pull request, where I:
a) addressed @dgibson 's the comments by simplifying the `-machine` argurment string for VFIO tests; also the virtio-net is attached to a higher PCIe slot number
b) added the commit https://github.com/kata-containers/tests/pull/3965/commits/3b286bf469afbf4eed483117fde4f3a4738afc8c

Vagrantfile
dgibson: Well, yes, but more to the point, so is `-machine pc-q35-5.1 -machine accel=kvm -machine usb=off -machine dump-guest-core=off -machine kernel-irqchip=split`.

Vagrantfile
```diff
@@ -44,6 +44,23 @@ Vagrant.configure("2") do |config|
     if host_arch == "x86_64"
       lv.machine_type = "q35"
     end
+    # The VM needs one additional virtio-net device and iommu enabled
+    # for the vfio tests.
+    if job == "VFIO" and host_arch == "x86_64"
```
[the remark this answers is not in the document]
wainersm: I had to rollback that idea... in Ubuntu host vagrant fails to detect reboot. If I always set the variables for VFIO tests then the ubuntu guest will always reboot (so it won't work for any CI_JOB).

wainersm:
@jodh-intel @dgibson can we get this merged? I think it is good enough... I can work on follow up PRs if needed.

dgibson (approved):

.ci/vagrant-cleaner.sh
```diff
@@ -0,0 +1,164 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 Red Hat, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+cidir="$(dirname $(readlink -f "$0"))"
+
+# Print message to stderr. Add the 'ERROR:' prefix.
+perror() {
```
jodh-intel, line 10: Spot the C hacker! :smile:

Should this call `exit 1` though? It looks like the majority of calls immediately exit.

.ci/vagrant-cleaner.sh
```diff
@@ -0,0 +1,164 @@
...
+# Use vagrant to gentle destroy the VM.
+#
+# Parameters:
+#   $1 - the VM name
+vagrant_destroy() {
+	local vm="$1"
+	if vagrant destroy --force "$vm"; then
+		echo "VM '$vm' destroyed: OK"
+	else
+		echo "VM '$vm' destroyed: FAILED"
```
jodh-intel, line 51: s/echo/perror/ ?

.ci/vagrant-cleaner.sh
```diff
@@ -0,0 +1,164 @@
...
+
+	if [ -n "$single_vm" ]; then
+		vm_wipeout "$single_vm" || ret=$?
+	else
+		for vm in $(get_vms); do
+			vm_wipeout $vm || ret=$?
+		done
+
+		[ $ret -eq 0 ] || \
+			perror "Failed the removal of some VM."
```
jodh-intel, line 148: Should the script die at this point?

.ci/vagrant-cleaner.sh
wainersm: You are right. I just updated the PR but I renamed `perror` to `die`  :(

.ci/vagrant-cleaner.sh
wainersm: Here I just want to info the user.

.ci/vagrant-cleaner.sh
wainersm: done.

jodh-intel (approved):
Thanks @wainersm.

wainersm:
Moved to wip to fix the job jenkins-ci-ubuntu-1804-containerd-k8s-e2e-minimal. This job is running on Ubuntu 18.04, which apparently hasn't the driverctl package (see commit a3d903a9b4a1d1029ff13d1c1300b93d92e2999a of this PR).

wainersm:
Updated the commit https://github.com/kata-containers/tests/pull/3965/commits/fe4bc7fe37ea0715ded0e5baee27092ac5d983bf so that driverctl package is only installed on Ubuntu 20.04 (or newer) because that isn't available on 18.04 thus some CI jobs failing.

wainersm:
Hi @jodh-intel , updated the commit https://github.com/kata-containers/tests/commit/fe4bc7fe37ea0715ded0e5baee27092ac5d983bf , mind to review again? :)

jodh-intel (approved):
Thanks @wainersm.

lgtm

Issue #3972, opened 2021-09-15
ARM CI is failing


GabyCT:
It seems that the ARM CI is failing
```
14:52:37 Active Platform          = /tmp/kata-containers/tmp.OoPrHOSyfF/edk2/ArmVirtPkg/ArmVirtQemu.dsc
14:52:37 .... done!
14:52:38 Building ... /tmp/kata-containers/tmp.OoPrHOSyfF/edk2/MdeModulePkg/Library/LzmaCustomDecompressLib/LzmaCustomDecompressLib.inf [AARCH64]
14:52:42 Building ... /tmp/kata-containers/tmp.OoPrHOSyfF/edk2/ArmPkg/Library/CompilerIntrinsicsLib/CompilerIntrinsicsLib.inf [AARCH64]
[...]

PR #3973, opened 2021-09-16
arm64: fix fail to download toolchain issue

jongwu:
SSL certification fail when download compiler toolchain for building edk2.
here we disable curl SSL certificate validating to let it go.

Fixes: #3972

### DIFF
--- a/.ci/aarch64/install_rom_aarch64.sh
+++ b/.ci/aarch64/install_rom_aarch64.sh
@@ -49,7 +49,7 @@ build_uefi()
 
 	mkdir toolchain
 	pushd toolchain/
-	curl -LO "${TOOLCHAIN_SOURCE_URL}" && tar -xf "${TOOLCHAIN_ARCHIVE}"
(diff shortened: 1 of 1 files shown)


jongwu:
help review @fidencio @GabyCT, this bug block the arm ci.

Jakob-Naucke:
@jongwu I'd like to unblock the ARM CI as much as you. OTOH, disabling secure connections seems like a _huge_ red flag to me. I can reproduce this and I'm finding it a little weird (curl & wget will complain the certificate is not trusted with recent ca-certificates installed, but in a browser, it works for me). Is there any particular reason for this?

jongwu:
Hello @Jakob-Naucke -, I have no exact idea about the reason. I guess that it's caused by recent update of that web site. I have requested help from arm internal but I'm not sure when will that issue can be fixed. Before that I have no better way to bypass this issue.

jongwu:
This issue has been fixed in Arm internal. thus I will close it.

Issue #3812, opened 2021-08-10
Agent shutdown test needs Rust explicitly


Jakob-Naucke:
The newly introduced agent shutdown test (#3495) needs Rust, but if we build the agent itself in a container, Rust will not be installed on the host.

cmaf:
Hey @Jakob-Naucke, we are reverting https://github.com/kata-containers/tests/commit/ca718a600954323cfe95624913e880f2c59543d9 due to CI failures (see https://github.com/kata-containers/tests/issues/3978), so I'm reopening this issue

PR #3979, opened 2021-09-17
agent: Revert "agent: Install Rust for shutdown test"

cmaf:
The changes break CI jobs which run test-agent-shutdown.sh.
This reverts commit ca718a600954323cfe95624913e880f2c59543d9.

Fixes #3978

Signed-off-by: Chelsea Mafrica <chelsea.e.mafrica@intel.com>

Issue #3978, opened 2021-09-17
ci failure related to tracing/test-agent-shutdown.sh


cmaf:
Cargo command is not found and there is an additional permissions error in 2 tests. I think it is related to https://github.com/kata-containers/tests/commit/ca718a600954323cfe95624913e880f2c59543d9 which was recently merged. @Jakob-Naucke 

jenkins-ci-pmem-ubuntu-18-04:
```
+ local cargo=/home/jenkins/.cargo
+ '[' -d /home/jenkins/.cargo ']'
+ sudo chown -R jenkins: /home/jenkins/.cargo /tmp/jenkins/workspace/kata-containers-2.0-pmem-ubuntu-20.04-PR/go/src/github.com/kata-containers/kata-containers
+ cargo --version
tracing/test-agent-shutdown.sh: line 774: cargo: command not found
+ /tmp/jenkins/workspace/kata-containers-2.0-pmem-ubuntu-20.04-PR/go/src/github.com/kata-containers/tests/tracing/../.ci/install_rust.sh
Install rust 1.54.0
info: downloading installer
error: could not write settings file: '/home/jenkins/.rustup/settings.toml': Permission denied (os error 13)
[install_rust.sh:25] ERROR: sh -s -- -y --default-toolchain 1.54.0
```

jenkins-ci-ubuntu-20-04:
```
+ local cargo=/home/jenkins/.cargo
+ '[' -d /home/jenkins/.cargo ']'
+ sudo chown -R jenkins: /home/jenkins/.cargo /tmp/jenkins/workspace/kata-containers-2.0-ubuntu-20.04-PR/go/src/github.com/kata-containers/kata-containers
+ cargo --version
tracing/test-agent-shutdown.sh: line 774: cargo: command not found
+ /tmp/jenkins/workspace/kata-containers-2.0-ubuntu-20.04-PR/go/src/github.com/kata-containers/tests/tracing/../.ci/install_rust.sh
Install rust 1.54.0
info: downloading installer
error: could not write settings file: '/home/jenkins/.rustup/settings.toml': Permission denied (os error 13)
[install_rust.sh:25] ERROR: sh -s -- -y --default-toolchain 1.54.0
```

cmaf:
see https://github.com/kata-containers/kata-containers/pull/2620

cmaf:
@GabyCT should we revert this commit until it can be further debugged?

GabyCT:
@Jakob-Naucke could you please take a look?

[...]

PR #3813, opened 2021-08-10
agent: Install Rust for shutdown test

Jakob-Naucke:
The newly introduced agent shutdown test (#3495) needs Rust, but if we
build the agent itself in a container, Rust will not be installed on the
host. Add a Makefile target for this.

Fixes: #3812
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

PR #3985, opened 2021-09-20
agent: Install Rust for shutdown test -- respin

Jakob-Naucke:
The newly introduced agent shutdown test (#3495) needs Rust, but if we
build the agent itself in a container, Rust will not be installed on the
host. Add a catch in the script. For agent-ctl, run permissions fix as
for cargo setup (this was relevant only on the initrd CI).

This is a respin of #3813 following #3979. The existence check of Cargo
was (erroneously) run _before_ the `source` that would put Cargo into
$PATH.

Fixes: #3812

/cc @cmaf @GabyCT
The problem is that while I'm certain this will pass regular tests, I cannot promise #3978 will not return because AFAICT that was due to a build cloning from main, which is why it wasn't caught in tests, cf. https://github.com/kata-containers/kata-containers/issues/2407. WDYT?


GabyCT:
@Jakob-Naucke I will put this as DNM label first until we solve the issue that this test is running in a bunch of CI jobs where it should not run

GabyCT:
@Jakob-Naucke it seems that the agent shutdown test is not working properly as you can see here https://github.com/kata-containers/tests/blob/main/tracing/test-agent-shutdown.sh#L754. This test should only run on CI_JOB:= 'CRI_CONTAINERD_K8S_MINIMAL' and it is disabled on that job and tracing is only run at that job https://github.com/kata-containers/tests/blob/8155bba5890cb4cb342236f94c36c21526f33241/.ci/run.sh#L62
/cc @jodh-intel

GabyCT:
@Jakob-Naucke so the issue of running on multiple CIs has been fixed but I am not sure how this PR will be test it if the test is not running in any job

Jakob-Naucke:
@GabyCT well, `make tracing` is in `CRI_CONTAINERD_K8S_MINIMAL`, which is run in e.g. `jenkins-ci-ubuntu-1804-containerd-k8s-e2e-minimal`, so we can remove DNM, right?

GabyCT:
@Jakob please see https://github.com/kata-containers/tests/issues/3990

GabyCT:
I will remove the DNM label just please coordinate with @cmaf and @jodh-intel about where this will be test it

devimc (approved):
thanks @Jakob-Naucke

cmaf (approved):
@Jakob-Naucke these changes LGTM, but since the test is effectively disabled right now, we won't be able to catch any failures right away if we merge it. 
https://github.com/kata-containers/tests/blob/621bf88c03a3f35abb0cc686fa5a16f2b3ff897d/tracing/test-agent-shutdown.sh#L754-L761
We could try enabling it in a different job, and until then, add the do-not-merge label.

Jakob-Naucke:
@cmaf I'm really just here for https://github.com/kata-containers/kata-containers/pull/2399 :grin:. Can we merge that one for the time being, even when the re-enabling of agent shutdown testing is going to depend on this then?

Jakob-Naucke:
@cmaf ^

cmaf:
@Jakob-Naucke I read through  kata-containers/kata-containers#2399, since the agent shutdown test is disabled now I don't see a problem merging it. I'll make sure to either merge this one or add your commit to a PR when we re-enable it.

PR #4005, opened 2021-09-23
snap: check if kata_repo_dir exist before go into

jongwu:
"kata_repo_dir" may not exist. So before go into it, do check
and get it if there is no such dir.

Depends-on: github.com/kata-containers/kata-containers#2706
Fixes: #4004

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -200,7 +200,8 @@ test_snap_build() {
 	if [[ "${ID}" == "ubuntu" && "$(uname -m)" != "x86_64" ]]; then
 		echo "Test snap build"
 		sudo apt install -y snapcraft
-		pushd ${cidir}/../../kata-containers
+		[ ! -d "${kata_repo_dir}" ] && go get -d "${kata_repo}" || true
+		pushd "${kata_repo_dir}"
 		sudo snapcraft -d snap --destructive-mode
 		# PREFIX is changed in snap build, change it back
 		PREFIX=


.ci/jenkins_job_build.sh
```diff
@@ -200,7 +200,8 @@ test_snap_build() {
 	if [[ "${ID}" == "ubuntu" && "$(uname -m)" != "x86_64" ]]; then
 		echo "Test snap build"
 		sudo apt install -y snapcraft
-		pushd ${cidir}/../../kata-containers
+		[ ! -d "${kata_repo_dir}" ] && go get -d "${kata_repo}" || true
```
fidencio, line 203: Do we enforce `GO111MODULE=auto` before calling `go get -d ...`?  If so, we're fine.  If not, we should, otherwise it'll be downloaded in a different place than the one expected.

jongwu: do you mean change it to "sudo -E GO111MODULE=auto go get -d "${kata_repo}"?

fidencio: No, nevermind.  I just double-checked and on line 167 we do:
```
source "${cidir}/lib.sh"
```

And that does:
```
# Support Golang 1.16.x.
# By default in Golang >= 1.16 GO111MODULE is set to "on",
# some subprojects in this repo may not support "go modules",
# set GO111MODULE to "auto" to enable module-aware mode only when
# a go.mod file is present in the current directory.
export GO111MODULE="auto"
```

So, we're good to go as it's here.

fidencio (approved):
lgtm, thanks @jongwu!

jongwu:
thanks @fidencio -. take a look at this @Jakob-Naucke

Jakob-Naucke (commented):
sure, whatever, but -- where did you observe this? Don't all of our jobs clone the main repo? How does one get to this point in the script without a main repo checkout?
e: ah, just read your issue. This is strange...

.ci/jenkins_job_build.sh
Jakob-Naucke: I don't think we'd ever need L. 203 and the fix only lies in L. 204 because `$cidir` is a relative path. @jongwu if you really feel more comfortable with it, we can keep it -- I guess it doesn't hurt much if the checkout already exists.

Amulyam24:
Thanks @jongwu! This fixes the issue on ppc64le as well.

jongwu:
> sure, whatever, but -- where did you observe this? Don't all of our jobs clone the main repo? How does one get to this point in the script without a main repo checkout?
> e: ah, just read your issue. This is strange...

yeah, it's strange for me too. But it happens so we must do something to avoid that crash.

.ci/jenkins_job_build.sh
Jakob-Naucke: @jongwu are you strongly in favour of keeping the `go get`?

jongwu:
What's wrong with "jenkins-ci-ubuntu-1804-clh-containerd-k8s-e2e-minimal"? I can't trigger it.  @likebreath

.ci/jenkins_job_build.sh
jongwu: keep it will not bring more overhead as we do check before "go get". Just in case the kata_repo is gone sometimes.

Jakob-Naucke (approved):

Jakob-Naucke:
Y'all need /test-k8s-clh-minimal (I did have to check the trigger phrase :smile:)

Issue #4045, opened 2021-09-30
CCv0: error in CCv0 part of install_kata_image.sh


stevenhorsman:
# Description of problem

I was looking through the [output of one of the CCv0 PR jobs](http://jenkins.katacontainers.io/view/CCv0/job/kata-containers-CCv0-ubuntu-20.04-PR/42/consoleFull) and noticed an issue with the ccv0 branch of the install_kata_image.sh script such that my custom ccv0 code wasn't been run and so the extra packages needed weren't being used:
`.ci/install_kata_image.sh: line 46: [: missing `]'`

As I've learnt a bit more about the CI code since then I also spotted that I'd missed out a bit of logic that is required for the fedora-like systems to use podman as well, so that code needs moving out of the if branch

# Expected result

When the ci job is run in the output you don't see the error, but see `Creating rootfs for fedora` which is our ccv0 image base OS

# Actual result

In the output 
```
.ci/install_kata_image.sh: line 46: [: missing `]'
Creating rootfs for ubuntu
```

stevenhorsman:
Closed by #4046

PR #4046, opened 2021-09-30
CCv0 | ci: Fix ccv0 install_install script

stevenhorsman:
- Fix the conditional, so it doesn't syntax error
- Move the `use_docker` logic out of the conditional, so both paths can
use it.

Fixes: #4045


.ci/install_kata_image.sh
```diff
@@ -45,6 +45,7 @@ build_rust_image() {
 	case "$build_method" in
 		"distro")
 			distro="${osbuilder_distro:-fedora}"
+			OS_VERSION="30"
```
c3d: Could you try switching to a more recent version of Fedora (the currently released one is 35 now)

stevenhorsman: @c3d - I'm currently looking into it, but unfourtunately with version 34 of fedora the containerd tests fail, but they pass with fedora 30. I'm hoping to end up with tests working on fedora 34, but that's proving a challenge at the moment.

PR #4038, opened 2021-09-28
kubernetes-e2e: improvements to the cleanup()

wainersm:
The main goal with this PR is to fix [kubernetes-e2e: cannot be executed more than one time](https://github.com/kata-containers/tests/issues/4037). I also included two improvements.

PR #3803, opened 2021-08-06
ci: Also test firecracker with jailer

fidencio:
FIRECRACKER_JAILER is a target introduced and used by the two* CIs,
which have been recently introduced to avoid regressions when running
firecracker with the jailer.

*: The CIs using this CI_JOB target:
   - kata-containers-2.0-tests-firecracker-jailer-ubuntu-PR
   - kata-containers-2.0-firecracker-jailer-ubuntu-PR

Fixes: #3802

Signed-off-by: Anastassios Nanos <ananos@nubificus.co.uk>
Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>

PR #4058, opened 2021-10-04
Add crio ci script

littlejawa:
Add an entry point for the kata CI job used on cri-o PRs.

This is a copy of the script run by http://jenkins.katacontainers.io/job/kata-containers-2-crio-PR/
When this is added to our repository, the job will be changed to fetch that script and run it, like our other jenkins jobs.
This will allow future changes to this job to benefit from versionning and peer review.


c3d (changes requested):
@littlejawa I like the direction this is taking and the code changes, but the commits seem to be under the wrong name. Please check.

littlejawa:
> @littlejawa I like the direction this is taking and the code changes, but the commits seem to be under the wrong name. Please check.

Right... Not sure where that's coming from. I'll check.

c3d (approved):
Looks good to me now that this is authored by you.

fgiudici (approved):
lgtm, thanks @littlejawa :-)

fidencio:
@littlejawa, this is a very good first step.
I wonder if we should add a new  EXTERNAL_CRIO` (or similar) target to our {ci_jobs_flags,run}.sh, similarly to what we did when adding the FIRECRACKER_JAILER target.  See as reference: https://github.com/kata-containers/tests/pull/3803/files

littlejawa:
> I wonder if we should add a new EXTERNAL_CRIO` (or similar) target to our {ci_jobs_flags,run}.sh, [...]

I didn't think about it, but yes, I could certainly use as much as possible from existing scripts, and reduce the size of this one.
Looking at it quickly, it wouldn't reduce much though.
I can add ci_jobs_flags.sh and run.sh changes to this commit - running run.sh at the end of ci_crio_entry_point.sh.

And then, maybe I could go further, and use the jenkins_job_build.sh ?
This would reduce my new script further I think.

Is it what you have in mind?

fidencio:
> > I wonder if we should add a new EXTERNAL_CRIO` (or similar) target to our {ci_jobs_flags,run}.sh, [...]
> 
> I didn't think about it, but yes, I could certainly use as much as possible from existing scripts, and reduce the size of this one. Looking at it quickly, it wouldn't reduce much though. I can add ci_jobs_flags.sh and run.sh changes to this commit - running run.sh at the end of ci_crio_entry_point.sh.
> 
> And then, maybe I could go further, and use the jenkins_job_build.sh ? This would reduce my new script further I think.

That's where I'd like us to get with your new script.
With this the external CRI-O CI will basically follow exactly the same pattern as we have for our "internal" CIs.

> 
> Is it what you have in mind?

That's exactly what I have in mind. :-)

.ci/ci_job_flags.sh
```diff
@@ -178,6 +178,14 @@ case "${CI_JOB}" in
 	export KUBERNETES="yes"
 	export experimental_kernel="true"
 	;;
+"EXTERNAL_CRIO")
```
fidencio, line 181: please, just ensure we have all thje CI flags initalised as expected by calling `init_ci_flags` as the first thing.

.ci/run.sh
```diff
@@ -92,6 +92,12 @@ case "${CI_JOB}" in
 		echo "INFO: Running complete e2e kubernetes tests"
 		sudo -E PATH="$PATH" CRI_RUNTIME="containerd" bash -c "make kubernetes-e2e"
 		;;
+	"EXTERNAL_CRIO")
+		echo "INFO: Running tests on cri-o PR"
+		sudo -E PATH="$PATH" bash -c "make kubernetes"
+		sudo -E PATH="$PATH" bash -c "make kubernetes-e2e"
+		sudo -E PATH="$PATH" bash -c "make crio"
```
fidencio, line 99: Are we fine running those 3 back-to-back?  I remember we had to do some cleanups (CNI / network related) between one and another.  Isn't this needed anymore?

fidencio (commented):
A few comments, but it's getting there.
Thanks for the patience, @littlejawa!

.ci/run.sh
littlejawa: No it's not :)
Our cleanup actually started to fail when #4038 was merged.
I had to fix it in the jenkins job, and it's actually why I decided to do this PR now :-)

fidencio: Perfect then!
Seems that I missed a lot of fun since the start of the vacation. :-)

.ci/ci_job_flags.sh
littlejawa: right - my bad.
Just fixed it.

fidencio (approved):
lgtm, thanks @littlejawa!

PR #4060, opened 2021-10-04
ci: Fix cleanup for Metrics CI baremetal

### LINKED ISSUE #4059: fix cleanup for metrics CI baremetal

GabyCT:
This PR removes unnecessary steps in the clean up for the metrics CI
baremetal in order to avoid errors like ERROR: packages not installed or
removing CRI-O dependencies when CRI-O is not even install in the system
and so on. This PR only leaves the clean up that is necessary for the metrics CI.

Fixes #4059

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -14,7 +14,7 @@ export KATA_ETC_CONFIG_PATH="/etc/kata-containers/configuration.toml"
 export kata_repo=${katacontainers_repo:="github.com/kata-containers/kata-containers"}
 export kata_repo_dir="${GOPATH}/src/${kata_repo}"
 export kata_default_branch="${kata_default_branch:-main}"
-
+export CI_JOB="${CI_JOB:-}"
 
 # Name of systemd service for the throttler
 KATA_KSM_THROTTLER_JOB="kata-ksm-throttler"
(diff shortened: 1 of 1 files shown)


devimc (approved):

.ci/lib.sh
```diff
@@ -324,30 +324,37 @@ cleanup_network_interface() {
...
-	stale_docker_mount_point_union=( "/var/lib/docker/containers" "/var/lib/docker/${docker_storage_driver}" )
-	stale_docker_dir_union=( "/var/lib/docker" )
-	stale_kata_dir_union=( "/var/lib/vc" "/run/vc" "/usr/share/kata-containers" "/usr/share/defaults/kata-containers" )
-
-	info "kill stale process"
-	kill_stale_process
-	if [ -z "${USE_PODMAN}" ]; then
-		info "delete stale docker resource under ${stale_docker_dir_union[@]}"
-		delete_stale_docker_resource
+	if [[ $CI_JOB != "METRICS" ]]; then
```
c3d, line 327: Could you add a comment explaining why METRICS is different here? Maybe something like

```
# The metrics CI on bare metal generates errors during cleanup because it does not install packages
```

c3d (approved):
Looks good to me, with maybe a little additional comment.

Issue #4067, opened 2021-10-08
ci: cri-o integration test "ctr logging [tty=true]" fails randomly


littlejawa:
When running cri-o integration tests with Kata, one of the test fails every now and then, without any relation to the source PR.
The test is "ctr logging [tty=true]".

At the moment, the logs from the failed run didn't help catch the problem.

In order to keep having smooth tests, I will add this test to the "skip" list, and investigate separately.

PR #4068, opened 2021-10-08
ci: skip "ctr logging [tty=true]"

littlejawa:
Adding this test to the skip list to prevent random failures of the CI.

Fixes: #4067


haircommander:
part of me feels like we should figure out why the test randomly flakes. If we do disable, I think an issue should be opened to track fixing

littlejawa:
> fine with me, can we just open an issue to track?

Done :-)
#4069

Issue #4087, opened 2021-10-14
waitForProcess is used before definition


Jakob-Naucke:
in `clean_env_ctr` in `lib/common.bash`, where `waitForProcess` is also defined, but _after_ use

Issue #4083, opened 2021-10-13
Random failures on init k8s for s390x


GabyCT:
I have seen multiple random failures on s390x CI with the init script on k8s
http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-s390x-PR/365/console
/cc @Jakob-Naucke

PR #4088, opened 2021-10-14
lib: Define waitForProcess before use

Jakob-Naucke:
`waitForProcess` is used in `lib/common.bash`'s `clean_env_ctr`, but
only defined thereafter. Put the definition in front.

Fixes: #4087

no backport required as stable does not have this issue


Jakob-Naucke:
let's /test-s390x for now also to see whether #4083 has been resolved

Jakob-Naucke:
cool. /test

PR #4091, opened 2021-10-14
stable-2.2 | k8s: add test to check capabilities of pod

egernst:
Check permitted capabilities of a given container process, as well as
from exec session into same container.

Fixes: #4081

Signed-off-by: Eric Ernst <eric_ernst@apple.com>


egernst:
That’s odd. If the log/grep is failing, it seems that maybe the container hadn’t run yet by the time I’m checking the log? I wouldn’t expect this would take so long. 

The print out of the value later shows that it had completed and had a value that should’ve matched the check.

PR #4137, opened 2021-11-01
metrics: Enable the check dockerfiles images for metrics Report

### LINKED ISSUE #4136: metrics: Enable the check dockerfiles images for metrics Report

GabyCT:
This PR brings back the check dockerfiles images function as it needs
to be used for the make report script for the metrics CI.

Fixes #4136

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -281,3 +281,36 @@ restart_docker_service() {
 
 	return 0
 }
+
+# This function verifies that the dockerfile version is
+# equal to the test version in order to build the image or
+# just run the test
+check_dockerfiles_images()
+{
+	local image="${1:-}"
+	local dockerfile_path="${2:-}"
+
+	if [ -z "$image" ] || [ -z "$dockerfile_path" ]; then
+		die "Missing image or dockerfile path variable"
+	fi
+
+	# Verify that dockerfile version is equal to test version
+	check_image=$(docker images "$image" -q)
+	if [ -n "$check_image" ]; then
+		# Check image label
+		local check_image_version=$(docker image inspect $image | grep -w DOCKERFILE_VERSION | head -1 | cut -d '"' -f4)
+		if [ -n "$check_image_version" ]; then
+			echo "$image is not updated"
+			build_dockerfile_image "$image" "$dockerfile_path"
+		else
+			# Check dockerfile label
+			local dockerfile_version=$(grep DOCKERFILE_VERSION= $dockerfile_path | cut -d '"' -f2)
+			if [ "$dockerfile_version" != "$check_image_version" ]; then
+				echo "$dockerfile_version is not equal to $check_image_version"
+				build_dockerfile_image "$image" "$dockerfile_path"
+			fi
+		fi
+	else
+		build_dockerfile_image "$image" "$dockerfile_path"
+	fi
+}


lib/common.bash
```diff
@@ -281,3 +281,36 @@ restart_docker_service() {
...
+{
+	local image="$1"
+	local dockerfile_path="$2"
+
+	if [ -z "$image" ] || [ -z "$dockerfile_path" ]; then
+		die "Missing image or dockerfile path variable"
+	fi
+
+	# Verify that dockerfile version is equal to test version
+	check_image=$(docker images "$image" -q)
```
wainersm, line 298: Hi @GabyCT , I know nothing about the metrics CI but let me try to help with review of this...

`docker images "$image" -q` can return more than one hash if the "$image" hasn't a specific tag. For example,

`
[wmoschet@wainer-laptop kubernetes]$ docker images "docker.io/library/ubuntu"
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
REPOSITORY                TAG         IMAGE ID      CREATED        SIZE
docker.io/library/ubuntu  18.04       2c047404e52d  11 months ago  65.6 MB
docker.io/library/ubuntu  20.04       d70eaf7277ea  12 months ago  75.3 MB
`

lib/common.bash
```diff
@@ -281,3 +281,36 @@ restart_docker_service() {
...
+	if [ -n "$check_image" ]; then
+		# Check image label
+		check_image_version=$(docker image inspect $image | grep -w DOCKERFILE_VERSION | head -1 | cut -d '"' -f4)
+		if [ -n "$check_image_version" ]; then
```
wainersm, line 302: Shouldn't it be `-z` (`if [ -z "$check_image_version" ]; then`) instead of `-n`?

IIUC if `$check_image_version` is empty then the image hasn't got `DOCKERFILE_VERSION`, thus it should be built from dockerfile.

lib/common.bash
```diff
@@ -281,3 +281,36 @@ restart_docker_service() {
...
+			echo "$image is not updated"
+			build_dockerfile_image "$image" "$dockerfile_path"
+		else
+			# Check dockerfile label
+			dockerfile_version=$(grep DOCKERFILE_VERSION $dockerfile_path | cut -d '"' -f2)
```
wainersm, line 307: maybe grep for `DOCKERFILE_VERSION=` just to be sure :)

wainersm: actually maybe make it like `grep DOCKERFILE_VERSION= $dockerfile_path | cut -d '=' -f2 | tr -d \"` so it can deal with double-quoted (or not) version. 

```
[wmoschet@wainer-laptop tests]$ dockerfile_path="metrics/network/iperf3_dockerfile/Dockerfile"; grep DOCKERFILE_VERSION= $dockerfile_path | cut -d '=' -f2 | tr -d \"
1.0
[wmoschet@wainer-laptop tests]$ echo "DOCKERFILE_VERSION=1.0" | grep DOCKERFILE_VERSION= | cut -d '=' -f2 | tr -d \"
1.0

```

lib/common.bash
GabyCT: @wainersm yes it retrieves several images but then it looks for a particular image with a specific tag

lib/common.bash
GabyCT: @wainersm that is what it does if the image is not there , then it builds the image

lib/common.bash
GabyCT: @wainersm changes applied

lib/common.bash
```diff
@@ -281,3 +281,36 @@ restart_docker_service() {
 
 	return 0
 }
+
+# This function verifies that the dockerfile version is
+# equal to the test version in order to build the image or
+# just run the test
+check_dockerfiles_images()
+{
+	local image="$1"
```
jodh-intel, line 290: Might be best to set this to default to handle the case where this script is sourced and the calling script has `set -o nounset`:

```suggestion
	local image="${1:-}"
```

Same comment for the variable below.

lib/common.bash
jodh-intel, line 298: Missing `local` here and below?

GabyCT:
@jodh-intel changes applied

jodh-intel (approved):
Thanks @GabyCT.

lgtm

devimc (approved):

PR #4140, opened 2021-11-02
Selective enablement of tests

littlejawa:
- Add some code to the test script to selectively enable tests based on the version of cri-o being tested
- Enable several tests that we were skipping, for specific versions of cri-o
- Also enable a test fixed in kata - as we are testing against "main", there is no need to check the version here

### DIFF
--- a/integration/cri-o/cri-o.sh
+++ b/integration/cri-o/cri-o.sh
@@ -74,6 +74,24 @@ do
 	sed -i '/'${testName}'/a skip \"'${skipCRIOTests[$testName]}'\"' "ctr_kata_integration_tests.bats"
 done
 
+# selectively skip tests depending on the version of cri-o we're testing with
+CRIO_VERSION=$(git status | head -n1 | awk '{print $NF}' | cut -d'-' -f 2)
+echo GOT CRIO VERSION $CRIO_VERSION
+if [ "$CRIO_VERSION" != "main" ]; then
+    if [ -z "$CRIO_VERSION" ]; then
+        echo "Unknown version of cri-o - skipping more tests"
+        CRIO_VERSION="1.21"
+    fi
+
+    for testName in "${!fixedInCrioVersion[@]}"
+    do
+        if [ "$(echo -e "$CRIO_VERSION\n${fixedInCrioVersion[$testName]}" | sort -V | head -n1)" != "${fixedInCrioVersion[$testName]}" ]; then
+            echo "Skipping $testName for cri-o v$CRIO_VERSION"
+            sed -i '/'${testName}'/a skip \"- fixed in cri-o v'${fixedInCrioVersion[$testName]}'\"' "ctr_kata_integration_tests.bats"
+        fi
+    done
+fi
+
 IFS=$OLD_IFS
 
 echo "Ensure crio service is stopped before running the tests"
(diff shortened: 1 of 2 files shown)


littlejawa:
I have a question that is not totally related to this PR, but more on the general git CI workflow.
When I make a change like this one, tests are being run, but as far as I can tell none of them is actually testing the thing that I modified (our test script is actually triggered only from cri-o PRs).
So it means that, even if the tests are successful, we have no way to tell whether I introduced a regression, right?

I'm not sure we want to run this script on every PR for the kata-containers/tests repository, but is there a way to check for regression at least when we modify something in that area?

fidencio:
> I have a question that is not totally related to this PR, but more on the general git CI workflow. When I make a change like this one, tests are being run, but as far as I can tell none of them is actually testing the thing that I modified (our test script is actually triggered only from cri-o PRs). So it means that, even if the tests are successful, we have no way to tell whether I introduced a regression, right?

You're totally right.

> I'm not sure we want to run this script on every PR for the kata-containers/tests repository, but is there a way to check for regression at least when we modify something in that area?

One way to do that is introducing one specific job that would do something similar than what CRI-O does, running this PR against the `main` CRI-O branch.  In order to do so, we'd need to clone one specific CI, change its action call, and call it whenever doing such changes.

In case you need help to get this set up, please, do not hesitate on contacting me and I'd be happy to assist this process.

integration/cri-o/cri-o.sh
```diff
@@ -74,6 +75,24 @@ do
 	sed -i '/'${testName}'/a skip \"'${skipCRIOTests[$testName]}'\"' "ctr_kata_integration_tests.bats"
 done
 
+# selectively skip tests depending on the version of cri-o we're testing with
+CRIO_VERSION=$(git status | head -n1 | awk '{print $NF}' | cut -d'-' -f 2)
+echo GOT CRIO VERSION $CRIO_VERSION
+if [ "$CRIO_VERSION" != "main" ]; then
+    if [ -z "$CRIO_VERSION" ]; then
+        echo "Unknown version of cri-o - skipping more tests"
+        CRIO_VERSION="1.21"
```
fidencio, line 84: I'd be more comfortable setting `CRIO_VERSION` to the latest release of the project.  In this case, 1.22.0 (see https://github.com/cri-o/cri-o/releases/tag/v1.22.0).

fidencio (commented):
I like a lot the approach taken, @littlejawa!
There's one suggestion I made about the CRI-O version we should fallback to, and once that gets fixed and we ensure it doesn't break CRI-O CI this one will be good to go.

Thanks!

integration/cri-o/cri-o.sh
littlejawa: Well, this is not the version we fall back to for running the test, this is the version we use to decide what test we enable or not.
1.21 is the first version we started testing with. And it is a version that we know have none of our fixes.

Using this number by default means we will skip all "unfixed" tests. If we're not sure what version of cri-o we're testing with, this is the safest. 

If we use "1.22" for instance, we will enable 3 more tests that are not fixed in 1.21, which means 3 failures if we're testing with a version of cri-o < 1.22

fidencio: Ack!  Thanks for the explanation.

fidencio (approved):

jodh-intel:
@littlejawa - *Quis custodiet ipsos custodes?* :smile: We try to mitigate that risk, where possible, by having tests for the tests. For example, see:

- https://github.com/kata-containers/tests/blob/main/.ci/ci-fast-return.sh#L390..L599

You might be able to find further opportunities to add more "test tests" if you look around though ;)

jodh-intel (approved):
Thanks @littlejawa.

lgtm

Issue #4141, opened 2021-11-02
Enable data race detection on ARM and Power


Jakob-Naucke:
It may not help with the s380x issue _directly_, but I did notice looking at that script that the `-race` logic is no longer correct:

- https://github.com/kata-containers/tests/blob/main/.ci/go-test.sh#L25..L27

`go help build` now shows:

```
-race
        enable data race detection.
        Supported only on linux/amd64, freebsd/amd64, darwin/amd64, windows/amd64,
        linux/ppc64le and linux/arm64 (only for 48-bit VMA).
```

... so it's probably worth inverting the logic so we add `-race` unless we're running on s390x, and hopefully that might give more clues as to the issue (which I doubt is actually architecture-specific).

_Originally posted by @jodh-intel in https://github.com/kata-containers/tests/issues/4131#issuecomment-957320902_

PR #4142, opened 2021-11-02
ci: Enable data race detection on ARM and Power

Jakob-Naucke:
Run `go test` with `-race` on architectures that support this by now.
This might also help uncover the root cause of #4131, even though that
has only been observed on s390x thus far.

Fixes: #4141

/cc @jodh-intel


fidencio (approved):
lgtm, thanks @Jakob-Naucke!

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

Issue #4165, opened 2021-11-09
cri-o: Remove what seems to be the address of an internal registry from Red Hat


fidencio:
@jodh-intel noticed a weird registry on https://github.com/kata-containers/tests/blob/main/.ci/install_crio.sh#L41 and pointed that out at my direction, as I'm the one who introduced that.

It seems that the `registry-proxy.engineering.redhat.com` is indeed an internal Red Hat registry and that was a wrong copy & paste from the developer's side.

PR #4166, opened 2021-11-09
crio: Remove internal Red Hat registry

fidencio:
This was mistakenly added as part of
66eea16d6301d7a9710e236bfa7834792c5f9de0.

Fixes: #4165

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -38,7 +38,7 @@ popd
 echo "Configure registries"
 sudo mkdir -p /etc/containers/registries.conf.d/
 cat <<EOF| sudo tee "/etc/containers/registries.conf.d/ciregistries.conf"
-unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.redhat.com", "registry.centos.org", "docker.io", "registry-proxy.engineering.redhat.com"]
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @fidencio.

lgtm

wainersm (approved):
Hi @fidencio, LGTM.

Orthogonal to this PR:

* could you explain me why we need this `/etc/containers/registries.conf.d/ciregistries.conf` file in the first place? I thought CRI-O install scripts would already configure some suitable registries

* Digging in the git logs for the `.ci/install_crio.sh` script  I saw a comment saying that the file changed of format from one version of cri-o to another. So isn't error prone to drop our own configuration?

Sorry, I know very few about CRI-O :)

GabyCT (approved):

fidencio:
> * Digging in the git logs for the `.ci/install_crio.sh` script  I saw a comment saying that the file changed of format from one version of cri-o to another. So isn't error prone to drop our own configuration?

@wainersm, previously we were simply `sed`ing the registries on crio.conf file, **that** was error prone. On 66eea16d6301d7a9710e236bfa7834792c5f9de0 I went through the recommended way to configure the registries, dropping a registry file to `/etc/containers/registries.conf.d/` with everything we need, and let CRI-O read it from there.

That's the less error-prone approach I can think of, as any version of CRI-O that doesn't support it has already reached its EOL.

Issue #4173, opened 2021-11-10
Move back to using cri-containerd on our tests


fidencio:
On 393e0714667fe45cf9feece5d3f1ce51f53f82fd we moved to using containerd instead of cri-containerd.  However, this turned out to cause issues when using "not too new" versions of containerd, which is the case with cloud providers services such AKS.

@bergwolf reverted the change on kata-containers side (see: https://github.com/kata-containers/kata-containers/issues/2999), let me revert the changes here on the tests side so his PR can land and unblock the release.

PR #3000, opened 2020-10-21
lib/common.bash: Fix overwrite of KATA_TESTS_BASEDIR

wainersm:
To overwrite the KATA_TESTS_BASEDIR variable it should be defined
another of same name, not KATA_TESTS_LOGDIR.

Fixes #2999

Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

PR #4174, opened 2021-11-10
tests: Revert "tests: use containerd to replace cri-containerd"

fidencio:
This reverts commit 393e0714667fe45cf9feece5d3f1ce51f53f82fd.

On the mentioned commit we moved to using containerd instead of
cri-containerd for parsing the crioptions.  However, it turned out to
cause issues when using "not too new enough" versions of containerd,
which happens to be the case on cloud providers services such as AKS.

Tao already proposed to revert the offending commit on the
`kata-containers` repo, but that requires the change reverted on the
tests repo as well, so let's do it.

Fixes: #4173
Depends-on: github.com/kata-containers/kata-containers#3000

### DIFF
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -25,9 +25,10 @@ source "${script_dir}/lib.sh"
 CONTAINERD_OS=$(go env GOOS)
 CONTAIENRD_ARCH=$(go env GOARCH)
 
-containerd_tarball_version=$(get_version "externals.containerd.version")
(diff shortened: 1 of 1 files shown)


liubin:
Hi @fidencio I don't think this commit needs to be reverted. This commit only changed the `contaienrd` config  name in `versions.yaml` for getting `containerd`'s version.

fidencio:
@liubin, you're correct.  I assumed a few other commits had to be reverted as part of #3000, which is **not** the case.
This one can be closed.

Issue #4188, opened 2021-11-15
Building Rust components with *newest* supported Rust version in the CI is unhelpful


dgibson:
# Description of problem

The `versions.yaml` in the main Kata repository has both oldest and newest supported versions for the Rust compiler:

```
  rust:
    description: "Rust language"
    notes: "'version' is the default minimum version used by this project."
    version: "1.45.0"
    meta:
      description: |
        'newest-version' is the latest version known to work when
        building Kata
      newest-version: "1.54.0"
```

The `.ci/install_rust.sh` script installs Rust based on the newest version listed.  AFAICT none of the scripts in either the main or tests repository examine the oldest version.  This isn't particularly useful:
- Rust has a strong commitment to backwards compatibility, any newer Rust version will generally be fine
- If there is a compatibility problem with a new Rust version, that's a bug in Rust - which means that it doesn't tell us anything about later or intervening versions, so "newest tested" doesn't really tell us anything except that that *specific* version has been tested.
- On the other hand, accidentally introducing things into the Kata codebase which require newer Rust features that the oldest one listed is absolutely posslble, and might not get caught by the CI (it may get caught by some of the extra cases which don't use `install_rust.sh`, but it's not tested systematically)

# Expected result

CI should build with the *oldest* supported Rust version, to ensure that all Kata code works with the features in the oldest supported Rust.

# Actual result

CI runs with the "newest known to work" Rust version.

dgibson:
Re-opening, since the original merge had to be reverted (#4192) due to a subtle problem.

dgibson:
@GabyCT you've marked this as completed, but I don't actually see anything indicating the change has been committed.  It *was* committed, but then reverted (in #4192), AFAIK it hasn't been re-applied (if it has been, the PR should be linked to this issue).

PR #4189, opened 2021-11-15
Install the oldest, rather than newest supported Rust by default

dgibson:
For the reasons described in #4188 this makes more sense.

### DIFF
--- a/.ci/install_rust.sh
+++ b/.ci/install_rust.sh
@@ -14,7 +14,7 @@ source "${cidir}/lib.sh"
 rustarch=$(${cidir}/kata-arch.sh --rust)
 version="${1:-""}"
 if [ -z "${version}" ]; then
-	version=$(get_version "languages.rust.meta.newest-version")
+	version=$(get_version "languages.rust.version")
 fi
 
 echo "Install rust ${version}"
--- a/.ci/install_rust.sh
+++ b/.ci/install_rust.sh
@@ -12,8 +12,6 @@ cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
 rustarch=$(${cidir}/kata-arch.sh --rust)
-# release="nightly"
-# recent functional version
 version="${1:-""}"
 if [ -z "${version}" ]; then
 	version=$(get_version "languages.rust.meta.newest-version")


jodh-intel (approved):
Thanks @dgibson

lgtm

/cc @liubin, @bergwolf.

Jakob-Naucke (approved):

fidencio:
All the tests here were using the latest rust version, and the way to go with this PR would have been opening a PR on the `kata-containers` repo, that would depend on this one, and then ensure all the tests would pass there.

This one should be reverted till we ensure `kata-containers` won't be affected by this.

fidencio:
See: https://github.com/kata-containers/tests/pull/4192

Jakob-Naucke:
@fidencio yes, seems "unit tests passing on !x86 in Jenkins" was too optimistic. Sorry!
oops, seems you were faster. Maybe you can still use my #4193 :)
oh, you are.

dgibson:
> All the tests here were using the latest rust version, and the way to go with this PR would have been opening a PR on the `kata-containers` repo, that would depend on this one, and then ensure all the tests would pass there.

If it's broken things, then reverting makes sense.  However, I'm confused by this statement.  Why would the tests here run with a recent Rust version?  The tests should be run from a repo with the tests PR applied, yes?  I don't see why juggling a kata-containers/kata-containers PR would change what's tested.

Do note that this change does *not* affect the Rust used for building within osbuilder - that will still be `languages.rust.meta.newest-version`.  That's a whole different issue, that I'm working on in https://github.com/kata-containers/kata-containers/pull/3043

jodh-intel:
@dgibson - I think @fidencio is saying that if we'd opened a "test PR" in the main repo (with that awkward `Cargo.lock` file) and use the [`Depends-on:` features](https://github.com/kata-containers/tests/blob/main/README.md#breaking-compatibility), when the PR tests ran for that test PR they would be tested with *this* fix, not the main code in the `tests` repo. That would have exposed the issue "in isolation" without affecting all future CI runs in the main repo.

The problem was subtle so anyone could have done this. But not to worry - no harm done! :smile:

dgibson:
> @dgibson - I think @fidencio is saying that if we'd opened a "test PR" in the main repo (with that awkward `Cargo.lock` file) and use the [`Depends-on:` features](https://github.com/kata-containers/tests/blob/main/README.md#breaking-compatibility), when the PR tests ran for that test PR they would be tested with _this_ fix, not the main code in the `tests` repo. That would have exposed the issue "in isolation" without affecting all future CI runs in the main repo.

I still don't follow.  Surely the CI on *this* PR must also run with this fix.  There's no point at all running the CI on tests repo PRs if they don't, you know, exercise the changes in the PR.  I'm trying to get clarity on what a CI run on the main repo would have exercised which the PR on this repo did not.

> The problem was subtle so anyone could have done this. But not to worry - no harm done! smile

It's true I didn't consider the effects on `Cargo.lock`

fidencio:
> > @dgibson - I think @fidencio is saying that if we'd opened a "test PR" in the main repo (with that awkward `Cargo.lock` file) and use the [`Depends-on:` features](https://github.com/kata-containers/tests/blob/main/README.md#breaking-compatibility), when the PR tests ran for that test PR they would be tested with _this_ fix, not the main code in the `tests` repo. That would have exposed the issue "in isolation" without affecting all future CI runs in the main repo.
> 
> I still don't follow. Surely the CI on _this_ PR must also run with this fix. There's no point at all running the CI on tests repo PRs if they don't, you know, exercise the changes in the PR. I'm trying to get clarity on what a CI run on the main repo would have exercised which the PR on this repo did not.

@dgibson, let me take a step back here, as I can see a few problems with the Ci that could be reported (if not yet).
* static-checks running on the `tests` and `kata-containers` repo are **not** doing the same thing.
  * and should they? that's a honest question ... maybe the tests performed there and here should be exactly the same so we can avoid such issues.  This is worth a new issue.
* I'm not sure static-checks (on the `kata-containers` repo) understand the `Depends-on` tag
  * @snir911, I remember at some point you also faced this ... was there any issue opened? 
  * **This** is the part that I mentioned that opening a PR against the `kata-containers` branch would have caught this.  But I think I am mistaken and it would not, as I think the `Depends-on` tag is not used / checked as part of the static-checks.
* can we ensure the version set is actually being used?
  * for this one it clearly seems a bug somewhere in the scripts, but keeping an eye that the changes related to the PR are actually doing what they're expected to do is also a good practice (for everyone involved, not only the submitter).

Anyways, then I installed the minimum version on my machine and locally tried to build the agent, I saw a bunch of failures.  Have you tried to do that?  Was the build clean for you?
```
error[E0658]: the `#[track_caller]` attribute is an experimental feature
   --> /home/ffidenci/.cargo/registry/src/github.com-1ecc6299db9ec823/rlimit-0.5.4/src/unix/rlim_type.rs:140:1
    |
140 | #[track_caller]
    | ^^^^^^^^^^^^^^^
    |
    = note: see issue #47809 <https://github.com/rust-lang/rust/issues/47809> for more information

error[E0658]: the `#[track_caller]` attribute is an experimental feature
   --> /home/ffidenci/.cargo/registry/src/github.com-1ecc6299db9ec823/rlimit-0.5.4/src/unix/rlim_type.rs:153:1
    |
153 | #[track_caller]
    | ^^^^^^^^^^^^^^^
    |
    = note: see issue #47809 <https://github.com/rust-lang/rust/issues/47809> for more information
```
^^^ those are the issues I faced, and it goes on and on and on.

Then I noticed those were not happening in our CI because the version of rust being used was still the newest one.  Have you check that when you opened the PR?  Of course, CI should just work, but it's worth keeping an eye on the logs to be sure the changes proposed are actually being applied.


Last but not least, if possible, please, go ahead and open some issues so we can track your experience here.

fidencio:
> Do note that this change does not affect the Rust used for building within osbuilder - that will still be languages.rust.meta.newest-version. That's a whole different issue, that I'm working on in kata-containers/kata-containers#3043

Aha!  That's the reason building the agent wouldn't fail, and that's something I missed.

I think the root-cause of this is that https://github.com/kata-containers/kata-containers/blob/e34893a0c470a79a42ca8724d8ea4994ae08520b/.github/workflows/static-checks.yaml#L78-L82 is only running (as a few other checks) as part of the `kata-containers` static checks, and that's exactly where we faced the failure do to this PR.

I think the question we can ask ourselves about this case is: Shall we run **exactly the same** `static-check` tests on **both** repos?

dgibson:
> I think the question we can ask ourselves about this case is: Shall we run **exactly the same** `static-check` tests on **both** repos?

I would say yes.  Of course, I'd also see that as a stop-gap measure before combining the repos together..

PR #4192, opened 2021-11-15
Revert "tests: Install the oldest, rather than the newest supported R…

### LINKED ISSUE #4188: Building Rust components with *newest* supported Rust version in the CI is unhelpful

fidencio:
…ust by default"

This reverts commit 7177a5f0e7ec3b15485a4dd44c7040cfec44b1fa.
```
tests: Install the oldest, rather than the newest supported Rust by default

If not given an explicit version install_rust.sh looks up versions.yaml to
find the newest Rust version that's known to work.  One of the main uses of
this script is in the CI, where installing the newest Rust version isn't
really what we want:
  - Rust has strong backwards compatibility, so breakage due to a newer
    Rust compiler version is very unlikely
  - However, breakage due to adding things to the Kata codebase which
    require new Rust features that aren't in the *oldests* supported version
    is very likely.

Therefore, it's much more useful to test with the oldest supported version
rather than the newest.  Change install_rust.sh to do this.

fixes #4188
```

While the intention is really good (and should be persued), right now
this change causes breaks in our CI.  This means that we should either
update the minimum version or fix the issues to ensure everything works
as expected with rust 1.45.0.

However, while this is not done, let's get `kata-containers` unblocked
by reverting this.

### DIFF
--- a/.ci/install_rust.sh
+++ b/.ci/install_rust.sh
@@ -14,7 +14,7 @@ source "${cidir}/lib.sh"
 rustarch=$(${cidir}/kata-arch.sh --rust)
 version="${1:-""}"
 if [ -z "${version}" ]; then
-	version=$(get_version "languages.rust.version")
+	version=$(get_version "languages.rust.meta.newest-version")
 fi
 
 echo "Install rust ${version}"


fidencio:
/cc @dgibson @jodh-intel @Jakob-Naucke

Also, David, please, don't get me wrong with the revert here.  As mentioned in the commit message I think the change is good and we should pursue it.  We also should have a better way to define what's the minimum & latest versions we are interested in supporting (mainly taking into consideration what downstreams are using when picking the minimum one).

For now, however, kata-containers PRs are being blocked due to this change and reverting https://github.com/kata-containers/tests/commit/7177a5f0e7ec3b15485a4dd44c7040cfec44b1fa is the easier / faster approach to take.

fidencio:
ARM error seems like a flaky / instability on the docker / ubuntu side:
```
17:37:15 Head "https://registry-1.docker.io/v2/library/ubuntu/manifests/20.04": unknown: unable to complete public repo access query from database: context deadline exceeded
17:37:16 ++ error_handler
17:37:16 ++ '[' 1 -eq 0 ']'
17:37:16 ++ '[' -n '' ']'
17:37:16 make: *** [Makefile:93: /home/jenkins/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/tools/osbuilder/.ubuntu_rootfs.done] Error 1
17:37:16 [install_kata_image.sh:50] ERROR: sudo -E USE_DOCKER=1 DISTRO=ubuntu make -e image
17:37:16 [install_kata.sh:22] ERROR: .ci/install_kata_image.sh 
17:37:16 [jenkins_job_build.sh:172] ERROR: .ci/setup.sh
17:37:16 Build step 'Execute shell' marked build as failure
```

jodh-intel (approved):
Thanks @fidencio.

lgtm

jodh-intel:
Related:

- https://github.com/kata-containers/kata-containers/pull/3046#issuecomment-968952093
- https://github.com/kata-containers/tests/pull/4189

dgibson:
> /cc @dgibson @jodh-intel @Jakob-Naucke
> 
> Also, David, please, don't get me wrong with the revert here. As mentioned in the commit message I think the change is good and we should go for it. We also should have a better way to define what's the minimum & latest versions we are interested in supporting (mainly taking into consideration what downstreams are using when picking the minimum one).
> 
> For now, however, kata-containers PRs are being blocked due to this change and reverting [7177a5f](https://github.com/kata-containers/tests/commit/7177a5f0e7ec3b15485a4dd44c7040cfec44b1fa) is the easier / faster approach to take.

That's entirely reasonable, however I would really like to know *how* my PR broke the CI - it's not clear to me.

jodh-intel:
Hi @dgibson - fwics the problem was that a PR updated the `Cargo.lock` file to the new "version 3" format. However, when we reverted the rust lang version in the CI, that lock file got regenerated as part of the build *without* the `version = 3` bit since the CI version of `cargo` didn't support it. However, since the lock file is under git control the static checker detected that the branch had uncommitted changes and was thus "dirty", which caused the CI to fail.

dgibson:
> Hi @dgibson - fwics the problem was that a PR updated the `Cargo.lock` file to the new "version 3" format. However, when we reverted the rust lang version in the CI, that lock file got regenerated as part of the build _without_ the `version = 3` bit since the CI version of `cargo` didn't support it. However, since the lock file is under git control the static checker detected that the branch had uncommitted changes and was thus "dirty", which caused the CI to fail.

So.. what I'm getting from this is that the tests repo doesn't run an identical suite of tests to the main repo?  If so, that seems broken on the face of it.  @fidencio, am I misunderstanding something?

fidencio:
> > Hi @dgibson - fwics the problem was that a PR updated the `Cargo.lock` file to the new "version 3" format. However, when we reverted the rust lang version in the CI, that lock file got regenerated as part of the build _without_ the `version = 3` bit since the CI version of `cargo` didn't support it. However, since the lock file is under git control the static checker detected that the branch had uncommitted changes and was thus "dirty", which caused the CI to fail.
> 
> So.. what I'm getting from this is that the tests repo doesn't run an identical suite of tests to the main repo? If so, that seems broken on the face of it. @fidencio, am I misunderstanding something?

Please,see: https://github.com/kata-containers/tests/pull/4189#issuecomment-971430610
There I mention ...
""''
I think the root-cause of this is that https://github.com/kata-containers/kata-containers/blob/e34893a0c470a79a42ca8724d8ea4994ae08520b/.github/workflows/static-checks.yaml#L78-L82 is only running (as a few other checks) as part of the kata-containers static checks, and that's exactly where we faced the failure do to this PR.

I think the question we can ask ourselves about this case is: Shall we run exactly the same static-check tests on both repos?
"""

There are more comments on https://github.com/kata-containers/tests/pull/4189#issuecomment-971405937, the two most relevant being:
"""
* static-checks running on the tests and kata-containers repo are not doing the same thing.
  * and should they? that's a honest question ... maybe the tests performed there and here should be exactly the same so we can avoid such issues. This is worth a new issue.
* I'm not sure static-checks (on the kata-containers repo) understand the Depends-on tag
  * @snir911, I remember at some point you also faced this ... was there any issue opened?
  * This is the part that I mentioned that opening a PR against the kata-containers branch would have caught this. But I think I am mistaken and it would not, as I think the Depends-on tag is not used / checked as part of the static-checks.
"""

Issue #4198, opened 2021-11-19
switch containerd cri integration tests from containerd/cri to containerd


lifupan:
# Get your issue reviewed faster

To help us understand the problem more quickly, please do the following:

Since the containerd/cri repo had been deprecated, thus we should
switch to integration tests from containerd repo instead of
containerd/cri.

PR #4199, opened 2021-11-19
integration: containerd cri integration tests using containerd repo

lifupan:
Since the containerd/cri repo had been deprecated, thus we should
switch to integration tests from containerd repo instead of
containerd/cri.

Fixes: #4198

### DIFF
--- a/integration/containerd/cri/integration-tests.sh
+++ b/integration/containerd/cri/integration-tests.sh
@@ -12,9 +12,12 @@ set -o errtrace
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/../../../lib/common.bash"
+source "${SCRIPT_PATH}/../../../.ci/lib.sh"
 
 # runc is installed in /usr/local/sbin/ add that path
 export PATH="$PATH:/usr/local/sbin"
+
+containerd_tarball_version=$(get_version "externals.containerd.version")
 
 # Runtime to be used for testing
 RUNTIME=${RUNTIME:-containerd-shim-kata-v2}
@@ -43,8 +46,7 @@ SNAP_CI=${SNAP_CI:-""}
 CI=${CI:-""}
 
 containerd_shim_path="$(command -v containerd-shim)"
-readonly cri_containerd_repo="github.com/containerd/containerd"
-readonly cri_repo="github.com/containerd/cri"
+readonly cri_containerd_repo=$(get_version "externals.containerd.url")
 
 #containerd config file
 readonly tmp_dir=$(mktemp -t -d test-cri-containerd.XXXX)
@@ -181,7 +183,7 @@ check_daemon_setup() {
 		FOCUS="TestImageLoad" \
 		RUNTIME="" \
 		CONTAINERD_CONFIG_FILE="$CONTAINERD_CONFIG_FILE" \
-		make -e test-integration
+		make -e cri-integration
 }
 
 testContainerStart() {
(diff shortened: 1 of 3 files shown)


amshinde (approved):

wllenyj:
It's great, we won't have to deal with the conflict problem of deploying `containerd/cri` and containerd at the same time.

lifupan:
install_cri_containerd.sh is only used for deploy the containerd, which could be used by any other test cases such k8s testcase and so on, and it would download  the ready binary with a backup of deploying from the src code. 

Of course, if install_cri_containerd.sh deployed container from the source code, then the source would be reused by this cri integration test since the go get <containerd git repo> would return directly and do nothing.

lifupan:
> It's great, we won't have to deal with the conflict problem of deploying `containerd/cri` and `containerd` at the same time.
> 
> But we already have an `install_cri_containerd.sh`, can we merge it to one ?

install_cri_containerd.sh is only used for deploy the containerd, which could be used by any other test cases such k8s testcase and so on, and it would download the ready binary with a backup of deploying from the src code.

Of course, if install_cri_containerd.sh deployed container from the source code, then the source would be reused by this cri integration test since the go get would return directly and do nothing.

wllenyj:
> > It's great, we won't have to deal with the conflict problem of deploying `containerd/cri` and `containerd` at the same time.
> > But we already have an `install_cri_containerd.sh`, can we merge it to one ?
> 
> install_cri_containerd.sh is only used for deploy the containerd, which could be used by any other test cases such k8s testcase and so on, and it would download the ready binary with a backup of deploying from the src code.
> 
> Of course, if install_cri_containerd.sh deployed container from the source code, then the source would be reused by this cri integration test since the go get would return directly and do nothing.

make senses， LGTM

wllenyj (approved):

integration/containerd/cri/integration-tests.sh
```diff
@@ -44,7 +47,6 @@ CI=${CI:-""}
 
 containerd_shim_path="$(command -v containerd-shim)"
 readonly cri_containerd_repo="github.com/containerd/containerd"
```
gkurz, line 49: Since this patch brings `.ci/lib.sh` in, this open-coded string could be converted to
 `$(get_version "externals.containerd.url")` like `install_cri_containerd.sh`
 already does.

Appart from that, this is a very nice simplification. Thanks !

gkurz:
I see there are other references to `github.com/containerd/cri` in the repo :
```
$ git grep github.com/containerd/cri
.ci/install_cri_containerd.sh:#https://github.com/containerd/cri/blob/master/docs/installation.md#release-tarball
.ci/lib.sh:     containerd_cri_dir="github.com/containerd/cri"
```

The doc url should be converted to `https://github.com/containerd/containerd/blob/main/docs/cri/installation.md#release-tarball`.

The second belongs to the `delete_containerd_cri_stale_resource()` function. Do we still need it after your changes ?

integration/containerd/cri/integration-tests.sh
lifupan: Hi @gkurz 

Thanks for your comments, updated.

lifupan:
> I see there are other references to `github.com/containerd/cri` in the repo :
> 
> ```
> $ git grep github.com/containerd/cri
> .ci/install_cri_containerd.sh:#https://github.com/containerd/cri/blob/master/docs/installation.md#release-tarball
> .ci/lib.sh:     containerd_cri_dir="github.com/containerd/cri"
> ```
> 
> The doc url should be converted to `https://github.com/containerd/containerd/blob/main/docs/cri/installation.md#release-tarball`.
> 
> The second one belongs to the `delete_containerd_cri_stale_resource()` function. Do we still need it after your changes ?

Updated.

gkurz (approved):
LGTM now. Thanks @lifupan !

@amshinde @wllenyj would you have another look ? Especially the changes in `.ci/lib.sh` that were added after your review. They look correct to me but better safe than sorry :smile:

wllenyj:
LGTM

PR #4216, opened 2021-11-24
k8s: cleanup environment before retry

snir911:
https://github.com/kata-containers/kata-containers/pull/3068#issuecomment-972810503

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -155,7 +155,10 @@ echo "Kubelet options:"
 sudo cat /var/lib/kubelet/kubeadm-flags.env
 sudo systemctl daemon-reload && sudo systemctl restart kubelet
 
-kubectl get nodes
+for i in $(seq 4)
+do
+        kubectl get nodes && break || sleep $i
+done
 kubectl get pods
 
 # default network plugin should be flannel, and its config file is taken from k8s 1.12 documentation
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -133,6 +133,7 @@ sed -i "s|CGROUP_DRIVER|${cgroup_driver}|" "${kubeadm_config_file}"
 trap 'sudo -E sh -c "rm -r "${kubeadm_config_file}""' EXIT
 
 if [ "${CI}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
+	grep -q zram /proc/swaps && echo "# zram swap disabled"  | sudo tee /etc/systemd/zram-generator.conf
 	sudo swapoff -a || true
 fi
 
--- a/integration/kubernetes/e2e_conformance/setup.sh
+++ b/integration/kubernetes/e2e_conformance/setup.sh
@@ -24,7 +24,7 @@ info "Setup env for K8s e2e testing"
 cd "${GOPATH}/src/github.com/kata-containers/tests/integration/kubernetes"
 if ! bash ./init.sh; then
 	info "k8s init failed trying again"
-	sudo systemctl restart "${CRI_RUNTIME}"
+	bash ./cleanup_env.sh
 	sleep "${wait_init_retry}"
 	bash ./init.sh
 fi


GabyCT (approved):

wainersm (commented):
Hi @snir911 , this looks good but could you write a better commit message? :)

snir911:
It looks like [it's not enough](http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-fedora-PR-crio-k8s-e2e-minimal/1043/console) as `kubeadm reset` `Performs a best effort revert of changes made to this host by 'kubeadm initPerforms a best effort revert of changes made to this host by 'kubeadm init`, should we stop the kubelet manually in this case maybe?

wainersm:
> It looks like [it's not enough](http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-fedora-PR-crio-k8s-e2e-minimal/1043/console) as `kubeadm reset` `Performs a best effort revert of changes made to this host by 'kubeadm initPerforms a best effort revert of changes made to this host by 'kubeadm init`, should we stop the kubelet manually in this case maybe?

I do not understand why it is failing on the first place. The baseline job is green for a while now, indicating it can initialize Kubernetes just fine. Also, locally, I've been able to run those e2e minimal tests by running the init / cleanup many times.

snir911:
IDK, TBH I just saw there's a retry there and assumed someone was expecting init.sh to fail at times, maybe @GabyCT know why it fails

GabyCT:
@snir911 not sure why we have the retry there, I noticed in that it seems the environment is not being properly clean
```
00:50:02 Init cluster using /var/run/crio/crio.sock
00:50:02 [init] Using Kubernetes version: v1.22.0
00:50:02 [preflight] Running pre-flight checks
00:50:02 error execution phase preflight: [preflight] Some fatal errors occurred:
00:50:02 	[ERROR Port-6443]: Port 6443 is in use
00:50:02 	[ERROR Port-10259]: Port 10259 is in use
00:50:02 	[ERROR Port-10257]: Port 10257 is in use
00:50:02 	[ERROR Port-2379]: Port 2379 is in use
00:50:02 	[ERROR Port-2380]: Port 2380 is in use
```

snir911:
@GabyCT @wainersm it seems that the non-required jobs were removed, thing is, i'm trying to fix some of those non-required (the k8s-fedora ones) that were unstable. The thing is i cannot test it now as those were removed and i wasn't able to reproduce it locally :P

does it make sense to leave also the non-required just on the tests repo? cc @ariel-adam

GabyCT:
the fedora now are required

integration/kubernetes/init.sh
```diff
@@ -155,6 +155,8 @@ echo "Kubelet options:"
 sudo cat /var/lib/kubelet/kubeadm-flags.env
 sudo systemctl daemon-reload && sudo systemctl restart kubelet
 
+sleep 5
```
GabyCT, line 158: why we need this sleep?

snir911: wip, just want to see if it helps to avoid the `Unable to connect to the server: net/http: TLS handshake timeout`

snir911: @GabyCT this error seems to be pretty common, mostly in single node clusters AFAICT, i couldn't find any solution for this.
In our case the sleep seems to help, i tried several runs and didn't get the error, hence, I suggested simple backoff sleep in the recent push

integration/kubernetes/init.sh
```diff
@@ -155,7 +156,10 @@ echo "Kubelet options:"
 sudo cat /var/lib/kubelet/kubeadm-flags.env
 sudo systemctl daemon-reload && sudo systemctl restart kubelet
 
-kubectl get nodes
+for i in 1 3 7
```
jodh-intel, line 159: ooi, why this set of numbers? Might be clearer to say:

```suggestion
for i in $(seq 3)
```

integration/kubernetes/init.sh
```diff
@@ -133,6 +133,7 @@ sed -i "s|CGROUP_DRIVER|${cgroup_driver}|" "${kubeadm_config_file}"
 trap 'sudo -E sh -c "rm -r "${kubeadm_config_file}""' EXIT
 
 if [ "${CI}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
+	cat /proc/swaps | grep -q zram && sudo touch /etc/systemd/zram-generator.conf # turn off zram swap if exist
```
jodh-intel, line 136: Nits: This could be simplified to remove the `cat`, but also, although that is the official way to disable zram, wouldn't it be clearer to write a comment to that file. Something like:

```suggestion
	grep -q zram /proc/swaps && echo "# zram swap disabled"  | sudo tee -a /etc/systemd/zram-generator.conf
```

integration/kubernetes/init.sh
snir911: just because i saw the number 10s mentioned somewhere so together it sums (although it might not really needed here)

fidencio (approved):
LGTM considering we get @jodh-intel's comments addressed.
Due to that, I'm adding a `do-not-merge` label so it doesn't get mistakenly merged.

integration/kubernetes/init.sh
snir911: I'll do that, actually i won't even append as if for some reason this file is not empty i guess it won't work

integration/kubernetes/init.sh
```diff
@@ -133,6 +133,7 @@ sed -i "s|CGROUP_DRIVER|${cgroup_driver}|" "${kubeadm_config_file}"
 trap 'sudo -E sh -c "rm -r "${kubeadm_config_file}""' EXIT
 
 if [ "${CI}" == true ] && [[ $(wc -l /proc/swaps | awk '{print $1}') -gt 1 ]]; then
+	grep -q zram /proc/swaps && echo "# zram swap disabled"  | sudo tee /etc/systemd/zram-generator.conf
```
jodh-intel, line 136: Nit: It would be safer to `tee -a` in case that file already exists.

jodh-intel (approved):
Thanks @snir911.

integration/kubernetes/init.sh
snir911: but then it won't disable swap no?

jodh-intel: I thought  if the file existed, swap was disabled? But if it already exists swap will already be disabled - the script doesn't check if the file exists or not. So if it does exist, it might have some other content so it would be safer to append to it.

snir911: It's mentioned that in order to disable `you can override the default config file with an empty one`
By default this file doesn't exist, if you do create it and set some custom zram configurations it will use those, hence, appending a comment to an existing custom configuration won't help disabling zram

(I plan to disable zram in kernel cmdline or by deleting the pkg from the f35 ci images and also update the relevant documentation)

GabyCT (approved):

integration/kubernetes/init.sh
```diff
@@ -155,7 +155,10 @@ echo "Kubelet options:"
 sudo cat /var/lib/kubelet/kubeadm-flags.env
 sudo systemctl daemon-reload && sudo systemctl restart kubelet
 
-kubectl get nodes
+for i in $(seq 4)
```
wainersm, line 159: Nit: the commit's subject for files in integration/kubernetes is usually "k8s" or "integration/kubernetes". The "ci" component is misleading.

IIUC it runs `kubectl get nodes` as a mean to check if the kubelet was restarted correctly (the `sudo systemctl restart kubelet` command).  I wonder if there is a better approach.

Anyway for the current refactored code I see the problems:
 * the loop never fail. You need to check if `i` is `4` then kill the script (because I could not get the pods).
 * it should "sleep" first then check. Otherwise `i==4` won't do any difference as it will falls off the loop afterwards

integration/kubernetes/init.sh
```diff
@@ -155,7 +155,10 @@ echo "Kubelet options:"
...
+do
+        kubectl get nodes && break || sleep $i
```
wainersm, line 161: @jodh-intel while working with Dockerfile delinting I came across an interesting pattern catch by ShellCheck, which is the `A && B || C`. If `A` is true and `B` is false then `C` is executed (in other words, it not always can be used as a replacement of `if A then B else C`). Here is the rule https://github-wiki-see.page/m/koalaman/shellcheck/wiki/SC2015

I see that pattern on my places. Maybe worth mentioning on the shell script advise document?

integration/kubernetes/init.sh
wainersm: @snir911 just noticed it was merged...it will need a follow up PR

integration/kubernetes/init.sh
wainersm: @snir911 does Kubernetes fail if zram swap is on?

integration/kubernetes/init.sh
snir911: Right! kid mistake @wainersm :P 
As for the first one, if getting the nodes will fail i expect also getting the pods afterwards will fail it

Issue #4237, opened 2021-11-30
docs: Moving from EOT to EOF


wzshiming:
Original discussion: https://github.com/kata-containers/kata-containers/issues/2550 https://github.com/kata-containers/tests/pull/3937

PR #3937, opened 2021-09-03
ci: Add static checks for anchor

wzshiming:
ci: Add static checks for anchor 

Only this uses EOT, the others are EOF, uniformly changed to EOF to
avoid confusion

Fixes: #4102

Signed-off-by: Shiming Zhang <wzshiming@foxmail.com>

PR #4238, opened 2021-11-30
docs: Moving from EOT to EOF

wzshiming:
uniformly changed to EOF to avoid confusion

Fixes: #4237

Signed-off-by: Shiming Zhang <wzshiming@foxmail.com>

### DIFF
--- a/.ci/ci-fast-return.sh
+++ b/.ci/ci-fast-return.sh
@@ -261,14 +261,14 @@ can_we_skip() {
 	# If we have any files that *must* be checked, then immediately return
 	if [ -n "$need_checking" ]; then
 		# stderr, so it does not get in our return value...
-		cat >&2 <<-EOT
+		cat >&2 <<-EOF
 		INFO: Cannot fastpath skip CI.
 		INFO: Some files present must be CI checked.
 		INFO: Files to check are:
 
 		$need_checking
 
-EOT
+EOF
 		echo "1"
 		return 0
 	else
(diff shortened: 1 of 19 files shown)


jodh-intel (commented):
Thanks @wzshiming. There are two other files that we should change (`EOC`, `SHELL`):

```bash
$ rg -s "<< *-*[A-Z][A-Z]*\b"|egrep -vi "eot|eof|vendor"
tests/Vagrantfile:  config.vm.provision "shell", env: {"CI_JOB" => job}, inline: <<-SHELL
tests/Vagrantfile:    fedora.vm.provision "shell", reboot: true, inline: <<-SHELL
tests/Vagrantfile:    fedora.vm.provision "shell", inline: <<-SHELL
tests/Vagrantfile:      ubuntu.vm.provision "shell", reboot: true, inline: <<-SHELL
tests/Vagrantfile:    ubuntu.vm.provision "shell", inline: <<-SHELL
tests/integration/kubernetes/k8s-exec.bats:     kubectl exec -i "$pod_name" -- sh <<EOC
tests/integration/kubernetes/k8s-exec.bats:     run bash -c "kubectl exec -i $pod_name -- sh <<EOC
```

I'm assuming we *can* change that Vagrantfile, but @wainersm can probably tell us.

Note the vendor exclude in the command above. There are a few files that "don't confirm" but we need to ensure the static checks don't fail due to vendored code.

wzshiming:
https://github.com/kata-containers/tests/blob/main/Vagrantfile#L68-L97
There may be exceptions here, <<-EOF is nested in <<-SHELL

jodh-intel (approved):
Thanks @wzshiming.

lgtm, but you might need to update the checker to ignore `Vagrantfile`'s (or include `SHELL` as a valid here-document marker (wdyt @wainersm?)

jodh-intel:
Related: https://github.com/kata-containers/tests/pull/3937.

liubin (approved):

Issue #4239, opened 2021-11-30
Add k8s integration test for nested volumes


egernst:
See https://github.com/kata-containers/kata-containers/issues/3140

We should add a integration test that verifies mounting  a secret under a configmap mount

PR #4240, opened 2021-11-30
integration/kubernetes: volume tests for nested configmap and secret

skaegi:
This test makes sure that pods that mount a secret volumes inside of configmap volume will work correctly.

Fixes: #4239

Signed-off-by: Simon Kaegi <simon.kaegi@gmail.com>

### DIFF
--- a/integration/kubernetes/k8s-nested-configmap-secret.bats
+++ b/integration/kubernetes/k8s-nested-configmap-secret.bats
@@ -1,0 +1,40 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2021 IBM Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
+	[ "${KATA_HYPERVISOR}" == "firecracker" ] && skip "test not working see: ${fc_limitations}"
+	
+	export KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
+	get_pod_config_dir
+
+	pod_name="nested-configmap-secret-pod"
+}
+
+@test "Nested mount of a secret volume in a configmap volume for a pod" {
+	# Creates a configmap, secret and pod that mounts the secret inside the configmap
+	kubectl create -f "${pod_config_dir}/pod-nested-configmap-secret.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"
+
+	# Check config/secret value are correct
+	[ "myconfig" == $(kubectl exec $pod_name -- cat /config/config_key) ]
+	[ "mysecret" == $(kubectl exec $pod_name -- cat /config/secret/secret_key) ]
+}
+
+teardown() {
+	[ "${KATA_HYPERVISOR}" == "firecracker" ] && skip "test not working see: ${fc_limitations}"
+
+	# Debugging information
+	kubectl describe "pod/$pod_name"
+
+	# Delete the configmap, secret, and pod used for testing
+	kubectl delete -f "${pod_config_dir}/pod-nested-configmap-secret.yaml"
+}
--- a/integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
@@ -1,0 +1,44 @@
+#
+# Copyright (c) 2021 IBM Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config
+data:
+  config_key: myconfig
+---
+apiVersion: v1
+data:
+  secret_key: bXlzZWNyZXQ= #mysecret
+kind: Secret
+metadata:
+  name: secret
+type: Opaque
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nested-configmap-secret-pod
+spec:
+  terminationGracePeriodSeconds: 0
+  runtimeClassName: kata
+  containers:
+    - name: test-container
+      image: quay.io/prometheus/busybox:latest
+      command: ["tail", "-f", "/dev/null"]
+      volumeMounts:
+        - mountPath: /config
+          name: config
+        - mountPath: /config/secret
+          name: secret
+  volumes:
+    - name: secret
+      secret:
+        secretName: secret
+    - name: config
+      configMap:
+        name: config
+  restartPolicy: Never
--- a/integration/kubernetes/run_kubernetes_tests.sh
+++ b/integration/kubernetes/run_kubernetes_tests.sh
@@ -36,6 +36,7 @@ K8S_TEST_UNION=("k8s-attach-handlers.bats" \
 	"k8s-limit-range.bats" \
 	"k8s-liveness-probes.bats" \
 	"k8s-memory.bats" \
+	"k8s-nested-configmap-secret.bats" \
 	"k8s-number-cpus.bats" \
 	"k8s-oom.bats" \
 	"k8s-optional-empty-configmap.bats" \


skaegi:
To pass I believe this test will need https://github.com/kata-containers/kata-containers/pull/3152 merged first.

egernst:
@skaegi -- you can add a "depends-on" in order to indicate that, and the CI will test against that branch instead of main.

liubin:
@skaegi as egernst said you can use [depends-on](https://github.com/kata-containers/tests/blob/3f2a663a53e1933b86d752d1f7fcd6788435e664/README.md#breaking-compatibility) in your commit message to run tests using a PR that has not merged yet.

And I also suggest two changes:

1. that to verify that the volumes are working, we should extract check the content of config/secret but not only check the exists of the files.

```bash
[ "myconfig" == $(kubectl exec -it nested-config-secret cat /config/config) ]
[ "mysecret" == $(kubectl exec -it nested-config-secret cat /config/secrets/secret) ]
```

2. move configmap/secrets to yaml file, that you can manage their lifetime with only `kubectl create/delete -f` command.

integration/kubernetes/k8s-nested-configmap-secret.bats
```diff
@@ -0,0 +1,43 @@
...
+	kubectl create secret generic "$secret_name" --from-literal=secret=mysecret
+
+	# Create a pod that mounts the secret inside the configmap
+	kubectl create -f "${pod_config_dir}/pod-nested-configmap-secret.yaml"
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"
+
+	# Check config/secret folder exists
+	kubectl exec $pod_name -- sh -c "ls /config/secret"
```
jodh-intel, line 32: Should this be instead?:

```suggestion
	kubectl exec $pod_name -- sh -c "ls /${configmap_name}/${secret_name}"
```

integration/kubernetes/k8s-nested-configmap-secret.bats
```diff
@@ -0,0 +1,43 @@
...
+
+setup() {
+	export KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
+	get_pod_config_dir
+}
+
+@test "Optional and Empty Secret Volume for a pod" {
+	configmap_name="config"
+	secret_name="secret"
+	pod_name="nested-configmap-secret-pod"
```
jodh-intel, line 19: These are globals, so why not put them in `setup()`, or just set them at the top of the file?

integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
```diff
@@ -0,0 +1,29 @@
...
+          name: config
+        - mountPath: /config/secret
+          name: secret
+  volumes:
+    - name: secret
+      secret:
+        secretName: secret
+    - name: config
+      configMap:
+        name: config
```
jodh-intel, line 28: Please could you consider removing the hard-coded values here by using `envsubst(1)`. If you renamed `pod-nested-configmap-secret.yaml` to `pod-nested-configmap-secret.yaml.in`, you could then change `pod-nested-configmap-secret.yaml.in` to be similar to:

```diff
diff --git a/integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml b/integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
index 018fe358..07592e23 100644
--- a/integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
+++ b/integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
@@ -17,13 +17,13 @@ spec:
       volumeMounts:
         - mountPath: /config
           name: config
-        - mountPath: /config/secret
+        - mountPath: /${configmap_name}/${secret_name}
           name: secret
   volumes:
     - name: secret
       secret:
-        secretName: secret
+        secretName: ${secret_name}
     - name: config
       configMap:
-        name: config
+        name: ${configmap_name}
   restartPolicy: Never
```

Then, the bats script can generate the final YAML file like this:

```bash
envsubst \
  < integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml.in \
  > integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
```

/cc @wainersm. We need to put that in the the shell script doc when we write it (https://github.com/kata-containers/kata-containers/issues/3128 :)

jodh-intel (commented):
Thanks for raising @skaegi. A few comments.

skaegi:
Honestly this is just needlessly complex. I'll take @liubin 's advice and make it so we can just `kubectl create/delete -f` and do away with most of the fancy variables.

jodh-intel:
@skaegi - I fail to see how a single extra command in your script is needlessly complex, particularly given the benefits. Further, by using variables in the YAML, it's clearer to the reviewer wtf you are doing as the variables stand out from the surrouding boiler plate noise.

Further comments:

- You've also hard-coded the container image provider which isn't ideal (we've changed this before and we'll probably change it again).
- You've also hard-coded `busybox:latest`. That makes sense most of the time, but not always: `latest` broke recently. If we generated the YAML files, this could be worked around by changing a single config file rather than having to consider an awkward recursive search and replace.

I know: *"it's just a YAML file"*. They are easy to write, but maintaining all these files isn't quite as easy as some might imagine... :smile:

My comment is just a suggestion. None of the other YAML files are generated today afaik and as a result they continue to cause headaches for those that maintain them. Your PR looked like a good candidate to demonstrate a better way of dealing with these files, particularly since doing so would be so easy to rework on this PR. Your call though. Let's see what others think too.

integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
wainersm: +1 for this suggestion. I'd like to just add one thing: you don't necessarily need to create a temporary file (the `integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml` file in the example above), as you can send the output directly to `kubectl`. For example,
```
envsubst \
  < integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml.in | kubectl -f -
```

@jodh-intel I added the use of `envsubst` in the list of advises in  https://github.com/kata-containers/kata-containers/issues/3128

jodh-intel: @wainersm - Nice: simpler is better! ;)

integration/kubernetes/k8s-nested-configmap-secret.bats
```diff
@@ -0,0 +1,43 @@
...
+# Copyright (c) 2021 IBM Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
+	export KUBECONFIG="${KUBECONFIG:-$HOME/.kube/config}"
```
wainersm, line 12: This initialization can be improved on *all tests* as suggested in https://github.com/kata-containers/tests/issues/4242 (commenting here just to advertise the issue :)

integration/kubernetes/runtimeclass_workloads/pod-nested-configmap-secret.yaml
skaegi: If we use `.in` as the file extension we lose some of the editor niceties and validation when it detects the file as yaml. Perhaps using a `.template.yaml` or `.tmpl.yaml` suffix would work better.

In some cases there is value in creating the file so you can do the `create` in the setup and a matching `delete` in the teardown. Let me update to show what I'm thinking.

skaegi: Here's what it looked like -- https://github.com/skaegi/tests/commit/b6aadfef674b7e3478eadd2c45a49c0118555eaa -- it works but it's actually fairly complex. I found you had to `export` all the shell variables and also had to be careful to quote values in your yaml.

integration/kubernetes/k8s-nested-configmap-secret.bats
```diff
@@ -0,0 +1,36 @@
...
+	get_pod_config_dir
+
+	pod_name="nested-configmap-secret-pod"
+
+	# Creates a configmap, secret and pod that mounts the secret inside the configmap
+	kubectl create -f "${pod_config_dir}/pod-nested-configmap-secret.yaml"
```
liubin, line 18: Normally, `setup` is only used to prepare test configs, the creation of pods should be in the test body(`@test`).

liubin:
@skaegi https://github.com/kata-containers/kata-containers/pull/3152 has been merged, if you will update your commit, you can remove the `depends-on` now.

Thanks.

integration/kubernetes/k8s-nested-configmap-secret.bats
skaegi: Fair enough -- I was trying a pattern where setup does the opposite of teardown.

skaegi:
/test
... but I don't think my `/test` does anything -- if someone could help out ;)

skaegi:
I think this is an unrelated flake but not sure...

egernst:
@skaegi actually, its because the test is running with Firecracker VMM.  We want to skip this test if Firecracker.

integration/kubernetes/k8s-nested-configmap-secret.bats
```diff
@@ -0,0 +1,36 @@
...
+#
+# Copyright (c) 2021 IBM Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../.ci/lib.sh"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
```
egernst, line 11:
```suggestion
setup() {
	[ "${KATA_HYPERVISOR}" == "firecracker" ] && skip "test not working see: ${fc_limitations}"

```

integration/kubernetes/k8s-nested-configmap-secret.bats
```diff
@@ -0,0 +1,36 @@
...
+	
+	# Check pod creation
+	kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"
+
+	# Check config/secret value are correct
+	[ "myconfig" == $(kubectl exec $pod_name -- cat /config/config_key) ]
+	[ "mysecret" == $(kubectl exec $pod_name -- cat /config/secret/secret_key) ]
+}
+
+teardown() {
```
egernst, line 30:
```suggestion
teardown() {
	[ "${KATA_HYPERVISOR}" == "firecracker" ] && skip "test not working see: ${fc_limitations}"

```

egernst (changes requested):
Sorry I didn't catch this before -- let's make sure we skip if using FC

egernst (approved):

egernst:
Looks like both Jenkins-ci-fedora-32-crio* tests failed for unrelated reasons/flakes. Restarted each of these. 🤞🏻

liubin (approved):

fidencio:
I am just adding the do-not-merge label to ensure James has the chance to check all his comments got addressed.

jodh-intel (approved):
The YAML changes on this PR aren't making life better for us, but they're no worse that what we already have, so I guess ack.

One final point: @skaegi: it looks slightly odd that you've submitted this with your personal email address but the new files show your company in the copyright. Is this intentional or do you need to rework using your company email address I wonder?

PR #4251, opened 2021-12-03
try to test vm factory

liubin:


liubin:
22:25:20 init vm template
22:25:20 Factory option enable_template requires an initrd image

GabyCT:
we do not have a initrd CI running anymore, the main reason was we did not have a mantainer for that so if these tests need an intird image they will not work on other jobs

PR #4260, opened 2021-12-06
 openshift-ci: adapt scripts to test like OpenShift Sandboxed Containers - respin

wainersm:
This is a respin of [openshift-ci: adapt scripts to test like OpenShift Sandboxed Containers](https://github.com/kata-containers/tests/pull/4164).

When I started working on the new job on OpenShift CI for OCP 4.10 I realized that the `centos:8` image had been updated to ship the EL 8.5 kernel, consequently kata's initrd now includes `4.18.0-348.2.1.el8_5.x86_64` modules. The OCP 4.10 worker nodes will be running with an EL 8.4 kernel, and that mismatch drives the Kata's VM to crash on boot. To solve that problem I added a code on the installer script to simply copy the host's kernel modules over the initrd in case of mismatch. This worked out on my test environment.

An alternative approach would be to use a buildroot which provides the EL 8.4 kernel but I couldn't find a suitable image, hence the solution I propose on that PR.

### DIFF
--- a/.ci/openshift-ci/images/Dockerfile.installer
+++ b/.ci/openshift-ci/images/Dockerfile.installer
@@ -7,7 +7,7 @@
 #
 FROM registry.centos.org/centos:8
 
-RUN yum install -y rsync
+RUN yum install -y rsync dracut
 
 # Load the installation files.
 COPY ./_out ./_out
--- a/.ci/openshift-ci/images/entrypoint.sh
+++ b/.ci/openshift-ci/images/entrypoint.sh
@@ -60,13 +60,48 @@ update_kernel_path()
 	sed -i 's#^kernel = ".*"#kernel = "'${kernel_path}'"#g' "$toml"
 }
 
+# In case the initrd was not created in a build root which kernel version
+# mismatch with the one of the OpenShift worker node, this function will
+# update the initrd accordingly.
+update_initrd_kernel_modules()
+{
+	local initrd_path="${SRC}/opt/kata/share/kata-containers/kata-containers-initrd.img"
+	local moddir="${HOST_MOUNT}/lib/modules"
+	# Modules directory relative to initrd root directory.
+	local initrd_moddir="usr/lib/modules"
+	local kernel=$(ls ${moddir} | head -1)
+	local rootfs_dir="${PWD}/rootfs"
+
+	if lsinitrd "${initrd_path}" | grep -q "${initrd_moddir}/${kernel}"; then
+		echo "initrd contain ${kernel} modules. Nothing to do."
+		return
+	fi
+
+	echo "Going to update ${initrd_path} for kernel ${kernel}"
+	rm -rf "${rootfs_dir}"
+	mkdir -p "${rootfs_dir}"
+
+	cat "${initrd_path}" | cpio --extract --preserve-modification-time \
+		--make-directories --directory="${rootfs_dir}"
+	cp -ar "${moddir}/${kernel}" "${rootfs_dir}/${initrd_moddir}" || true
+
+	# Make a copy of the old initrd just in case debug is needed.
+	cp ${initrd_path} ${initrd_path}.old.$$
+	( cd "${rootfs_dir}" && find . | cpio -H newc -o | gzip -9 ) > ${initrd_path}
+}
+
 if [ "$(id -u)" -ne 0 ]; then
 	echo "ERROR: $0 must be executed by privileged user"
 	terminate
 fi
 
 [ -n "$QEMU_PATH" ] && update_qemu_path
-[ "$USE_HOST_KERNEL" == "yes" ] && update_kernel_path
+
+# Configure to use the host kernel.
+if [ "$USE_HOST_KERNEL" == "yes" ]; then
+	update_kernel_path
+	update_initrd_kernel_modules
+fi
 
 # Some files are copied over /usr which on Red Hat CoreOS (rhcos) is mounted
 # read-only by default. So re-mount it as read-write, otherwise files won't
--- a/.ci/openshift-ci/test.sh
+++ b/.ci/openshift-ci/test.sh
@@ -26,3 +26,4 @@ info "Run test suite: $suite"
 test_status='PASS'
 ${script_dir}/run_${suite}_test.sh || test_status='FAIL'
 info "Test suite: $suite: $test_status"
+[ "$test_status" == "PASS" ]
--- a/versions.yaml
+++ b/versions.yaml
@@ -86,4 +86,4 @@ externals:
 openshift-ci:
   dracut_kernel:
     description: "kernel that dracut should be pulling modules in the initramfs"
-    version: "4.18.0-305.19.1.el8_4.x86_64"
+    version: "4.18.0-348.2.1.el8_5.x86_64"


fidencio:
> This is a respin of [openshift-ci: adapt scripts to test like OpenShift Sandboxed Containers](https://github.com/kata-containers/tests/pull/4164).
> 
> When I started working on the new job on OpenShift CI for OCP 4.10 I realized that the `centos:8` image had been updated to ship the EL 8.5 kernel, consequently kata's initrd now includes `4.18.0-348.2.1.el8_5.x86_64` modules. The OCP 4.10 worker nodes will be running with an EL 8.4 kernel, and that mismatch drives the Kata's VM to crash on boot. To solve that problem I added a code on the installer script to simply copy the host's kernel modules over the initrd in case of mismatch. This worked out on my test environment.
> 
> An alternative approach would be to use a buildroot which provides the EL 8.4 kernel but I couldn't find a suitable image, hence the solution I propose on that PR.

@wainersm, maybe you could stick to https://hub.docker.com/layers/centos/library/centos/8.4.2105/images/sha256-a1801b843b1bfaf77c501e7a6d3f709401a1e0c83863037fa3aab063a7fdb9dc?context=explore ?

snir911:
Following Fabiano's comment would it be possible to use the centos:<version that is pulled from version.yaml> container?

wainersm:
> 

@fidencio I tried to pin to a old version of the image but it is still pulling the EL 8.5 kernel. See below:
```
[wmoschet@xxxxxx openshift-ci]$ podman pull docker.io/centos@sha256:a1801b843b1bfaf77c501e7a6d3f709401a1e0c83863037fa3aab063a7fdb9dc
Trying to pull docker.io/library/centos@sha256:a1801b843b1bfaf77c501e7a6d3f709401a1e0c83863037fa3aab063a7fdb9dc...
Getting image source signatures
Copying blob a1d0c7532777 [--------------------------------------] 0.0b / 0.0b
Copying config 5d0da3dc97 done  
Writing manifest to image destination
Storing signatures
5d0da3dc976460b72c77d94c8a1ad043720b0416bfc16c52c45d4847e53fadb6
[wmoschet@xxxxxx openshift-ci]$ podman run --rm docker.io/library/centos@sha256:a1801b843b1bfaf77c501e7a6d3f709401a1e0c83863037fa3aab063a7fdb9dc bash -c 'dnf list --showduplicates kernel'
CentOS Linux 8 - AppStream                      393 kB/s | 8.2 MB     00:21    
CentOS Linux 8 - BaseOS                         559 kB/s | 3.5 MB     00:06    
CentOS Linux 8 - Extras                          12 kB/s |  10 kB     00:00    
Available Packages
kernel.x86_64                    4.18.0-348.el8                           baseos
kernel.x86_64                    4.18.0-348.2.1.el8_5                     baseos
```

It seems the repo is configured to follow CentOS 8 Stream:
```
[root@69948cf6ca5b /]# cat /etc/yum.repos.d/CentOS-Linux-BaseOS.repo 
# CentOS-Linux-BaseOS.repo
#
# The mirrorlist system uses the connecting IP address of the client and the
# update status of each mirror to pick current mirrors that are geographically
# close to the client.  You should use this for CentOS updates unless you are
# manually picking other mirrors.
#
# If the mirrorlist does not work for you, you can try the commented out
# baseurl line instead.

[baseos]
name=CentOS Linux $releasever - BaseOS
mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=BaseOS&infra=$infra
#baseurl=http://mirror.centos.org/$contentdir/$releasever/BaseOS/$basearch/os/
gpgcheck=1
enabled=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
```

An alternative seems to be using the archived repository: https://vault.centos.org/8.4.2105/BaseOS/x86_64/os/

Let me do some tests....I will hold this PR.

Thanks for the insight @fidencio !

wainersm:
@fidencio @snir911 okay, I did some experiments yesterday and it seems I managed to install the kernel from EL 8.4. However I'm still trying to figure out how to use a image other than `centos:8` on the openshift-ci as the framework keeps a list of restricted images which can be used.  Hence, I might still need this PR landing....let's see, If the case I let you know.

Regarding the use of `centos:8.4.2105`, below is the dockerfile which I used. 

```
FROM quay.io/centos/centos@sha256:96a58892e20a62e264ccf432da7007871a427103804d82cb5029df5499f1d82f

ENV GOPATH /go

# We need to use the archived in vault.centos.org repositories to pick up the
# packages from EL 8.4. Note that we cannot use `yum configure-manager --add-repo`
# to simply add the repositories because the tests/.ci/setup_env_centos.sh script
# make assumptions about the names of repositories, thus here it is instead
# replaced the urls.
RUN for repo in BaseOS AppStream PowerTools; do sed -i -e 's%^#baseurl=.*%baseurl=https://vault.centos.org/8.4.2105/'${repo}'/$basearch/os/%' -e 's%^mirrorlist=.*%%' "/etc/yum.repos.d/CentOS-Linux-${repo}.repo"; done && \
    sed -i -e 's%^#baseurl=.*%baseurl=https://vault.centos.org/8.4.2105/extras/$basearch/os/%' -e 's%^mirrorlist=.*%%' "/etc/yum.repos.d/CentOS-Linux-Extras.repo" && \
    yum -y update && \
    yum -y install \
        git \
        sudo \
        wget && \
    yum clean all
```

fidencio:
@wainersm, one thing that I was wondering is, considering this whole thing runs under the Red Hat infra, using the `ubi8/ubi:8.4` container would be an option, wouldn't it?

wainersm:
> @wainersm, one thing that I was wondering is, considering this whole thing runs under the Red Hat infra, using the `ubi8/ubi:8.4` container would be an option, wouldn't it?

The problem is that ubi doesn't provide kernel packages. An alternative would be to use a developer's subscription to RHEL, but it is getting too complex.

In the end I will need to include the kernel modules at the time of installation because the environment I am running the tests is changing the kernel version very often. Example, the other day `4.18.0-305.19.1.el8_4` was all I needed, today I just checked it is running `4.18.0-305.28.1.el8_4.x86_64`; and the `centos:8.4.2105` image (using `vault.centos.org` repos) gives me up to the `4.18.0-305.25.1.el8_4` kernel. In order words, it's been a fast moving target...

wainersm:
@snir911 @fidencio my conclusion is that I still need the changes of this PR. There are some remediation that I can do, but for now the best to be done is to update the modules on install time. Thus, can I have you ack on this?

fidencio:
@wainersm, I will leave the last approval for Red Hatters, cc @kata-containers/redhat 

A note, testing with 8.5 kernel deviates from the idea of testing with the same environment used for OCP (which is based on 8.4).  So, you may hit issues prematurely (which is good), but you also may not hit real issues.

fidencio (approved):
lgtm, with the caveat that it's deviating from testing the OCP content as the OCP content is.

snir911:
> In the end I will need to include the kernel modules at the time of installation because the environment I am running the tests is changing the kernel version very often. Example, the other day `4.18.0-305.19.1.el8_4` was all I needed, today I just checked it is running `4.18.0-305.28.1.el8_4.x86_64`; and the `centos:8.4.2105` image (using `vault.centos.org` repos) gives me up to the `4.18.0-305.25.1.el8_4` kernel. In order words, it's been a fast moving target...

@wainersm IIUC according to your comment it just a matter of days till it will break again no? I'm fine with having it for now, but it doesn't worth much AFAIU.
BTW pining the ocp env for this job is also not possible?

wainersm:
> @wainersm, I will leave the last approval for Red Hatters, cc @kata-containers/redhat
> 
> A note, testing with 8.5 kernel deviates from the idea of testing with the same environment used for OCP (which is based on 8.4). So, you may hit issues prematurely (which is good), but you also may not hit real issues.

Actually with the initrd updater I am introducing on commit 7e359ac9e2c5e3cee5e7f5773aaae4a5925f2fb2 it will copy the kernel modules from the OCP node at installation time, so it will be using the 8.4 modules in OCP 4.10.

wainersm:
> > In the end I will need to include the kernel modules at the time of installation because the environment I am running the tests is changing the kernel version very often. Example, the other day `4.18.0-305.19.1.el8_4` was all I needed, today I just checked it is running `4.18.0-305.28.1.el8_4.x86_64`; and the `centos:8.4.2105` image (using `vault.centos.org` repos) gives me up to the `4.18.0-305.25.1.el8_4` kernel. In order words, it's been a fast moving target...
> 
> @wainersm IIUC according to your comment it just a matter of days till it will break again no? I'm fine with having it for now, but it doesn't worth much AFAIU. BTW pining the ocp env for this job is also not possible?

As I said above the inird updater will ensure the host's modules are copied into the initrd so it won't break.

The job is pined to OCP 4.10. What I don't know yet is if I can pin to a specific build of OCP 4.10....I'd like to have the job running so I can fine tune those details, and for that I need this PR merged. So can I get you ack @snir911 ? :D

snir911 (approved):
I see, LGTM

GabyCT (approved):

PR #4267, opened 2021-12-07
ci: Remove rhel support for CI scripts

### LINKED ISSUE #4266: ci: Remove rhel support for CI scripts

GabyCT:
This PR removes rhel support as currently for kata 2.0 we do not have
a rhel CI running and this PR makes cleaner the CI scripts by leaving
the current distros that are running on the CI.

Fixes #4266


devimc (approved):

fidencio:
@Jakob-Naucke, are you running any of your tests on RHEL?
Also, cc @wainersm

Jakob-Naucke (changes requested):
Thanks for the heads-up @fidencio. Yes, the only Red Hat-based distro available on the LinuxONE Community Cloud (which we use for s390x CI) is RHEL. Fedora exists for s390x, but isn't available there, and CentOS will only exist from Stream 9 on (ClefOS 7 was a thing as a substitute, but no 8; also I don't think we'll see support on Alma/Rocky).
I had begun some [work](http://jenkins.katacontainers.io/job/kata-containers-2.0-crio-rhel-initrd-s390x-PR/) on this, but wasn't able to conclude over the past weeks yet. However, I still plan to do so.
Therefore, while I think your cleanup work is very noble @GabyCT :), I want to request changes here.
What we could still do to shrink the codebase is use the script we currently call `.ci/setup_env_centos.sh` for both CentOS and RHEL. I can make the necessary changes there in the future, but I cannot guarantee there will be no RHEL alternative paths; the diff doesn't look too scary though. I wouldn't be sure about a "branding-conformant" way to name the script though, not sure if having a "CentOS" script for both is okay.

wainersm:
Red Hat is not running RHEL on CI. But @c3d recently has used RHEL for development, he may still need those scripts. Christophe?

Ok, a more general thought...I think we should have maintainers for the platforms where Kata is tested and supported; just like we did with the CI jobs. If a given platform hasn't a maintainer nor CI jobs testing on it then we should remove it from the repo. I would expand the same idea to the osbuilder.

Cc @kata-containers/architecture-committee

GabyCT:
@Jakob-Naucke the only problem that I have is that I do not have a RHEL system so I can not test the changes

GabyCT:
> Red Hat is not running RHEL on CI. But @c3d recently has used RHEL for development, he may still need those scripts. Christophe?
> 
> Ok, a more general thought...I think we should have maintainers for the platforms where Kata is tested and supported; just like we did with the CI jobs. If a given platform hasn't a maintainer nor CI jobs testing on it then we should remove it from the repo. I would expand the same idea to the osbuilder.
> 
> Cc @kata-containers/architecture-committee

yes in case that @c3d is using the CI scripts we should maintain them if he is installing kata on RHEL on a different way we should remove them

Jakob-Naucke:
> are you using the CI scripts to install kata on that system?

@GabyCT I intend to do so once I re-enable it (maybe even this year, but I can't promise anything).

Issue #4283, opened 2021-12-13
CI's podman is failing to run with crun on fedora 35


snir911:
# Description of problem
CI podman is failing to run on f35 on jenkins with the following error

# Expected result

run as usual 

# Actual result

```
error running container: error from /usr/bin/crun creating container for [/bin/sh -c apt-get -y update &&     apt-get -y upgrade &&     apt-get -y install stress]: sd-bus call: Transport endpoint is not connected
```

PR #4284, opened 2021-12-13
env: install runc before podman

snir911:
so that podman will use runc and not crun, as it seems we
mess some crun configuration earlier

Fixes: #4283

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -56,11 +56,12 @@ install_container_engine() {
 		# Podman is the primary container engine on Fedora-likes
 		# Remove Docker repo to avoid its runc, see https://github.com/containers/podman/issues/8764
 		sudo rm -f /etc/yum.repos.d/docker-ce.repo
+		# if crun is installed, remove it (it will also remove podman if it's the only runtime)
+		command -v crun && sudo dnf remove -y crun
+		# Then install runc from the distribution repo so that podman will use it
+		sudo dnf install -y runc
 		# Try reinstalling to fix CNI configuration, allow erasing incompatible containerd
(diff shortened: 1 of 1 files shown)


snir911:
AFAIU when podman is installed it's trying to look for installed oci runtimes, if non were found it will install crun.
I think the reason some of the jobs appears to install runc is because docker was installed on the image/pre-script (docker depends on runc which causes a podman installation without crun)

Jakob-Naucke (approved):
Slightly late to the party, I had to relive some things... I experimented a little in #4289 and I think when we have migrated to Fedora 35 this can be simplified a little. It looks janky (don't get me wrong, it was janky since I began it), but now I think it's actually minimal for supporting all cgroups and distro versions.

> yes, I added the --nobotest as it has some issues in some CIs that was installing `crun` instead of `runc`, maybe the best thing to do is install `crun` and remove `crun` and `podman` only for the specific CI that is need it to avoid other CIs failure

@GabyCT I think that was me? This was related to some containerd that would get pulled in with the Docker repo (which we should remove from pre-setup IMO but that's another PR). Doesn't seem to be required when installing `runc` first.

snir911:
For the record:
* in Fedora 35 image i removed the docker setup so probably this repo file deletion can be removed (at least for the fedoras, we should see about centos)
* although using runc fixes the issue it would probably be a good idea also to figure out how we are mess crun so that it fails to run (https://github.com/kata-containers/tests/issues/4283). I couldn't reproduce this issue with crun locally, but it consistently happens in Jenkins (i suspect it's something related with the jenkins user is used there, not sure)

Jakob-Naucke:
@snir911 okay, merge when you're ready :)

Issue #4179, opened 2021-11-11
ARM and s390x test cases use an older rustc version


dgibson:
# Description of problem

The ARM and s390x CI tests use Rust 1.47.0, whereas the bulk of the tests use Rust 1.54.0.  Obviously the current Rust code works with the older version, but it prevents us using some useful features that have been introduced more recently.  For example the in proress PR https://github.com/kata-containers/kata-containers/pull/2927 uses features from Rust 1.52.0 (`str::split_once`) and Rust 1.53.0 (`IntoIterator` implementation for arrays).

The patch could be rewritten to avoid the newer features, but since we already have a newer Rust on x86 it seems more sensible to just update the version on ARM and s390x to match.

# Expected result

All CI tests use the same Rust version.

# Actual result

ARM and s390x have older Rust versions than x86.

dgibson:
To expand on this: for x86, we appear to take the Rust version from `languages.rust.meta.newest-version` in the `versions.yaml` file of the kata-containers repository, in the `.ci/install_rust.sh` script.  But for some reason, the ARM and s390x environments must not use the same script and somehow end up with a different version.

Jakob-Naucke:
There are a bunch of scripts in which we run `{kata-containers/,tests/.}ci/install_rust.sh`.

1. `kata-containers/tools/osbuilder/rootfs-builder/rootfs.sh` for building the guest if Rust is not installed. (Most of the time, this is run in a container which has Rust installed through some custom logic in `osbuilder/scripts/lib.sh`, which is also a bit redundant. Probably a separate issue.) 
2. `kata-containers/.github/workflows/static-checks.yaml` which generally runs on freshly provisioned GHA instances.
3. `tests/.ci/jenkins_job_build.sh` for running static checks on non-x86 (whose CI instances are generally not freshly provisioned) **if Rust is not installed**. (This is why those are on 1.47, that's the version that got installed at some point in the past).
4. `tests/tracing/test-agent-shutdown.sh` if Rust is not installed, which is required for x86 Jenkins CI, since Rust would only be installed in a container from (1).

Also, we have `languages.rust.version`, which I thought we enforced as a minimum, so you could just increment that. _This is no longer the case, we only_ used _to do that in osbuilder._ We could decide to not maintain this (and remove that), or introduce checking logic. Again, probably a separate issue.

I listed all those because we should be careful not to screw up a developer's setup, which has some implications on when we can just blindly install `newest-version`.
[...]

Issue #3998, opened 2021-09-22
"Found unexpected /usr/local/bin/containerd-shim-kata-v2 present" in s390x


Jakob-Naucke:
```
ok 2 Hugepages and sandbox cgroup # skip test not working see: https://github.com/kata-containers/runtime/issues/2172
~/workspace/kata-containers-2.0-ubuntu-s390x-PR/go/src/github.com/kata-containers/tests
[reset] Reading configuration from the cluster...
[reset] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -o yaml'
[preflight] Running pre-flight checks
[reset] Removing info for node "kata02" from the ConfigMap "kubeadm-config" in the "kube-system" Namespace
[reset] Stopping the kubelet service
[reset] Unmounting mounted directories in "/var/lib/kubelet"
[reset] Deleting contents of config directories: [/etc/kubernetes/manifests /etc/kubernetes/pki]
[reset] Deleting files: [/etc/kubernetes/admin.conf /etc/kubernetes/kubelet.conf /etc/kubernetes/bootstrap-kubelet.conf /etc/kubernetes/controller-manager.conf /etc/kubernetes/scheduler.conf]
[reset] Deleting contents of stateful directories: [/var/lib/etcd /var/lib/kubelet /var/lib/dockershim /var/run/kubernetes /var/lib/cni]

The reset process does not clean CNI configuration. To do so, you must remove /etc/cni/net.d

The reset process does not reset or clean up iptables rules or IPVS tables.
If you wish to reset iptables, you must do so manually by using the "iptables" command.

If your cluster was setup to utilize IPVS, run ipvsadm --clear (or similar)
to reset your system's IPVS tables.

The reset process does not clean your kubeconfig files and you must remove them manually.
Please, check the contents of the $HOME/.kube/config file.
[cleanup_env.sh:203] INFO: restart docker service
[cleanup_env.sh:90] INFO: remove local registry server: kata-registry
3864409
[cleanup_env.sh:114] ERROR: Found unexpected /usr/local/bin/containerd-shim-kata-v2 present
```

observed e.g. in http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-PR/615/console
testing a `waitForProcess` internally

Jakob-Naucke:
[...]

PR #4288, opened 2021-12-14
ci: Always install newest Rust version

Jakob-Naucke:
In the unit tests and the agent shutdown test, we install Rust iff it
isn't installed. On CI instances that are not reinstated for every job
(e.g. ARM & s390x), this leads to Rust never getting updated (1.47 at
the moment). The Rust installation script already does what we should
do: It installs the required version of Rust unless it is already
installed. Use that without checks.

Fixes: #4179

/cc @dgibson @stevenhorsman

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -174,8 +174,8 @@ popd
 # Run unit tests on non x86_64
 if [ "$arch" != "x86_64" ]; then
 	echo "Running unit tests"
-	cargo_env="$HOME/.cargo/env"
-	[ -e "${cargo_env}" ] || "${ci_dir_name}/install_rust.sh" && source "${cargo_env}"
+	sudo chown -R "$USER" "$HOME/.cargo" || true
+	"$ci_dir_name/install_rust.sh" && source "$HOME/.cargo/env"
 	pushd "${GOPATH}/src/${katacontainers_repo}"
 	echo "Installing libseccomp library from sources"
 	libseccomp_install_dir=$(mktemp -d -t libseccomp.XXXXXXXXXX)
--- a/tracing/test-agent-shutdown.sh
+++ b/tracing/test-agent-shutdown.sh
@@ -765,14 +765,7 @@ setup()
 		exit 0
 	}
 
-	# Ensure rust commands are found
-	local cargo="$HOME/.cargo"
-	[ -d "$cargo" ] && sudo chown -R "${USER}:" \
-		"$cargo" \
-		"$kata_repo_dir"
-
-	local file="${cargo}/env"
-	[ -e "$file" ] || "${SCRIPT_PATH}/../.ci/install_rust.sh" && source "$file"
+	"${SCRIPT_PATH}/../.ci/install_rust.sh" && source "$HOME/.cargo/env"
 
 	trap cleanup EXIT
 


stevenhorsman (approved):
The code LGTM, thanks Jakob! I'll kick of the tests and check the output tomorrow

dgibson (approved):
So, I think this problem arises because of some other broken conditions:

- Having the test scripts manage their own environment is super fragile
- Installing the *latest* rust version, rather than the earliest required is basically asking for people to add things that don't actually work with the supposedly earliest supported version.

Nonetheless this addresses a real problem in the short term, and at least makes the scripts simpler and behaviour more consistent across test cases.  So, I approve.

stevenhorsman:
It looks like the change worked and s390x and ARM jobs do `Install rust 1.54.0`, though both are still failing

Jakob-Naucke:
As for the test failures on ARM & s390x: Those instances have rustup installed into a user home as root. This is something I'd like to eliminate altogether. I'm going to go the admittedly cringeworthy route of running them with a temporary commit that removes rustup. Adding DNM for the time being. /test-arm /test-s390x

> So, I think this problem arises because of some other broken conditions:
> * Having the test scripts manage their own environment is super fragile

@dgibson you mean it would be better to provision environments without the test scripts, and have the test scripts not install anything? I'm sure it would work, it is a major undertaking though.

> * Installing the _latest_ rust version, rather than the earliest required is basically asking for people to add things that don't actually work with the supposedly earliest supported version.

We are already doing this on all other jobs. DYT we should change this to the oldest supported version? (Probably a separate PR if we do this). https://github.com/kata-containers/tests/blob/8c62e42167a923c2e2317c8bba0efee514bd0172/.ci/install_rust.sh#L17

Jakob-Naucke:
Should have pushed _before_... :facepalm: /test-arm /test-s390x

Jakob-Naucke:
Okay, both are past that point. Removing the "cleanup" commit and DNM. Interestingly, I don't need to run most tests because they have already been run on that log. /test-arm /test-s390x

Jakob-Naucke:
Missed a `source` if I'm not mistaken -- /test-s390x

Jakob-Naucke:
forgot to stage, sorry for the noise -- /test-s390x

Jakob-Naucke:
another `source`, /test-s390x

Jakob-Naucke:
s390x:
```
169713
[cleanup_env.sh:129] ERROR: Found unexpected /usr/local/bin/containerd-shim-kata-v2 present
[run_kubernetes_tests.sh:1] ERROR: popd
```

this is #3998, /test

Jakob-Naucke:
ARM:
```
Step 1/2 : FROM quay.io/libpod/ubuntu
received unexpected HTTP status: 504 Gateway Time-out
```

s390x: #3998 again (?)
VFIO:
```
E: Failed to fetch https://packages.cloud.google.com/apt/dists/kubernetes-xenial-unstable/main/binary-amd64/by-hash/SHA256/bb15f9e88b01ede4d1d4701f0ed95125e5535ab248cddbde4784557018ec5c44  404  Not Found [IP: 2607:f8b0:4005:80a::200e 443]
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

/retest-arm /retest-s390x /retest-vfio

Jakob-Naucke:
ARM still has
```
Set environment variables for the libseccomp crate to link the libseccomp library statically
utils.mk:142: "WARNING: aarch64-musl needs extra symbols from libgcc"
make -C src/agent
make[1]: Entering directory '/home/jenkins/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/src/agent'
../../utils.mk:142: "WARNING: aarch64-musl needs extra symbols from libgcc"
/bin/sh: 1: cargo: not found
```

even after all the sourcing, and s390x, which also runs the unit tests in the Jenkins job, does not have this issue. I can't figure out the difference (this still seems to work with `secure_path` set). /cc @jongwu maybe?

jongwu:
> ARM still has
> 
> ```
> Set environment variables for the libseccomp crate to link the libseccomp library statically
> utils.mk:142: "WARNING: aarch64-musl needs extra symbols from libgcc"
> make -C src/agent
> make[1]: Entering directory '/home/jenkins/workspace/kata-containers-2.0-tests-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/src/agent'
> ../../utils.mk:142: "WARNING: aarch64-musl needs extra symbols from libgcc"
> /bin/sh: 1: cargo: not found
> ```
> 
> even after all the sourcing, and s390x, which also runs the unit tests in the Jenkins job, does not have this issue. I can't figure out the difference (this still seems to work with `secure_path` set). /cc @jongwu maybe?

Maybe the ci machine is not clear. I remove the ".cargo" and the ci passes.

Jakob-Naucke:
> Maybe the ci machine is not clear. I remove the ".cargo" and the ci passes.

@jongwu hmmm, okay. Thanks! Just got to fix #3998 now for s390x...

Jakob-Naucke:
Ah. Because there can be `root`-owned `~/.cargo`s left, we should always `chown`. /test

PR #4289, opened 2021-12-15
ci: Simplify Podman installation

Jakob-Naucke:
We introduced a lot of steps for installing Podman, which are not
necessary any more due to packaging changes.

Fixes: #4283

@snir911 this is in response to #4284, I think things can be simplified a bit. It works on my machines, I need to see in CI though.

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -56,11 +56,10 @@ install_container_engine() {
 		# Podman is the primary container engine on Fedora-likes
 		# Remove Docker repo to avoid its runc, see https://github.com/containers/podman/issues/8764
 		sudo rm -f /etc/yum.repos.d/docker-ce.repo
-		# Try reinstalling to fix CNI configuration, allow erasing incompatible containerd
-		sudo dnf reinstall -y podman || sudo dnf install -y --allowerasing podman
-		# runc is not always a dependency, but required for the tests
-		# When it is a dependency, explicitly installing it leads to version conflicts, so use --nobest
-		sudo dnf install -y --nobest runc
+		# Remove crun (cgroups issues) and Podman (reinstall CNI configuration)
+		sudo dnf remove -y crun podman
+		# Use runc instead (also required for some tests), allow erasing and co-existing alongside incompatible containerd
+		sudo dnf install -y --nobest --allowerasing runc podman
 		return
 	fi
 


snir911:
@Jakob-Naucke this is failing because runc is not installed first (actually I'm not sure what exactly messes crun from working)
I didn't remove the docker repo and the reinstalling as i couldn't verify the issues they supposed to solve were fixed, if you're positive about that `dnf install -y runc podman `should work

Jakob-Naucke:
Nope -- on Fedora 32:
```
error running container: error creating container for [/bin/sh -c apt-get update && apt-get --no-install-recommends install -y     apt-utils     autoconf     automake     binutils     build-essential     ca-certificates     chrony     coreutils     curl     debianutils     debootstrap     g++     gcc     git     libc6-dev     libstdc++-8-dev     m4     make     musl     musl-dev     musl-tools     protobuf-compiler     sed     systemd     tar     vim     wget]: container_linux.go:345: starting container process caused "error adding seccomp rule for syscall socket: permission denied"
```

Cannot reproduce on Fedora 35 @snir911.
On CentOS 8:
```
 Problem: package podman-3.3.1-9.module_el8.5.0+988+b1f0b741.x86_64 requires runc >= 1.0.0-57, but none of the providers can be installed
  - package containerd.io-1.3.7-3.1.el8.x86_64 conflicts with runc provided by runc-1.0.2-1.module_el8.5.0+911+f19012f9.x86_64
```

Can only reproduce when a Docker CE repo is still present, which shouldn't happen on a fresh instance?!

Jakob-Naucke:
@snir911 only saw your comment now. Let me try explicit runc. /test-fedora

snir911:
@Jakob-Naucke oh wait, sorry, It's a different issue, I'm not sure that's the same reason (but give it a try we'll see :))
Jenkins pre-script is setting this repo and installing docker, one I'll finish setting the f35 I'll check what is going on with centos and will try to remove that

Jakob-Naucke:
@snir911 no, that's not it. Anyway, what my PR did not take care of is that we (at least temporarily) need a codebase that supports Fedora 32 and 35 simultaneously, which your PR seems to achieve (although I want to play a little with it still). I think the Docker pre-provision should be removed though.

Jakob-Naucke:
Still think there is some room for simplification and I want to know, so /test-fedora /test-centos

PR #4298, opened 2021-12-21
Backport ci: Always install newest Rust version

Jakob-Naucke:
In the unit tests and the agent shutdown test, we install Rust iff it
isn't installed. On CI instances that are not reinstated for every job
(e.g. ARM & s390x), this leads to Rust never getting updated (1.47 at
the moment). The Rust installation script already does what we should
do: It installs the required version of Rust unless it is already
installed. Use that without checks.

Fixes: #4179

Backport of #4288


fidencio (approved):
lgtm, thanks @Jakob-Naucke!

GabyCT (approved):

PR #4306, opened 2021-12-27
ci: exclude the SPDX-License check for some proto files

lifupan:
Since we'd plan to move the protocols crate to upper libs
directory, thus rename the excluded spdx license checking
files.

Fixes: [#3348](https://github.com/kata-containers/kata-containers/issues/3348)


liubin (approved):

fidencio (approved):
lgtm, thanks @lifupan!

jodh-intel (approved):
Thanks @lifupan.

lgtm. I do wonder if we could just change this to `--exclude="*.proto"` though?

lifupan:
> Thanks @lifupan.
> 
> lgtm. I do wonder if we could just change this to `--exclude="*.proto"` though?

Yeah, of course.

Issue #4318, opened 2022-01-05
Possible flake observed for capability integration test


egernst:
Observed:

```
08:16:51 not ok 1 Check capabilities of pod
08:16:56 # (in test file k8s-caps.bats, line 36)
08:16:56 #   `kubectl logs "$pod_name" | grep -q "$expected"' failed
08:16:56 # [bats-exec-test:23] INFO: k8s configured to use runtimeclass
08:16:56 # pod/pod-caps created
08:16:56 # pod/pod-caps condition met
08:16:56 # expected capability mask:
08:16:56 # CapPrm.*..0..5fb$
08:16:56 # observed: 
08:16:56 # CapInh:	00000000a80425fb
08:16:56 # CapPrm:	00000000a80425fb
08:16:56 # CapEff:	00000000a80425fb
08:16:56 # CapBnd:	00000000a80425fb
08:16:56 # CapAmb:	0000000000000000
08:16:56 # I0105 16:16:56.111840   24951 log.go:184] (0xc0007aa000) (0xc0007a6000) Create stream
08:16:56 # I0105 16:16:56.111884   24951 log.go:184] (0xc0007aa000) (0xc0007a6000) Stream added, broadcasting: 1
08:16:56 # I0105 16:16:56.113578   24951 log.go:184] (0xc0007aa000) Reply frame received for 1
08:16:56 # I0105 16:16:56.113611   24951 log.go:184] (0xc0007aa000) (0xc0005a0000) Create stream
08:16:56 # I0105 16:16:56.113619   24951 log.go:184] (0xc0007aa000) (0xc0005a0000) Stream added, broadcasting: 3
08:16:56 # I0105 16:16:56.114146   24951 log.go:184] (0xc0007aa000) Reply frame received for 3
08:16:56 # I0105 16:16:56.114172   24951 log.go:184] (0xc0007aa000) (0xc0007a60a0) Create stream
08:16:56 # I0105 16:16:56.114179   24951 log.go:184] (0xc0007aa000) (0xc0007a60a0) Stream added, broadcasting: 5
08:16:56 # I0105 16:16:56.114678   24951 log.go:184] (0xc0007aa000) Reply frame received for 5
08:16:56 # I0105 16:16:56.383949   24951 log.go:184] (0xc0007aa000) Data frame received for 3
08:16:56 # I0105 16:16:56.383971   24951 log.go:184] (0xc0005a0000) (3) Data frame handling
[...]

PR #4319, opened 2022-01-05
k8s: ensure capability test container logs are available before checking

egernst:
I noticed infrequent failures of the capability check test where the
resulting logs observed during teardown match our expected value.

Let's add a 1 second delay between pod readiness and our check of the running
container logs to ensure  the logs are available.

Fixes: #4318

Signed-off-by: Eric Ernst <eric_ernst@apple.com>

### DIFF
--- a/integration/kubernetes/k8s-caps.bats
+++ b/integration/kubernetes/k8s-caps.bats
@@ -33,7 +33,15 @@ setup() {
         kubectl create -f "${pod_config_dir}/pod-caps.yaml"
         # Check pod creation
         kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"
-        kubectl logs "$pod_name" | grep -q "$expected"
+
+        # Verify expected capabilities for the running container. Add retry to ensure
+        # that the container had time to execute:
+        wait_time=5
+        sleep_time=1
+        cmd="kubectl logs $pod_name | grep -q $expected"
+        waitForProcess "$wait_time" "$sleep_time" "$cmd"
+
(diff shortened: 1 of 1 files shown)


fengwang666 (approved):

integration/kubernetes/k8s-caps.bats
```diff
@@ -33,6 +33,12 @@ setup() {
         kubectl create -f "${pod_config_dir}/pod-caps.yaml"
         # Check pod creation
         kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"
+
+        # Small wait to ensure the container completed execution and has the expected logs
+        # available:
+        sleep 1
```
wainersm, line 39: Hi @egernst , one second might still be not sufficient... and we will need to increase it... what if you use `waitForProcess` with a sleep time of 1 second and wait time of, say, 10 seconds?

egernst: I see.  In my case, I'm not waiting for a particular process to complete, since the container itself will keep on executing after it'll output logs. 

I agree, sprinkling "sleep" in a test is generally horrible. In this case, its a pretty infrequent flake - I was surprised to see the container not start executing immediately once the sandbox is marked ready.

I guess a solution could be to check the logs, and retry after a second if we don't see any of the expected log output?

egernst:
Looks like s390x has been in pain for a while now. @Jakob-Naucke - this expected?

jodh-intel (approved):
Thanks @egernst.

lgtm

wainersm (approved):

PR #4313, opened 2022-01-04
CCv0: Merge main into CCv0 branch 

stevenhorsman:
Merge remote-tracking branch 'upstream/main' into CCv0

Fixes: #4312
Depends-on: github.com/kata-containers/kata-containers#3377

Signed-off-by: stevenhorsman <steven@uk.ibm.com>

PR #4282, opened 2021-12-13
CCv0: Merge main into CCv0 branch

stevenhorsman:
Weekly merge of new commits in `main` into `CCv0`

Fixes: #4281
Signed-off-by: stevenhorsman <steven@uk.ibm.com>

PR #4320, opened 2022-01-06
DO NOT MERGE - Investigate multiple depends-on problem

stevenhorsman:
Investigation into the issues we have with multiple depends-on labels when merging between branches
eg https://github.com/kata-containers/tests/pull/4313#issuecomment-1004865821 https://github.com/kata-containers/tests/pull/4282


liubin:
@stevenhorsman Is using the first depends is safe enough? The first may be the oldest commit.

My suggestion is to ignore all depends, skip the `apply_depends_on` totally, and control it by other conditions other than branch name, for now, `runtime-rs` branch will face the same issue.

For example, we can add a new label named `ignore-depends-on` to PRs, and pass this value to CI jobs by ENVs.

stevenhorsman:
> @stevenhorsman Is using the first depends is safe enough? The first may be the oldest commit.

Hi @liubin. I _think_ the first depends-on should be okay as the depends-on list is created by parsing `git log`, so my assumption was that the newest depends-on would be at the top of the log, but I might have missed something?

> My suggestion is to ignore all depends, skip the `apply_depends_on` totally, and control it by other conditions other than branch name, for now, `runtime-rs` branch will face the same issue.
> 
> For example, we can add a new label named `ignore-depends-on` to PRs, and pass this value to CI jobs by ENVs.

I've been chatting to @fidencio and he had the same suggestion about using a label to skip depends-on, which I think it a good one, but I think we actually need to utilise the depends-on feature when we merge `main` into `CCv0`.

An example of when this is required, so when a change goes into `main` that updates some code and a test. When we create the PR merge those into the `CCv0`, without the tests merge PR depending on the kata-container merge PR we'd have a scenario that the merge tests PR is running against the base `CCv0` without the code change (so testing something that doesn't exist) and will fail.
If the kata-containers change is breaking then we'll have the same problem the other way around too, which results in both PRs being broken and us not being able to merge them.

At the moment my process is to make the tests merge PR depend on the kata-containers merge PR, so then we can get a clean run and merge the tests PR and then run the kata-containers CI which will pick up the merged new test code so then work. (AFAIK you can't have both PRs depend on each other as you'd get into cycles, but I'm happy to try that out).

I hope this makes some senses and if you can spot holes in my logic it would be great to hear.

Thanks!

stevenhorsman:
What I failed to mention/highlight in the scenario I talked about above is that in the case that the kata-containers commits we are merging from main have depends-on labels against a test PR we'd want to use the ignore-depends-on label there, so although I think it's not the sole answer, it is part of is, so I'll start looking at adding in a PoC for this feature to this PR

liubin:
@stevenhorsman  I understood your case.

From another point of view, if I can ensure that both the depends-on pair in kata-containers/tests repo are mereged into the `maini` branch, I can unconditionally use the `ignore_depends_on_label` label without causing any issues, right?

fidencio:
@stevenhorsman, I am closing this one as we've worked this around, right?
Please, feel free to re-open it in case you want to re-visit this problem.

Issue #4334, opened 2022-01-07
`git rebase --rebase-merges` not working on Ubuntu 18.04


stevenhorsman:
As per discussion in https://github.com/kata-containers/tests/pull/4330#issuecomment-1007583285 `git 2.17` which is the version on Ubuntu 18.04 doesn't support the `--rebase-merges` flag, but instead uses the now deprecated `--preserve-merges` to do a similar thing.

PR #4330, opened 2022-01-07
ci: do not drop merge commits when rebasing

fidencio:
As "by default, a rebase will simply drop merge commits from the todo
list, and put the rebased commits into a single, linear branch", let's
ensure we pass `--rebase-merges` to `git rebase`.

Reference: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt--r

Fixes: #4329

Signed-off-by: Fabiano Fidêncio <fabiano.fidencio@intel.com>

PR #4336, opened 2022-01-07
ci: Use different rebase flag for ubuntu 18.04

stevenhorsman:
- Ubuntu 18.04 uses an old version of git (2.17)
- That doesn't support `--rebase-merges`
- Instead we use `--preserve-merges`

Fixes: #4334

### DIFF
--- a/.ci/ci_entry_point.sh
+++ b/.ci/ci_entry_point.sh
@@ -14,6 +14,8 @@ set -o nounset
 set -o pipefail
 set -o errtrace
 
+source /etc/os-release || source /usr/lib/os-release
+
 script_name=${0##*/}
 script_dir=$(dirname "$(readlink -f "$0")")
 CI_JOB="${CI_JOB:-}"
@@ -74,7 +76,11 @@ if [ "${repo_to_test}" == "${tests_repo}" ]; then
 		pr_branch="PR_${pr_number}"
 		git fetch origin "pull/${pr_number}/head:${pr_branch}"
 		git checkout "${pr_branch}"
-		git rebase --rebase-merges "origin/${ghprbTargetBranch}"
+		local rebase_merge_flag="--rebase-merges"
+		if [[ ${ID} == "ubuntu" && ${VERSION_ID} == "18.04" ]]; then
+			rebase_merge_flag="--preserve-merges"
+		fi
+		git rebase "${rebase_merge_flag}" "origin/${ghprbTargetBranch}"
 	fi
 fi
 


fidencio:
/test

Let's re-trigger this one in order to skip the tests.  We need to do that as the CI entry point used is the one from `main`, which is the problematic one on Ubuntu 18.04 machines.

fidencio (approved):
lgtm, thanks @stevenhorsman!

(and sorry for the breakage!)

egernst (approved):

fidencio:
Oh, wow, force-skip-ci happens too late in the process and we'll have to mark all the ubuntu-18-04 CIs as momentarily non-required. (@stevenhorsman, isn't this a deja vu?)

stevenhorsman:
> Oh, wow, force-skip-ci happens too late in the process and we'll have to mark all the ubuntu-18-04 CIs as momentarily non-required. (@stevenhorsman, isn't this a deja vu?)

Haha - that probably also explains why we didn't spot this in the tests of PR #4330 as it was testing the old version

fidencio:
> 

Exactly!

fidencio:
> Haha - that probably might explain why we didn't spot this in the tests of PR #4330 as it was still testing the `main` version of .ci_entry_point.sh as it hadn't updated yet?

Exactly!

jodh-intel:
Thanks for handling this @stevenhorsman! I appreciate I'm rather (too) late to the party, but I think checking `git` itself might be safer as the fix would then work on any distro, not just Ubuntu 18.04:

```bash
have_rebase_merges=$(git rebase --help 2>&1|grep -- --rebase-merges || true)
```

stevenhorsman:
> Thanks for handling this @stevenhorsman! I appreciate I'm rather (too) late to the party, but I think checking `git` itself might be safer as the fix would then work on any distro, not just Ubuntu 18.04:
> 
> ```shell
> have_rebase_merges=$(git rebase --help 2>&1|grep -- --rebase-merges || true)
> ```

@jodh-intel, Hi James, thanks for the feedback. Yes, we were in a bit of a rush on Friday to get the jobs back working. We think that only Ubuntu 18.04 is impacted by the issues, but you are correct that a more encompassing fix would be better. I've created https://github.com/kata-containers/tests/pull/4345 for this change, so please take a look and see if it is what you expected when you have a chance. Thank you!

jodh-intel:
Thanks @stevenhorsman - ack, I understand this was an "emergency fix" :smile:

Issue #4344, opened 2022-01-10
ci: Update `--rebase-merges` to support all distros


stevenhorsman:
As part of https://github.com/kata-containers/tests/pull/4336 we made a fix to support the fact that `git rebase --rebase-merges` wasn't supported in git versions prior to `2.18`, for example `2.17.1` which is used in our Ubuntu 18.04 tests.

We believe that Ubuntu 18.04 is the only platform currently impacted by this issue, and were in a bit of a rush to fix the jobs, but as James points out in https://github.com/kata-containers/tests/pull/4336#issuecomment-1008678246 we could probably do a more complete fix.

PR #4345, opened 2022-01-10
ci: Improve `--rebase-merges` compatibility

stevenhorsman:
- Improve the logic to support distros other than Ubuntu 18.04

Fixes: #4344

### DIFF
--- a/.ci/ci_entry_point.sh
+++ b/.ci/ci_entry_point.sh
@@ -77,7 +77,8 @@ if [ "${repo_to_test}" == "${tests_repo}" ]; then
 		git fetch origin "pull/${pr_number}/head:${pr_branch}"
 		git checkout "${pr_branch}"
 		rebase_merge_flag="--rebase-merges"
-		if [[ ${ID} == "ubuntu" && ${VERSION_ID} == "18.04" ]]; then
+		has_rebase_merges=$(git rebase --help 2>&1|grep -- --rebase-merges || true)
+		if [ -z "${has_rebase_merges}" ]; then
 			rebase_merge_flag="--preserve-merges"
 		fi
 		git rebase "${rebase_merge_flag}" "origin/${ghprbTargetBranch}"


stevenhorsman:
It is necessary to point out that *I think* the code in `.ci_entry_point.sh` runs before the new PR's code is checked out (as it is the bit that does the checking out). So any test runs we do here will actually be running the version that is already in `main` rather than these new changes. I'm not sure if there is another way to test this code other than merge it, but happy to hear other suggestions.

.ci/ci_entry_point.sh
```diff
@@ -77,7 +77,8 @@ if [ "${repo_to_test}" == "${tests_repo}" ]; then
 		git fetch origin "pull/${pr_number}/head:${pr_branch}"
 		git checkout "${pr_branch}"
 		rebase_merge_flag="--rebase-merges"
-		if [[ ${ID} == "ubuntu" && ${VERSION_ID} == "18.04" ]]; then
+		has_rebase_merges=$(git rebase --help 2>&1|grep -- --rebase-merges || true)
+		if [[ -n "${has_rebase_merges}" ]]; then
```
jodh-intel, line 81: - Nit: Only one set of square brackets is required here.
- Don't you mean `[ -z "${has_rebase_merges}" ] && rebase_merge_flag="--preserve-merges"` ?

stevenhorsman: Yes, I definitely meant `-z` rather than `-n`! I'd love to blame Monday morning, but it's not the first time my brain has got those two muddled up! Should be fixed now, thanks for the correction!

jodh-intel (approved):
Thanks @stevenhorsman - lgtm! :smile:

jodh-intel:
One more ack and this PR can land..

jodh-intel:
Could you tal @fidencio ?

fidencio (approved):
lgtm, thanks @stevenhorsman!

(and sorry for the delay getting to this, I've been with my head down in a different issue).

Issue #4346, opened 2022-01-10
ci: ci_crio_entry_point.sh is not workiing for the baseline CI job


littlejawa:
Our baseline CI job[1] is meant to perform the same tests as the PR-triggered job [2], but the entry point we're using expects a PR number that is not used in the baseline.
We need to modify the script to allow using it on both PR and baseline, so that they really work the same.

[1] http://jenkins.katacontainers.io/view/Daily%20baseline/job/crio-kata-containers-2.0-fedora-32-main-baseline/
[2] http://jenkins.katacontainers.io/job/kata-containers-2-crio-PR/

PR #4347, opened 2022-01-10
ci: do not fail if no PR number is given

littlejawa:
The entry point script for cri-o testing should not fail when no PR number is
given. This allow local testing of the script, and make it possible to use
the same script on daily runs (from main/HEAD) and on PR-triggered jobs.

Fixes: #4346

### DIFF
--- a/.ci/ci_crio_entry_point.sh
+++ b/.ci/ci_crio_entry_point.sh
@@ -18,13 +18,12 @@
 # bash ci_crio_entry_point.sh
 
 set -o errexit
-set -o nounset
 set -o pipefail
 set -o errtrace
 
 if [ -z "$PULL_NUMBER" ]; then
-	echo "ERROR: PULL_NUMBER missing"
-	exit 1
+	echo "No pull number given: testing with HEAD"
+	PULL_NUMBER="NONE"
 fi
 
 # set defaults for required variables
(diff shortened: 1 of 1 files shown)


jodh-intel:
I wonder if we should change the script to handle these two scenarios explicitly? The local run can be handled by checking if [dev mode](https://github.com/kata-containers/tests#developer-mode) is enabled and we could / should probably set a new variable for daily GH action runs.

littlejawa:
> I wonder if we should change the script to handle these two scenarios explicitly? The local run can be handled by checking if [dev mode](https://github.com/kata-containers/tests#developer-mode) is enabled and we could / should probably set a new variable for daily GH action runs.

I like the idea of explicitly handling the daily CI case - it is certainly better than making assumptions out of missing environment variables :-) And I'm pretty sure this could be used elsewhere.

For the the dev mode, I'm not sure: we may actually want to test locally with a specific pull request, in which case both PULL_NUMBER and KATA_DEV_MODE would need to be set. We can't assume that "dev mode" means testing with HEAD only.

jodh-intel (approved):
Thanks @littlejawa - yep, on reflection I think this is the best approach for now.

lgtm

/cc @GabyCT for thoughts.

GabyCT (approved):

Issue #4350, opened 2022-01-10
CCv0 | ci: Only use first depends-on tag 


stevenhorsman:
In the weekly merge of `main` into `CCv0`, we frequently have multiple commits with `depends-on` labels in them. When this happens, the default behaviour is to to do:
```
echo "Repository $repo was already defined in a 'Depends-on:' tag."
echo "Only one repository per tag is allowed."
return 1
```
which can make sense for normal development, but not in our bulk branch -> branch merge scenario.

There are a couple of ways we can solve this:
- Ignore all depends-on tags. This is probably part of the solution and I'm planning to work on a label that gives this option soon.
- The common approach I take (as mentioned [here](https://github.com/kata-containers/tests/pull/4320#issuecomment-1007304261)) in the merges is to make the 'tests' merge PR depend on the 'kata-containers' merge PR, so we can be confident we are going to test against the 'next' `CCv0` branch after the merges are done, rather than testing the current one. As such we need to use the depends-on tag from the merge commit, but ignore the rest. The way that depends-on tags are processed is from scanning the git log. This means that the first 'depends-on' tag should be the most recent, and the one from the merge commit, and we can therefore process that and ignore the rest.

stevenhorsman:
Closed with #4351 merging

PR #4351, opened 2022-01-10
CCv0 | ci: Only use first depends-on tag

stevenhorsman:
- Add logic to ignore multiple depends-on for CCv0 PRs, but just use the first one

Fixes: #4350


jodh-intel (approved):
Thanks @stevenhorsman.

lgtm.

Issue #4356, opened 2022-01-11
Update Vagrantfile to use the default distro as the rootfs


fidencio:
Instead of using the same distro as the host, let's just use the default
one.

The reasoning behind this change is because support for some of the
distros used as rootfs has been removed.

Related: kata-containers/kata-containers#3420

PR #4357, opened 2022-01-11
vagrant: Use the default distro with osbuilder

fidencio:
Instead of using the same distro as the host, let's just use the default
one.

The reasoning behind this change is because support for some of the
distros used as rootfs has been removed.

Fixes: #4356
Depends-on: github.com/kata-containers/kata-containers#3420

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -113,8 +113,6 @@ EOF
     fedora.vm.provision "shell", inline: <<-SHELL
       source "#{guest_env_file}"
       cd "${GOPATH}/src/github.com/kata-containers/tests"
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Nice! Thanks @fidencio!

lgtm

devimc (approved):
lgtm - @wainersm ptal

wainersm:
@fidencio thanks for taking care of it!

wainersm (approved):

wainersm:
/test-fedora

The job hit the timeout of 45min...maybe we should increase it.

Issue #4384, opened 2022-01-19
Force skip CI check runs after static checks


Jakob-Naucke:
> Both ARM and s390x CIs are not respecting the `force-skip-ci` label and that's the reason they're failing, and this failure is expected.
> Let me cc @Jakob-Naucke and @jongwu for awareness here, as this is something that should be fixed on their side.
> 
> _Originally posted by @fidencio in https://github.com/kata-containers/kata-containers/issues/3468#issuecomment-1016435100_

We run the force skip CI check _after_ the static check, which is obviously wrong.

PR #4385, opened 2022-01-19
ci: force-skip-ci check before static checks

Jakob-Naucke:
no point in forcing skip if we're still running those

Fixes: #4384
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

PR #4386, opened 2022-01-19
stable-2.3 | ci: force-skip-ci check before static checks

Jakob-Naucke:
no point in forcing skip if we're still running those

Fixes: #4384

Backport of #4385

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -127,21 +127,6 @@ fi
 # Resolve kata dependencies
 "${GOPATH}/src/${tests_repo}/.ci/resolve-kata-dependencies.sh"
 
-# Run the static analysis tools
-if [ "${METRICS_CI}" = "false" ]; then
-	# We run static checks on GitHub Actions for x86_64, 
-	# hence run them on Jenkins for non-x86_64 only.	
-	if [ "$arch" != "x86_64" ]; then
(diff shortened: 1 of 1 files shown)


devimc (approved):

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

bergwolf:
Why do we need `force-skip-ci` in the first place?

fidencio:
> Why do we need `force-skip-ci` in the first place?

We need it in cases we actually have to force skip the CI, and those are quite extreme cases, but every now and then we have those.  The most recent one being moving GoVMM to the kata-containers repo.

On that specific PR, the CI would totally fail as we check for a subsystem being part of a commit message.  As we imported everything from GoVMM **keeping the history**, what they had there wouldn't match with our checks, hence we force skip the CI.

There are a few other situations where this is useful, such as simple fixes to unblock things here and there.

I know, we should have some written guideline on how and when to use those, but that's a different discussion.

Last but not least, if you ask me whether we could solve this just by changing the jobs from `Required` to `Non Required` as part of the GitHub configuration, that's also an option, but that ... IMHO, opens the possibility to have other PRs merged on the timeframe where the jobs are not marked as required (as that would be an action for the entire repo, while the `force-skip-ci` is an action for the current PR only), plus the time it takes to do such changes.

Issue #4404, opened 2022-01-27
http respond with "429" when check URL 


jongwu:
I see lots of check URL errors recently in arm baseline ci:

08:08:08 [static-checks.sh:540] INFO: Checking URL http://www.qemu-project.org
08:08:08 [static-checks.sh:584] ERROR: found HTTP error status codes for URL https://github.com/kata-containers/kata-containers/releases (429)
08:08:42 [static-checks.sh:584] ERROR: found HTTP error status codes for URL https://github.com/kata-containers/kata-containers/tree/main/src/agent (429)
08:08:42 [static-checks.sh:584] ERROR: found HTTP error status codes for URL https://github.com/kata-containers/kata-containers/tree/main/src/runtime/pkg/katatestutils (429)
08:08:42 [static-checks.sh:584] ERROR: found HTTP error status codes for URL https://github.com/kata-containers/kata-containers/releases/tag/2.0.0 (429)
08:08:42 [static-checks.sh:584] ERROR: found HTTP error status codes for URL https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/kata-deploy#kubernetes-quick-start (429)
08:08:42 [static-checks.sh:584] ERROR: found HTTP error status codes for URL https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/kata-deploy (429)

The HTTP 429 Too Many Requests response status code indicates the user has sent too many requests in a given amount of time ("rate limiting").
we should do something with "429", ignore or retry. But I see the retry time in the response header. it's regularly 60s which is so long a time. For now I just have only one idea of skip it.

jodh-intel:
Thanks for raising @jongwu  but this his is a duplicate of #4304, so I'll close this one...

jongwu:
hello @jodh-intel -, I'm not sure if this issue is resolved. The "429" issue always occurs for baseline ci test on arm nowadays. see http://jenkins.katacontainers.io/view/Daily%20baseline/job/kata-containers-2.0-ubuntu-20.04-ARM-main-baseline/145/console

PR #4405, opened 2022-01-27
static-check: ignore http response of 429 when check URL

jongwu:
When check URL, "429" is sometimes returned.
"429 Too Many Requests" response status code indicates the user has
sent too many requests in a given amount of time. We can't predict
when the http server return this and it's not a fatal error. Also the
retry time lies in the "retry-after" of the header of response is too
long. So we can let it go and give a warning instead.

Fixes: #4404


jongwu:
@GabyCT @egernst

Jakob-Naucke:
@jongwu what about #4409?

jongwu:
thanks @Jakob-Naucke , I'll close this.

Issue #4433, opened 2022-02-09
ci: cri-o integration tests should allow testing more than one test file


littlejawa:
The cri-o integration tests are currently focused on running the test file "ctr.bats" from cri-o repository.
There are more tests to run though, and we started running them locally with kata to see which were stable enough to consider inclusion to our CI.

We would like to add some of those tests to the CI, to increase our test coverage. But it requires modifying our test script to check for multiple files from a validated list, and not just ctr.bats.

PR #4434, opened 2022-02-09
ci: make it possible to use multiple test files

littlejawa:
The cri-o integration test script is focusing on ctr.bats only.
This commit makes it possible to define a list of validated test files
that we want to use during testing.

Fixes: #4433

### DIFF
--- a/integration/cri-o/cri-o.sh
+++ b/integration/cri-o/cri-o.sh
@@ -79,21 +79,23 @@ else
 fi
 
 for batsfile in ${bats_files_list[@]}; do
-    testfile=${batsfile%.*}_kata_integration_tests.bats
-    cp ${batsfile} ${testfile}
-    for testName in "${!skipCRIOTests[@]}"
+    testfile=${batsfile}_kata_integration_tests.bats
+    cp ${batsfile}.bats ${testfile}
+    declare -n skip_list=${batsfile}_skipCRIOTests
+    for testName in "${!skip_list[@]}"
     do
         echo "Skipping $testName in $testfile"
-        sed -i '/'${testName}'/a skip \"'${skipCRIOTests[$testName]}'\"' "${testfile}"
+        sed -i '/'${testName}'/a skip \"'${skip_list[$testName]}'\"' "${testfile}"
     done
 
     # selectively skip tests depending on the version of cri-o we're testing with
     if [ "$CRIO_VERSION" != "main" ]; then
-        for testName in "${!fixedInCrioVersion[@]}"
+	declare -n skip_list=${batsfile}_fixedInCrioVersion
+        for testName in "${!skip_list[@]}"
         do
-            if [ "$(echo -e "$CRIO_VERSION\n${fixedInCrioVersion[$testName]}" | sort -V | head -n1)" != "${fixedInCrioVersion[$testName]}" ]; then
+            if [ "$(echo -e "$CRIO_VERSION\n${skip_list[$testName]}" | sort -V | head -n1)" != "${skip_list[$testName]}" ]; then
                 echo "Skipping $testName in $testfile for cri-o v$CRIO_VERSION"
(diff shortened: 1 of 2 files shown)


integration/cri-o/cri-o.sh
```diff
@@ -69,29 +69,35 @@ IFS=''
...
+    CRIO_VERSION="1.21"
+else
+    echo GOT CRIO VERSION $CRIO_VERSION
+fi
 
-    for testName in "${!fixedInCrioVersion[@]}"
+for batsfile in ${bats_files_list[@]}; do
+    testfile=${batsfile%.*}_kata_integration_tests.bats
+    cp ${batsfile} ${testfile}
+    for testName in "${!skipCRIOTests[@]}"
```
fgiudici, line 84: We are preparing for multiple bats file here (yay)!
The code here will try to skip all the skipCRIOTests from each bat file.
While this will work (as if the test is not found nothing bad will happen) maybe we can have separate skipCRIOTests arrays for each bats file (e.g., ctr_skipCRIOTests) and just skip the test for each bats file (so also in crio_skip_tests.sh we will have tests to skip better grouped there, i.e., in different array vars).
We can do that with something  like:
`eval skip_list=\$\{\!${batsfile%.*}_skipCRIOTests\[\@\]\}`
`for testName in $skip_list`

fgiudici (commented):
Overall lgtm!
Nice to have support to multiple bats files :-D 
I would anyway consider keeping different arrays for different bats files, mainly to have a better organization of the skipped tests. WDYT?

littlejawa:
> I would anyway consider keeping different arrays for different bats files, mainly to have a better organization of the skipped tests. WDYT?

My idea was to use comments in the array itself to separate skipped tests.
But I can imagine separating the tests more formally.

I don't want to duplicate code though, so it would require to have the array name built from the bats file's name.
I will have a look.

integration/cri-o/cri-o.sh
```diff
@@ -69,29 +69,35 @@ IFS=''
...
 CRIO_VERSION=$(echo ${PULL_BASE_REF} | cut -d'-' -f 2)
-echo GOT CRIO VERSION $CRIO_VERSION
-if [ "$CRIO_VERSION" != "main" ]; then
-    if [ -z "$CRIO_VERSION" ]; then
-        echo "Unknown version of cri-o - skipping more tests"
-        CRIO_VERSION="1.21"
-    fi
+if [ -z "$CRIO_VERSION" ]; then
+    echo "Unknown version of cri-o - skipping more tests"
+    CRIO_VERSION="1.21"
```
jodh-intel, line 76: Might be worth logging this hard-coded version (by switching the order of these lines) to help with debugging.

integration/cri-o/cri-o.sh
littlejawa: Thanks @fgiudici 
I ended up using an alias (declare -n skip_list=${batsfile}_skipCRIOTests) and using "skip_list" everywhere after.
It required less \ characters :-)
I'm also doing it for the "fixedInCrioVersion" list.
I'll send the commit after testing.

integration/cri-o/cri-o.sh
littlejawa: Thanks - I will add it.
Note that I'm also logging it when editing the test files (the logs will show "skipped for crio vX.Y.Z")

fgiudici (dismissed):
Thanks @littlejawa , lgmt!

littlejawa:
NOTE: I added another commit to fix some bugs I found while testing #4444 
- do not error out when the bats file is missing : this will happen on some branches of cri-o
- do not use '-' in array names : if there is a '-' in the test file name, ignore it

fidencio (approved):
lgtm, thanks @littlejawa!

fgiudici (approved):
lgtm!

PR #4440, opened 2022-02-10
docs: fix vm setup instructions

### LINKED ISSUE #4439: VM setup instructions
# Description of problem

The command in .ci/README.md checking for cgroups2 existence and disabling is wrong

# Expected result

if cgroupsv2 is used set grub to disable it

# Actual result

cgroupv2 is still in use

snir911:
the command for checking cgroups v2 and disable it is wrong

Fixes #4439


.ci/README.md
```diff
@@ -131,7 +131,7 @@ The CI can run on a local VM using the following instructions:
     $ sudo -H -u $USER bash -c 'echo export USE_PODMAN=true >> ~/.bashrc'
     $ sudo -H -u $USER bash -c 'echo cd ~/go/src/github.com/kata-containers/tests >> ~/.bashrc'
     $ sudo -H -u $USER bash -c 'echo source .ci/ci_job_flags.sh >> ~/.bashrc'
-    $ (! mount -t cgroup2 | grep "/sys/fs/cgroup\ ") && sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
+    $ (! mount -t cgroup2 | grep "/sys/fs/cgroup\ ") || sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
```
c3d, line 134: I would have simply removed the `!`

c3d (commented):
Works as is, but I feel like this is a complicated way to express the idea 😄

.ci/README.md
snir911: /o\ will fix

.ci/README.md
```diff
@@ -131,7 +131,7 @@ The CI can run on a local VM using the following instructions:
     $ sudo -H -u $USER bash -c 'echo export USE_PODMAN=true >> ~/.bashrc'
     $ sudo -H -u $USER bash -c 'echo cd ~/go/src/github.com/kata-containers/tests >> ~/.bashrc'
     $ sudo -H -u $USER bash -c 'echo source .ci/ci_job_flags.sh >> ~/.bashrc'
-    $ (! mount -t cgroup2 | grep "/sys/fs/cgroup\ ") && sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
+    $ (mount -t cgroup2 | grep "/sys/fs/cgroup\ ") && sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
```
jodh-intel, line 134: It's not strictly part of this PR, but ideally we'd also remove the match on the separator. Further, the path isn't guaranteed to match _only_ for `/sys/fs/cgroup` (Something like `/foo/sys/fs/cgroup` would also match), so I'm wondering about something like the following, which is robust:

```bash
mount -t cgroup2 | awk '{print $3}' | grep -E "^/sys/fs/cgroup\>"
```

snir911: The idea was more about not to match "/sys/fs/cgroup/unified" because AFAIU when you have both v1 and v2, and v2 is disabled it usually mounted under /unified, and v1 is used is instead under /sys/fs/cgroup.
TBH I don't like to check it like that, but i couldn't find a better way, i'm open for ideas

GabyCT (approved):

.ci/README.md
snir911: @jodh-intel WDYT^?

jodh-intel: How about:

```suggestion
    $ mount -t cgroup2 | grep -q "/sys/fs/cgroup[^/]" && sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
```

Jakob-Naucke (approved):

fgiudici (approved):
lgtm!

wainersm (approved):

fidencio (approved):
lgtm, let's have it merged!

Issue #4447, opened 2022-02-14
Remove RUST_AGENTs


liubin:
There are still some RUST_AGENTs in script files, which could be deleted.

PR #4448, opened 2022-02-14
tests: remove RUST_AGENTs variable

liubin:
From 2.0 there will only one Rust agent,
so the RUST_AGENT is not needed anymore.

Fixes: #4447


jodh-intel (approved):
Thanks @liubin.

lgtm

devimc (approved):

PR #4456, opened 2022-02-14
ci: install podman-docker when using podman

fidencio:
When using podman we need to also install `podman-docker`, as that
provides a compatibility layer which is very much needed as outside
scripts (as in, scripts from our dependencies) may not be aware of
`podman`.

Fixes: #4552
Depends-on: github.com/kata-containers/kata-containers#3664


devimc (approved):
thanks @fidencio

wainersm (approved):

Issue #3983, opened 2021-09-20
CI: add basic testing for kata-monitor


fgiudici:
**Which feature do you think can be improved?**
We currently miss CI tests for the kata-monitor binary

**How can it be improved?**
We could start adding few basics checks for kata-monitor in our CI, to catch regressions early in the PRs.
An environment with just a container engine and kata would be enough. 
We could start pods via crictl, run the kata-monitor binary and scrape its endpoints checking the output matches the expected values.
We have also to identify the CI job where we can add the kata-monitor tests.

fgiudici:
I already prepared a simple script that would start a couple of containers and check that kata-monitor correctly picks them up and exports the metrics in its own endpoints.
Looking for the best suited place to add those few and lightweight tests. Any advice? :-)

wainersm:
Hi @fgiudici , thanks for opening this!

Based on your description and how I see the organization of the tests repository, I think that test should live under the `tests/functional` directory. @GabyCT @devimc any idea?

Regarding the CI jobs, does the kata-monitor contain code paths dependent of the CRI? If it is CRI agnostic then it could be ran on `CRI_CONTAINERD` as that job runs most of the tests suites.

fgiudici:
> Hi @fgiudici , thanks for opening this!
> 
> Based on your description and how I see the organization of the tests repository, I think that test should live under the `tests/functional` directory. @GabyCT @devimc any idea?
> 
> Regarding the CI jobs, does the kata-monitor contain code paths dependent of the CRI? If it is CRI agnostic then it could be ran on `CRI_CONTAINERD` as that job runs most of the tests suites.

Hey, thanks @wainersm... yeah, kata-monitor is container engine agnostic (as long the container engine is CRI compliant). So testing with containerd would be perfect :-)

fidencio:
> Regarding the CI jobs, does the kata-monitor contain code paths dependent of the CRI? If it is CRI agnostic then it could be ran on `CRI_CONTAINERD` as that job runs most of the tests suites.

I am slightly afraid that if it only runs on `CRI_CONTAINERD` we may not have someone from CRI-O side to help maintaining it.

While I'd like to see tests about kata-monitor, I'd like to have an agreement that this should be maintained by both CRI-O & containerd folks.  Thus, maybe, run those on both CRI-O and CONTAINERD CIs, to ensure we don't break integration with one or another.

fgiudici:
[...]

PR #4461, opened 2022-02-15
ci: add basic tests for kata-monitor

fgiudici:
This commit is meant to address #3983: add a new script to allow basic testing of kata-monitor

### DIFF
--- a/functional/kata-monitor/run.sh
+++ b/functional/kata-monitor/run.sh
@@ -1,0 +1,282 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Red Hat
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test file will test kata-monitor for basic functionality (retrieve kata sandboxes)
+# It will assume an environment where:
+# - a CRI container manager (container engine) will be up and running
+# - crictl is installed and configured
+# - the kata-monitor binary is available on the host
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[ -n "${BASH_VERSION:-}" ] && set -o errtrace
+[ -n "${DEBUG:-}" ] && set -o xtrace
+
+readonly MONITOR_HTTP_ENDPOINT="127.0.0.1:8090"
+# we should collect few hundred metrics, let's put a reasonable minimum
+readonly MONITOR_MIN_METRICS_NUM=200
+CRI_RUNTIME=${CRI_RUNTIME:-"crio"}
+RUNTIME=${RUNTIME:-"containerd-shim-kata-v2"}
+KATA_MONITOR_BIN=${KATA_MONITOR_BIN:-$(command -v kata-monitor)}
+IAM=$(whoami)
+TMPATH=$(mktemp -d -t kata-monitor-test-XXXXXXXXX)
+METRICS_FILE="${TMPATH}/metrics.txt"
+MONITOR_LOG_FILE="${TMPATH}/kata-monitor.log"
+CACHE_UPD_TIMEOUT_SEC=${CACHE_UPD_TIMEOUT_SEC:-20}
+POD_ID=""
+CID=""
+RUNC_POD_ID=""
+RUNC_CID=""
+CURRENT_TASK=""
+
+FALSE=1
+TRUE=0
+
+trap error_with_msg ERR
+
+title() {
+	local step="$1"
+	echo -e "\n* STEP: $step"
+}
+
+echo_ok() {
+	local msg="$1"
+
+	echo "OK: $msg"
+}
+
+# quiet crictrl
+qcrictl() {
+	crictl "$@" > /dev/null
+}
+
+# this is just an hash of current date (+ nanoseconds)
+gen_unique_id() {
+	date +%T:%N | md5sum | cut -d ' ' -f 1
+}
+
+error_with_msg() {
+	local msg=${1:-"cannot $CURRENT_TASK"}
+
+	trap - ERR
+	echo -e "\nERROR: $msg"
+	echo -e "\nkata-monitor logs:\n----------------"
+	cat "$MONITOR_LOG_FILE"
+	echo -e "\nkata-monitor testing: FAILED!"
+	cleanup
+	exit 1
+}
+
+cleanup() {
+	stop_workload
+	stop_workload "$RUNC_CID" "$RUNC_POD_ID"
+
+	[ -d "/proc/$KATA_MONITOR_PID" ] \
+		&& kill -9 "$KATA_MONITOR_PID"
+
+	rm -rf "$TMPATH"
+}
+
+create_sandbox_json() {
+	local uid_name_suffix="$(gen_unique_id)"
+	local sbfile="$TMPATH/sandbox-$uid_name_suffix.json"
+
+	cat <<OEM >$sbfile
+{
+	"metadata": {
+		"name": "nginx-$uid_name_suffix",
+		"namespace": "default",
+		"attempt": 1
+	},
+	"logDirectory": "/tmp",
+	"linux": {
+	}
+}
+OEM
+	echo "$sbfile"
+}
+
+create_container_json() {
+	local uid_name_suffix="$(gen_unique_id)"
+	local cntfile="$TMPATH/container-$uid_name_suffix.json"
+
+	cat <<OEM >$cntfile
+{
+	"metadata": {
+		"name": "busybox"
+	},
+	"image":{
+		"image": "busybox"
+	},
+	"command": [
+		"top"
+	],
+	"log_path":"busybox.log",
+	"linux": {
+	}
+}
+OEM
+	echo "$cntfile"
+}
+
+start_workload() {
+	local runtime=${1:-}
+	local args=""
+	local sbfile=""
+	local cntfile=""
+
+	[ -n "$runtime" ] && args="-r $runtime"
+
+	sbfile="$(create_sandbox_json)"
+	cntfile="$(create_container_json)"
+
+	POD_ID=$(crictl runp $args $sbfile)
+	CID=$(crictl create $POD_ID $cntfile $sbfile)
+	qcrictl start $CID
+}
+
+stop_workload() {
+	local cid="${1:-$CID}"
+	local pod_id="${2:-$POD_ID}"
+	local check
+
+	check=$(crictl pods -q -id $pod_id)
+	[ -z "$check" ] && return
+
+	qcrictl stop $cid
+	qcrictl rm $cid
+
+	qcrictl stopp $pod_id
+	qcrictl rmp $pod_id
+}
+
+is_sandbox_there() {
+	local podid=${1}
+	local sbs s
+
+	sbs=$(curl -s ${MONITOR_HTTP_ENDPOINT}/sandboxes)
+	if [ -n "$sbs" ]; then
+		for s in $sbs; do
+			if [ "$s" = "$podid" ]; then
+				return $TRUE
+				break
+			fi
+		done
+	fi
+	return $FALSE
+}
+
+is_sandbox_there_iterate() {
+	local podid=${1}
+
+	for i in $(seq 1 $CACHE_UPD_TIMEOUT_SEC); do
+		is_sandbox_there "$podid" && return $TRUE
+		echo -n "."
+		sleep 1
+		continue
+	done
+
+	return $FALSE
+}
+
+is_sandbox_missing_iterate() {
+	local podid=${1}
+
+	for i in $(seq 1 $CACHE_UPD_TIMEOUT_SEC); do
+		is_sandbox_there "$podid" || return $TRUE
+		echo -n "."
+		sleep 1
+		continue
+	done
+
+	return $FALSE
+}
+
+main() {
+	###########################
+	title "pre-checks"
+
+	[ "$IAM" != "root" ] &&
+		error_with_msg "run this script as root please (user is \"${IAM}\")"
+
+	CURRENT_TASK="connect to the container engine"
+	qcrictl pods
+	echo_ok "$CURRENT_TASK"
+
+	###########################
+	title "create workloads"
+
+	CURRENT_TASK="start workload (runc)"
+	start_workload
+	RUNC_POD_ID="$POD_ID"
+	RUNC_CID="$CID"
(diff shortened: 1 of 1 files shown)


fgiudici:
This commit just adds a script to do some basic testing of kata-monitor.
Looking for some advice on where/how to make good use of it:
the script runs for no more than a minute to perform those basic tests. Should I add a new entry in our Makefile to allow to run the kata-monitor tests only? Wouldn't be the setup time overkill (would need installing kata + a container manager + crictl)? Is there a recommended way to do the basic setup with our scripts?

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+# This test file will test kata-monitor for basic functionality (retrieve kata sandboxes)
+# It will assume an environment where:
+# - a CRI container manager (container engine) will be up and running
+# - crictl is installed and configured
+# - the kata-monitor binary is available on the host
+#
+
+CRI_RUNTIME=${CRI_RUNTIME:-"crio"}
+RUNTIME=${RUNTIME:-"containerd-shim-kata-v2"}
+KATA_MONITOR_BIN=${KATA_MONITOR_BIN:-$(type -p kata-monitor)}
```
jodh-intel, line 16: Unlikely though it is, this would fail if `/bin/sh` was zsh so I'd consider using `command -v kata-monitor` as the output of that command is consistent between shells.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+	crictl stopp "$PODID" "$VANILLA_PODID" > /dev/null 2>&1
+	crictl rmp "$PODID" "$VANILLA_PODID"   > /dev/null 2>&1
+	kill -9 "$KATA_MONITOR_PID" > /dev/null 2>&1
+	rm -rf "$TMPATH"
+}
+
+create_sandbox_json() {
+	local sbfile="$1"
+
+	POD_NAME="$(date | md5sum | cut -d ' ' -f 1)"
```
jodh-intel, line 63: Nits:

- `POD_NAME` isn't `local`.
- consistency: why are some of the locals upper-case and others lower-case? I thought you might be using upper-caps for globals, but it seems not.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+
+CRI_RUNTIME=${CRI_RUNTIME:-"crio"}
+RUNTIME=${RUNTIME:-"containerd-shim-kata-v2"}
+KATA_MONITOR_BIN=${KATA_MONITOR_BIN:-$(type -p kata-monitor)}
+IAM=$(whoami)
+TMPATH=$(mktemp -d -t kata-monitor-test-XXXXXXXXX)
+METRICS_FILE="${TMPATH}/metrics.txt"
+MONITOR_LOG_FILE="${TMPATH}/kata-monitor.log"
+CACHE_UPD_TIMEOUT_SEC=${CACHE_UPD_TIMEOUT_SEC:-10}
+PODID=""
```
jodh-intel, line 22: Nit: `POD_ID` would be clearer.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+CONID=""
```
jodh-intel, line 23: Again, this isn't very clearly named, and in fact I read this as "connection ID", but it seems it's actually a container ID. You could just use `CID` as that's a common acronym.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test file will test kata-monitor for basic functionality (retrieve kata sandboxes)
+# It will assume an environment where:
+# - a CRI container manager (container engine) will be up and running
+# - crictl is installed and configured
+# - the kata-monitor binary is available on the host
+#
+
```
jodh-intel, line 13: I'd consider adding in the "boiler plate" code for consistency and to avoid surprises:

```bash
set -o errexit
set -o nounset
set -o pipefail
 
[ -n "$BASH_VERSION" ] && set -o errtrace

```

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+	crictl stop "$CONID" "$VANILLA_CONID"  > /dev/null 2>&1
+	crictl rm "$CONID" "$VANILLA_CONID"    > /dev/null 2>&1
+	crictl stopp "$PODID" "$VANILLA_PODID" > /dev/null 2>&1
+	crictl rmp "$PODID" "$VANILLA_PODID"   > /dev/null 2>&1
+	kill -9 "$KATA_MONITOR_PID" > /dev/null 2>&1
+	rm -rf "$TMPATH"
+}
+
+create_sandbox_json() {
+	local sbfile="$1"
```
jodh-intel, line 61: Nit: No check on param.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+	echo -e "\nkata-monitor logs:\n----------------"
+	cat "$MONITOR_LOG_FILE"
+	cleanup
+	echo -e "\nkata-monitor testing: FAILED!"
+	exit 1
+}
+
+cleanup() {
+	crictl stop "$CONID" "$VANILLA_CONID"  > /dev/null 2>&1
+	crictl rm "$CONID" "$VANILLA_CONID"    > /dev/null 2>&1
```
jodh-intel, line 53: Nit: It's rather distracting having the redirects on every call. Maybe you could use `chronic` or create a `run_command()` function that handles the redirect?

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+	[ $? -ne 0 ] && error_with_msg "cannot stop pod $PODID"
+	crictl rmp $PODID > /dev/null
+	[ $? -ne 0 ] && error_with_msg "cannot remove pod $PODID"
+}
+
+is_sandbox_there() {
+	local podid=${1}
+	local sbs s
+
+	sbs=$(curl -s 127.0.0.1:8090/sandboxes)
```
jodh-intel, line 134: This URL is used multiple times so I'd consider creating a `readonly` global for it.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+
+###########################
+title "kata-monitor metrics retrieval"
+
+curl -s 127.0.0.1:8090/metrics > "$METRICS_FILE"
+[ $? -ne 0 ] && error_with_msg "cannot retrieve metrics"
+
+FILESIZE=$(stat -c%s "$METRICS_FILE")
+echo "Metrics file ($FILESIZE bytes): ${METRICS_FILE}"
+[ $FILESIZE -le 90000 ] && error_with_msg "metrics file looks too short"
```
jodh-intel, line 223: Where does this value come from? Please can you create a variable for this and add a comment.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,245 @@
...
+		echo -n "."
+		sleep 1
+		continue
+	done
+
+	return $FALSE
+}
+
+###########################
+title "pre-checks"
```
jodh-intel, line 174: Since the rest of the script is nicely divided into functions, I'd consider adding a `main` for completeness and consistency:

```bash
main()
{
    # ...
}

main "$@"
```

jodh-intel (commented):
Thanks @fgiudici - a few comments...

functional/kata-monitor/run.sh
fgiudici: ups... I missed that when reworking the script before sharing. Yeah, idea was to keep upper case var names for global vars only.  nice catch, thanks!

functional/kata-monitor/run.sh
fgiudici: yep, makes sense.
Will need to trap ERR than and change a bit the logic of error reporting to not lose that.

functional/kata-monitor/run.sh
fgiudici: If we set -u unset, I think we could leave it as is (no param passed will cause an error). How would that look?

functional/kata-monitor/run.sh
fgiudici: I redirect /dev/null only for crictl... so created a `qcrictl()` function (quiet crictl)!

functional/kata-monitor/run.sh
fgiudici: Well, checking the file size doesn't make much sense as we also count the number of metrics for a kata pod: removing this check  and adding a check on the number of metrics retrieved.

fgiudici:
@jodh-intel , many thanks for the review!
Updated following your suggestions

jodh-intel (commented):
Thanks @fgiudici.

functional/kata-monitor/run.sh
fgiudici: You are right.
Looking back at this I also think that would be even better if the function which creates the *returns* the filename!
Will update

fgiudici:
Many thanks @jodh-intel : followed your suggestions and re-pushed :-)

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,282 @@
...
+}
+
+# quiet crictrl
+qcrictl() {
+	crictl "$@" > /dev/null
+}
+
+# this is just an hash of current date (+ nanoseconds)
+gen_unique_id() {
+	date +%T:%N | md5sum | cut -d ' ' -f 1
```
jodh-intel, line 61: That is pretty cute :smile:

jodh-intel (approved):
Thanks @fgiudici.

lgtm

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,282 @@
...
+	echo -e "\nkata-monitor logs:\n----------------"
+	cat "$MONITOR_LOG_FILE"
+	echo -e "\nkata-monitor testing: FAILED!"
+	cleanup
+	exit 1
+}
+
+cleanup() {
+	stop_workload
+	stop_workload "$RUNC_CID" "$RUNC_POD_ID"
```
c3d, line 78: You probably only want to run that test if `$RUNC_CID` and $RUNC_POD_ID` are set, otherwise you are going to see extra dubious errors.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,282 @@
...
+	echo_ok "$CURRENT_TASK"
+
+	###########################
+	CURRENT_TASK="cleanup"
+	cleanup
+
+	echo -e "\nkata-monitor testing: PASSED!\n"
+}
+
+main "@"
```
c3d, line 282: Add newline at end of file?

c3d (dismissed):
Looks good to me, although I must admit I did not run it.

functional/kata-monitor/run.sh
```diff
@@ -0,0 +1,282 @@
...
+
+	qcrictl stopp $pod_id
+	qcrictl rmp $pod_id
+}
+
+is_sandbox_there() {
+	local podid=${1}
+	local sbs s
+
+	sbs=$(curl -s ${MONITOR_HTTP_ENDPOINT}/sandboxes)
```
jodh-intel, line 163: @fgiudici, @liubin - I cannot get `/sandboxes` to work - it doesn't generate an HTTP error, but also doesn't show any output even when a sb is running?!?

jodh-intel: @liubin - thanks for checking. I see the problem - if you run it _after_ a container is started, you get nothing. I've raised https://github.com/kata-containers/kata-containers/issues/3704 as we should add some docs for this tool.

fgiudici: > @liubin - thanks for checking. I see the problem - if you run it _after_ a container is started, you get nothing.

Well, actually when kata-monitor starts it does collect already running kata sandboxes, so it should work.
What may happen is that if you start kata-monitor and no kata sbs have ever been started from the host boot, the /run/vc/sbs dir will not exist - so kata-monitor will not be able to watch the dir.
In this particular case, kata-monitor will poll every 60 seconds to see if the directory has been created, so if you start your very first ever kata pod __after__ kata-monitor has been started, you may get a delay up to 60 seconds to have it reported by kata-monitor.

> I've raised [kata-containers/kata-containers#3704](https://github.com/kata-containers/kata-containers/issues/3704) as we should add some docs for this tool.

Yep, thank you, since I have done few changes I think it would be fair if I take care of that. I would like anyway to do it after  https://github.com/kata-containers/kata-containers/pull/3553 gets merged.

jodh-intel: Thanks @fgiudici ! :smile:

jodh-intel: @fgiudici - I'm using containerd and even though I have 3 sbs's in `/run/vc/sbs/`, starting `kata-monitor` in trace mode shows zero sandboxes in its logs and nothing is returned from `curl -s localhost:8090/sandboxes`. It looks like it's not scanning for existing directory entries on startup which would be a bug I guess.

functional/kata-monitor/run.sh
fgiudici: Yep, good catch, I fixed that in the very last push I did!
I put some logic to suppress errors in the stop_workload() func that will manage also the empty params case; I also just added RUNC_CID="" and RUNC_POD_ID="" at the very beginning of the script to ensure the would be set (and not trigger the `nounset`.

functional/kata-monitor/run.sh
fgiudici: @jodh-intel, I would try to run:
`kata-monitor --log-level trace`

jodh-intel: Yep - I've got the output of that on https://github.com/kata-containers/kata-containers/issues/3705.

jodh-intel (approved):
Thanks @fgiudici.

lgtm

liubin (dismissed):
Thank you @fgiudici.

Maybe you need to rebase main to let the CI jobs pass.

fgiudici:
Thanks @jodh-intel , thanks @liubin !
Just rebased on main :-)

liubin (approved):
/test

Issue #4462, opened 2022-02-15
clh: Re-use the kata-containers' build script here


fidencio:
Let's avoid code duplication and re-use here the very same script that's used to build the kata-deploy binaries.

By doing this we:
* Reduce code duplication / complexity;
* Test the same build scripts used for kata-deploy on every single CI job;

PR #4463, opened 2022-02-15
clh: Use kata-containers' static build script

fidencio:
Let's avoid code duplication and re-use here the very same script that's
used to build the kata-deploy binaries.

By doing this we:
* Reduce code duplication / complexity;
* Test the same build scripts used for kata-deploy on every single CI
  job;

Fixes: #4462

### DIFF
--- a/.ci/install_cloud_hypervisor.sh
+++ b/.ci/install_cloud_hypervisor.sh
@@ -15,63 +15,15 @@ readonly script_dir=$(dirname $(readlink -f "$0"))
 cidir=$(dirname "$0")
 arch=$("${cidir}"/kata-arch.sh -d)
 source "${cidir}/lib.sh"
-# Where real kata build script exist, via docker build to avoid install all deps
-latest_build_url="${jenkins_url}/job/cloud-hypervisor-nightly-$(uname -m)/${cached_artifacts_path}"
-clh_bin_name="cloud-hypervisor"
-clh_install_path="/usr/bin/${clh_bin_name}"
-cloud_hypervisor_repo=$(get_version "assets.hypervisor.cloud_hypervisor.url")
-go_cloud_hypervisor_repo=${cloud_hypervisor_repo/https:\/\//}
-
-install_clh() {
-	[ -n "$cloud_hypervisor_repo" ] || die "failed to get cloud_hypervisor repo"
-	export cloud_hypervisor_repo
-
-	# Get version for cloud_hypervisor from runtime/versions.yaml
-	cloud_hypervisor_version=$(get_version "assets.hypervisor.cloud_hypervisor.version")
-	[ -n "$cloud_hypervisor_version" ] || die "failed to get cloud_hypervisor version"
-	export cloud_hypervisor_version
-
-	# Get cloud_hypervisor repo
-	go get -d "${go_cloud_hypervisor_repo}" || true
-	# This may be downloaded before if there was a depends-on in PR, but 'go get' wont make any problem here
-	clone_katacontainers_repo
-	pushd  $(dirname "${GOPATH}/src/${go_cloud_hypervisor_repo}")
-	# packaging build script expects run in the hypervisor repo parent directory
-	# It will find the hypervisor repo and checkout to the version exported above
-	"${katacontainers_repo_dir}/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh"
-	sudo install -D "cloud-hypervisor/${clh_bin_name}"  "${clh_install_path}"
-	popd
-}
-
-install_prebuilt_clh() {
-	local checksum_file="sha256sum-cloud-hypervisor"
-	go get -d "${go_cloud_hypervisor_repo}" || true
-	pushd  $(dirname "${GOPATH}/src/${go_cloud_hypervisor_repo}")
-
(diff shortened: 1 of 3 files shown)


fidencio:
/cc @wainersm, @GabyCT, what do you think about this approach?
Ideally we could do the very same for kernel and QEMU, at least.

jodh-intel (dismissed):
Thanks @fidencio.

lgtm

.ci/install_cloud_hypervisor.sh
```diff
@@ -15,63 +15,15 @@ readonly script_dir=$(dirname $(readlink -f "$0"))
...
 	clone_katacontainers_repo
-	pushd  $(dirname "${GOPATH}/src/${go_cloud_hypervisor_repo}")
-	# packaging build script expects run in the hypervisor repo parent directory
-	# It will find the hypervisor repo and checkout to the version exported above
-	"${katacontainers_repo_dir}/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh"
-	sudo install -D "cloud-hypervisor/${clh_bin_name}"  "${clh_install_path}"
-	popd
-}
-
-install_prebuilt_clh() {
```
wainersm, line 46: It is supposed to use a cached clh, otherwise build it. However the clh cache job was disabled a long time ago (apparently around Sept 2020). I just want to double-check if you still want to keep building clh on every job....

fidencio: @wainersm, we usually don't have to, as cloud-hypervisor provides a static binary as part of its release.
We'll have to build it in case we don't use their release, as we do right now.

Do you want to bring back the cloud-hypervisor cache? I wouldn't mind, but that's something to be done out of the scope of this PR.

.ci/install_cloud_hypervisor.sh
```diff
@@ -15,63 +15,15 @@ readonly script_dir=$(dirname $(readlink -f "$0"))
...
-	curl -fsOL --progress-bar "${latest_build_url}/${clh_bin_name}" || return 1
-	curl -fsOL "${latest_build_url}/${checksum_file}" || return 1
-
-	info "Verify download checksum"
-	sudo sha256sum -c "${checksum_file}" || return 1
-
-	info "installing ${clh_bin_name}" "${clh_install_path}"
-	sudo install -D ${clh_bin_name} "${clh_install_path}"
+	pushd $katacontainers_repo_dir
+	sudo -E PATH=$PATH make cloud-hypervisor-tarball
```
jodh-intel, line 24: - Shouldn't this be something like:
  ```bash
  make -C tools/packaging/kata-deploy/local-build -n cloud-hypervisor-tarball
  ```
- Do we actually need `sudo` to build the archive file?

jodh-intel (commented):
Thanks @fidencio.

.ci/install_cloud_hypervisor.sh
wainersm: Sorry, I am not used to the clh build process and scripts. Looking at https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh now I realize it gets the built clh or build from source. Great, so the clean up you did make complete sense. Ignore my comment please.

.ci/install_cloud_hypervisor.sh
jodh-intel: Ack - thanks for the explanation @fidencio !

.ci/install_runtime.sh
```diff
@@ -85,6 +85,7 @@ case "${KATA_HYPERVISOR}" in
 		;;
 	"cloud-hypervisor")
 		enable_hypervisor_config "${PKGDEFAULTSDIR}/configuration-clh.toml"
+		sudo sed -i 's|/usr/bin/cloud-hypervisor|/opt/kata/bin/cloud-hypervisor|g' "${runtime_config_path}"
```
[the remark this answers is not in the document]
jodh-intel: @wainersm - that's a good idea but would probably requires updates all over the place so you might want to explore changing this carefully :smile:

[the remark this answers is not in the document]
fidencio: @wainersm, what I'd prefer doing is actually changing bit by bit to use the kata-deploy scripts and then change the script used to build kata-containers to actually look for the artefacts on `/opt/kata/`.  But I see that as the next step after getting, for instance, kernel, and QEMU built using the kata-deploy script.

WDYT?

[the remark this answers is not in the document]
jodh-intel: Incremental changes sound good to me to avoid :boom: :boom: :boom: ... which obviously would only get triggered late on a Friday afternoon :smile:

devimc (approved):

wainersm (approved):

fidencio:
I've pushed a new commit due to failures faced on the VFIO CI.

jodh-intel (approved):
Thanks @fidencio.

lgtm

wainersm:
@fidencio vfio still failing so I will defer my approval because any update to this PR will reset it anyway

jodh-intel (approved):
Thanks @fidencio.

lgtm

devimc (approved):

Issue #4435, opened 2022-02-09
containerd integration tests override config


Jakob-Naucke:
#4406 hasn't been fully resolved, e.g. http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-s390x-PR/757/. This is because the containerd nullify its temporary fix.

PR #4436, opened 2022-02-09
integration/containerd: Do not override config

Jakob-Naucke:
Instead of blindly copying the default Kata config, use the existing if
it is available, and restore it afterwards; as there might be important
changes, e.g. #4407.

Also, do not change the default config.

PR #4407, opened 2022-01-27
Switch to virtio-9p on s390x

Jakob-Naucke:
until leftover shims under virtio-fs (#3998) are resolved.

Fixes: #4406
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

PR #4469, opened 2022-02-16
stable-2.3 | integration/containerd: Do not override config

Jakob-Naucke:
Instead of blindly copying the default Kata config, use the existing if
it is available, and restore it afterwards; as there might be important
changes, e.g. #4407.

Also, do not change the default config.

Fixes: #4435

Backport of #4436.

### DIFF
--- a/integration/containerd/cri/integration-tests.sh
+++ b/integration/containerd/cri/integration-tests.sh
@@ -55,11 +55,13 @@ readonly CONTAINERD_CONFIG_FILE="${tmp_dir}/test-containerd-config"
 readonly default_containerd_config="/etc/containerd/config.toml"
 readonly default_containerd_config_backup="$CONTAINERD_CONFIG_FILE.backup"
 readonly kata_config="/etc/kata-containers/configuration.toml"
+readonly kata_config_backup="$kata_config.backup"
 readonly default_kata_config="/usr/share/defaults/kata-containers/configuration.toml"
 
 ci_config() {
 	sudo mkdir -p $(dirname "${kata_config}")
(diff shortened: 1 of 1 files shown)


GabyCT (dismissed):

integration/containerd/cri/integration-tests.sh
```diff
@@ -98,15 +99,8 @@ ci_cleanup() {
...
-	if [ "$ID" == ubuntu ]; then
-		if [ -n "${SNAP_CI}" ]; then
-			# restore default configuration
-			sudo cp "${default_kata_config}" "${kata_config}"
-		elif [ -n "${CI}" ] ;then
-			[ -f "${kata_config}" ] && sudo rm "${kata_config}"
-		fi
-	fi
+	[ -f "$kata_config_backup" ] && sudo mv "$kata_config_backup" "$kata_config" || \
+		sudo cp "$default_kata_config" "$kata_config"
```
c3d, line 103: Why is the copy of the default kata config needed here?

This diverges from what was done before, which was dependent on `$SNAP_CI` and `$CI` values.

Jakob-Naucke: Hmm, good call. I'd actually like to keep the simplified logic suggested here rather than checking `$CI` etc. and just create a backup if a non-default config exists and restore it. However, it would still be better to remove the config in this case.

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

snir911 (approved):
LGTM thanks!

Issue #4471, opened 2022-02-16
jenkins_job_build.sh comments suggest we should run the static checks before fast return


Jakob-Naucke:
> https://github.com/kata-containers/tests/blob/003f84394ba8a3a18c19e15cb60ff1e8ce39f6aa/.ci/jenkins_job_build.sh#L130-L132

Oops.

WDYT? Should we keep the `force-skip-ci` label checks separate from other "doesn't touch anything the integration tests would catch"-style checks?

_Originally posted by @Jakob-Naucke in https://github.com/kata-containers/tests/issues/4385#issuecomment-1019991100_

In fact, we run `--only-arch`, so GHA takes care of it and this doesn't matter.

fidencio:
So, let's take a look at the static checks https://github.com/kata-containers/tests/blob/3132ae24cfa70fe66771b760982238879ca992aa/.ci/static-checks.sh#L1330-L1351 ...

We currently call them with `--all` as part of the scripts, but we should have different steps inside the same action calling `--commits`, `--docs`, `--dockerfiles`, `--files`, ... and once we have those we could have the `force-skip-ci` label skipping selected ones.

But in order to do that, we need to change https://github.com/kata-containers/kata-containers/blob/main/ci/static-checks.sh#L12

If someone is up to take the bullet of implementing this, I'd be happy to see it landing.

fidencio:
And then we need to update the githab actions. :-)

Jakob-Naucke:
@fidencio at this point, in the !x86 Jenkins runs, we run `--only-arch`, which would not encompass those anyhow.

PR #4472, opened 2022-02-16
ci: Remove some jenkins_job_build.sh cruft

Jakob-Naucke:
- Remove error message about missing spec of the Kata containers
  repository. nounset takes care of this.
- Remove comment about running fast return check before static checks.
  We have stopped doing this and there is no need to because we run the
  static checks `--only-arch` and other checks are run in GHA.

Fixes: #4471


cmaf (approved):

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

Issue #4484, opened 2022-02-18
install_qemu.sh broken for non-cached builds


Jakob-Naucke:
such as the ones on s390x, but also reproducible with `FORCE_BUILD_QEMU`:
```
$ FORCE_BUILD_QEMU=yes .ci/install_qemu.sh 
[install_qemu.sh:141] INFO: current QEMU version: v6.1.0
[install_qemu.sh:142] INFO: cached QEMU version: v6.1.0
fatal: destination path '/home/jakob/go/src' already exists and is not an empty directory.
[install_qemu.sh:85] ERROR: git clone --branch v6.1.0 --single-branch --depth 1 --shallow-submodules https://github.com/qemu/qemu /home/jakob/go/src/
```

PR #4485, opened 2022-02-18
ci: Set qemu_repo after QEMU_REPO_URL

Jakob-Naucke:
For non-cached builds, a `git clone` into QEMU_REPO is attempted, but
QEMU_REPO_URL is no longer set before it. Set it before use.

Fixes: #4484

/cc @devimc

### DIFF
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -21,8 +21,6 @@ KATA_DEV_MODE="${KATA_DEV_MODE:-}"
 
 CURRENT_QEMU_VERSION=""
 QEMU_REPO_URL=""
-# Remove 'https://' from the repo url to be able to git clone the repo
-QEMU_REPO=${QEMU_REPO_URL/https:\/\//}
 QEMU_ARCH=$(${cidir}/kata-arch.sh -d)
 PACKAGING_DIR="${katacontainers_repo_dir}/tools/packaging"
 ARCH=$("${cidir}"/kata-arch.sh -d)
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

fidencio (dismissed):
lgtm, thanks @Jakob-Naucke!

fidencio:
```
17:14:52 .ci/install_qemu.sh: line 104: QEMU_REPO: unbound variable
17:14:52 [jenkins_job_build.sh:172] ERROR: .ci/setup.sh
```

fidencio:
> lgtm, thanks @Jakob-Naucke!

Let me take it back, sorry, but it seems there's still something wrong here.

.ci/install_qemu.sh
```diff
@@ -225,7 +223,9 @@ main() {
 			;;
 	esac
 
-	# this variables are used by static-build scripts
+	# Strip "https://" to clone into $GOPATH
+	gopath_qemu_repo=${QEMU_REPO_URL/https:\/\//}
```
jodh-intel, line 227: Nit: This is cute but doesn't anchor the pattern so although it's highly unlikely to be an issue, it would be safer to use something like `sed 's,^https://,,g`

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

.ci/install_qemu.sh
Jakob-Naucke: Hmm. I kind of like not calling externals and I think we can call this impossible to be an issue rather than just highly unlikely, because that would mean we'd want to have `…https:/…` in a path and if that ever becomes a use case, we took a wrong turn IMO :)

GabyCT (approved):

Issue #4502, opened 2022-02-24
ci: kubernetes tests fail on host with uppercase letters in hostname


agamdua:
# Get your issue reviewed faster

Tests were run on macOS + parallels.
```
$ cat /etc/*-release  | grep -i pretty
PRETTY_NAME="Ubuntu 20.04.4 LTS"

$ uname -r
5.13.0-30-generic

$ export KATA_DEV_MODE=true
$ sudo -E PATH=$PATH make kubernetes
```

# Description of problem

While running the integration tests for kubernetes, tests that depend on the node being untainted for scheduling would fail, as it depends on the output of the `hostname` command.
Another test in `integration/kubernetes/k8s-block-volume.bats:"Block Storage Support` is also impacted with the `hostname` command being used within.

On debugging, I realized:

`kubectl get nodes` returns all node names in lower case.
The `hostname` command returns the host name with capitalization honored.

Error message seen:
```
0/1 nodes are available: 1 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.
```

Example:
```
parallels@parallels-Parallels-Virtual-Platform:~/go/src/github.com/kata-containers/tests$ hostname
parallels-Parallels-Virtual-Platform
```
While debugging:
```
root@parallels-Parallels-Virtual-Platform:/home/parallels/go/src/github.com/kata-containers/tests# kubectl get nodes
NAME                                   STATUS   ROLES                  AGE   VERSION
parallels-parallels-virtual-platform   Ready    control-plane,master   25s   v1.23.1
```

# Expected result

The test should pass.

# Actual result

```
0/1 nodes are available: 1 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.
```

# Further information

PR incoming.

PR #4503, opened 2022-02-24
ci: Fixed: convert nodename to lower case to match kubectl

agamdua:
`kubectl get nodes` returns all node names in lower case.
`hostname` returns the true host name, with capitalization honored.

This converts the hostname to lower case so that the master node
is successfully untainated for scheduling. Without this, the node
name is not recognized and the test fails with the error:

```
0/1 nodes are available: 1 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.
```

Fixes: #4502

Signed-off-by: Agam Dua <agam_dua@apple.com>

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -27,7 +27,7 @@ CRI_RUNTIME="${CRI_RUNTIME:-crio}"
 
 untaint_node() {
 	# Enable the master node to be able to schedule pods.
-	local node_name="$(hostname)"
+	local node_name="$(hostname | awk '{print tolower($0)}')"
 	local get_taints="kubectl get 'node/${node_name}' -o jsonpath='{.spec.taints}'"
 	if eval $get_taints | grep -q 'NoSchedule'; then
 		info "Taint 'NoSchedule' is found. Untaint the node so pods can be scheduled."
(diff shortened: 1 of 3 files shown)


Jakob-Naucke (commented):
@agamdua thanks for your PR. Should this be updated as well? https://github.com/kata-containers/tests/blob/41f2317f447fe467d7667fdc48f979646420318f/integration/kubernetes/vfio.sh#L188

agamdua:
Thanks for taking a look @Jakob-Naucke! On reading the code it indeed seems like we could benefit from an update there as well, good callout.
I was looking only at the tests in `K8S_TEST_UNION` in `run_kubernetes_tests.sh` since I only ran the `kubernetes` target from the `Makefile`.

I am trying to figure out what actually gets `vfio.sh` to run so I can confirm the behavior on the same VM.

Jakob-Naucke:
@agamdua I think you need VFIO-capable hardware to test this, which you might not have on hand :/

agamdua:
Thanks @Jakob-Naucke, I updated the code as suggested, good eye to spot that.
Also, based on what I see in Github, it looks like there's a `jenkins-vfio-host` which runs the tests!

Jakob-Naucke:
@agamdua oh yes, should have mentioned that. It doesn't have a mixed-case hostname though.

Jakob-Naucke (approved):

agamdua:
@Jakob-Naucke good eye, first time contributor, done it now!

Jakob-Naucke (approved):

fidencio (approved):
lgtm, thanks @agamdua!

GabyCT (approved):

PR #4508, opened 2022-02-24
tests: add image service_offload feature test

wainersm:
This introduces a test case for the image service_offload feature (i.e. the agent ability to pull the container image from inside the guest) when Kata Containers is built for Confidential Containers. My ultimate aim with this is to provide a backbone for more Confidential Containers tests be created by the community. 

This is a adaptation of some methods and workflows implemented in @stevenhorsman 's [`ccv0.sh`](https://github.com/kata-containers/kata-containers/blob/CCv0/docs/how-to/ccv0.sh) script. BTW, thanks @stevenhorsman  for all the help and please let me know if I misplaced the copyright statements and also if I can add your signed-off-by.

Also It was created a new CI job type (`CC_CRI_CONTAINERD`) to allow us add another Jenkins Job to run that and other cc + crictl + containerd tests. I added a [job](http://jenkins.katacontainers.io/job/wainer-test-CC) on Jenkins just to test this PR; and you can see a successful execution [here](http://jenkins.katacontainers.io/job/wainer-test-CC/17/console).

If you want to run this in your local host then you can use kata's vagrantfile (see instructions [here](https://github.com/kata-containers/tests/blob/main/README.md#using-vagrant-to-test-your-code-changes)). Do:
```
$ export GOPATH=~/go
$ export CI_JOB="CC_CRI_CONTAINERD"
< Ensure you have kata-containers on CCv0 branch and tests pointing to this PR.>
$ vagrant up ubuntu
$ vaggrant ssh ubuntu
<inside the VM>
$ ./.ci/run.sh
```
There are some work still to be done on the code (see the TODO markers) but I would prefer to address them in follow up PRs for two reasons:
  1) I need people to help me, and I want to allow those interested on CC tests/CI to start getting involved in code activity
  2) This test is already good enough to be run to test pull requests to the CCv0 branch.

Things that aren't TODO marked and I think they should be in focus next:

1) Ensure the test run in Fedora
2) Ensure the test passes with Cloud Hypervisor (I tested only with QEMU)
3) DNS resolution on guest side is still problematic. Currently it is using 8.8.8.8 as a workaround; ideally it should be using the same nameserver as the host and the bridge networking be working fine.
4) There must be a checker to see if the image was really pulled in the guest 

Fixes https://github.com/kata-containers/tests/issues/4509

Cc @c3d @fidencio @sameo @Jakob-Naucke


integration/containerd/confidential/agent_image.bats
```diff
@@ -0,0 +1,76 @@
...
+
+	echo "Prepare containerd for Confidential Container"
+	SAVED_CONTAINERD_CONF_FILE="/etc/containerd/config.toml.$$"
+	configure_cc_containerd "$SAVED_CONTAINERD_CONF_FILE"
+
+	echo "Reconfigure Kata Containers"
+	switch_image_service_offload on
+}
+
+@test "[cc][agent] Test can pull an unencrypted image inside the guest" {
```
stevenhorsman, line 21: Maybe specify that crictl is being used in this test description as I image we'll want to add a Kubernetes version (albeit in a different test file probably, but it might be helpful in reviewing the output to know quickly)?

integration/containerd/confidential/agent_image.bats
```diff
@@ -0,0 +1,76 @@
...
+	local pod_config="${FIXTURES_DIR}/pod-config.yaml"
+	local container_config="${FIXTURES_DIR}/container-config.yaml"
+
+	# TODO: add a disable_full_debug to revert the changes on teardown.
+	enable_full_debug
```
stevenhorsman, line 26: There might be an argument that the debug config shouldn't be used in the pipeline, especially as it introduces potential security holes? It depends if we need it to grab required logs long term?

stevenhorsman (approved):
This is some great work Wainer. The copyrights look good to me, so feel free to add me as a author/DCO sign-off.
I'd probably add another TODO of trying to reduce the copied code from ccv0.sh by updating it to use these methods, but that's probably an action on me and I'm fine with it not being part of this PR.
Thanks!

integration/containerd/confidential/agent_image.bats
wainersm: That's a good idea. I will change it.

BTW, using "tags"  on test description is an idea I got from the kubernetes e2e tests. Other Kata Containers tests doesn't use tags but I think they should.

integration/containerd/confidential/agent_image.bats
wainersm: Good point.

Maybe I should break this `enable_full_debug` into small enable_FOO_debug methods. For this test in particular, while running on CI (i.e. `CI=true`) we only want the agent logs for debugging in case the test fails. However, if you as a developer wants to debug a problem locally, enabling other debug like runtime is really appreciated.

I will refactor that code and send a v2.

wainersm:
/test

Updated this PR:
 * rebased with CCv0
 * split `enable_full_debug` in `enable_agent_debug`, `enable_agent_console` and `enable_agent_runtime`
 * if running with `CI=true` it doesn't enable agent console. If DEBUG=true and CI=false (devel mode) it won't  revert the environment

Tested those changes in http://jenkins.katacontainers.io/job/wainer-test-CC/18

@stevenhorsman thanks for the review! and let me know if with those changes it is okay to merge as is.

stevenhorsman (approved):
LGTM

Issue #4518, opened 2022-02-28
Add Alibaba Cloud Linux support to the Kata Containers CI scripts


fidencio:
```
17:41:32 + case "$distro" in
17:41:32 + die 'invalid distro: '\''alinux'\'''
17:41:32 + local 'msg=invalid distro: '\''alinux'\'''
17:41:32 + echo 'ERROR: invalid distro: '\''alinux'\'''
17:41:32 ERROR: invalid distro: 'alinux'
17:41:32 + exit 1
```

alinux seems to be the Alibaba Cloud Linux, which could be easily added as a RHEL / CentOS like distro.

fidencio:
Note: "not known distro" means, not known by our CI. :-)

Jakob-Naucke:
FYI I updated the labels, CI seems operational. Will still need script support though, idk if we want to keep this open.

fidencio:
I have re-purposed this one, at least we keep the history here.

liubin:
Just a memo:

```
$ cat /etc/os-release 
NAME="Alibaba Cloud Linux"
VERSION="3 (Soaring Falcon)"
ID="alinux"
ID_LIKE="rhel fedora centos"
VERSION_ID="3"
PLATFORM_ID="platform:al8"
PRETTY_NAME="Alibaba Cloud Linux 3 (Soaring Falcon)"
ANSI_COLOR="0;31"
HOME_URL="https://www.aliyun.com/"
```

PR #4519, opened 2022-02-28
setup: Add "alinux" support

fidencio:
Alibaba Cloud Linux seems to be a RHEL / CentOS like distro.  Let's try
to treat it the same as CentOS and hope for the best.

This is needed as it seems the ARM CIs switched to using Alibaba Cloud
Linux.

Fixes: #4518


fidencio:
Nah, closing this one, version_id is different and someone will actually have to invest time to get it added.
For now, I'd recommend we disable the ARM CI till Alibaba Cloud Linux is fully supported.

Issue #4531, opened 2022-03-02
Replace `io.containerd.run.kata.v2` by `io.containerd.kata.v2`


liubin:
Normally runtime name includes 4 parts, `io.containerd.run.kata.v2` has 5 parts.

But it works, since containerd compose the binary name only use the last 5 parts:

```golang
func BinaryName(runtime string) string {
	// runtime name should format like $prefix.name.version
	parts := strings.Split(runtime, ".")
	if len(parts) < 2 {
		return ""
	}

	return fmt.Sprintf("containerd-shim-%s-%s", parts[len(parts)-2], parts[len(parts)-1])
}

```
https://github.com/containerd/containerd/blob/2a3f1094a4684222ffc1078a171fda9bc650eac5/runtime/v2/shim/util.go#L83-L93

Use the normal pattern `io.containerd.kata.v2` can reduce user confusion.

PR #4532, opened 2022-03-02
tests: unify runtime type to io.containerd.kata.v2

liubin:
Generally speaking, the runtime type consists of 4 parts,
such as `io.containerd.runc.v1`, but `io.containerd.run.kata.v2`
may mislead users about how to name runtime type.

Fixes: #4531


Jakob-Naucke (approved):

devimc (approved):
thanks @liubin

Issue #4535, opened 2022-03-02
`priority=1` is constantly being added to the repo file


liubin:
When setup AlibabaCloud(based CentOS:8) node, I found after the CI job runs, the package repo file is constantly appending `priority=1`:

```bash
$ tail -10 /etc/yum.repos.d/AliYun.repo 

priority=1
priority=1
priority=1
priority=1
priority=1
priority=1
priority=1
priority=1
priority=1
```

The source:

https://github.com/kata-containers/tests/blob/f3d17866e411f1246898bd7e6c7e649ff64a67a8/.ci/setup_env_centos.sh#L40

CentOS may have the same issue maybe.

To fix it, can only append when the last line is not `priority=1`.

/cc @GabyCT

fidencio:
@liubin, I guess the main reason it's never been exposed on CentOS CIs is because there's not baremetal running CentOS (AFAICT).  Nice catch!

liubin:
Note: this line has the same issue.

```
echo "skip_missing_names_on_install=0" | sudo tee -a /etc/yum.conf
```

PR #4567, opened 2022-03-10
setup: avoid appending duplicated lines to config files

liubin:
setup_env_centos.sh appends some lines to package manager
config files, it makes too many duplicated lines in the
config files.

Since these config files are not edited manually too often,
so check the last line and append new lines only when needed.

Fixes: #4535


jodh-intel (approved):
Thanks @liubin.

lgtm

GabyCT (approved):

PR #4573, opened 2022-03-11
ci: add CC_SKOPEO_CRI_CONTAINERD CI job type

### LINKED ISSUE #4572: tests | CCv0: Non-TEE, CC pod creation with various combinations of (un)signed image vs (un)protected registry
Implement the I4-I7 test cases defined in https://github.com/confidential-containers/community/issues/3 . They are:

 * Test that we can create a Kata CC pod by pulling a sample signed, unencrypted image, from an protected registry, inside the guest
 * Test that we can create a Kata CC pod by pulling a sample unsigned, unencrypted image, from an unprotected registry, inside the guest
 * Test that we cannot create a Kata CC pod by pulling a sample unsigned, unencrypted image, from an protected registry, inside the guest
 * Test that we cannot create a Kata CC pod by pulling a sample signed, unencrypted image, from an protected registry but with a incorrect gpg public key, inside the guest

Currently the image signature checking and protection feature relies on `skopeo`. There will be needed to add a new `CI_JOB` configuration and job on Jenkins so that Kata's guest image is built with skopeo in.

wainersm:
Currently the kata-agent integrates with skopeo to pull and verify
container images inside the guest. On the absence of skopeo, the agent
falls back to a (still) limited implementation. For example, without skopeo
the container policy file is not ignored, so images cannot be verified.

This adds the CC_SKOPEO_CRI_CONTAINERD job type which extends the
CC_CRI_CONTAINERD (added in commit 0678babef540884) configuration
so that skopeo is installed in the guest image. We will use that configuration
to test features which relies on skopeo.

Fixes #4572

### DIFF
--- a/.ci/ci_job_flags.sh
+++ b/.ci/ci_job_flags.sh
@@ -93,7 +93,7 @@ case "${CI_JOB}" in
 	export KATA_HYPERVISOR="qemu"
 	export KUBERNETES="no"
 	;;
-"CRI_CONTAINERD"|"CRI_CONTAINERD_K8S"|"CC_CRI_CONTAINERD")
+"CRI_CONTAINERD"|"CRI_CONTAINERD_K8S"|"CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD")
 	# This job only tests containerd + k8s
 	init_ci_flags
 	export CRI_CONTAINERD="yes"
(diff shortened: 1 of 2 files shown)


.ci/ci_job_flags.sh
```diff
@@ -103,12 +103,14 @@ case "${CI_JOB}" in
...
-		"CC_CRI_CONTAINERD")
+		"CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD")
 			# Export any CC specific environment variables
 			export CCV0="yes"
-			#export SKOPEO=${SKOPEO:-}
 			export UMOCI=yes
 			export SECCOMP=yes
+			if [ "${CI_JOB}" == "CC_SKOPEO_CRI_CONTAINERD" ]; then
+				export SKOPEO=yes
+			fi
```
jodh-intel, line 113: Nit: This could be simplified:

```suggestion
			[ "${CI_JOB}" = 'CC_SKOPEO_CRI_CONTAINERD' ] && export SKOPEO=yes
```

jodh-intel (approved):
Thanks @wainersm.

.ci/ci_job_flags.sh
wainersm: True. I will send a v2. Thanks @jodh-intel !

stevenhorsman (approved):
LGTM. Thanks Wainer!

.ci/ci_job_flags.sh
wainersm: hmmm maybe I should not make that change otherwise the statement will return 1 in case of `CI_JOB` != `CC_SKOPEO_CRI_CONTAINER`. Or am I missing something @jodh-intel ?

Issue #4580, opened 2022-03-14
CCv1: Runt the CC specific tests with Cloud Hypervisor as well.


fidencio:
Right now the tests running are only being run with QEMU.  Although this is not exactly a huge problem as the tests running right now are non-TEE specific, we could already start checking no difference in behaviour happens depending on the VMM we plan to support.

PR #4581, opened 2022-03-14
CCv1 | ci: Add cloud-hypervisor to the CC specific jobs

fidencio:
As we're running those with QEMU only, let's improve the coverage and
add cloud-hypervisor to the possible tests.

Note: This is only addin the possibility to run the jobs, but the jobs
are not yet created.

Fixes: #4580

### DIFF
--- a/.ci/ci_job_flags.sh
+++ b/.ci/ci_job_flags.sh
@@ -114,6 +114,20 @@ case "${CI_JOB}" in
 			;;
 	esac
 	;;
+"CC_CRI_CONTAINERD_CLOUD_HYPERVISOR"|"CC_SKOPEO_CRI_CONTAINERD_CLOUD_HYPERVISOR")
+	# This job only tests containerd + k8s
+	init_ci_flags
+	export CRI_CONTAINERD="yes"
+	export CRI_RUNTIME="containerd"
+	export KATA_HYPERVISOR="cloud-hypervisor"
+	# Export any CC specific environment variables
+	export CCV0="yes"
+	export UMOCI=yes
+	export SECCOMP=yes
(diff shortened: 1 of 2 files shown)


stevenhorsman (approved):
LGTM

.ci/ci_job_flags.sh
```diff
@@ -114,6 +114,24 @@ case "${CI_JOB}" in
 			;;
 	esac
 	;;
+"CC_CRI_CONTAINERD_CLOUD_HYPERVISOR"|"CC_SKOPEO_CRI_CONTAINERD_CLOUD_HYPERVISOR")
+	# This job only tests containerd + k8s
+	init_ci_flags
+	export CRI_CONTAINERD="yes"
+	export CRI_RUNTIME="containerd"
+	export KATA_HYPERVISOR="cloud-hypevisor"
```
wainersm, line 122: s/hypevisor/hypervisor/

.ci/ci_job_flags.sh
```diff
@@ -114,6 +114,24 @@ case "${CI_JOB}" in
...
+	case "${CI_JOB}" in
```
wainersm, line 123: Don't need this case statement; it will be either CC_CRI_CONTAINERD_CLOUD_HYPERVISOR or CC_SKOPEO_CRI_CONTAINERD_CLOUD_HYPERVISOR anyway.

Somewhat unrelated with this PR: I just realized we have configuration to test CLH + K8s + (containerd or cri-o) but we don't have to test "pure" CLH + containerd. Is it intentional?

wainersm:
hi @fidencio , thanks for working on the support for CLH too! I only have those comments.

.ci/ci_job_flags.sh
fidencio: It's not intentional, what's the QEMU equivalent test that I missed? :-)

wainersm (approved):

Issue #4598, opened 2022-03-17
Disable agent shutdown test on s390x until more stable


Jakob-Naucke:
We should disable it for now.

_Originally posted by @Jakob-Naucke in https://github.com/kata-containers/tests/issues/4597#issuecomment-1070999722_

Issue #4597, opened 2022-03-17
Agent shutdown test unstable on s390x


Jakob-Naucke:
`tracing/test-agent-shutdown.sh` on s390x:
```
INFO: Waiting for agent VSOCK server
INFO: Kata VM running
INFO: Testing agent: shutdown test type: 'with-debug', agent test type: vm
INFO: Stopping agent
[test-agent-shutdown.sh:1060] ERROR: no VSOCK connections found
[test-agent-shutdown.sh:1060] ERROR: no VSOCK connections found
[test-agent-shutdown.sh:1087] ERROR: agent_addr=
WARNING: ERROR: Test failed
WARNING:
WARNING: Not cleaning up to help debug failure:
WARNING:
INFO: agent-ctl log file   : /home/linux1/go/src/github.com/kata-containers/tests/agent-ctl.log
INFO: agent log file       : /home/linux1/go/src/github.com/kata-containers/tests/kata-agent.log
INFO: OCI bundle directory : /tmp/tmp.LoSJ28D0u2
make: *** [Makefile:95: agent-shutdown] Error 1
```

which I guess implies the agent was unreachable during shutdown, which was supposed to have been fixed? Seems timing-related, I see this not as often on internal premises which have better time slices.

Jakob-Naucke:
We should disable it for now.

PR #4599, opened 2022-03-17
tracing: Fix or skip on s390x

Jakob-Naucke:
tracing: Remove Jaeger container before starting

can be left over on permanent instances

---

tracing: Skip agent shutdown test on s390x

is unstable, see https://github.com/kata-containers/tests/issues/4597

Fixes: #4598

### DIFF
--- a/tracing/test-agent-shutdown.sh
+++ b/tracing/test-agent-shutdown.sh
@@ -743,9 +743,9 @@ setup()
 		exit 0
 	}
 
-	# Do not run on ppc64le for now
-	[ "$(uname -m)" = "ppc64le" ] && {
-		info "Exiting, do not run on ppc64le"
+	# Do not run on ppc64le/s390x for now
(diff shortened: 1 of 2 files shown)


Jakob-Naucke:
> thanks @Jakob-Naucke, could you please move all the disabled tests/conditions to the same section? There are others disabled in `setup` in `test-agent-shutdown.sh`.

Ah, thanks for the heads-up -- I'd missed those. Done.

> Also could you skip s390x in `tracing-test.sh` too?
> 
> (Unless you can fix it)

`tracing-test.sh` appears to work with the minor fix already given.

> /test-s390x
> 
> actually...does this trigger still work?

s390x has been disabled, but this is hopefully the last step. The keyword is `/debug-s390x`, but it didn't pull in the changes from the PR last time I checked.

cmaf (approved):

Issue #4600, opened 2022-03-18
Add a CI target & entry point for testing cloud hypervisor from its own repo


fidencio:
Similarly to what we do with CRI-O, let's try to test each Cloud Hypervisor PR, on their repo, against kata-containers `main`branch.

This will allow us to catch disruptive changes before they happen and act on changes needed as soon as a PR is opened on the Cloud Hypervisor side.

PR #4601, opened 2022-03-18
ci: Add an external job for testing cloud-hypervisor

fidencio:
This will allow to test each cloud-hypervisor PR, and give us a head's
up on changes that will break Kata Containers, as we've faced in the
past.

Fixes: #4600

### DIFF
--- a/.ci/ci_clh_entry_point.sh
+++ b/.ci/ci_clh_entry_point.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script is called by our jenkins instances, triggered by PRs on Cloud Hypervisor.
+# It relies on the following environment variables being set:
+# REPO_OWNER    - owner of the source repository (default: cloud-hypervisor)
+# REPO_NAME     - repository name (default: cloud-hypervisor)
+# PULL_BASE_REF - name of the branch where the pull request is merged to (default: main)
+# PULL_NUMBER   - pull request number (REQUIRED)
+#
+# (see: http://jenkins.katacontainers.io/job/kata-containers-2-clh-PR/)
+#
+# Usage:
+# curl -OL https://raw.githubusercontent.com/kata-containers/tests/main/.ci/ci_clh_entry_point.sh
+# bash ci_clh_entry_point.sh
+
+set -o errexit
+set -o pipefail
+set -o errtrace
+set -o nounset
+[ -n "${DEBUG:-}" ] && set -o xtrace
+
+if [ -z "$PULL_NUMBER" ]; then
+	echo "No pull number given: testing with HEAD"
+	PULL_NUMBER="NONE"
+fi
+
+# set defaults for required variables
+export REPO_OWNER=${REPO_OWNER:-"cloud-hypervisor"}
+export REPO_NAME=${REPO_NAME:-"cloud-hypervisor"}
+export PULL_BASE_REF=${PULL_BASE_REF:-"main"}
+
+# Export all environment variables needed.
+export CI_JOB="EXTERNAL_CLOUD_HYPERVISOR"
+export INSTALL_KATA="yes"
+export GO111MODULE=auto
+
+export cloud_hypervisor_pr="${PULL_NUMBER}"
+export cloud_hypervisor_pull_ref_branch="${PULL_BASE_REF}"
+
+export ghprbGhRepository="${REPO_OWNER}/${REPO_NAME}"
+export GOROOT="/usr/local/go"
+
+# Put our go area into the Jenkins job WORKSPACE tree
+export GOPATH=${WORKSPACE}/go
+export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:/sbin:${PATH}
+mkdir -p "${GOPATH}"
+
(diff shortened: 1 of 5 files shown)


.ci/ci_clh_entry_point.sh
```diff
@@ -0,0 +1,97 @@
...
+#
+# (see: http://jenkins.katacontainers.io/job/kata-containers-2-clh-PR/)
+#
+# Usage:
+# curl -OL https://raw.githubusercontent.com/kata-containers/tests/main/.ci/ci_clh_entry_point.sh
+# bash ci_clh_entry_point.sh
+
+set -o errexit
+set -o pipefail
+set -o errtrace
```
jodh-intel, line 22: Could you add the following for extra safety and debugability:

```bash
 set -o nounset
```

.ci/ci_clh_entry_point.sh
```diff
@@ -0,0 +1,97 @@
...
+export INSTALL_KATA="yes"
+export GO111MODULE=auto
+
+export cloud_hypervisor_pr="${PULL_NUMBER}"
+export cloud_hypervisor_pull_ref_branch="${PULL_BASE_REF}"
+
+export ghprbGhRepository="${REPO_OWNER}/${REPO_NAME}"
+export GOROOT="/usr/local/go"
+
+export JOBS=1
```
jodh-intel, line 45: What is this for?

.ci/ci_clh_entry_point.sh
```diff
@@ -0,0 +1,97 @@
...
+mkdir -p $(dirname "${katacontainers_repo_dir}")
+[ -d "${katacontainers_repo_dir}" ] || git clone "https://${katacontainers_repo}.git" "${katacontainers_repo_dir}"
+
+# Run kata-containers setup
+cd "${tests_repo_dir}"
+.ci/setup.sh
+
+pushd "${katacontainers_repo_dir}"
+	echo "Rebuild Kata Containers' runtime using the latest cloud-hypervisor.yaml file"
+	cp tools/packaging/kata-deploy/local-build/build/cloud-hypervisor/builddir/cloud-hypervisor/vmm/src/api/openapi/cloud-hypervisor.yaml src/runtime/virtcontainers/pkg/cloud-hypervisor/cloud-hypervisor.yaml
```
jodh-intel, line 88: I wonder if it's worth adding a check to make sure that the dest file already exists? That would ensure that if we do any further repo restructuring, this test will fail.

jodh-intel, line 88: Nit: 3 mentions of `cloud-hypervisor.yaml` so I'd consider creating a variable for it.

jodh-intel (commented):
Thanks @fidencio - a few comments...

.ci/ci_clh_entry_point.sh
fidencio: Done!

.ci/ci_clh_entry_point.sh
fidencio: Leftover :-)

.ci/ci_clh_entry_point.sh
fidencio: The `dest` file exists anyways, and we're only replacing it by the one from cloud-hypervisor repo.

I'm not sure what's the best approach here, as any repo restructuring on either cloud-hypervisor or kata-containers could cause this to break.

Ideas?

jodh-intel: > The `dest` file exists anyways, and we're only replacing it by the one from cloud-hypervisor repo.

Exactly. It's supposed to exist, but as long as the directory that file is supposed to exist in exists, this script could silently fail I think? This might be overkill / paranoia, but the script could check if the destination exists. If it does, great, overwrite it. But if it doesn't, fail as that's an unexpected scenario and implies that if we went ahead with the copy (which would probably work), the test would not be testing what we think it should be.

fidencio: This should be addressed now.

jodh-intel (approved):
Thanks @fidencio.

lgtm

.ci/install_cloud_hypervisor.sh
```diff
@@ -17,12 +17,15 @@ arch=$("${cidir}"/kata-arch.sh -d)
 source "${cidir}/lib.sh"
 
 main() {
+	local buildscript="${katacontainers_repo_dir}/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh"
```
GabyCT, line 20: why we are using the binaries? this will affect other CI jobs that installs clh

fidencio: This is exactly what we're already doing for cloud-hypervisor, @GabyCT, but instead of calling the script, we call `sudo -E PATH=$PATH make cloud-hypervisor-tarball`, which would then be translated to `${katacontainers_repo_dir}/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh" --build=cloud-hypervisor`.

This was introduced not so long ago (see: https://github.com/kata-containers/tests/pull/4463) and ensures that we rely on a single way to build cloud-hypervisor, respecting the needed flags, across all the different parts of the project (CIs, snap, kata-deploy, release binaries, etc).

Does this cover your concern?

GabyCT: great thanks @fidencio for the explanation

GabyCT (approved):

Issue #4603, opened 2022-03-21
edk2 build fail due to the bug of its component


jongwu:
acpica is one of the buildkit for building edk2 on arm64. but there is bug in the lastest code of it, which lead to failure of arm ci.
see http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-ARM-PR/3536/console

PR #4606, opened 2022-03-22
backport: uefi/arm64: optimize uefi installation

jongwu:
It's better clone source code, including acpica and edk2, from a stable
tag.
As there maybe some unknown issue about qemu efi from linaro, which doesn't
work well with kata. So, only create qemu efi from souce code, until the
issue is fixed.

Fixes: #4603


GabyCT (approved):

fidencio (approved):
lgtm, thanks @jongwu!

Issue #4391, opened 2022-01-20
add ctr based bats suite


egernst:
We have a nice collection of k8s integration tests now. It would be useful to also run tests at a lower level -- I think copying the ./integration/kubernetes pattern and add set of bats based on ctr would be helpful. 

Scenarios where testing at this level is better include:
- exercise standalone container logic (not a pod). For example: https://github.com/kata-containers/kata-containers/pull/2941
- exercise tests which require a known sandbox/container id - for example: https://github.com/kata-containers/tests/issues/4305
- run stress tests which cycle through container lifecycle at high rate (ie, something like containerd-stress).

PR #4622, opened 2022-03-24
ci: Added: tests for OCI prestart and poststart hooks.

### LINKED ISSUE #4305: Create guest OCI hooks tests
It seems that **guest** OCI hooks are broken, but about to be fixed on https://github.com/kata-containers/kata-containers/pull/3351.

We need an explicit (bats?) test for guest hooks that does something like:

- Share a directory into the contain as `/usr/share/oci/hooks/`.
- Create a single executable shell script hook in that directory that write a file to `/tmp/`.
- Enable `guest_hook_path=/usr/share/oci/hooks/` in the config file.
- Create a container that `cat`'s the file the hook should have created in `/tmp/`.
- Fail if the file doesn't exist.

agamdua:
Tests were created in a new `integration/ctr` directory,
which sets the stage for more work related to #4391.

The tests create a file on the guest vm which is shared
 with the container and adds the string `foo` to it.
A grep validates the string exists.
This was more helpful while debugging than a straight
presence check/stat of the file existing.

Resolves #4305.

Signed-off-by: Agam Dua <agam_dua@apple.com>

TODO:
- [ ] need to trigger this in CI.


integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bats
```
fidencio, line 1: You're missing a copyright statement here.
Please, check https://github.com/kata-containers/tests/blob/8fa30acf6d5a2f1f137a6e47c4de21c45dc9b2e3/integration/kubernetes/k8s-inotify.bats#L1-L7 for instance

fidencio (commented):
This is not an official review, but rather a way to get CI passing ...

You need to use `Fixes: #NNNN`, where NNNN is the issue number, rather than `Resolves ....`
The script checks specifically for that.

Apart from this, you'll face some issues due to this file missing a copyright statement, see the comment inline.

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,94 @@
...
+	# add guest image to config file and enable oci hooks.
+	mkdir -p /etc/kata-containers
+	sudo sed -i -r "s#^(image =).*#\1 \"$tmp_guest_img_path\"#" "$KATA_CONFIG"
+	sudo sed -i '/^#guest_hook_path /s/^#//' "$KATA_CONFIG"
+
+	# pull an image with ctr to run and test for oci hooks
+	sudo ctr image pull "$IMAGE"
+	[ $? != 0 ] && die "Unable to get image $IMAGE"
+
+	run sudo ctr run --rm --runtime="$CTR_RUNTIME" "$IMAGE" "$CONTAINER_NAME" sh -c "grep foo /tmp/test-oci-hooks-file"
```
jodh-intel, line 59: Could you create a variable for `/tmp/test-oci-hooks-file` as it's used multiple times.

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,94 @@
...
+}
+
+
+@test "Ensure poststart OCI hooks work" {
+	local tmp_guest_img_path="$HOST_TMP_TEST_DIR/kata-containers.img"
+	local tmp_hooks_filepath="$HOST_TMP_TEST_DIR/hooks.sh"
+	local tmp_mount_point="$HOST_TMP_TEST_DIR/mnt"
+
+	# leveraging an oci poststart hook so that we can inspect the file after the container is up and running.
+	local tmp_oci_hooks_dir="$tmp_mount_point/usr/share/oci/hooks/poststart"
```
jodh-intel, line 36: This is the only line different between this test function and the prestart test function below. I'd consider creating a generic test function and passing "prestart" or "poststart" to it from each of the `@test` functions to avoid all the duplication.

jodh-intel (commented):
Thanks for raising @agamdua! A few comments...

agamdua:
Thanks @fidencio for suggestions on getting CI to pass. I have incorporated them in the latest commit!
Thanks @jodh-intel for taking a look at this PR. I 100% agree with both your comments and have simplified the code with your suggestions. Further feedback welcome!

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,78 @@
...
+
+function test_hook_works() {
+	hook_type=$1
+
+	local tmp_guest_img_path="$HOST_TMP_TEST_DIR/kata-containers.img"
+	local tmp_hooks_filepath="$HOST_TMP_TEST_DIR/hooks.sh"
+	local tmp_mount_point="$HOST_TMP_TEST_DIR/mnt"
+	local tmp_oci_hooks_file="/tmp/test-oci-hooks-file"
+
+	# leveraging an oci hook so that we can inspect the file after the container is up and running.
```
jodh-intel, line 42: Nit:

```suggestion
	# leveraging an OCI hook so that we can inspect the file after the container is up and running.
```

jodh-intel (approved):
Thanks @agamdua.

lgtm

wainersm:
Hi @agamdua , first of all, thanks for this contribution! :)

One thing that might sound nit but it might not be in the end: it seems to me that `integration/containerd` is the best directory to hold the future ctr-based tests (https://github.com/kata-containers/tests/issues/4391) as they are really containerd integration tests (ctr happens to be just a tool).

To be fair, I find myself in doubt every time I need to choose where to put a new test on tests' repo. The other day I put "confidential" under "integration/containerd" but after a while I regretted, I am almost to move it to `functional`. On the other hand, there are tests under `integration` (e.g. `rootless`, `vcpu`) that seems to fit better in `functional`. @jodh-intel are there documented guidelines on where to put tests?

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,78 @@
...
+	# prepare guest image with hooks
+	mkdir $tmp_mount_point
+	cp /usr/share/kata-containers/kata-containers.img $tmp_guest_img_path
+	sudo mount -o loop,offset=$((512*6144)) $tmp_guest_img_path $tmp_mount_point
+	sudo mkdir -p $tmp_oci_hooks_dir
+	sudo cp $tmp_hooks_filepath $tmp_oci_hooks_dir
+	sudo umount $tmp_mount_point
+
+	# add guest image to config file and enable oci hooks.
+	mkdir -p /etc/kata-containers
```
wainersm, line 58: I think `/etc/kata-containers`  directory was created on `setup`.

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2022 Apple Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+IMAGE="${IMAGE:-docker.io/library/busybox:latest}"
```
wainersm, line 8: It would be nice if you could use an image from quay.io or a registry other than docker.io. As people behind a firewall can very often hit the limit of pull requests imposed by docker.io

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,78 @@
...
+CONTAINER_NAME="${CONTAINER_NAME:-kataocihookstest}"
+CTR_RUNTIME="${CTR_RUNTIME:-io.containerd.kata.v2}"
```
wainersm, line 10: Do you need to test any other runtime than `io.containerd.kata.v2`? If not, I would hard-code that value, because otherwise you might end up on testing a different runtime by mistake (suppose the environment have `CTR_RUNTIME` exported with an unexpected value).

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,78 @@
...
+function teardown() {
+	sudo rm -rf $HOST_TMP_TEST_DIR
+	[ -f "$KATA_CONFIG_BACKUP" ] && sudo mv "$KATA_CONFIG_BACKUP" "$KATA_CONFIG" || \
+		sudo cp "$DEFAULT_KATA_CONFIG" "$KATA_CONFIG"
+}
+
+function setup() {
+	HOST_TMP_TEST_DIR="$(mktemp -d)"
+	sudo mkdir -p $(dirname "${KATA_CONFIG}")
+	[ -f "$KATA_CONFIG" ] && sudo cp "$KATA_CONFIG" "$KATA_CONFIG_BACKUP" || \
```
wainersm, line 30: Let `KATA_CONFIG=/etc/kata-containers/configuration.toml` and `$KATA_CONFIG` file doesn't exist. Then:
1. `setup` copies `/usr/share/defaults/kata-containers/configuration.toml` to `$KATA_CONFIG`
2. This test changes `$KATA_CONFIG`
3. `teardown` copies `/usr/share/defaults/kata-containers/configuration.toml` to `$KATA_CONFIG`
4. This test exited and left a `/etc/kata-containers/configuration.toml` behind.
5. The next test doesn't change the `/etc/kata-containers/configuration.toml` because this file is unexpected on CI by default. Instead it looks at `/usr/share/defaults/kata-containers/configuration.toml` but changes take no effect anymore.

It seems error prone. Am I wrong and/or too paranoiac?

Maybe we could turn this method https://github.com/kata-containers/tests/blob/main/functional/tdx/run.sh#L39  generic then on this test you get the default config, save it on setup, change on test, and restore on teardown.

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,78 @@
...
+#
+
+IMAGE="${IMAGE:-docker.io/library/busybox:latest}"
+CONTAINER_NAME="${CONTAINER_NAME:-kataocihookstest}"
+CTR_RUNTIME="${CTR_RUNTIME:-io.containerd.kata.v2}"
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o errtrace
```
wainersm, line 15: hmmm... I don't know the (bad, if any) consequences of setting those options on bats tests.

integration/ctr/oci_hooks_tests.bats
```diff
@@ -0,0 +1,78 @@
...
+	# leveraging an oci hook so that we can inspect the file after the container is up and running.
+	local tmp_oci_hooks_dir="$tmp_mount_point/usr/share/oci/hooks/$hook_type"
+
+	# add file with greppable phrase to file that is shared by guest vm and container.
+	printf "#!/bin/sh\necho \"foo\" > /run/kata-containers/%s/rootfs/$tmp_oci_hooks_file\n" "$CONTAINER_NAME" > $tmp_hooks_filepath
+	chmod +x $tmp_hooks_filepath
+
+	# prepare guest image with hooks
+	mkdir $tmp_mount_point
+	cp /usr/share/kata-containers/kata-containers.img $tmp_guest_img_path
```
wainersm, line 51: What happens if kata was built with initrd guest instead?

integration/ctr/oci_hooks_tests.bats
agamdua: thanks for this suggestion, I didn't consider the limits with docker.io.

I see `quay.io/prometheus/busybox:latest` used in a few places in the repo. I have tested it locally -- ill push that up as a change, but if you have a more opinionated recommendation happy to try it out.

integration/ctr/oci_hooks_tests.bats
agamdua: I picked this up from another file as a "pattern" I assumed, so happy to switch back to hardcoding it to `io.containerd.kata.v2`!

integration/ctr/oci_hooks_tests.bats
agamdua: good eye 👀 👍

jodh-intel:
> @jodh-intel are there documented guidelines on where to put tests?

Not that I'm aware of. Feel free to raise an issue /PR to discuss that though ;)

wainersm:
> > @jodh-intel are there documented guidelines on where to put tests?
> 
> Not that I'm aware of. Feel free to raise an issue /PR to discuss that though ;)

@jodh-intel , just created an issue for that: https://github.com/kata-containers/tests/issues/4645

agamdua:
Thanks for the bump @jodh-intel — I’m going to get back to this in earnest next week, and I’ll definitely rebase with the latest then!

integration/ctr/oci_hooks_tests.bats
agamdua: I think you may have a point -- I do want to mention I used the same pattern as [here](https://github.com/kata-containers/tests/blob/2.4.1/integration/containerd/cri/integration-tests.sh#L65-L66), and I trusted the assumptions made there. I can't see the function you're linking to, since `main` has changed!

jodh-intel:
@agamdua - the CI's are still failing with what looks like conflicts:

```
12:41:36 .git/rebase-apply/patch:110: new blank line at EOF.
12:41:36 +
12:41:36 warning: 1 line adds whitespace errors.
12:41:36 Using index info to reconstruct a base tree...
12:41:36 M	.ci/run.sh
12:41:36 M	Makefile
12:41:36 Falling back to patching base and 3-way merge...
12:41:36 Auto-merging Makefile
12:41:36 Auto-merging .ci/run.sh
12:41:36 CONFLICT (content): Merge conflict in .ci/run.sh
12:41:36 error: Failed to merge in the changes.
```

Issue #4632, opened 2022-03-28
kata-log-parser fails because of missing PID


littlejawa:
When using logs from the containerd-shim-v2, kata-log-parser fails because some of the logs have no PID.
While kata-log-parser is used in our CI's teardown script, this makes it useless, as it actually fails on every run.

We would need to make kata-log-parser more tolerant over the log format.

wainersm:
Hi @littlejawa always bothered me the fact that kata-log-parser runs on the teardown and it doesn't fail the job. I suspect it is because the tool is used, in CI, as a mere log collector, i.e., the intention is not really to test the tool. This should be changed IMO.

PR #4646, opened 2022-03-31
tests: kata-log-parser should not make an error for missing fields

littlejawa:
Adding a parameter to kata-log-parser to make it accept lines with
missing fields (name, source, level or pid).
Other checks are kept untouched.

Some tests have been added to verify this new behavior.

Fixes: #4632

### DIFF
--- a/cmd/log-parser/logentry.go
+++ b/cmd/log-parser/logentry.go
@@ -115,7 +115,7 @@ func (le LogEntry) Fields() []string {
 }
 
 // Check runs basic checks on the LogEntry to ensure it is valid.
-func (le LogEntry) Check() error {
+func (le LogEntry) Check(ignoreMissingFields bool) error {
 	if le.Filename == "" {
 		return fmt.Errorf("missing filename: %+v", le)
 	}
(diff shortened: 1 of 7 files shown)


cmd/log-parser/logentry.go
```diff
@@ -134,24 +134,26 @@ func (le LogEntry) Check() error {
...
-	}
+	if !ignoreMissingFields {
+		if le.Pid == 0 {
+			return fmt.Errorf("missing pid: %+v", le)
+		}
 
-	if le.Pid < 0 {
-		return fmt.Errorf("invalid pid: %+v", le)
-	}
+		if le.Level == "" {
```
snir911, line 142: I'm not sure if the coloring here means something is wrong with indentation

littlejawa: I don't think so. I effectively added a level of indentation for each line by embedding the existing code inside an "if" block.
The diff is trying to show that, but it is slightly off because I moved one of the existing "if" outside of my new block.

GabyCT (approved):

PR #4655, opened 2022-04-01
integration/rootless: add more asserts

### LINKED ISSUE #4654: integration/rootless: add more asserts to make it more robust

The rootless test (`integration/rootless`).

**How can it be improved?**

Currently it starts a pod and assume rootless is working. However, to make it more robust, it should be checking the process and files are really owned by a non-root user.

wainersm:
This added two more asserts to the rootless test:
- that the QEMU process is owned by the temporary non-root user
- that the QEMU files are owned by that user too.

Fixes #4654

### DIFF
--- a/integration/rootless/rootless_test.sh
+++ b/integration/rootless/rootless_test.sh
@@ -17,6 +17,7 @@ source /etc/os-release || source /usr/lib/os-release
 pod_id=""
 
 setup() {
+	extract_kata_env
 	getent group kvm &>/dev/null || sudo groupadd --system kvm
 	sudo chown root:kvm /dev/kvm
 	sudo chmod g+rw /dev/kvm
@@ -27,19 +28,47 @@ setup() {
 
 cleanup() {
 	sudo sed -i -e 's/^.*\(rootless\)/# \1/g' /usr/share/defaults/kata-containers/configuration.toml
-	sudo crictl stopp "$pod_id" || true
-	sudo crictl rmp "$pod_id" || true
+	sudo crictl stopp "$pod_id" &>/dev/null || true
+	sudo crictl rmp "$pod_id" &>/dev/null || true
 }
 
 run() {
 	pod_id="$(sudo crictl runp -r kata "${dir_path}/rootless-pod.json")"
 	waitForProcess 15 3 "sudo crictl inspectp "$pod_id" | jq '.status.state' | grep 'SANDBOX_READY'"
+
+	# There must be created one and only one temporary user.
+	local grep_uid="grep -e '^kata-[0-9]\+' /etc/passwd"
+	waitForProcess 5 1 "$grep_uid >/dev/null"
+	[ "$(eval $grep_uid | wc -l)" -eq 1 ] || \
+		die "Unexpected more than one kata temporary UID"
+	local kata_uid="$(eval $grep_uid | cut -d: -f3)"
+	local kata_gid="$(eval $grep_uid | cut -d: -f4)"
+	[[ -n "$kata_uid" && -n "$kata_gid" ]] || \
+		die "Unable to find the kata temporary UID"
+
+	# Assert the QEMU process is owned by that user.
+	waitForProcess 10 1 "ps -u "$kata_uid" -o cmd --no-header | \
+		grep -q "^${HYPERVISOR_PATH:-qemu}"" || \
+		die "Expects the QEMU process be owned by a non-root user"
+
+	# Assert the QEMU files are owned by that user too.
+	local pod_dir="/run/user/${kata_uid}/run/vc/vm/${pod_id}"
+	[ -d "$pod_dir" ] || \
+		die "Unable to find the pod directory: $pod_dir"
+	local pod_files=(console.sock qmp.sock pid vhost-fs.sock)
+	for file_name in ${pod_files[@]}; do
+		local file="${pod_dir}/${file_name}"
+		waitForProcess 10 1 "test -e "$file"" || \
+			die "File didn't show up: $file"
+		[ "$(sudo stat -c %u-%g $file)" == "${kata_uid}-${kata_gid}" ]
+	done
 }
 
 main() {
 	trap cleanup EXIT QUIT KILL
 	setup
 	run
+	echo "rootless test: PASSED"
 }
 
 main


devimc:
thanks @wainersm but afaik the CI job where this test runs was removed

wainersm:
Hi @devimc ,

> thanks @wainersm but afaik the CI job where this test runs was removed

It's executed in kata-containers-fedora-35-x86_64-CRIO_K8S-PR at least. For example, http://jenkins.katacontainers.io/view/PR%20Triggered/job/kata-containers-fedora-35-x86_64-CRIO_K8S-PR/341/console : 
```
08:16:37 INFO: Running rootless tests
08:16:37 bash -f integration/rootless/rootless_test.sh
08:16:37 "SANDBOX_READY"
08:16:39 Stopped sandbox 93a477025da9f9c1ff3c3a605336457b16b20db0568808429a6d39e62b5bda4b
08:16:39 Removed sandbox 93a477025da9f9c1ff3c3a605336457b16b20db0568808429a6d39e62b5bda4b
```

Am I missing something?

devimc (approved):
thanks @wainersm

integration/rootless/rootless_test.sh
```diff
@@ -27,19 +27,35 @@ setup() {
...
 	pod_id="$(sudo crictl runp -r kata "${dir_path}/rootless-pod.json")"
 	waitForProcess 15 3 "sudo crictl inspectp "$pod_id" | jq '.status.state' | grep 'SANDBOX_READY'"
+
+	# There must be created a temporary user.
+	local kata_uid="$(grep -e '^kata-[0-9]\+' /etc/passwd | cut -d: -f3)"
+	local kata_gid="$(grep -e '^kata-[0-9]\+' /etc/passwd | cut -d: -f4)"
+	[[ -n "$kata_uid" && -n "$kata_gid" ]]
+
+	# Assert the QEMU process is owned by that user.
+	ps -u "$kata_uid" | grep "qemu" &>/dev/null || \
```
snir911, line 44: should we validate qemu process exists first?

wainersm: I'm not sure. And if there is a running qemu process, it is hard to tell for sure it belongs to this test case. However your comment made me review the code and I found some flaws....gonna send a v2.

wainersm:
Update this PR with:
 * added a check for the case there are more than one kata UID

dborquez (approved):
LGTM, thanks @wainersm. I saw an improvement but then you fixed it.

wainersm:
> LGTM, thanks @wainersm. I saw an improvement but then you fixed it.

hey @dborquez , thanks for reviewing!

snir911 (approved):
thanks!

integration/rootless/rootless_test.sh
```diff
@@ -27,19 +27,39 @@ setup() {
...
+	local grep_uid="grep -e '^kata-[0-9]\+' /etc/passwd"
+	[ "$(eval $grep_uid | wc -l)" -eq 1 ] || \
+		die "Unexpected more than one kata temporary UID"
+	local kata_uid="$(eval $grep_uid | cut -d: -f3)"
+	local kata_gid="$(eval $grep_uid | cut -d: -f4)"
+	[[ -n "$kata_uid" && -n "$kata_gid" ]] || \
+		die "Unabled to find the kata temporary UID"
+
+	# Assert the QEMU process is owned by that user.
+	ps -u "$kata_uid" | grep "qemu" &>/dev/null || \
```
gkurz, line 48: Is it possible that some other unrelated process gets caught by this `grep "qemu"`  ? What about extracting the name of the QEMU executable out of `configuration.toml` and passing its `basename` to something like `ps -C ${qemu_binary} -o uid --no-headers` ?

c3d: Isn't that the case that if there is more than one process, it's likely to be another qemu instance? How would your ps help? IOW, what are you trying to filter out? A "qemu-img" process for example?

c3d: Ah, unless you were thinking about the frequent case where a `ps | grep` finds the `grep` process. But then that cannot happen here because the `grep` does not run as user `$kata_uid`.

wainersm: > Is it possible that some other unrelated process gets caught by this `grep "qemu"` ? What about extracting the name of the QEMU executable out of `configuration.toml` and passing its `basename` to something like `ps -C ${qemu_binary} -o uid --no-headers` ?

The way I understood rootless, kata is going to create a temporary user (kata-`RANDOM_ID`) for every and each sandbox. The QEMU process and some control files should be owned by that user. Here I list all the processes owned by the that temporary user, and checking there is a match to "qemu". It is hard to tell exactly what is the binary name because the build scripts allow users to overwrite it; for example, there are jobs using "qemu-tdx". An alternative is to get the path to the hypervisor set on kata's config file.

Do you think it should be more accurate?

wainersm: ah, there is also another approach: call the `check_processes()` on setup and fail if any QEMU process is actually running at test time. It should work fine with CI but it mightl make developers unhappy when they are running the tests locally.

gkurz: > > Is it possible that some other unrelated process gets caught by this `grep "qemu"` ? What about extracting the name of the QEMU executable out of `configuration.toml` and passing its `basename` to something like `ps -C ${qemu_binary} -o uid --no-headers` ?
> 
> The way I understood rootless, kata is going to create a temporary user (kata-`RANDOM_ID`) for every and each sandbox. The QEMU process and some control files should be owned by that user. Here I list all the processes owned by the that temporary user, and checking there is a match to "qemu". It is hard to tell exactly what is the binary name because the build scripts allow users to overwrite it; for example, there are jobs using "qemu-tdx". An alternative is to get the path to the hypervisor set on kata's config file.

My suggestion is precisely to get the path to QEMU from the config file :smiley:

> 
> Do you think it should be more accurate?

Yes.

wainersm:
Updated this PR with:
 - call `extract_kata_env` on `setup` then use `HYPERVISOR_PATH` to grep the process with better accuracy (address @gkurz 's suggestion)

integration/rootless/rootless_test.sh
wainersm: @gkurz ok, what do you think now?

integration/rootless/rootless_test.sh
```diff
@@ -27,19 +28,40 @@ setup() {
...
 	waitForProcess 15 3 "sudo crictl inspectp "$pod_id" | jq '.status.state' | grep 'SANDBOX_READY'"
+
+	# There must be created one and only one temporary user.
+	local grep_uid="grep -e '^kata-[0-9]\+' /etc/passwd"
+	[ "$(eval $grep_uid | wc -l)" -eq 1 ] || \
+		die "Unexpected more than one kata temporary UID"
+	local kata_uid="$(eval $grep_uid | cut -d: -f3)"
+	local kata_gid="$(eval $grep_uid | cut -d: -f4)"
+	[[ -n "$kata_uid" && -n "$kata_gid" ]] || \
+		die "Unabled to find the kata temporary UID"
```
gkurz, line 46: s/Unabled/Unable

integration/rootless/rootless_test.sh
```diff
@@ -27,19 +28,40 @@ setup() {
...
+
+	# Assert the QEMU process is owned by that user.
+	ps -u "$kata_uid" -o cmd --no-header | \
+		grep -q "^${HYPERVISOR_PATH:-qemu}" || \
```
gkurz, line 50: If the QEMU path isn't present in the config file, `kata-runtime kata-env` falls back on the hard-coded default value `qemu-system-XXXX`  => `HYPERVISOR_PATH` is always set and non-null AFAICT.

integration/rootless/rootless_test.sh
```diff
@@ -27,19 +28,40 @@ setup() {
...
+		die "Expects the QEMU process be owned by a non-root user"
+
+	# Assert the QEMU files are owned by that user too.
```
gkurz, line 53: How are we sure that QEMU has created any file at this point ?

integration/rootless/rootless_test.sh
wainersm: I was worried about a bug on `extract_kata_env` that would export a empty string.

integration/rootless/rootless_test.sh
wainersm: you are right. Gonna fix it.

wainersm:
Updated this PR:
 - To address the comment of @gkurz about the necessity to check if the QEMU files really exist.

integration/rootless/rootless_test.sh
```diff
@@ -27,19 +28,45 @@ setup() {
...
+	local pod_dir="/run/user/${kata_uid}/run/vc/vm/${pod_id}"
+	[ -d "$pod_dir" ] || \
+		die "Unable to find the pod directory: $pod_dir"
+	local pod_files=($(sudo find "$pod_dir"))
+	[ "${#pod_files[@]}" -gt 0 ] || \
+		die "No pod files found in the directory: $pod_dir"
```
gkurz, line 59: Shouldn't we `waitForProcess` for the pod dir and files to show up ?

wainersm: At this point the files should already exist otherwise something is wrong because on line 37 we wait the pod to show up (https://github.com/kata-containers/tests/pull/4655/commits/b1716d9b5d2823984797110946782708a7a9d586#diff-c969ae5cb95b0b60400857625d05b7e33264a94d3fadbef39ca9db6c151d299fR37)

gkurz: @wainersm are you saying that a POD reaching SANDBOX_READY guarantees that the QEMU process already created its pid file, QMP socket ... etc... ?

gkurz: > @wainersm are you saying that a POD reaching SANDBOX_READY guarantees that the QEMU process already created its pid file, QMP socket ... etc... ?

@jodh-intel @egernst @fidencio any insights to share ?

fidencio: Take this with a grain of salt, okay?

My understanding is that when we reach SANDBOX_READY we're good to go, it means the SANDBOX was created but the containers inside it are not yet started, but from the sandbox perspective (VMM and its helper processes) we're fine.

fidencio: Just to complement: https://github.com/kata-containers/kata-containers/blob/604a7950738ac7b6d2d621fa262f6ada7f718aef/src/runtime/virtcontainers/sandbox.go#L474-L516 is where we set the StateReady for the sandbox.

Earlier in that function, we call: https://github.com/kata-containers/kata-containers/blob/604a7950738ac7b6d2d621fa262f6ada7f718aef/src/runtime/virtcontainers/sandbox.go#L518-L607, which is responsible for creating the VM (see: https://github.com/kata-containers/kata-containers/blob/604a7950738ac7b6d2d621fa262f6ada7f718aef/src/runtime/virtcontainers/sandbox.go#L597-L600)

So, yeah, we're good to go.

gkurz: > So, yeah, we're good to go.

I'm not that sure because `createSanbox()` is [called **before** QEMU is even started](https://github.com/kata-containers/kata-containers/blob/604a7950738ac7b6d2d621fa262f6ada7f718aef/src/runtime/virtcontainers/api.go#L61-L91) by `startVM()`. If the `StateReady` state is visible in the meantime then we cannot assume that QEMU has done anything at all.

fidencio: You're right, @gkurz.

We should be looking for `StateRunning` in the Sandbox, as `s.hypervisor.CreateVM()` will build the QEMU command line and **not** start the VM, my bad here.

wainersm: Thanks @gkurz and @fidencio ! this discussion has been very educational to me. I will change the script to wait on the files to show up with a small timeout.

wainersm:
Updated this PR with:
 * Added  loop through known files (console.sock qmp.sock pid vhost-fs.sock) and `waitForProcess` to wait the file be created before checking its ownership

@gkurz Does it address your concerns?

integration/rootless/rootless_test.sh
```diff
@@ -27,19 +28,46 @@ setup() {
...
+	local pod_dir="/run/user/${kata_uid}/run/vc/vm/${pod_id}"
+	[ -d "$pod_dir" ] || \
+		die "Unable to find the pod directory: $pod_dir"
+	local pod_files=(console.sock qmp.sock pid vhost-fs.sock)
+	for file_name in ${pod_files[@]}; do
+		local file="${pod_dir}/${file_name}"
+		waitForProcess 10 1 "test -e "$file"" || \
+			die "File didn't show up: $file"
+		[ "$(sudo stat -c %u-%g $file)" == "${kata_uid}-${kata_gid}" ]
+	done
```
gkurz, line 63: > Updated this PR with:
> 
>     * Added  loop through known files (console.sock qmp.sock pid vhost-fs.sock) and `waitForProcess` to wait the file be created before checking its ownership
> 
> 
> @gkurz Does it address your concerns?

Yes for the files but according to our findings, it is possible that the QEMU process hasn't even started yet. It seems that the `ps` check should also go through `waitForProcess`.

wainersm: oh, True. Forgot that.

wainersm:
Updated this PR with:
 * added a couple of `waitForProcess` to ensure resources are available

gkurz (approved):
This looks fine and race-free now :smiley:. Thanks @wainersm !

wainersm:
> This looks fine and race-free now smiley. Thanks @wainersm !

Thanks for the very good reviews, @gkurz !

Issue #4653, opened 2022-04-01
Intermittent failures in integration/kubernetes/k8s-replication.bats


gkurz:
Sometimes the test fails because it doesn't find the replication controller even though the traces of the teardown path show it is present:

```
not ok 1 Replication controller
# (in test file k8s-replication.bats, line 30)
#   `grep "replication-controller"' failed		<<< no "replication-controller" in the output of "kubectl describe"
# INFO: k8s configured to use runtimeclass
# replicationcontroller/replicationtest created		<<< replication controller is officially created
# Name:         replicationtest				<<< "kubectl describe" on the teardown path
# Namespace:    default
# Selector:     app=nginx-rc-test
# Labels:       app=nginx-rc-test
# Annotations:  <none>
# Replicas:     1 current / 1 desired
# Pods Status:  0 Running / 1 Waiting / 0 Succeeded / 0 Failed
# Pod Template:
#   Labels:  app=nginx-rc-test
#   Containers:
#    nginxtest:
#     Image:        quay.io/sjenning/nginx:1.15-alpine
#     Port:         80/TCP
#     Host Port:    0/TCP
#     Environment:  <none>
#     Mounts:       <none>
#   Volumes:        <none>
# Events:
#   Type    Reason            Age   From                    Message
#   ----    ------            ----  ----                    -------
#   Normal  SuccessfulCreate  0s    replication-controller  Created pod: replicationtest-mdmjs
                                    ^^^^^^^^^^^^^^^^^^^^^^^
                                  would have made grep happy
# replicationcontroller "replicationtest" deleted
[run_kubernetes_tests.sh:109] ERROR: bats k8s-replication.bats
```

c3d:
Also saw that with the local CI script.

```
not ok 1 Replication controller
# (in test file k8s-replication.bats, line 30)
#   `grep "replication-controller"' failed
# INFO: k8s configured to use runtimeclass
# replicationcontroller/replicationtest created
# Name:         replicationtest
# Namespace:    default
# Selector:     app=nginx-rc-test
# Labels:       app=nginx-rc-test
# Annotations:  <none>
# Replicas:     0 current / 1 desired
# Pods Status:  0 Running / 1 Waiting / 0 Succeeded / 0 Failed
# Pod Template:
#   Labels:  app=nginx-rc-test
#   Containers:
#    nginxtest:
#     Image:        quay.io/sjenning/nginx:1.15-alpine
#     Port:         80/TCP
[...]

PR #4659, opened 2022-04-04
k8s: Wait for k8s replication controller to show up

c3d:
This avoids intermittent failures in the CI

Fixes: #4653

Signed-off-by: Christophe de Dinechin <dinechin@redhat.com>
Suggested-by: Greg Kurz <groug@kaod.org>

### DIFF
--- a/integration/kubernetes/k8s-replication.bats
+++ b/integration/kubernetes/k8s-replication.bats
@@ -26,8 +26,8 @@ setup() {
 	kubectl create -f "${pod_config_dir}/test-replication-controller.yaml"
 
 	# Check replication controller
-	kubectl describe replicationcontrollers/"$replication_name" | \
-		grep "replication-controller"
+	local cmd="kubectl describe replicationcontrollers/$replication_name | grep replication-controller"
+	waitForProcess "$wait_time" "$sleep_time" "$cmd"
 
 	number_of_replicas=$(kubectl get replicationcontrollers/"$replication_name" \
 		--output=jsonpath='{.spec.replicas}')


c3d:
@gkurz This is essentially  your fix with a different way to set the variable

c3d:
@gkurz @wainersm I had added this to my local CI in a loop running over the week-end. but the laptop panicked (twice, actually), so I don't know for sure how good it is.

integration/kubernetes/k8s-replication.bats
```diff
@@ -26,8 +26,8 @@ setup() {
 	kubectl create -f "${pod_config_dir}/test-replication-controller.yaml"
 
 	# Check replication controller
-	kubectl describe replicationcontrollers/"$replication_name" | \
-		grep "replication-controller"
+        kc_rc_chk="kubectl describe replicationcontrollers/$replication_name | grep replication-controller"
+        waitForProcess "$wait_time" "$sleep_time" "$kc_rc_chk"
```
gkurz, line 30: Rest of the script is indented with tabs. Appart from that, LGTM.

integration/kubernetes/k8s-replication.bats
```diff
@@ -26,8 +26,8 @@ setup() {
 	kubectl create -f "${pod_config_dir}/test-replication-controller.yaml"
 
 	# Check replication controller
-	kubectl describe replicationcontrollers/"$replication_name" | \
-		grep "replication-controller"
+        kc_rc_chk="kubectl describe replicationcontrollers/$replication_name | grep replication-controller"
```
fidencio, line 29: What does `kc_rc_chk` stands for?
I'd go for a more descriptive variable name here, @c3d.

Apart from this, LGTM.

fidencio (changes requested):

integration/kubernetes/k8s-replication.bats
wainersm: or just call that `cmd` as in many other scripts. Ah, and make it a `local` variable :)

c3d: Done. `cmd` is actually used a few lines below.

c3d:
@fidencio Changed. To answer your question, `kc_rc_chk` was short `kubectl replication-controller check`. But since `cmd` is used in the same function, let's stick with that.

fidencio (approved):
lgtm, thanks @c3d!

wainersm (approved):
Thanks @c3d !

wainersm:
The job jenkins-ci-centos-8-stream-x86_64-crio_k8s  failed with same error as I've seen from time to time on OpenShift CI. It seems that use Stream 8 for CI isn't going to work out as this is a moving target (so subject to intermittent fails). @c3d did you create a issue about using almaLinux/Rocky instead, right?

c3d:
@wainersm Yes, there is some ongoing work with AlmaLinux, but it's not finished yet (I spent my time today working around a libguestfs bug that impacts Ubuntu 20.04)

gkurz (approved):
LGTM. Thanks @c3d !

Issue #4667, opened 2022-04-05
Enable TEE related CI


arronwy:
Add TEE_TYPE flag to generate TEE related guest kernel, hypervisor and related runtime config.

PR #4668, opened 2022-04-05
ci: Add TEE_TYPE flag for TEE enabled kernel/hypervisor

arronwy:
Fixes: #4667

Signed-off-by: Arron Wang <arron.wang@intel.com>

### DIFF
--- a/.ci/install_kata.sh
+++ b/.ci/install_kata.sh
@@ -13,15 +13,27 @@ cidir=$(dirname "$0")
 tag="${1:-""}"
 source /etc/os-release || source /usr/lib/os-release
 source "${cidir}/lib.sh"
+KATA_BUILD_KERNEL_TYPE="${KATA_BUILD_KERNEL_TYPE:-vanilla}"
+KATA_BUILD_QEMU_TYPE="${KATA_BUILD_QEMU_TYPE:-vanilla}"
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 experimental_qemu="${experimental_qemu:-false}"
+TEE_TYPE="${TEE_TYPE:-}"
+
+if [ -n "${TEE_TYPE}" ]; then
+	echo "Install with TEE type: ${TEE_TYPE}"
+fi
+
+if [ "${TEE_TYPE:-}" == "tdx" ]; then
+	KATA_BUILD_KERNEL_TYPE="${KATA_BUILD_KERNEL_TYPE:-tdx}"
+	KATA_BUILD_QEMU_TYPE="${KATA_BUILD_QEMU_TYPE:-tdx}"
+fi
 
 echo "Install Kata Containers Image"
 echo "rust image is default for Kata 2.0"
 "${cidir}/install_kata_image.sh" "${tag}"
 
 echo "Install Kata Containers Kernel"
-"${cidir}/install_kata_kernel.sh" "${tag}"
+"${cidir}/install_kata_kernel.sh" -t "${KATA_BUILD_KERNEL_TYPE}"
 
 install_qemu(){
 	echo "Installing qemu"
@@ -29,7 +41,7 @@ install_qemu(){
 		echo "Install experimental Qemu"
 		"${cidir}/install_qemu_experimental.sh"
 	else
-		"${cidir}/install_qemu.sh"
+		"${cidir}/install_qemu.sh" -t "${KATA_BUILD_QEMU_TYPE}"
 	fi
 }
 
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -23,6 +23,7 @@ PREFIX="${PREFIX:-/usr}"
 DESTDIR="${DESTDIR:-/}"
 TEST_INITRD="${TEST_INITRD:-}"
 USE_VSOCK="${USE_VSOCK:-yes}"
+TEE_TYPE="${TEE_TYPE:-}"
 
 arch=$("${cidir}"/kata-arch.sh -d)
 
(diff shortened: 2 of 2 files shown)


.ci/setup.sh
```diff
@@ -100,7 +101,10 @@ enable_nested_virtualization() {
 install_kata() {
 	if [ "${INSTALL_KATA}" == "yes" ]; then
 		echo "Install Kata sources"
-		bash -f ${cidir}/install_kata.sh
+		if [ -n ${TEE_TYPE} ]; then
```
wainersm, line 104: Nit: missing double quotes (i.e. `"${TEE_TYPE}"`)

wainersm:
Hi @arronwy , first, thanks for splitting this out of https://github.com/kata-containers/tests/pull/4662 .

.ci/install_runtime.sh
```diff
@@ -162,3 +163,20 @@ if [ "$arch" == "aarch64" -a "${KATA_HYPERVISOR}" == "qemu" -a "${ENABLE_ARM64_U
 	#enable pflash
 	sudo sed -i 's|#pflashes|pflashes|' "${runtime_config_path}"
 fi
+
+if [ "$TEE_TYPE" == "tdx" ]; then
```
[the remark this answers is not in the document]
arronwy: Yes, infact we try to test TDX features at different level,  for the tests in `functional` directory, it may ensure TDX functon is ok at kata runtime level. For the tests in `integration` directory, we want test the feature in E2E stack which can support different combination with different config: 
https://github.com/kata-containers/ci/tree/main/jenkins/jobs

We want test TDX stack works well in confidential container E2E pipeline like image downloading in guest/ image decryption/signature verification and more tests to be added.

.ci/setup.sh
arronwy: Yes, thanks, fixed now :-)

.ci/install_runtime.sh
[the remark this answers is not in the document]
wainersm: Hi @arronwy, I think that now I understood the goal of this PR. I'm okay with having the installation and configuration at the job setup stage.

.ci/install_kata.sh
```diff
@@ -10,7 +10,7 @@ set -o nounset
 set -o pipefail
 
 cidir=$(dirname "$0")
-tag="${1:-""}"
+tag="${1:-"vanilla"}"
```
wainersm, line 13: I don't remember where I mentioned that originally this `tag` option was meant to be a version of kata (i.e. a tag in the git repository). Here we are changing the semantic of `tag` completely.

What if we do the following:
 - Introduce two new variables: `KATA_BUILD_KERNEL_TYPE`  and `KATA_BUILD_QEMU_TYPE` (don't need to be exactly those names). They can have the values of `vanilla` (default if not set), `experimental` and `tdx`.
 - In `install_kata.sh` have a block to initialize those (code below is just a draft):
   ```
   KATA_BUILD_KERNEL_TYPE="${KATA_BUILD_KERNEL_TYPE:-vanilla}"
   KATA_BUILD_QEMU_TYPE="${KATA_BUILD_QEMU_TYPE:-vanilla}"
   if [ "${TEE_TYPE:-}" == "tdx" ]; then
       # This allow one to even export a different TDX type, like KATA_BUILD_KERNEL_TYPE=my-tdx-version
       KATA_BUILD_KERNEL_TYPE="${KATA_BUILD_KERNEL_TYPE:-tdx}"
       ....
       ....
   fi
   ```
 - Finally call the respective install scripts with `-t` ( `-t ${KATA_BUILD_KERNEL_TYPE}`, for example).
 - Kill the `experimental_qemu` and `experimental_kernel` variables. Okay, this can be done in a follow up PR as involve many other considerations I believe.

What do you think @arronwy ?
Any opinion on that @devimc, @jcvenegas ?

.ci/install_runtime.sh
```diff
@@ -162,3 +163,20 @@ if [ "$arch" == "aarch64" -a "${KATA_HYPERVISOR}" == "qemu" -a "${ENABLE_ARM64_U
...
+        sudo sed -i -e 's/^# confidential_guest/confidential_guest/' "${runtime_config_path}"
+        sudo sed -i -e 's/^kernel_params = "\(.*\)"/kernel_params = "\1 force_tdx_guest tdx_disable_filter "/g' "${runtime_config_path}"
+
+        case "${KATA_HYPERVISOR}" in
+		"cloud-hypervisor")
+			sudo sed -i -e 's/^firmware = ".*"/firmware = "\/usr\/share\/td-shim\/final-pe.bin"/' "${runtime_config_path}"
+			;;
+
+		"qemu")
+			sudo sed -i -e 's/^firmware = ".*"/firmware = "\/usr\/share\/qemu\/OVMF.fd"/' "${runtime_config_path}"
```
devimc, line 179: I think `firmware_volume` is also needed, isn't it?

.ci/setup.sh
```diff
@@ -100,7 +101,10 @@ enable_nested_virtualization() {
 install_kata() {
 	if [ "${INSTALL_KATA}" == "yes" ]; then
 		echo "Install Kata sources"
-		bash -f ${cidir}/install_kata.sh
+		if [ -n "${TEE_TYPE}" ]; then
+			echo "Using with TEE type: ${TEE_TYPE}"
+		fi
+		bash -f ${cidir}/install_kata.sh "${TEE_TYPE}"
```
wainersm, line 107: Hi @arronwy ! Let me ask you one last thing... here I think you should continue passing the `tag` and remove https://github.com/kata-containers/tests/pull/4668/commits/19aa47e84581228edc0b7d81fc6720c833793fe8#diff-488b8054ced3f482b39437df10e2189b4acef7470bfb5d0ce4b70a7cff5cbbedR13 . The value of `TEE` is exported and the `install_kata.sh` handles it; so no need to pass a script parameter.

arronwy: @wainersm , thanks, fixed now.

.ci/install_runtime.sh
arronwy: Yes, it is optional, but we should also add it.

fidencio:
@arronwy, this change LGTM.

I will leave the final approval to @wainersm!

wainersm (approved):

wainersm:
The ppc64le job failed because another process had locked apt. Maybe because concurrent jobs running? Anyway, I am gonna re-run the job.

/retest-power

PR #4675, opened 2022-04-06
Prepare to merge CCv0 into main: introduce the KATA_BUILD_CC variable

wainersm:
In order to merge the Confidential Containers (CC) CCv0 branch into main (https://github.com/kata-containers/tests/issues/4441), there will be needed to differentiate the "vanilla" and "CC" builds of Kata Containers because the later depends on a [fork of containerd](https://github.com/confidential-containers/containerd). In other words, in some situations the scripts will need to tweak the code for CC. Thus, this pull requests introduces the `KATA_BUILD_CC` variable, which should be use exported in the environment (`export KATA_BUILD_CC=yes`) to indicate the build target is for CC.

Apart from this new variable, there are created:

1. the `CC_CRI_CONTAINERD` and `CC_CRI_CONTAINERD_CLOUD_HYPERVISOR`  CI job configurations. They will be used on near future to test more merges, being #4627 likely the first. Those jobs are configured to run the current containerd tests (i.e. run the same tests as the `CRI_CONTAINERD` configuration) plus CC specific tests (at this point no specific tests are merged on main, so it will do nothing).
2. the `cc-containerd` makefile target meant to run CC specific tests with CRI+Containerd. As mentioned above, there aren't specific tests into the main branch yet, so this make target is just a placeholder for now.

Cc @kata-containers/architecture-committee @stevenhorsman @fidencio @jodh-intel 

Fixes #4626

Issue #4628, opened 2022-03-25
ci: Add integration test for the confidential-containers changes scenarios


stevenhorsman:
After https://github.com/kata-containers/tests/issues/4627 and https://github.com/kata-containers/kata-containers/issues/3996 are complete we should have the CI/CD jobs set-up to run against the CC parameter (just regression testing and using the containerd fork until the runtime changes are merged) and all of the development changes from the `CCv0` branch into `main`. At this point we should be able to start adding in our confidential-containers integration tests, wherever we are up to based on [the plan](https://github.com/confidential-containers/community/issues/2) into `main` and they should run.

See https://github.com/kata-containers/tests/issues/4441 for some overall discussion of the merge

PR #4677, opened 2022-04-06
tests: add test case for kata agent API configuration

### LINKED ISSUE #4676: Add test cases for the kata agent API configuration
The kata agent can be configured to block the access API endpoints.

There aren't test cases for that functionality though. Implement e2e tests for the following scenarios:

"Check that when a Kata CC pod is created with an agent api config that doesn't restrict anything that the endpoints work"
"Check that when a Kata CC pod is created with an agent api config that blocked certain endpoints that requests to these are rejected"

Those are originally reported on issue https://github.com/confidential-containers/community/issues/3

wainersm:
Added a e2e test to check that when a pod is created with an agent API
config that blocked certain endpoints that requests to these are rejected.

Fixes #4676

### DIFF
--- a/integration/containerd/confidential/agent_api.bats
+++ b/integration/containerd/confidential/agent_api.bats
@@ -5,19 +5,8 @@
 #
 
 load "${BATS_TEST_DIRNAME}/lib.sh"
+load "${BATS_TEST_DIRNAME}/asserts.sh"
 load "${BATS_TEST_DIRNAME}/../../../lib/common.bash"
-
-# Check the containerd logged messages on host have a given message.
-#
-# Parameters:
-#      $1 - the message
-#
-# Note: get the logs since the global $start_date.
-#
-assert_containerd_logs_contain() {
-	local message="$1"
-	journalctl -x -t containerd --since "$start_date" | grep "$message"
-}
 
 setup() {
 	start_date=$(date +"%Y-%m-%d %H:%M:%S")
@@ -78,7 +67,7 @@ setup() {
 	# operation should fail.
 	! crictl exec "$container_id" cat /proc/cmdline
 
-	assert_containerd_logs_contain "ExecProcessRequest is blocked"
+	assert_logs_contain "ExecProcessRequest is blocked"
 }
 
 teardown() {
--- a/integration/containerd/confidential/agent_api.bats
+++ b/integration/containerd/confidential/agent_api.bats
@@ -39,13 +39,9 @@ setup() {
 
 	# Check that the agent allow ExecProcessRequest requests by default.
 	#
+	echo "Check can create a container and exec a command"
 	crictl_create_cc_pod "$pod_config"
-	crictl_create_cc_container "$sandbox_name" "$pod_config" \
-		"$container_config"
-
-	pod_id=$(crictl pods --name "$sandbox_name" -q)
-	container_id=$(crictl ps --pod ${pod_id} -q)
-	crictl exec "$container_id" cat /proc/cmdline
+	assert_container "$container_config"
 
 	# Check that the agent endpoints can be restricted. In this case it will
 	# have ExecProcessRequest blocked.
@@ -61,12 +57,11 @@ setup() {
 	crictl_create_cc_container "$sandbox_name" "$pod_config" \
 		"$container_config"
 
-	pod_id=$(crictl pods --name "$sandbox_name" -q)
-	container_id=$(crictl ps --pod ${pod_id} -q)
-	# The endpoint ExecProcessRequest is not allowed so this exec
+	# The endpoint ExecProcessRequest is not allowed so any exec
 	# operation should fail.
-	! crictl exec "$container_id" cat /proc/cmdline
-
+	echo "Check cannot exec on container"
+	! assert_can_exec_on_container
+	echo "Check failed to exec because the endpoint is blocked"
 	assert_logs_contain "ExecProcessRequest is blocked"
 }
 
--- a/integration/containerd/confidential/agent_api.bats
+++ b/integration/containerd/confidential/agent_api.bats
@@ -1,0 +1,99 @@
+#!/usr/bin/env bats
+# Copyright (c) 2022 Red Hat
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/lib.sh"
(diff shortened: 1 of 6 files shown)


wainersm:
Hi! I'm labeling 'rfc' this PR because so far my goal is to collect feedback.

I'm introducing a method (`cp_to_guest_img()`) to allow us to copy files to the guest image. Currently, on CCv0 branch we build the image with some files for testing included, but we likely must stop doing that when merge the branches. In a particular chat with @fidencio he suggested the use of volumes (IIUC pod volumes) but I should have the file inside the guest at the time the agent starts. Did you like that approach?

I noticed that when I enable the "CC" handler in containerd's configuration and I *don't* enable image offloading on kata side then it fails to start the container. Is it expected behavior? i.e. once you add the handler, you are completely disabling the image pulls from containerd (delegated to the kata agent) so it is mandatory to enable image offload on kata?

My last remark is that it seems we don't necessarily need to test agent API configuration with CC. If that is the case maybe I should try to merge that test case straight to the main branch. Or am I missing something?

integration/containerd/confidential/agent_api.bats
```diff
@@ -0,0 +1,66 @@
...
+	crictl_create_cc_container "$sandbox_name" "$pod_config" \
+		"$container_config"
+
+	local pod_id=$(crictl pods --name "$sandbox_name" -q)
+        local container_id=$(crictl ps --pod ${pod_id} -q)
+	# The endpoint "ExecProcessRequest" is not allowed so this exec
+	# operation should fail
+	! crictl exec "$container_id" cat /proc/cmdline
+
+	# TODO: grep the logs to be sure if failed because of the blocked endpoint.
```
amshinde, line 53: @wainersm Are you planning to add this check for logs ?

wainersm: Hi @amshinde , yes, I will lift the rfc label and send a new version of this PR where I am going to introduce that check.

wainersm:
Updates to this pull request:
 * added code to save/restore the guest image.

wainersm:
Updates to this pull request:
  * Added a checker on logs to ensure the cause of the failure was the API blocked
  * Added a checker for the default case: no passing a configuration file, the exec API should work
  * Created a shared library (asserts.sh) to hold all assert functions
  * using those asserts in both agent_image and agent_api tests

integration/containerd/confidential/agent_api.bats
```diff
@@ -0,0 +1,83 @@
...
+	echo "Delete any existing ${sandbox_name} pod"
+	crictl_delete_cc_pod_if_exists "$sandbox_name"
+
+	echo "Prepare containerd for Confidential Container"
+	SAVED_CONTAINERD_CONF_FILE="/etc/containerd/config.toml.$$"
+	configure_cc_containerd "$SAVED_CONTAINERD_CONF_FILE"
+
+	echo "Reconfigure Kata Containers"
+	clear_kernel_params
+	switch_image_service_offload on
```
stevenhorsman, line 26: As this code matches the `agent_image.bats` I wonder if there is merit in moving it to a shared library to avoid the duplication?

stevenhorsman (approved):
Hi Wainer, sorry I missed this PR. I think it looks good and having a method to add files to the guest image is going to be really helpful and something we should switch to using in `ccv0.sh`.
With respect to your question about where to test agent API - I think you are correct that the function went into `main` and your test case would be valid in a non CC environment.

wainersm:
Update this PR with what I think the last missing bits: added handler for initrd in `cp_to_guest_img()` (https://github.com/kata-containers/tests/pull/4677/commits/40d7f5d0a9428bc2ef2bc455bcb83d13ff1dbfe0). Sending in a separate commit just to ease the reviews but I can squash into b306e7a726d6f60322236ae34faf38f6f6228177 before the merge.

integration/containerd/confidential/agent_api.bats
wainersm: Yeah, It does make sense to remove that duplication. I wish it was Python (or any OO language) where we could use heritage. Mind if I send it in a follow up PR?

wainersm:
Hi @stevenhorsman ,

> Hi Wainer, sorry I missed this PR. I think it looks good and having a method to add files to the guest image is going to be really helpful and something we should switch to using in `ccv0.sh`. With respect to your question about where to test agent API - I think you are correct that the function went into `main` and your test case would be valid in a non CC environment.

Can we adapt `ccv0.sh` after merging this one? Thinking aloud...maybe it can be done alongside the merge to `main`. Once we have https://github.com/kata-containers/tests/pull/4675 merged and ported to `CCv0`, we can work on preparing the osbuilder (**) which, among other changes, it will need to remove the signature and config files being added in the image at the build time.

** it seems we don't have an issue to track that work. At least the closest that I found was about removing skopeo/umoci (https://github.com/kata-containers/kata-containers/issues/3970)

---

Regarding where to merge this PR...I think it is too late to merge on `main` as it is too tied with the code in `CCv0`. Mind if I port it alongside the other CC tests when the time comes for https://github.com/kata-containers/tests/issues/4628?

stevenhorsman:
Hey Wainer, sorry for the lack of clarity, my comments weren't supposed to be merge stoppers, just something to consider.

> Can we adapt `ccv0.sh` after merging this one? Thinking aloud...maybe it can be done alongside the merge to `main`. Once we have #4675 merged and ported to `CCv0`, we can work on preparing the osbuilder (**) which, among other changes, it will need to remove the signature and config files being added in the image at the build time.

Agreed, I meant this is great code we should try and switch to using in `ccv0.sh` in a separate PR after this as refactoring to remove the hard-coded policy files from the dev code will be great. I don't think that should be in this PR though :)

> Regarding where to merge this PR...I think it is too late to merge on `main` as it is too tied with the code in `CCv0`. Mind if I port it alongside the other CC tests when the time comes for #4628?

Yep, that's fine by me. My comment was indented to mean that the scenario would be valid with the current `main` dev, but can be ported as part of the exisiting plans.

stevenhorsman:
Hey @wainersm - as you mentioned that it was missing, I've created https://github.com/kata-containers/kata-containers/issues/4111 to cover using the new `cp_to_guest_img` to inject the image container image signature verification files into the guest image. Please take a look and add an info & fix my mistakes. This might be something that someone at IBM can do soon, but no promises yet until we've done some planning :)

Issue #4679, opened 2022-04-07
allow kata-log-parser to unpack all it's fields


snir911:
**Which feature do you think can be improved?**

kata-log-parser

**How can it be improved?**

when agent msgs are unpacked currently only the level, msg, pid, source and level fields are unpacked, the rest of the fields are contained withing the original json string (which is printed)

PR #4631, opened 2022-03-28
tools: Enable agent log parsing in kata-log-parser

littlejawa:
kata-log-parser can't parse the json format logs from the agent.
Adding a way to show the agent's log as part of the general logs.

Also put the vmconsole's content in place of the generic message
"reading guest console" for messages that are not from the agent.

Fixes #4630

Signed-off-by: Julien Ropé <jrope@redhat.com>

PR #4680, opened 2022-04-07
log-parser: unpack all agent fields

snir911:
and remove general non-agent fields

Fixes: #4679

### DIFF
--- a/cmd/log-parser/agent.go
+++ b/cmd/log-parser/agent.go
@@ -114,21 +114,35 @@ func unpackAgentLogEntry_v2(le LogEntry) (agent LogEntry, err error) {
 		agent.Source = "vmconsole"
 		return agent, nil
 	}
+	// we were able to unpack agent msg, let's drop previous Data (runtime entries)
+	agent.Data = make(map[string]string)
 
-	pid, err := strconv.Atoi(result["pid"])
-	if err != nil {
-		return LogEntry{}, fmt.Errorf("failed to convert pid")
+	for k, v := range result {
+		switch k {
+		case "container-id", "cid": // better to ensure consistency in the agent
+			agent.Container = v
+		case "level":
+			agent.Level = strings.ToLower(v)
+		case "msg":
+			agent.Msg = v
+		case "name":
+			agent.Name = v
+		case "pid":
+			pid, err := strconv.Atoi(v)
+			if err != nil {
+				return LogEntry{}, fmt.Errorf("failed to convert pid")
+			}
+			agent.Pid = pid
+		case "source":
+			agent.Source = v
+		default:
+			// NOTE: we do not take the agent's timestamp into account, because there is a ~1sec delay
+			// for the agent's log to get through. Using the agent's timestamp would then make its logs
+			// appear out of order compared to other logs from the guest.
+			// The agent's logs timestamp is still visible for reference in the Data section of the logs.
+			agent.Data[k] = v
+		}
 	}
-
-	// NOTE: we do not take the agent's timestamp into account, because there is a ~1sec delay
-	// for the agent's log to get through. Using the agent's timestamp would then make its logs
-	// appear out of order compared to other logs from the guest.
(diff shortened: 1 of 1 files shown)


cmd/log-parser/agent.go
```diff
@@ -114,21 +114,36 @@ func unpackAgentLogEntry_v2(le LogEntry) (agent LogEntry, err error) {
...
+				return LogEntry{}, fmt.Errorf("failed to convert pid")
+			}
+			agent.Pid = pid
+		case "container-id", "cid": // better to ensure consistency in the agent, same for exec-id/eid
+			agent.Container = v
+		// NOTE: we do not take the agent's timestamp into account, because there is a ~1sec delay
+		// for the agent's log to get through. Using the agent's timestamp would then make its logs
+		// appear out of order compared to other logs from the guest.
+		// The agent's logs timestamp is still visible for reference in the Data section of the logs.
+		case "ts":
```
littlejawa, line 142: As it is, the timestamp will not be visible in the Data section (as explained in the comment above).
You need to add "fallthrough" here, or comment/remove the 'case "ts":' line.
I don't think we should totally remove the agent's timestamp though.

snir911: Right, done

jodh-intel (commented):
Thanks for raising @snir911. A few comments...

Also, could you expand the tests for this change?

cmd/log-parser/agent.go
```diff
@@ -114,21 +114,35 @@ func unpackAgentLogEntry_v2(le LogEntry) (agent LogEntry, err error) {
...
-
-	pid, err := strconv.Atoi(result["pid"])
-	if err != nil {
-		return LogEntry{}, fmt.Errorf("failed to convert pid")
+	// we were able to unpack agent msg, let's drop previous Data (runtime entries)
+	agent.Data = make(map[string]string)
+
+	for k, v := range result {
+		switch k {
+		case "container-id", "cid": // better to ensure consistency in the agent, same for exec-id/eid
```
gkurz, line 122: > same for exec-id/eid

Does this mean there should be a `case "exec-id", "eid":` as well or I'm missing something ?

cmd/log-parser/agent.go
```diff
@@ -114,21 +114,35 @@ func unpackAgentLogEntry_v2(le LogEntry) (agent LogEntry, err error) {
...
+			pid, err := strconv.Atoi(v)
+			if err != nil {
+				return LogEntry{}, fmt.Errorf("failed to convert pid")
+			}
+			agent.Pid = pid
+		case "container-id", "cid": // better to ensure consistency in the agent, same for exec-id/eid
+			agent.Container = v
+		default:
+			// NOTE: we do not take the agent's timestamp into account, because there is a ~1sec delay
+			// for the agent's log to get through. Using the agent's timestamp would then make its logs
```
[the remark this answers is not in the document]
littlejawa: Yes, this is the "over the wire" delay : it takes ~1sec for the log to go from the guest to the host.
So if I take the agent's time into account, but don't change the rest of the guest's logs timestamp, I can end up having boot logs from the guest coming up AFTER the agent saying it started successfully - which can be confusing for troubleshooting :-/ That's why I decided to keep it in the "Data" field, and not update the timestamp.

See #4631

cmd/log-parser/agent.go
snir911: Well, you could, however the parser always has a fixed "**Container**" field for each log entry, and **exec-id** is logged only if exist (and under Data).
I could combine **eid** and **exec-id** to be mapped to the same key but i thought it may be too intrusive and may confuse a developer who is grepping for the specific naming, however i plan to change it on the agent side.

gkurz: Then maybe just don't mention **eid/exec-id** in this comment ?

gkurz (approved):
LGTM. Thanks @snir911 !

GabyCT (approved):

Issue #4692, opened 2022-04-13
Install qemu error: unsafe repository ( qemu/qemu is owned by someone else)


liubin:
>GIT     ui/keycodemapdb meson dtc
warn: ignoring non-existent submodule ui/keycodemapdb
warn: ignoring non-existent submodule meson
warn: ignoring non-existent submodule dtc
fatal: unsafe repository ('/home/jenkins/workspace/kata-containers-2.0-ppc64le-containerd-k8s-ubuntu-20-04-PR/go/src/github.com/qemu/qemu' is owned by someone else)
To add an exception for this directory, call:
>
>	git config --global --add safe.directory /home/jenkins/workspace/kata-containers-2.0-ppc64le-containerd-k8s-ubuntu-20-04-PR/go/src/github.com/qemu/qemu
/home/jenkins/workspace/kata-containers-2.0-ppc64le-containerd-k8s-ubuntu-20-04-PR/go/src/github.com/qemu/qemu/scripts/git-submodule.sh: failed to update modules

PR #4693, opened 2022-04-13
qemu: use sudo to clone and make QEMU

liubin:
New git will check git repo's mode/permission, and will return error:

fatal: unsafe repository ('/some/path' is owned by someone else)

See:

https://github.blog/2022-04-12-git-security-vulnerability-announced/

for more information.

We have 3 repos run into this issue: kata-containers, containerd and
QEMU(only build from source for aarch64/ppc64le).

For the first two repos,
we can just use `git config --global --add safe.directory`.

But QEMU have
some submodules, we need to add all submodules to the safe
directory, so unlike kata-containers and containerd repos,
for QEMU we add `sudo -E` before commands run in the repo.

Fixes: https://github.com/kata-containers/tests/issues/4692

### DIFF
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -77,12 +77,12 @@ install_cached_qemu() {
 
 clone_qemu_repo() {
 	# check if git is capable of shadow cloning
-        git_shadow_clone=$(check_git_version "${GIT_SHADOW_VERSION}")
+	git_shadow_clone=$(check_git_version "${GIT_SHADOW_VERSION}")
 
 	if [ "$git_shadow_clone" == "true" ]; then
-		git clone --branch "${CURRENT_QEMU_VERSION}" --single-branch --depth 1 --shallow-submodules "${QEMU_REPO_URL}" "${GOPATH}/src/${gopath_qemu_repo}"
+		sudo -E git clone --branch "${CURRENT_QEMU_VERSION}" --single-branch --depth 1 --shallow-submodules "${QEMU_REPO_URL}" "${GOPATH}/src/${gopath_qemu_repo}"
 	else
-		git clone --branch "${CURRENT_QEMU_VERSION}" --single-branch --depth 1 "${QEMU_REPO_URL}" "${GOPATH}/src/${gopath_qemu_repo}"
+		sudo -E git clone --branch "${CURRENT_QEMU_VERSION}" --single-branch --depth 1 "${QEMU_REPO_URL}" "${GOPATH}/src/${gopath_qemu_repo}"
 	fi
 }
 
@@ -99,12 +99,12 @@ build_and_install_qemu() {
 	clone_qemu_repo
 
 	pushd "${GOPATH}/src/${gopath_qemu_repo}"
-	git fetch
-	[ -n "$(ls -A capstone)" ] || git clone https://github.com/qemu/capstone.git capstone
-	[ -n "$(ls -A ui/keycodemapdb)" ] || git clone  https://github.com/qemu/keycodemapdb.git ui/keycodemapdb
+	sudo -E git fetch
+	[ -n "$(ls -A capstone)" ] || sudo -E git clone https://github.com/qemu/capstone.git capstone
+	[ -n "$(ls -A ui/keycodemapdb)" ] || sudo -E git clone  https://github.com/qemu/keycodemapdb.git ui/keycodemapdb
 
 	# Apply required patches
-	${PACKAGING_DIR}/scripts/patch_qemu.sh ${CURRENT_QEMU_VERSION} ${PACKAGING_DIR}/qemu/patches
+	sudo -E ${PACKAGING_DIR}/scripts/patch_qemu.sh ${CURRENT_QEMU_VERSION} ${PACKAGING_DIR}/qemu/patches
 
 	echo "Build QEMU"
 	# Not all distros have the libpmem package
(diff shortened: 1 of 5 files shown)


gkurz (commented):
Hi @liubin !

Would you mind fixing your Signed-off-by: and From: to `Bin Liu <bin@hyper.sh>` ?

Apart from that LGTM.

gkurz:
@liubin I think the problem is that the `dtc`, `meson` and `ui/keycodemapdb` submodules are cloned from the `configure` script :

```
Build Qemu
Using './build' as the directory for build output
Submodule 'dtc' (https://gitlab.com/qemu-project/dtc.git) registered for path 'dtc'
Submodule 'meson' (https://gitlab.com/qemu-project/meson.git) registered for path 'meson'
Submodule 'ui/keycodemapdb' (https://gitlab.com/qemu-project/keycodemapdb.git) registered for path 'ui/keycodemapdb'
Cloning into '/home/jenkins/workspace/kata-containers-2.0-tests-ppc64le-containerd-k8s-ubuntu-20-04-PR/go/src/github.com/qemu/qemu/dtc'...
Cloning into '/home/jenkins/workspace/kata-containers-2.0-tests-ppc64le-containerd-k8s-ubuntu-20-04-PR/go/src/github.com/qemu/qemu/meson'...
Cloning into '/home/jenkins/workspace/kata-containers-2.0-tests-ppc64le-containerd-k8s-ubuntu-20-04-PR/go/src/github.com/qemu/qemu/ui/keycodemapdb'...
```

You might need that extra change:

```
-		fi | xargs ./configure --prefix=${PREFIX}
+		fi | sudo xargs ./configure --prefix=${PREFIX}
```

liubin:
@gkurz I'm debugging it now.

Or maybe the simplest is just what the error recommending `git config --global --add safe.directory`.

gkurz:
> @gkurz I'm debugging it now.
> 
> Or maybe the simplest is just what the error recommending `git config --global --add safe.directory`.

@liubin yes it is maybe more reasonable than adding `sudo`s all over the place. Could this be a `git config --local ...` ?

wainersm:
> > @gkurz I'm debugging it now.
> > Or maybe the simplest is just what the error recommending `git config --global --add safe.directory`.
> 
> @liubin yes it is maybe more reasonable than adding `sudo`s all over the place. Could this be a `git config --local ...` ?

I would prefer `--local` to avoid writing on the user's files. However `--local` doesn't fix the problem as @stevenhorsman noted in https://github.com/kata-containers/tests/pull/4699#issuecomment-1098219120

liubin:
**Note**: `git config --global --add safe.directory` will always append the target to `.gitconfig` without checking the existence of the target, so it's better to check it manually before adding the target repo.

gkurz (approved):
I would have personally split that into two patches: one for safe.directory and one for QEMU, but not worth holding up this PR for that.

Thanks for the investigation and fix @liubin !

.ci/install_qemu.sh
```diff
@@ -77,12 +77,12 @@ install_cached_qemu() {
 
 clone_qemu_repo() {
```
wainersm, line 78: Hi @liubin ,

IIUC you added `sudo` calls in `clone_qemu_repo()` because that function is called by  `lib_install_qemu_aarch64.sh` and `lib_install_qemu_ppc64le.sh`. Happens that `clone_qemu_repo()` is also callable by `build_and_install_qemu()` from `install_qemu.sh` when  `FORCE_BUILD_QEMU=yes`. Therefore, `build_and_install_qemu()` should have added the `sudo` calls too.

BTW, I don't know what is the use case for `FORCE_BUILD_QEMU`. It seems none CI job uses it. Anyone has any idea?

gkurz: @devimc what's the use case for `FORCE_BUILD_QEMU` ?

liubin: I searched `tests` and `ci` repos, not found anywhere used `FORCE_BUILD_QEMU, only aarch64, and ppc64le install QEMU from the source.

And `clone_qemu_repo` only be called in `lib_install_qemu_aarch64.sh`.

gkurz: > I searched `tests` and `ci` repos, not found anywhere used `FORCE_BUILD_QEMU, only aarch64, and ppc64le install QEMU from the source.
> 
> And `clone_qemu_repo` only be called in `lib_install_qemu_aarch64.sh`.

Right but the generic implementation of `build_and_install_qemu()` is also used by s390x. Let's fix this so that the s390x CI  doesn't hit the issue when it is re-enabled. WDYT @liubin ?

liubin: > > I searched `tests` and `ci` repos, not found anywhere used `FORCE_BUILD_QEMU, only aarch64, and ppc64le install QEMU from the source.
> > And `clone_qemu_repo` only be called in `lib_install_qemu_aarch64.sh`.
> 
> Right but the generic implementation of `build_and_install_qemu()` is also used by s390x. Let's fix this so that the s390x CI doesn't hit the issue when it is re-enabled. WDYT @liubin ?

Fixed, please take a look again.

gkurz (approved):
> > > I searched `tests` and `ci` repos, not found anywhere used `FORCE_BUILD_QEMU, only aarch64, and ppc64le install QEMU from the source. And `clone_qemu_repo`only be called in`lib_install_qemu_aarch64.sh`.
> > 
> > 
> > Right but the generic implementation of `build_and_install_qemu()` is also used by s390x. Let's fix this so that the s390x CI doesn't hit the issue when it is re-enabled. WDYT @liubin ?
> 
> Fixed, please take a look again.

LGTM. Approving again ! :smile: 

Now waiting for the tests to pass and another approval and some PRs will get unblocked at last.

Thanks very much for the fix @liubin !

fgiudici (approved):
LGTM, thanks @liubin

Issue #4700, opened 2022-04-13
log-parser: various naming to the same log-level


snir911:
# Description of problem

when running the log-parser on a bunch of logs you can get the following log levels:
error,erro,warning,warn,debug,debg,info and maybe others

# Expected result

consistent log level naming

# Actual result

multiple names per level

# Further information

this is because of log libs defaults logurs(go) and slog(rust)

PR #4701, opened 2022-04-13
log-parser: match agent log levels

snir911:
The agent logging carte "slog" is using short naming for its log leveling,
this is causing the parser to log redundant/duplicated log levels such as
"erro"/"error", "warn"/"warning"... this patch groups the common log levels

See: https://docs.rs/slog/2.0.6/slog/static.LOG_LEVEL_SHORT_NAMES.html

Fixes: #4700

### DIFF
--- a/cmd/log-parser/agent.go
+++ b/cmd/log-parser/agent.go
@@ -122,7 +122,24 @@ func unpackAgentLogEntry_v2(le LogEntry) (agent LogEntry, err error) {
 		case "container-id", "cid": // better to ensure consistency in the agent
 			agent.Container = v
 		case "level":
-			agent.Level = strings.ToLower(v)
+			switch v {
+			// match agent's slog short version log leveling to the common log levels
+			case "CRIT":
+				agent.Level = "critical"
+			case "DEBG":
+				agent.Level = "debug"
+			case "ERRO":
+				agent.Level = "error"
+			case "INFO":
+				agent.Level = "info"
+			case "TRCE":
+				agent.Level = "trace"
+			case "WARN":
+				agent.Level = "warning"
+			default:
+				agent.Level = v
+			}
+			agent.Data[k] = v // in any case let's leave original level under Data
 		case "msg":
 			agent.Msg = v
 		case "name":


GabyCT (approved):

littlejawa:
LGTM (but I don't think I have "reviewer" rights, so you still need someone else's review)

cmd/log-parser/agent.go
```diff
@@ -122,7 +122,22 @@ func unpackAgentLogEntry_v2(le LogEntry) (agent LogEntry, err error) {
...
+			case "ERRO":
+				agent.Level = "error"
+			case "INFO":
+				agent.Level = "info"
+			case "TRCE":
+				agent.Level = "trace"
+			case "WARN":
+				agent.Level = "warning"
+			default:
+				agent.Level = v
```
gkurz, line 138: @snir911 @littlejawa shouldn't "CRIT" be mapped to "critical", like the others ?

snir911: @gkurz i wasn't sure about it, [logrus log levels](https://github.com/sirupsen/logrus#level-logging) aren't 1:1 matching to the slog levels, at first I thought CRIT may be translated to **Fatal** (or **Panic**?) but then decided to stick only with the more common levels. I'm open to other thoughts though..

gkurz: @snir911 There's code in the agent to translate from logrus to slog :

```
// Map logrus (https://godoc.org/github.com/sirupsen/logrus)
// log level to the equivalent slog log levels.
//
// Note: Logrus names are used for compatability with the previous
// golang-based agent.
#[instrument]
fn logrus_to_slog_level(logrus_level: &str) -> Result<slog::Level> {
    let level = match logrus_level {
        // Note: different semantics to logrus: log, but don't panic.
        "fatal" | "panic" => slog::Level::Critical,

        "critical" => slog::Level::Critical,
        "error" => slog::Level::Error,
        "warn" | "warning" => slog::Level::Warning,
        "info" => slog::Level::Info,
        "debug" => slog::Level::Debug,

        // Not in logrus
        "trace" => slog::Level::Trace,

        _ => bail!(ERR_INVALID_LOG_LEVEL),
    };

    Ok(level)
}
```
IIUC CRIT is slog's highest level and it doesn't panic, so I guess it should be fine to translate it to "critical".

snir911: @gkurz that's odd, it doesn't seem that logrus has critical...
I'll add it

gkurz (approved):
LGTM. Thanks @snir911 !

Issue #4702, opened 2022-04-13
Remove duplicated die / info / warn functions


jodh-intel:
The `die()`, `info()` and `warn()` functions are defined in:

- https://github.com/kata-containers/tests/blob/main/lib/common.bash

However, lots of scripts like to provide their own implementations, for no obvious reason.

Rework the scripts to use the one true definition.

PR #4703, opened 2022-04-13
scripts: Remove duplicated die/info/warn functions

jodh-intel:
Rework the scripts that used to define their own `die()`, `info()`
and/or `warn()` functions to use the standard versions defined in
`lib/common.bash`.

Fixes: #4702.

### DIFF
--- a/.ci/aarch64/install_rom_aarch64.sh
+++ b/.ci/aarch64/install_rom_aarch64.sh
@@ -8,6 +8,9 @@ set -e
 
 source /etc/os-release || source /usr/lib/os-release
 
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${script_dir}/../../lib/common.bash"
+
 EDK2_REPO="https://github.com/tianocore/edk2.git"
 EDK2_PLAT_REPO="https://github.com/tianocore/edk2-platforms.git"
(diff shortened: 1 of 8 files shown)


.ci/aarch64/install_rom_aarch64.sh
```diff
@@ -112,8 +112,7 @@ install_uefi_flash()
 clean_up_and_die()
 {
 	clean_up
-	echo "ERROR: $*" >&2
-	exit 1
+	die "$*"
```
wainersm, line 115: Hi @jodh-intel , don't you need to source `lib/common.bash` in this script?

cmaf (commented):
@jodh-intel a bunch of the tests are failing with the following:
```
[manage_ctr_mgr.sh:79] ERROR: Docker is already installed. Please use -f flag to force new installation
```
Perhaps a missing `source` or other env variable?

.ci/aarch64/install_rom_aarch64.sh
jodh-intel: Good catch - fixed.

jodh-intel:
@cmaf - thanks. I think I've worked out what was happening - see the new 2nd commit ;)

cmaf (approved):
thanks @jodh-intel

fidencio (approved):
lgtm, thanks @jodh-intel!

PR #4711, opened 2022-04-19
ci: update scripts to support OpenShift 4.11

### LINKED ISSUE #4707: CI: Update job to OpenShift 4.11

We have a CI job running on OpenShift CI to test the update Kata Containers on Openshift. The job in question is https://prow.ci.openshift.org/?job=periodic-ci-kata-containers-kata-containers-main-4.10-e2e-tests and is currently running on OpenShift 4.10.

**How can it be improved?**

Update the job to run on 4.11 so we test on the to-be-the-next release.

**Additional Information**

The changes on the OpenShift CI side are in https://github.com/openshift/release/pull/27778

However the job fails with the following error:
```
Going to update //_out/build_install/opt/kata/share/kata-containers/kata-containers-initrd.img for kernel 4.18.0-348.20.1.el8_5.x86_64
88037 blocks
390218 blocks
INFO: re-mount /host/usr on rw mode
INFO: copy the Kata Containers installation over /host
ln: failed to create symbolic link '/var/usrlocal/bin/containerd-shim-kata-v2': File exists
ln: failed to create symbolic link '/var/usrlocal/bin/kata-collect-data.sh': File exists
ln: failed to create symbolic link '/var/usrlocal/bin/kata-monitor': File exists
ln: failed to create symbolic link '/var/usrlocal/bin/kata-runtime': File exists
ln: failed to create symbolic link '/var/usrlocal/bin/qemu-system-x86_64': File exists
[install_kata.sh:204] ERROR: Kata Containers seems not installed on some nodes
[test.sh:23] ERROR: Failed to install kata-containers
```

Therefore, it will be needed to fix the installer script at kata containers side.

wainersm:
In order to update the job on OpenShift CI to test on 4.11 there is needed to adapt the installer script (among a few changes).

Fixes #4707

### DIFF
--- a/.ci/openshift-ci/cluster/install_kata.sh
+++ b/.ci/openshift-ci/cluster/install_kata.sh
@@ -35,6 +35,10 @@ export DAEMONSET_NAME="kata-deploy"
 #
 export DAEMONSET_LABEL="kata-deploy"
 
+# The OpenShift Server version.
+#
+ocp_version="$(oc version -o json | jq -r '.openshiftVersion')"
+
 # Wait all worker nodes reboot.
 #
 # Params:
@@ -152,7 +156,16 @@ if [ "${KATA_WITH_SYSTEM_QEMU}" == "yes" ]; then
 	enable_sandboxedcontainers_extension
 	# Export additional information for the installer to deal with
 	# the QEMU path, for example.
-	oc apply -f ${deployments_dir}/configmap_installer_qemu.yaml
+	case $ocp_version in
+		4.10*)
+			oc apply -f ${deployments_dir}/configmap_installer_qemu_4_10.yaml
+			;;
+		4.11*)
+			oc apply -f ${deployments_dir}/configmap_installer_qemu.yaml
+			;;
+		*)
+			die "Unsupported OpenShift version: $ocp_version"
+	esac
 fi
 
 if [ "${KATA_WITH_HOST_KERNEL}" == "yes" ]; then
--- a/.ci/openshift-ci/images/entrypoint.sh
+++ b/.ci/openshift-ci/images/entrypoint.sh
@@ -130,7 +130,7 @@ rsync -O -a . "$HOST_MOUNT"
 # Ensure the binaries are searcheable via PATH. Notice it expects that
 # kata has been built with PREFIX=/opt/kata.
 #
-chroot "$HOST_MOUNT" sh -c 'for t in /opt/kata/bin/*; do ln -s "$t" /var/usrlocal/bin/; done'
+chroot "$HOST_MOUNT" sh -c 'for t in /opt/kata/bin/*; do ln --force -s "$t" /var/usrlocal/bin/; done'
 
 # Check the installation is good (or not).
 echo "INFO: run kata-check to check the installation is fine"
--- a/.ci/openshift-ci/cluster/deployments/configmap_installer_qemu_4_10.yaml
+++ b/.ci/openshift-ci/cluster/deployments/configmap_installer_qemu_4_10.yaml
@@ -10,5 +10,5 @@ kind: ConfigMap
 metadata:
   name: ci.kata.installer.qemu
 data:
-  qemu_path: /usr/libexec/qemu-kvm
+  qemu_path: /usr/libexec/qemu-kiwi
   host_kernel: "yes"
--- a/.ci/openshift-ci/images/Dockerfile.installer
+++ b/.ci/openshift-ci/images/Dockerfile.installer
@@ -5,7 +5,7 @@
 # Build the image which wraps the kata-containers installation along with the
 # install script. It is used on a daemonset to deploy kata on OpenShift.
 #
-FROM registry.centos.org/centos:8
+FROM quay.io/centos/centos:stream8
 
 RUN yum install -y rsync dracut && \
   yum clean all


GabyCT (approved):

wainersm:
@kata-containers/redhat volunteers to review this? :)

cpmeadors (commented):
What is the easiest way to run this manually to verify the changes?  Or do we have ci runs for 4.11 using this PR branch?

.ci/openshift-ci/images/entrypoint.sh
```diff
@@ -90,6 +90,31 @@ update_initrd_kernel_modules()
...
+handle_symlinks() {
+	local create_links_script="/tmp/create_links.sh"
+	cat <<-EOF > "${HOST_MOUNT}/${create_links_script}"
+	for t in /opt/kata/bin/*; do
+		executable="\$(basename \$t)"
+		executable_found="\$(which "\$executable" 2>/dev/null)"
+		if [[ -n "\$executable_found" && \
+			"\$(readlink "\$executable_found")" = "\$t" ]]; then
+			echo "Found \$executable_found"
+			continue
```
snir911, line 103: Does "/var/usrlocal/bin/" have precedence over any other PATH? and if it does, would it make sense to just forcefully link all the executables regardless to whether executable_found or not?
Also do you recall what links were existed? do you know if this is expected?

.ci/openshift-ci/cluster/install_kata.sh
```diff
@@ -152,7 +152,11 @@ if [ "${KATA_WITH_SYSTEM_QEMU}" == "yes" ]; then
 	enable_sandboxedcontainers_extension
 	# Export additional information for the installer to deal with
 	# the QEMU path, for example.
-	oc apply -f ${deployments_dir}/configmap_installer_qemu.yaml
+	if oc version | grep -q "^Server Version: 4.10"; then
```
jodh-intel, line 155: This looks fragile:

- What happens for v4.11?
- What if `oc` gets i18n'd? You could set `LC_ALL` / `LANG` to protect against that.

.ci/openshift-ci/images/entrypoint.sh
```diff
@@ -90,6 +90,31 @@ update_initrd_kernel_modules()
 	( cd "${rootfs_dir}" && find . | cpio -H newc -o | gzip -9 ) > ${initrd_path}
 }
 
+# Manage symlinks on host.
+handle_symlinks() {
+	local create_links_script="/tmp/create_links.sh"
+	cat <<-EOF > "${HOST_MOUNT}/${create_links_script}"
```
jodh-intel, line 96: Would it be simpler to just create this script in-tree and copy it into the correct location before executing it? All the quoting is going to make maintaining it awkward.

.ci/openshift-ci/cluster/install_kata.sh
wainersm: Hi @jodh-intel , currently only 4.11 and 4.10 is supported but I will include `die` in case of a different version. Also I will use the json output of `oc` to parse the version so to avoid any problem with i18n. Thanks for the recommendations!

wainersm:
Update this PR to address @jodh-intel 's comments:
 * `create_links.sh` is now a stand-alone script which is copied to the image
 * using `oc version -o json` (i.e. output in json) to avoid any error related with i18n
 * If the cluster isn't 4.10 nor 4.11 then the installation bails out

.ci/openshift-ci/images/entrypoint.sh
wainersm: Hi @snir911 , yes,  `/var/usrlocal/bin/` is mounted in `/usr/local/bin` and it should have 1st precedence on PATH.

.ci/openshift-ci/images/entrypoint.sh
```diff
@@ -90,6 +90,15 @@ update_initrd_kernel_modules()
 	( cd "${rootfs_dir}" && find . | cpio -H newc -o | gzip -9 ) > ${initrd_path}
 }
 
+# Manage symlinks on host.
+handle_symlinks() {
+	local create_links_script="/tmp/create_links.sh"
+	cp /create_links.sh "${HOST_MOUNT}/${create_links_script}"
+
+	chmod +x "${HOST_MOUNT}/${create_links_script}"
```
jodh-intel, line 98: Nit: You could combine these by using `install(1)` (which might be safer as you could also set perms and ownership explicitly).

.ci/openshift-ci/images/create_links.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Red Hat Inc.
```
jodh-intel, line 3: Just wanted to check that the two copyright dates on this PR are correct?

jodh-intel (approved):
Thanks @wainersm. A couple of comments but overall...

lgtm

.ci/openshift-ci/images/create_links.sh
wainersm: Guest what? Copy-and-paste when you rushing with too many things... :)
I will fix it.

.ci/openshift-ci/images/entrypoint.sh
wainersm: Good idea. Old habits are hard to die, `install(1)` is the right tool here. Thanks for the tip @jodh-intel !

wainersm:
Updated this PR to fix the wrong license statement in `create_links.sh` as well as now using `install(1)` to copy that script over the host partition.

.ci/openshift-ci/images/entrypoint.sh
snir911: @wainersm wouldn't `chroot "$HOST_MOUNT" sh -c 'for t in /opt/kata/bin/*; do ln --force -s "$t" /var/usrlocal/bin/; done'` be good enough?

wainersm:
Just updated this PR to simplify things as suggested by @snir911 : now using `ln --force -s` to force the creation of the link.

snir911 (approved):
thanks!

Issue #4714, opened 2022-04-20
add git repo as safe dir before use it


jongwu:
recently, git give an enhancement on security, addressing a pair of security vulnerabilities. It affect us that maybe we can't do some change when work in git repo whose user is different from who am I. At this point, we must explictly add the git repo to safe dir to avoid the issue.

one of the example is in arm ci log:
16:04:59 warn: ignoring non-existent submodule tests/fp/berkeley-softfloat-3
16:04:59 warn: ignoring non-existent submodule dtc
16:04:59 fatal: unsafe repository ('/home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/qemu/qemu' is owned by someone else)
16:04:59 To add an exception for this directory, call:
16:04:59 
16:04:59 	git config --global --add safe.directory /home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/qemu/qemu
16:04:59 /home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/qemu/qemu/scripts/git-submodule.sh: failed to update modules
16:04:59 
16:04:59 Unable to automatically checkout GIT submodules ''.
16:04:59 If you require use of an alternative GIT binary (for example to
16:04:59 enable use of a transparent proxy), then please specify it by
16:04:59 running configure by with the '--with-git' argument. e.g.

you can refer to https://github.blog/2022-04-12-git-security-vulnerability-announced/ to find more.

PR #4715, opened 2022-04-20
scripts: fix git security issue when install qemu for arm64

jongwu:
There is a enhancement on security for git recently. It affects
git working on multi-user machines. eg. When working in git repo
whose user is non-root as a root, git may stop us to do something
affect other dirs out of this repo. We can add this repo explicitly
as a safe dir to avoid this issue.

see [1] to find more.

[1] https://github.blog/2022-04-12-git-security-vulnerability-announced/

Fixes: #4714

### DIFF
--- a/.ci/aarch64/lib_install_qemu_aarch64.sh
+++ b/.ci/aarch64/lib_install_qemu_aarch64.sh
@@ -34,6 +34,7 @@ build_and_install_qemu() {
         sudo -E make -j $(nproc)
 
         echo "Install Qemu"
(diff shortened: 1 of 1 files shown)


wainersm:
Hi @jongwu ! Isn't that problem solved by @liubin 's commit e3450f12f668b1eae98c19f572c7df9b778877c0 ?

jongwu:
@wainersm , Yes, this is fixed, but should be also back port to stable-2.4

jongwu:
Raise a backport PR, see https://github.com/kata-containers/tests/pull/4723

PR #4723, opened 2022-04-21
backport-2.4: tests: fix unsafe repository for new versions of git

jongwu:
New git will check git repo's mode/permission, and will return error:

fatal: unsafe repository ('/some/path' is owned by someone else)

See:

https://github.blog/2022-04-12-git-security-vulnerability-announced/

for more information.

We have 3 repos run into this issue: kata-containers, containerd and
QEMU(only build from source for aarch64/ppc64le).

For the first two repos,
we can just use `git config --global --add safe.directory`.

But QEMU have
some submodules, we need to add all submodules to the safe
directory, so unlike kata-containers and containerd repos,
for QEMU we add `sudo -E` before commands run in the repo.

Fixes: #4692

### DIFF
--- a/.ci/aarch64/lib_install_qemu_aarch64.sh
+++ b/.ci/aarch64/lib_install_qemu_aarch64.sh
@@ -22,16 +22,16 @@ build_and_install_qemu() {
 	clone_katacontainers_repo
 
         pushd "${GOPATH}/src/${QEMU_REPO}"
-        git fetch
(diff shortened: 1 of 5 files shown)


jongwu:
@wainersm @fidencio

fidencio (approved):
lgtm, thanks @jongwu!

stevenhorsman (approved):
LGTM. Thanks!

Issue #4557, opened 2022-03-08
tests | CCv0: Non-TEE, CC pod creation with Kubernetes from unencrypted image


stevenhorsman:
**As a** Kata CC contributor
**I want to** have automated testing of pod creation with Kubernetes on PRs
**so that** I can ensure that I haven't introduced any regressions

### Description

This is test case i2 from https://github.com/confidential-containers/community/issues/3: *Test we can use Kubernetes to create a Kata CC pod by pulling an unencrypted image inside the guest in a non-TEE environment*

Building on the work that Wainer did in https://github.com/kata-containers/tests/pull/4508 for a `crictl` environment and the ccv0 [script](https://github.com/kata-containers/kata-containers/blob/CCv0/docs/how-to/ccv0.sh) and [doc](https://github.com/kata-containers/kata-containers/blob/CCv0/docs/how-to/how-to-build-and-test-ccv0.md) we want to automate the test case 'Using Kubernetes for end-to-end provisioning of a Kata confidential containers pod with an unencrypted image' to do the same tests, but creating a Kubernetes pod rather than `crictl` one.

### Task list

- [x] Add a new CC Kubernetes job type
- [x] Add a job for the new CC Kubernetes job type: https://github.com/kata-containers/ci/issues/471
- [x] Add a new bats test of create a Kubernetes CC pod from an unsigned, unencrypted image
- [ ] ~Refactor `ccv0.sh` out into new integration test scripts to drive test~ Covered in https://github.com/kata-containers/kata-containers/issues/4132
- [x] Refactor common code in the `critctl` tests and Kubernetes into a shared library

stevenhorsman:
@Megan-Wright and @GeorginaKin are actually working on this, but as they aren't currently project members I've assigned this to myself as a contact point

stevenhorsman:
The tests and job changes have been completed in https://github.com/kata-containers/tests/pull/4729 and https://github.com/kata-containers/ci/pull/472. The remaining item to finish up `ccv0.sh` compatibility is covered under https://github.com/kata-containers/kata-containers/issues/4132

PR #4729, opened 2022-04-25
tests: CC pod creation with Kubernetes

GeorginaKin:
Adding in the new tests to pull an unencrypted image and create a CC pod
Also the ci job flags - `CC_SKOPEO_CRI_CONTAINERD_K8S` and `CC_CRI_CONTAINERD_K8S` for these tests

Fixes https://github.com/kata-containers/tests/issues/4557
Signed-off-by: Georgina Kinge <georgina.kinge@ibm.com>
Co-authored-by: Megan Wright <Megan.Wright@ibm.com>


GabyCT (approved):

stevenhorsman:
@GeorginaKin - can you fix the DCO check please - https://github.com/kata-containers/tests/pull/4729/commits/97131e7fbd1368f7cfe4ba337fcf9b3fe63fc8d2 shows:
```
Signed-off-by: Megan Wright <megan.wright@ibm.com>
Co-authored-by: Georgina Kinge <georgina.kinge@ibm.com>
```
but as you pushed the commit I think you need to flip those around.

GeorginaKin:
Thank you @stevenhorsman! I have changed these around so the DCO check should pass now.

wainersm:
Hi @GeorginaKin , thanks for this contribution! I will try to run it locally then in a job in CI that I will prepare just to test this PR. I post the results here.

.ci/run.sh
```diff
@@ -65,6 +65,10 @@ case "${CI_JOB}" in
 		echo "INFO: Running Confidential Container tests"
 		sudo -E PATH="$PATH" CRI_RUNTIME="containerd" bash -c "make cc-containerd"
 		;;
+	"CC_CRI_CONTAINERD_K8S"|"CC_SKOPEO_CRI_CONTAINERD_K8S")
+		echo "INFO: Running Confidential Container tests"
```
jodh-intel, line 69: This is in keeping with the existing code but we should switch to the `info()` call at some point as it's already available to this script.

integration/confidential/lib.sh
```diff
@@ -0,0 +1,177 @@
+#!/bin/bash
+# Copyright (c) 2021, 2022 IBM Corporation
+# Copyright (c) 2022 Red Hat
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This provides generic functions to use in the tests.
+#
+set -e
```
jodh-intel, line 9: I'd consider using this set of options, as they've become pretty standard for Kata now and make the scripts a lot more robust:

```bash
set -o errexit
set -o errtrace
set -o nounset
set -o pipefail

```

integration/confidential/lib.sh
```diff
@@ -0,0 +1,177 @@
...
+#	$1: "on" to activate the service, or "off" to turn it off.
+#
+# Environment variables:
+#	RUNTIME_CONFIG_PATH - path to kata's configuration.toml. If it is not
+#			      export then it will figure out the path via
+#			      `kata-runtime env` and export its value.
+#
+switch_image_service_offload() {
+	# Load the RUNTIME_CONFIG_PATH variable.
+	load_RUNTIME_CONFIG_PATH
```
jodh-intel, line 27: Could you make this all lower-case?

integration/confidential/lib.sh
```diff
@@ -0,0 +1,177 @@
...
+				"$RUNTIME_CONFIG_PATH"
+			;;
+		"off")
+			sudo sed -i -e 's/^\(service_offload\).*=.*$/#\1 = false/g' \
+				"$RUNTIME_CONFIG_PATH"
+
+			;;
+		*)
+			echo "Unknown option $1"
+			return 1
```
jodh-intel, line 41: Can you use our standard `die()` function here:

```suggestion
			die "Unknown option '$1'"
```

integration/confidential/lib.sh
```diff
@@ -0,0 +1,177 @@
...
+#	$1..$N - list of parameters
+#
+# Environment variables:
+#	RUNTIME_CONFIG_PATH - path to kata's configuration.toml. If it is not
+#			      export then it will figure out the path via
+#			      `kata-runtime env` and export its value.
+#
+add_kernel_params() {
+	local params="$@"
+	load_RUNTIME_CONFIG_PATH
```
jodh-intel, line 58: Is it really necessary for every function to make this call? I'd consider a `setup()` function that is called once instead.

integration/confidential/lib.sh
```diff
@@ -0,0 +1,177 @@
...
+#	$1 - (Optional) file path to where save the current containerd's config.toml
+#
+configure_cc_containerd() {
+	local saved_containerd_conf_file="${1:-}"
+	local containerd_conf_file="/etc/containerd/config.toml"
+
+	# Even if we are not saving the original file it is a good idea to
+	# restart containerd because it might be in an inconsistent state here.
+	sudo systemctl stop containerd
+	sleep 5
```
jodh-intel, line 139: Why is a sleep required after containerd is stopped? Either way, this is going to be fragile. Can you use `WaitForProcess()` or similar here too?

integration/confidential/lib.sh
```diff
@@ -0,0 +1,177 @@
...
+	[ -n "$saved_containerd_conf_file" ] && \
+		cp -f "$containerd_conf_file" "$saved_containerd_conf_file"
+	sudo systemctl start containerd
+	waitForProcess 30 5 "sudo crictl info >/dev/null"
+
+	# Ensure the cc CRI handler is set.
+	local cri_handler=$(sudo crictl info | \
+		jq '.config.containerd.runtimes.kata.cri_handler')
+	if [[ ! "$cri_handler" =~ cc ]]; then
+		sudo sed -z -i 's/\([[:blank:]]*\)\(runtime_type = "io.containerd.kata.v2"\)/\1\2\n\1cri_handler = "cc"/' \
```
jodh-intel, line 149: Is `-z` actually needed here?

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -0,0 +1,166 @@
...
+	pod_id=""
+
+	echo "Delete any existing ${sandbox_name} pod"
+	kubernetes_delete_cc_pod_if_exists "$sandbox_name"
+
+	echo "Prepare containerd for Confidential Container"
+	SAVED_CONTAINERD_CONF_FILE="/etc/containerd/config.toml.$$"
+	configure_cc_containerd "$SAVED_CONTAINERD_CONF_FILE"
+
+
```
jodh-intel, line 69: Nit: One blank line is sufficient here.

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -0,0 +1,166 @@
...
+# Check the logged messages on host have a given message.
+# Parameters:
+#      $1 - the message
+#
+# Note: get the logs since the global $start_date.
+#
+assert_logs_contain() {
+	local message="$1"
+    journalctl -x -t kata --since "$start_date" | grep "$message"
```
jodh-intel, line 83: Nit: Inconsistent indents.

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -0,0 +1,166 @@
...
+	create_test_pod
+
+	echo "Check the image was not pulled in the host"
+	local pod_id=$(kubectl get pods -o jsonpath='{.items..metadata.name}')
+	retrieve_sandbox_id
+	rootfs=($(find /run/kata-containers/shared/sandboxes/${sandbox_id}/shared \
+		-name rootfs))
+	[ ${#rootfs[@]} -eq 1 ]
+
+	# TODO: confirm the image was pulled in the guest. `kata-runtime exec`
```
jodh-intel, line 98: I'd consider putting a GitHub issue here so it's less likely we'll forget to do this ;)

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -0,0 +1,166 @@
...
+	#       can be used to get a shell prompt which is not exactly what we
+	#       want but can be used to with `expect` to implement a
```
jodh-intel, line 100:
```suggestion
	#       want but can be used with `expect(1)` to implement a
```

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -0,0 +1,166 @@
...
+#      $1 - the message
+#
+# Note: get the logs since the global $start_date.
+#
+assert_logs_contain() {
+	local message="$1"
+    journalctl -x -t kata --since "$start_date" | grep "$message"
+}
+
+@test "[cc][agent][cri][containerd] Test can pull an unencrypted image inside the guest" {
```
jodh-intel, line 86: All the tests start with `[cc][agent][cri][containerd] ` so how about creating a variable to avoid the duplication? Something like:

```bash

test_tag='[cc][agent][cri][containerd]'

# ...

@test "${test_tag} Test can pull an unencrypted image inside the guest" {
```

jodh-intel (commented):
Thanks @GeorginaKin - a few comments...

.ci/ci_job_flags.sh
```diff
@@ -96,23 +96,21 @@ case "${CI_JOB}" in
...
-"CRI_CONTAINERD"|"CRI_CONTAINERD_K8S"|"CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD")
+"CRI_CONTAINERD"|"CRI_CONTAINERD_K8S"|"CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD"|"CC_CRI_CONTAINERD_K8S"|"CC_SKOPEO_CRI_CONTAINERD_K8S")
 	# This job only tests containerd + k8s
 	init_ci_flags
 	export CRI_CONTAINERD="yes"
 	export CRI_RUNTIME="containerd"
 	export KATA_HYPERVISOR="qemu"
 	case "${CI_JOB}" in
-		"CRI_CONTAINERD_K8S")
+		"CRI_CONTAINERD_K8S"|"CC_CRI_CONTAINERD_K8S"|"CC_SKOPEO_CRI_CONTAINERD_K8S")
```
wainersm, line 106: I was trying to run it locally with `export CI_JOB=CC_CRI_CONTAINERD_K8S` but the `.ci/install_kata_image.sh` was failing complaining the `CCV0` variable wasn't set.

It took me while to understand that adding `CC_CRI_CONTAINERD_K8S` and `CC_SKOPEO_CRI_CONTAINERD_K8S` in this case entry is wrong. The way it is, the code path enters here (and the `KUBERNETES` variable is set) then it returns out of the case statement, i.e., the next condition (`"CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD"|"CC_CRI_CONTAINERD_K8S"|"CC_SKOPEO_CRI_CONTAINERD_K8S"`) is never evaluated and, thus, `CCV0` is never set.

It needs something like this:
```
diff --git a/.ci/ci_job_flags.sh b/.ci/ci_job_flags.sh
index ec36a93e..b8f02e83 100755
--- a/.ci/ci_job_flags.sh
+++ b/.ci/ci_job_flags.sh
@@ -103,7 +103,7 @@ case "${CI_JOB}" in
        export CRI_RUNTIME="containerd"
        export KATA_HYPERVISOR="qemu"
        case "${CI_JOB}" in
-               "CRI_CONTAINERD_K8S"|"CC_CRI_CONTAINERD_K8S"|"CC_SKOPEO_CRI_CONTAINERD_K8S")
+               "CRI_CONTAINERD_K8S")
                        export KUBERNETES="yes"
                        ;;
                "CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD"|"CC_CRI_CONTAINERD_K8S"|"CC_SKOPEO_CRI_CONTAINERD_K8S")
@@ -111,6 +111,7 @@ case "${CI_JOB}" in
                        export CCV0="yes"
                        export UMOCI=yes
                        [[ "${CI_JOB}" =~ SKOPEO ]] && export SKOPEO=yes
+                       [[ "${CI_JOB}" =~ K8S ]] && export KUBERNETES=yes
                        ;;
        esac
        ;;

```

GeorginaKin: Thank you @wainersm! We didn't notice this but have just pushed the above change you suggested so now CCV0 should get set!

wainersm: cool, thanks @GeorginaKin ! I will give it a try again.

integration/kubernetes/confidential/run_tests.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# This is a thin wrapper to call the confidential containers related tests.
+#
+set -e
+
+script_dir=$(dirname "$(readlink -f "$0")")
+cidir="${script_dir}/../../../.ci"
+source "${cidir}/lib.sh"
+
+main() {
```
wainersm, line 14: The tests fail because Kubernetes is not initialized at this point.

This `run_tests.sh` script should be almost identical to https://github.com/kata-containers/tests/blob/main/integration/kubernetes/run_kubernetes_tests.sh, except that the list of tests will be different.

At glance I can see two ways of fixing it without duplicating the `run_kubernetes_tests.sh` script code:

 1.  Make `integration/kubernetes/confidential/run_tests.sh` to call `integration/kubernetes/run_kubernetes_tests.sh`. The former script will need to export the `K8S_TEST_UNION` variable in such a way that only the CC bats tests are executed.
 2. Alike the option 1) it will be exporting `K8S_TEST_UNION`properly and calling `integration/kubernetes/run_kubernetes_tests.sh` however from the `cc-kubernetes` Makefile target. I.e. we don't have this `run_tests.sh` script anymore and just call `run_kubernetes_tests.sh` from the makefile target directly.
 
If you have a better solution, please don't hesitate to share. Otherwise I tend to think option 2) would be better.

GeorginaKin: Hi @wainersm we've done what you suggested for option 2 and deleted the `run_tests.sh` script, and then also referenced `run_kubernetes_tests.sh` in the makefile instead! Hopefully this change was what you were looking for 😄

Makefile
```diff
@@ -89,6 +89,11 @@ cri-containerd:
 cc-containerd:
 	bash integration/containerd/confidential/run_tests.sh
 
+# Run the Confidential Containers tests for kubernetes.
+cc-kubernetes:
+	export K8S_TEST_UNION="integration/kubernetes/confidential/agent_image.bats"
+	bash integration/kubernetes/confidential/run_kubernetes_tests.sh
```
wainersm, line 95: Hi @GeorginaKin , I've manually created a job on Jenkins so we can test this PR in the CI environment: http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S

And, as you can see in http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/2/console it failed. The path `integration/kubernetes/confidential/run_kubernetes_tests.sh` doesn't exist, you will need to fix that. Also the K8S_TEST_UNION should be relative (I guess) to the `integration/kubernetes` directory, so it should be `confidential/run_kubernetes_tests.sh`

GeorginaKin: Hey @wainersm thank you for the review again! We've fixed the path issue of `run_kubernetes_tests.sh` not being found and also made the path relative for `K8S_TEST_UNION` too 🙂

Makefile
```diff
@@ -89,6 +89,11 @@ cri-containerd:
 cc-containerd:
 	bash integration/containerd/confidential/run_tests.sh
 
+# Run the Confidential Containers tests for kubernetes.
+cc-kubernetes:
```
wainersm, line 93: We are almost there! :)

See in http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/3/console that the execution failed because bats was not installed in the environment:
```
08:15:26 integration/kubernetes/run_kubernetes_tests.sh: line 109: bats: command not found
08:15:26 [run_kubernetes_tests.sh:109] ERROR: bats k8s-attach-handlers.bats
```
You will need to install bats before run the tests. See the `kubernetes` Makefile target as an example.

GeorginaKin: Awesome!!
We've added in the `install_bats.sh` line now 🤞 😄

Makefile
```diff
@@ -89,6 +89,12 @@ cri-containerd:
...
+	export K8S_TEST_UNION="confidential/agent_image.bats"
+	bash -f .ci/install_bats.sh
+	bash integration/kubernetes/run_kubernetes_tests.sh
```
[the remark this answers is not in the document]
GeorginaKin: Ah okay not to worry - we've added in the suggested line (we were unsure if export worked in bash too so makes sense)! Hopefully this will allow the variable of `K8S_TEST_UNION` to be set 😄

integration/kubernetes/confidential/agent_image.bats
GeorginaKin: Good shout - I've opened an issue for this TODO here: https://github.com/kata-containers/tests/issues/4747 😄

.ci/run.sh
GeorginaKin: Thank you - we've used the `info()` call instead for line 69! Definitely something that can be applied to the rest of the file at some point tho 👍

integration/confidential/lib.sh
GeorginaKin: This is a good point! We've now added this function to the existing `setup()` function which runs before each test.

integration/confidential/lib.sh
GeorginaKin: We were using the sleep to ensure the stop process had finished, but using the `WaitForProccess()` one makes more sense so we've switched to use that now instead 🙂

integration/confidential/lib.sh
GeorginaKin: We've tested and it's not - so it's been removed, thanks!

integration/kubernetes/confidential/agent_image.bats
GeorginaKin: We've now added a `test_tag` variable as you suggested that is being passed in for each test.

GeorginaKin:
> Thanks @GeorginaKin - a few comments...

Thank you so much for your comments @jodh-intel! We've made changes based on these and welcome any further comments 😃

wainersm:
I was looking at the failure on http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/6/console .... It seems some bash configuration has messed with bats but I am not sure.

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env bats
+# Copyright (c) 2022 Red Hat
```
stevenhorsman, line 2: I wonder if this should have an IBM copyright as I think it's been modified from the crictl one quite a bit?

wainersm: I think so @stevenhorsman

wainersm:
> I was looking at the failure on http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/6/console .... It seems some bash configuration has messed with bats but I am not sure.

@GeorginaKin , I think I found the problem... in `integration/confidential/lib.sh` it is `set` a bunch of bash properties that mess with bats (bats will set bash properties on its own...). I changed the file as below and the tests ran on my local environment:
```
vagrant@ubuntu2004:~/go/src/github.com/kata-containers/tests$ git diff
diff --git a/integration/confidential/lib.sh b/integration/confidential/lib.sh
index a5da273f..61f0778d 100644
--- a/integration/confidential/lib.sh
+++ b/integration/confidential/lib.sh
@@ -6,12 +6,6 @@
 #
 # This provides generic functions to use in the tests.
 #
-set -o errexit
-set -o errtrace
-set -o nounset
-set -o pipefail
-
-[ -n "${DEBUG:-}" ] && set -o xtrace
 
 source "${BATS_TEST_DIRNAME}/../../../lib/common.bash"
 FIXTURES_DIR="${BATS_TEST_DIRNAME}/fixtures"
```
My suggestion is to set those properties only and only if not in bats. We could use one of the bats exported variables to decide if the lib is loaded on context of bats.

wainersm:
> > I was looking at the failure on http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/6/console .... It seems some bash configuration has messed with bats but I am not sure.
> 
> @GeorginaKin , I think I found the problem... in `integration/confidential/lib.sh` it is `set` a bunch of bash properties that mess with bats (bats will set bash properties on its own...). I changed the file as below and the tests ran on my local environment:
> 
> ```
> vagrant@ubuntu2004:~/go/src/github.com/kata-containers/tests$ git diff
> diff --git a/integration/confidential/lib.sh b/integration/confidential/lib.sh
> index a5da273f..61f0778d 100644
> --- a/integration/confidential/lib.sh
> +++ b/integration/confidential/lib.sh
> @@ -6,12 +6,6 @@
>  #
>  # This provides generic functions to use in the tests.
>  #
> -set -o errexit
> -set -o errtrace
> -set -o nounset
> -set -o pipefail
> -
> -[ -n "${DEBUG:-}" ] && set -o xtrace
>  
>  source "${BATS_TEST_DIRNAME}/../../../lib/common.bash"
>  FIXTURES_DIR="${BATS_TEST_DIRNAME}/fixtures"
> ```
> 
> My suggestion is to set those properties only and only if not in bats. We could use one of the bats exported variables to decide if the lib is loaded on context of bats.

Thinking about it a little further...IMO we should touch in bash properties in library files. The executable caller script (with `main()`) that should be setting those things instead... @jodh-intel do you agree with that?

jodh-intel:
@wainersm - I think you're saying this?:

- Ensure our bash "lib" scripts don't call `set -o` if the caller is a bats script.
- Have each script make the "boiler plate" `set -o` calls.

If so, I agree. In fact, I'm vaguely wondering if we should add a `kata_bash_setup()` or similar to our lib scripts so that if a script should call `set -o` (aka it's not a bats script), we take a consistent approach. That might not work for 100% of our scripts, but it would be a "one liner" to get a standard setup and we could document that easily.

integration/kubernetes/confidential/agent_image.bats
GeorginaKin: Awesome I've changed this now thanks both 😄

.ci/ci_job_flags.sh
```diff
@@ -106,13 +106,12 @@ case "${CI_JOB}" in
...
-		"CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD")
+		"CC_CRI_CONTAINERD"|"CC_SKOPEO_CRI_CONTAINERD"|"CC_CRI_CONTAINERD_K8S"|"CC_SKOPEO_CRI_CONTAINERD_K8S")
 			# Export any CC specific environment variables
 			export CCV0="yes"
 			export UMOCI=yes
-			if [ "${CI_JOB}" == "CC_SKOPEO_CRI_CONTAINERD" ]; then
-				export SKOPEO=yes
-			fi
+			[[ "${CI_JOB}" =~ SKOPEO ]] && export SKOPEO=yes
+			[[ "${CI_JOB}" =~ K8S ]] && export KUBERNETES=yes
```
stevenhorsman, line 114: So after many, many test builds and debugging I found that these lines seemed to be causing the bats tests to not run. I'm not sure why and if there is something strange with the unset variables, but when I replaced them with:
```
if [ "${CI_JOB}" =~ SKOPEO ]; then
	export SKOPEO=yes
fi
if [ "${CI_JOB}" =~ K8S ]; then
	export KUBERNETES=yes
fi
```
the tests ran e.g. http://jenkins.katacontainers.io/job/tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD-PR/81/console
In that case two of the skopeo skips didn't seem to work, but at least it's slight progress? I'd love to know if anyone has a reason why this might make a difference?

stevenhorsman: The other thing I noticed is that this PR is 45 commits behind the `CCv0` branch, which includes a whole bunch of refactors and re-working, so it is possible that this is causing the issue (when I tested my change I did a re-base on latest)

stevenhorsman: Oops my bad, after looking through the latest output I realised I made a typo with my suggested replacement code as `=~` only works inside double square brackets, so it should have been:
```
if [[ "${CI_JOB}" =~ SKOPEO ]]; then
	export SKOPEO=yes
fi
if [[ "${CI_JOB}" =~ K8S ]]; then
	export KUBERNETES=yes
fi
```

stevenhorsman:
@wainersm - I think we are finally there with getting all the tests to run now. We had to add the `sleep 5` after stopping containerd in as removing it was causing the service restart to fail, but we can always handle that in a separate issue as it wasn't new code introduced here. You test job also looks good now: http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/ :)

wainersm:
Hi @stevenhorsman ,

> @wainersm - I think we are finally there with getting all the tests to run now. We had to add the `sleep 5` after stopping containerd in as removing it was causing the service restart to fail, but we can always handle that in a separate issue as it wasn't new code introduced here. You test job also looks good now: http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/ :)

That's okay to for me to fix that `sleep 5` in a follow up PR (I can fix it as I added that command IIRC).

But, unfortunately, I cannot say that the job  http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S is working fine because the bats execution is messy. We should have the [`set -o xtrace`](https://github.com/kata-containers/tests/pull/4729/commits/94a2f4c3986264a87834601869687df65973d86c#diff-d9fa72994d19f3acd66a01c6ee36939cbe4e2751b2f0c402bb9e39491385b837R11) command out of `integration/confidential/lib.sh`.

stevenhorsman:
> We should have the [`set -o xtrace`](https://github.com/kata-containers/tests/pull/4729/commits/94a2f4c3986264a87834601869687df65973d86c#diff-d9fa72994d19f3acd66a01c6ee36939cbe4e2751b2f0c402bb9e39491385b837R11) command out of `integration/confidential/lib.sh`.

Are you thinking that we should move it to somewhere like `integration/containerd/confidential/run_tests.sh` and `integration/kubernetes/run_kubernetes_tests.sh`?

wainersm:
> > We should have the [`set -o xtrace`](https://github.com/kata-containers/tests/pull/4729/commits/94a2f4c3986264a87834601869687df65973d86c#diff-d9fa72994d19f3acd66a01c6ee36939cbe4e2751b2f0c402bb9e39491385b837R11) command out of `integration/confidential/lib.sh`.
> 
> Are you thinking that we should move it to somewhere like `integration/containerd/confidential/run_tests.sh` and `integration/kubernetes/run_kubernetes_tests.sh`?

Sorry, I didn't explain correctly.

In `integration/confidential/lib.sh` it is called `set -o xtrace` in case `DEBUG=true`. All our CI jobs set `DEBUG=true`, therefore   `set -o xtrace` is always called. That command mess with bats execution (see in [job 31](http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/13/console)  after `12:41:25 INFO: Run tests`), thus we should not call that command in any hypothesis along side bats.

I created another job to test the execution without `set -o xtrace` and it works pretty fine. See after `18:03:45 INFO: Run tests` in http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_SKOPEO_CRI_CONTAINERD_K8S/2/console

stevenhorsman:
> > > We should have the [`set -o xtrace`](https://github.com/kata-containers/tests/pull/4729/commits/94a2f4c3986264a87834601869687df65973d86c#diff-d9fa72994d19f3acd66a01c6ee36939cbe4e2751b2f0c402bb9e39491385b837R11) command out of `integration/confidential/lib.sh`.
> > 
> > 
> > Are you thinking that we should move it to somewhere like `integration/containerd/confidential/run_tests.sh` and `integration/kubernetes/run_kubernetes_tests.sh`?
> 
> Sorry, I didn't explain correctly.
> 
> In `integration/confidential/lib.sh` it is called `set -o xtrace` in case `DEBUG=true`. All our CI jobs set `DEBUG=true`, therefore `set -o xtrace` is always called. That command mess with bats execution (see in [job 31](http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S/13/console) after `12:41:25 INFO: Run tests`), thus we should not call that command in any hypothesis along side bats.
> 
> I created another job to test the execution without `set -o xtrace` and it works pretty fine. See after `18:03:45 INFO: Run tests` in http://jenkins.katacontainers.io/job/wainer-tests-CCv0-ubuntu-20.04-x86_64-CC_SKOPEO_CRI_CONTAINERD_K8S/2/console

Ah ok, so we just need to delete that line. I spotted that the bats output was a mess, but didn't realise that `DEBUG=true` for all CI jobs. Thanks!

stevenhorsman:
/test/test

stevenhorsman:
@wainersm - the output is looking better now :)

wainersm (approved):
Thanks for this contribution!

Issue #4732, opened 2022-04-26
Add backtrace and env metadata when the CI fails


jodh-intel:
Although the [CI jobs](https://github.com/kata-containers/ci) are run with `DEBUG=true` (which enables `set -x` / `set -o xtrace`), it is still difficult to understand some failures as we still only log partial details.

Ideally, when a script failed it would display:

- A full backtrace (particularly useful as we `source` a lot of scripts).
- Additional environment information (current list of processes, _etc_).

PR #4733, opened 2022-04-26
scripts: Make die show stacktrace and other info under CI

jodh-intel:
Enhance the standard `die()` function to dump lots of information (in
YAML format) after displaying the error message when running under the
CI. The dump includes a full stacktrace and lots of environment details
to help with debugging.

> **Note:**
>
> The conventional way of displaying this additional information is by
> defining a `trap` handler. However, that technique cannot be used here
> as this is effectively a library and the callers of this library can
> (and do) define their own `traps` (and only one `trap` can be defined
> per signal).

Fixes: #4732.

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -7,6 +7,9 @@
 # This file contains common functions that
 # are being used by our metrics and integration tests
 
+_script_dir=$(dirname "${BASH_SOURCE[0]}")
+source "${_script_dir}/error.sh"
+
 # Place where virtcontainers keeps its active pod info
 VC_POD_DIR="${VC_POD_DIR:-/run/vc/sbs}"
 
@@ -28,9 +31,44 @@ KATA_TESTS_CACHEDIR="${KATA_TESTS_CACHEDIR:-${KATA_TESTS_BASEDIR}/cache}"
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 experimental_qemu="${experimental_qemu:-false}"
 
+# Display a message to stderr, a dump of lots of useful debug
+# information (including a full stacktrace) and exit 1.
 die() {
 	local msg="$*"
-	echo -e "[$(basename $0):${BASH_LINENO[0]}] ERROR: $msg" >&2
+
+	if [ -z "${KATA_TEST_VERBOSE:-}" ] && [ -z "${CI:-}" ]
+	then
+		echo -e "[$(basename $0):${BASH_LINENO[0]}] ERROR: $msg" >&2
+		exit 1
+	fi
+
+	echo >&2 "ERROR: $msg"
+
+	# We are running in the CI or the user has requested verbose
+	# failure details, so dump as much information about
+	# the environment that generated the failure as possible.
+
+	# This function is called to indicate a fatal error occurred, so
+	# the caller of this function is the site of the detected error.
+	local error_location
+	error_location=$(caller 0)
+
+	local line
+	local func
+	local file
+
+	line=$(echo "$error_location"|awk '{print $1}')
+	func=$(echo "$error_location"|awk '{print $2}')
+	file=$(echo "$error_location"|awk '{print $3}')
+
+	local path
+	path=$(resolve_path "$file")
+
+	dump_details \
+		"${line}" \
+		"${func}" \
+		"${path}"
+
 	exit 1
 }
 
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -52,7 +52,7 @@ handle_error() {
 }
 trap 'handle_error $LINENO' ERR
 
-waitForProcess(){
+waitForProcess() {
 	wait_time="$1"
 	sleep_time="$2"
 	cmd="$3"
@@ -71,7 +71,7 @@ waitForProcess(){
 # Kata runtime. Of course, the end user can choose any name they
 # want in reality, but this function knows the names of the default
 # and recommended Kata docker runtime install names.
-is_a_kata_runtime(){
+is_a_kata_runtime() {
 	if [ "$1" = "containerd-shim-kata-v2" ] || [ "$1" = "io.containerd.kata.v2" ]; then
 		echo "1"
 	else
@@ -81,7 +81,7 @@ is_a_kata_runtime(){
 
 
 # Try to find the real runtime path for the docker runtime passed in $1
-get_docker_kata_path(){
+get_docker_kata_path() {
 	local jpaths=$(sudo docker info --format "{{json .Runtimes}}" || true)
 	local rpath=$(jq .\"$1\".path <<< "$jpaths")
 	# Now we have to de-quote it..
@@ -92,7 +92,7 @@ get_docker_kata_path(){
 
 # Gets versions and paths of all the components
 # list in kata-env
-extract_kata_env(){
+extract_kata_env() {
 	RUNTIME_CONFIG_PATH=$(kata-runtime kata-env --json | jq -r .Runtime.Config.Path)
 	RUNTIME_VERSION=$(kata-runtime kata-env --json | jq -r .Runtime.Version | grep Semver | cut -d'"' -f4)
 	RUNTIME_COMMIT=$(kata-runtime kata-env --json | jq -r .Runtime.Version | grep Commit | cut -d'"' -f4)
--- a/lib/error.sh
+++ b/lib/error.sh
@@ -1,0 +1,218 @@
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Description: Error handling functions.
+
+# Write specified message to stderr.
+stderr()
+{
+	local msg="$*"
+
+	echo >&2 "$msg"
+}
+
+# Simplified version of die that should be called by functions that
+# could fail as part of the normal "die()" code path. Required to
+# avoid infinite recursion.
+_fatal()
+{
+	local msg="$*"
+
+	echo >&2 "FATAL: $msg"
+	exit 1
+}
+
+# Canonicalize the specified path (which must be valid).
+resolve_path()
+{
+	local file="${1:-}"
+	[ -z "$file" ] && _fatal "need file to resolve"
+
+	local path
+
+	path=$(readlink --canonicalize-existing "$file" || \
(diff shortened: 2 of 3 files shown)


jodh-intel:
Currently, if a script fails (by calling `die()`, you'll get a one line error message prefixed with `ERROR:` like this:

```bash
$ cmd/github-labels/github-labels.sh foo
ERROR: Invalid command: 'foo'
$ echo $?
1
```

However, with this PR, when running under the CI, you'll see that same message and a block of YAML that further describes the error and the environment. Something like the following...

> **Notes:**
>
> - I've truncated the output for some sections.
> - NB the `stacktrace:` section.
> - There is some duplication: we could remove it but it might be worth retaining as each has its own value IMHO:
>   - processes:
>     - The `processes:` section shows *all* processes.
>     - The `process-hierarchy:` section shows just the processes related to the failing script.
>   - Failure details:
>     - The `stacktrace:` section shows all related scripts.
>     - The `failure:` section shows a brief summary.

```bash
$ CI=true cmd/github-labels/github-labels.sh foo
ERROR: Invalid command: 'foo'

script:
  name: 'cmd/github-labels/github-labels.sh'
  pid: 16853
  directory: '/home/james/go/src/github.com/kata-containers/tests'
  details: 'drwxr-x--- 15 james kata ? 4096 Apr 26 11:10 /home/james/go/src/github.com/kata-containers/tests'
failure:
  function: 'main'
  file: '/home/james/go/src/github.com/kata-containers/tests/cmd/github-labels/github-labels.sh'
  line: 171
  name: 'cmd/github-labels/github-labels.sh'
stacktrace:
  0: {function: 'dump_details', file: '/home/james/go/src/github.com/kata-containers/tests/lib/error.sh', line: 168}
  1: {function: 'die', file: '/home/james/go/src/github.com/kata-containers/tests/lib/common.bash', line: 63}
  2: {function: 'main', file: '/home/james/go/src/github.com/kata-containers/tests/cmd/github-labels/github-labels.sh', line: 171, current='yes'}
  3: {function: 'main', file: '/home/james/go/src/github.com/kata-containers/tests/cmd/github-labels/github-labels.sh', line: 176}
process-hierarchy:
  0: {pid: 16853, command: '/bin/bash cmd/github-labels/github-labels.sh foo', current='yes'}
  1: {pid: 2685, command: '-zsh'}
  2: {pid: 2345, command: 'tmux attach'}
  3: {pid: 1, command: '/sbin/init splash'}
time: '2022-04-26T12:06:19+01:00'
runtime-seconds: 0
host:
  name: 'hostname'
  uname: 'Linux hostname 5.15.0-25-generic #25-Ubuntu SMP Wed Mar 30 15:54:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux'
locale:
  LANG: 'en_GB.UTF-8'
  LANGUAGE: 'en_GB:en'

  << truncated for GitHub output >>

user:
  uid: {value: 1000, name: 'james'}
  euid: {value: 1000 , name: 'james'}
  groups: 'uid=1000(james) gid=1000(james) groups=1000(james),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),120(lpadmin),132(lxd),133(sambashare),136(docker),1001(kata)'
bash:
  version: '5.1.16(1)-release'
  version-info: '5 1 16 1 release x86_64-pc-linux-gnu'
environment:
  CI: 'true'
  HOME: '/home/james'

  << truncated for GitHub output >>

mounts: |
  sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
  proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)

  << truncated for GitHub output >>

processes: |
  UID          PID    PPID  C    SZ   RSS PSR STIME TTY          TIME CMD
  root           1       0  0 41735 11884  11 11:16 ?        00:00:01 /sbin/init splash
  root           2       0  0     0     0   0 11:16 ?        00:00:00 [kthreadd]
  root           3       2  0     0     0   0 11:16 ?        00:00:00 [rcu_gp]

  << truncated for GitHub output >>

$ echo $?
1
```

lib/error.sh
```diff
@@ -0,0 +1,217 @@
...
+
+	local line
+	local func
+	local file
+
+	local -i i=0
+
+	stderr "stacktrace:"
+
+	while true
```
c3d, line 103: Why not a `for ((i = 0; ; i++))` ?

lib/common.bash
```diff
@@ -28,9 +31,39 @@ KATA_TESTS_CACHEDIR="${KATA_TESTS_CACHEDIR:-${KATA_TESTS_BASEDIR}/cache}"
...
 
+# Display a message to stderr, a dump of lots of useful debug
+# information (including a full stacktrace) and exit 1.
 die() {
 	local msg="$*"
-	echo -e "[$(basename $0):${BASH_LINENO[0]}] ERROR: $msg" >&2
+
+	echo >&2 "ERROR: $msg"
+
+	[ -z "${CI:-}" ] && exit 1
```
c3d, line 41: If you run something manually, don't you expect to see the same kind of stack trace? What about having a dedicated variable for stack traces, which each developer can set the way they want, and that we set the right way for CI builds?

Also, this means that you print less information than before in that case, which I think is undesirable.

lib/common.bash
```diff
@@ -28,9 +31,39 @@ KATA_TESTS_CACHEDIR="${KATA_TESTS_CACHEDIR:-${KATA_TESTS_BASEDIR}/cache}"
...
+
+	# We are running in the CI so dump as much information about
+	# the environment that generated the failure as possible.
+
+	# This function is called to indicate a fatal error occurred, so
+	# the caller of this function is the site of the detected error.
+	local error_location
+	error_location=$(caller 0)
```
c3d, line 49: Why split it on a separate line?

lib/common.bash
```diff
@@ -28,9 +31,39 @@ KATA_TESTS_CACHEDIR="${KATA_TESTS_CACHEDIR:-${KATA_TESTS_BASEDIR}/cache}"
...
+
+	local line
+	local func
+	local file
+
+	line=$(echo "$error_location"|awk '{print $1}')
```
c3d, line 55: Same here, why split the declaration and initialization on separate lines?

lib/error.sh
```diff
@@ -0,0 +1,217 @@
...
+
+	echo >&2 "$msg"
+}
+
+# Simplified version of die that should be called by functions that
+# could fail as part of the normal "die()" code path. Required to
+# avoid infinite recursion.
+_fatal()
+{
+	echo >&2 "FATAL: $msg"
```
c3d, line 20: You need to initialize `msg` with `$*` here, no?

lib/error.sh
```diff
@@ -0,0 +1,217 @@
...
+	exit 1
+}
+
+# Canonicalize the specified path (which must be valid).
+resolve_path()
+{
+	local file="${1:-}"
```
c3d, line 27: Looks to me like this could be simplified as:

```
local path=$(readlink --canonicalize-existing "$1" || _fatal "need file to resolve, got $1")
echo $path
```

* Testing the empty case is not necessary, since readlink itself errors out in that case
* Checking for `/` is not necessary, since `-e` is a shortcut for `--canonicalize-existing`, which itself does canonicalize

I also used the extended option name, usually clearer in scripts.

lib/error.sh
```diff
@@ -0,0 +1,217 @@
...
+{
+	local pid="${$}"
+
+	local details
+
+	stderr "process-hierarchy:"
+
+	local -i i=0
+
+	while true
```
c3d, line 52: As I've asked elsewhere, is there any good reason for not using a `for ((i=0; ; i++))` loop here?

lib/error.sh
```diff
@@ -0,0 +1,217 @@
...
+		local details
+		local msg
+
+		local current=''
+
+		[ "${pid}" = "${$}" ] && current=", current='yes'"
+
+		details=$(ps --no-headers -p "$pid" -o ppid,cmd)
+
+		local ppid=$(echo "$details" | awk '{print $1}')
```
c3d, line 63: Could you elaborate on what you are trying to do here? This looks like you are trying to strip whitespace?

lib/error.sh
jodh-intel: Fixed.

lib/common.bash
jodh-intel: Yep - I did think about this. I've now changed the function so that verbose mode is enabled if:

- Running under the CI (`CI=true`)
- Manually requested (`KATA_TEST_VERBOSE=true`)

lib/common.bash
jodh-intel: `shellcheck(1)` complains otherwise:

```
SC2155: Declare and assign separately to avoid masking return values.
```

lib/error.sh
jodh-intel: Agreed - fixed.

lib/common.bash
jodh-intel: Plus https://github.com/kata-containers/tests/pull/4733/files#r858730128 explains the message.

lib/error.sh
```diff
@@ -0,0 +1,218 @@
...
+	stderr "  groups: '$(id)'"
+
+	stderr "bash:"
+	stderr "  version: '${BASH_VERSION}'"
+	stderr "  version-info: '${BASH_VERSINFO[*]}'"
+
+	stderr "environment:"
+
+	# Remove bash functions (that can span multiple lines)
+	env |\
```
c3d, line 200: 😳

I searched longe and hard to find something slightly simpler, and the simplest I can come up with is:

```
env $(env | sed -e "s/\(BASH_FUNC_.*%%\)=().*/-u \1/g;t;d" ) 
```

The way this works is
* Run the original env through a sed that replaces any `BASH_FUNC_foo%%=...` with `-u BASH_FUNC_foo%%`
* Pass these various `-u` options to `env`…
* without a command, so it still prints the environment.

c3d (approved):
Overall, I find that this is a vast improvement. Thanks @jodh-intel 

Still a minor nit, but in any case, approving because it's better.

fidencio (approved):
@jodh-intel, lgtm!

I do *not* see this as a backport candidate, let me know if you think differently.

PR #4758, opened 2022-05-03
Reorganize the tests - part 3 (move tracing to functional)

### LINKED ISSUE #4645: Re-think the organization of our tests
### Background

Our tests are spread over the `integration`, `functional`, `metrics`, and `tracing` directories so that they can be grouped by type of test.

I wasn't here when those directories were created so I can only take some assumptions. And I assume they were created with the following purposes:

 * integration: to hold tests that assert the integration with external entities. For example, runtimes (containerd, CRI-O,...), engines (docker, podman), systems (Kubernetes). Very often the tests consist of running the external entity suites of tests with Kata Containers configured as the default runtime, so asserting the entity work fine. 
 * functional: contain tests that assert a specific feature . For example, the ability run a rootless VMM (see https://github.com/kata-containers/kata-containers/blob/main/docs/how-to/how-to-run-rootless-vmm.md).
 * metrics: to hold tests that ensure no performance regressions are introduced. This is the home of the metrics CI.
 * tracing: I don't know why it exists, maybe it belongs to functional.

### The problem

We didn't document the purpose of those directories nor provide clear guidelines about where new tests better fit in. As a result, over the time we end up with most of the tests landing in the `integration` directory although many are actually functional tests. Also what seems to be new categories (the stability tests) also ended up in `integration`.  

### Proposing a reorganization

Here I propose two things:

1. that we re-organize the tests
1. that we document the new organization and provide guidelines for the developers

As for 1), I propose that we:
 * [ ] create a new directory `stability` out of `integration` to hold the stability tests
 * [ ] move `tracing` to `functional` as it is actually a functional test
 * [ ] move `agent`, `ksm`,  `pmem`, `rootless`,  `sandbox_cgroup`, `vcpus` from `integration` to `functional`
 * keep `containerd`, `cri-o`, `kubernetes`, **`nydus`, **`qat`  in `integration`
 * [ ] rename `functional/tdx` to `functional/confidential_guest` as likely soon there will be tests for other technologies like AMD SEV.

(**) I still don't have a strong option about `integration/nydus` and `integration/qat` but maybe they logically belong to integration.

Regarding 2), I propose that we:
 * create a new advice-style document for the non-unit tests. For unit tests we already have advises: https://github.com/kata-containers/kata-containers/blob/main/docs/Unit-Test-Advice.md

wainersm:
Continuing part 1 (#4751)  and 2 (#4752)...

The tracing tests are conceptually functional testing, so moving them to the
right directory.

Fixes #4645

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,12 +100,12 @@ qat:
 	bash integration/qat/qat_test.sh
 
 agent-shutdown:
-	bash tracing/test-agent-shutdown.sh
+	bash functional/tracing/test-agent-shutdown.sh
 
 # Tracing requires the agent to shutdown cleanly,
 # so run the shutdown test first.
 tracing: agent-shutdown
-	bash tracing/tracing-test.sh
+	bash functional/tracing/tracing-test.sh
 
 vcpus:
 	bash -f functional/vcpus/default_vcpus_test.sh
--- a/functional/tracing/test-agent-shutdown.sh
+++ b/functional/tracing/test-agent-shutdown.sh
@@ -41,7 +41,7 @@ set -o pipefail
 set -o errtrace
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
-source "${SCRIPT_PATH}/../.ci/lib.sh"
+source "${SCRIPT_PATH}/../../.ci/lib.sh"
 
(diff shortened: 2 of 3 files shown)


jodh-intel (approved):
Thanks @wainersm.

lgtm

/cc @cmaf

snir911 (approved):

cmaf (approved):

Issue #4788, opened 2022-05-13
virtiofsd: Only install the rust version when testing on x86_64


fidencio:
Previously, we've been installing the C and the rust version of the virtiofsd daemon so we would be able to test whether the rust version would actually work as expected.

https://github.com/kata-containers/kata-containers/pull/4253 was the first CI running only the rust daemon for x86_64 and, hey, it's green. :-)

With this in mind, let's stop installing the C version of virtiofsd when testing on x86_64.

PR #4789, opened 2022-05-13
virtiofsd: Do not install the C version of the daemon (x86_64)

fidencio:
For x86_64, instead of installing both the C and the rust versions of
the daemon, let's just install the rust one as we have the tests passing
with it.

Fixes: #4788
Depends-on: github.com/kata-containers/kata-containers#4253


cmaf (approved):
lgtm @fidencio, ci failures seem to be unrelated

sameo (approved):

Issue #4791, opened 2022-05-16
CIs are breaking due to leftover code looking for the C version of virtiofsd (x86_64)


fidencio:
For some reason it was not caught by the CI before, but the `install_qemu` script looks for the virtiofsd in the location where it should be if built and installed via QEMU.

This is not needed anymore, at least for x86_64, as we're using the rust version of virtiofsd.

PR #4792, opened 2022-05-16
ci: Don't look for the C version of virtiofsd (x86_64)

fidencio:
For some reason it was not caught by the CI before, but the install_qemu
script looks for the virtiofsd in the location where it should be if
built and installed via QEMU.

This is not needed anymore, at least for x86_64, as we're using the rust
version of virtiofsd.

Fixes: #4791


fengwang666 (approved):

cmaf (approved):

fidencio:
ARM CI failure, unrelated to this PR.
```
23:56:54 error execution phase preflight: [preflight] Some fatal errors occurred:
23:56:54 	[ERROR Port-6443]: Port 6443 is in use
```

/retest-arm

bergwolf (approved):

PR #4799, opened 2022-05-18
Adapt Openshift CI for the virtiofsd-rs

### LINKED ISSUE #4798: Recent migration to virtiofsd-rs broke the jobs in OpenShift CI
The script `.ci/openshift-ci/build_install.sh` doesn't call `.ci/install_kata.sh`, instead components' s scripts are called separated. As virtiofsd is no longer built with QEMU, the `build_install.sh` should call the `.ci/install_virtiofsd.sh`.

wainersm:
This fixes the scripts so that virtiofsd (rust) is proper installed in the OpenShift CI environment.

Fixes #4798

### DIFF
--- a/.ci/install_virtiofsd.sh
+++ b/.ci/install_virtiofsd.sh
@@ -13,6 +13,8 @@ set -o errtrace
 cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
+DESTDIR="${DESTDIR:-/}"
+
 main() {
 	local buildscript="${katacontainers_repo_dir}/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh"
 
@@ -21,8 +23,14 @@ main() {
 
 	pushd $katacontainers_repo_dir
 	sudo -E PATH=$PATH bash ${buildscript} --build=virtiofsd
-	sudo tar -xvJpf build/kata-static-virtiofsd.tar.xz -C /
-	sudo ln -sf /opt/kata/libexec/virtiofsd /usr/libexec/virtiofsd
+	sudo tar -xvJpf build/kata-static-virtiofsd.tar.xz -C "${DESTDIR}"
+	# Kata CI requires the link but this isn't true to all scenarios,
+	# for example, on OpenShift CI everything should be installed under
+	# /opt/kata. So do not try to create the link unless the directory
+	# exist.
+	[ -d "${DESTDIR}/usr/libexec" ] && \
+		sudo ln -sf "${DESTDIR}/opt/kata/libexec/virtiofsd" \
+			"${DESTDIR}/usr/libexec/virtiofsd"
 	popd
 }
 
--- a/.ci/openshift-ci/build_install.sh
+++ b/.ci/openshift-ci/build_install.sh
@@ -152,6 +152,8 @@ yum install -y dracut
 
 "${cidir}/install_runtime.sh"
 
+"${cidir}/install_virtiofsd.sh"
+
 # The resulting kata installation will be merged in rhcos filesystem, and
 # symlinks are troublesome. So instead let's convert them to in-place files.
 for ltarget in $(find ${DESTDIR} -type l); do


.ci/install_virtiofsd.sh
```diff
@@ -21,8 +23,10 @@ main() {
 
 	pushd $katacontainers_repo_dir
 	sudo -E PATH=$PATH bash ${buildscript} --build=virtiofsd
-	sudo tar -xvJpf build/kata-static-virtiofsd.tar.xz -C /
-	sudo ln -sf /opt/kata/libexec/virtiofsd /usr/libexec/virtiofsd
+	sudo tar -xvJpf build/kata-static-virtiofsd.tar.xz -C "${DESTDIR}"
+	[ -d "${DESTDIR}/usr/libexec/virtiofsd" ] && \
```
snir911, line 27: Shouldn't it be "-x" ?

.ci/install_virtiofsd.sh
```diff
@@ -13,6 +13,8 @@ set -o errtrace
 cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
+DESTDIR="${DESTDIR:-/}"
```
gkurz, line 16: Heh a skeptical DESTDIR ;-)

.ci/install_virtiofsd.sh
gkurz: The check itself doesn't quite make sense since `${DESTDIR}/usr/libexec/virtiofsd` isn't created yet at this point. I guess you mean `${DESTDIR}/opt/kata/libexec/virtiofsd`.

BTW the intent behind this check is to fail early if the `virtiofsd` binary isn't found, right ? If so, I'd rather make this explicit with something like `check || return 1`

wainersm: ops, made a mistake... I wanted to check the existence of `/usr/libexec` directory. It might happen that directory doesn't exist on `$DESTDIR`, that's the case on openshift ci build environment, so we won't attempt to create the link. I will fix that.

wainersm: IIRC on Kata CI the path to virtiofsd in `configuration.toml` is `/usr/libexec/virtiofsd` so the need of the link. Whereas on openshift-ci, everything kata-related should be installed under `/opt/kata` and `configuration.toml` already have the path to virtiofsd correctly, thus it doesn't need `virtiofsd` under `/usr/libexec`. More than that, the kata installation script on openshift CI cannot copy files over `/usr` as it is a read-only directory, and so having that link will introduce more problem.

wainersm: done

wainersm:
> @wainersm, I'd wait till v1.2.1 is out, otherwise you'll hit issues with the OpenShift CI as it runs, or at least should run, with SELinux enabled. See: https://gitlab.com/virtio-fs/virtiofsd/-/issues/49

Currently the CI runs kata with SELinux disabled so it should not be a problem. Thanks for the heads up @fidencio !

.ci/install_virtiofsd.sh
```diff
@@ -21,8 +23,14 @@ main() {
...
 	pushd $katacontainers_repo_dir
 	sudo -E PATH=$PATH bash ${buildscript} --build=virtiofsd
-	sudo tar -xvJpf build/kata-static-virtiofsd.tar.xz -C /
-	sudo ln -sf /opt/kata/libexec/virtiofsd /usr/libexec/virtiofsd
+	sudo tar -xvJpf build/kata-static-virtiofsd.tar.xz -C "${DESTDIR}"
+	# Kata CI requires the link but this isn't true to all scenarios,
+	# for example, on OpenShift CI everything should be installed under
+	# /opt/kata. So do not try to create the link unless the directory
+	# exist.
+	[ -d "${DESTDIR}/usr/libexec" ] && \
```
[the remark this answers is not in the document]
snir911: ahh, right, sorry, LGTM

snir911 (approved):

gkurz (approved):
LGTM. Thanks @wainersm !

PR #4805, opened 2022-05-24
ci/lib.sh: fix sha256sum_from_files() not picking all files

### LINKED ISSUE #4311: Only run URL checks in github workflow
Not sure how many of the Jenkins CI jobs are doing it right now, but apparently they fall into the category of static tests and should be done **ONCE** in github workflow, instead of repeating the same tests in different Jenkins jobs.

Example of URL checks in Jenkins jobs:

http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-PR/1238/console

http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-ARM-PR/3008/console

wainersm:
The calc_qemu_files_sha256sum() passes a list of files and directories
to sha256sum_from_files() so that the sha256sum of QEMU build scripts is
calculated. However, sha256sum_from_files() is picking only the first parameter,
ie., in the end it calculates the hash of only one file/directory and,
therefore, the QEMU cache sha256sum is computed wrongly.

Fixes #4311


fidencio (approved):
lgtm, thanks @wainersm!
/cc @egernst

PR #4807, opened 2022-05-24
stable 2.4 | Fix QEMU cache resolution

wainersm:
Fix the problem of getting the QEMU from cache (main branch)

* ci/lib.sh: fix sha256sum_from_files() not picking all files
* README: fix broken urls

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -482,7 +482,7 @@ get_pr_changed_file_details()
 # Gets a list of files and/or directories to build a SHA-256 from their contents.
 # Returns the SHA-256 hash if succeeded, otherwise an empty string.
 sha256sum_from_files() {
-	local files_in=${1:-}
+	local files_in=${@:-}
 	local files=""
 	local shasum=""
 
--- a/README.md
+++ b/README.md
@@ -16,11 +16,11 @@ We provide several tests to ensure Kata-Containers run on different scenarios
 and with different container managers.
 
 1. Integration tests to ensure compatibility with:
-   - [Kubernetes](https://github.com/kata-containers/tests/tree/main/integration/kubernetes)
-   - [Containerd](https://github.com/kata-containers/tests/tree/main/integration/containerd)
-2. [Stability tests](https://github.com/kata-containers/tests/tree/main/integration/stability)
-3. [Metrics](https://github.com/kata-containers/tests/tree/main/metrics)
-4. [VFIO](https://github.com/kata-containers/tests/tree/main/functional/vfio)
+   - [Kubernetes](./integration/kubernetes)
+   - [Containerd](./integration/containerd)
+2. [Stability tests](./integration/stability)
+3. [Metrics](./metrics)
+4. [VFIO](./functional/vfio)
 
 ## CI Content
 


.ci/ci-fast-return.sh
```diff
@@ -27,7 +27,7 @@ source "${cidir}/lib.sh"
 
 # If no branch specified, compare against the main branch.
 # The 'branch' var is required by the get_pr() lib functions.
-branch=${branch:-main}
+branch=${branch:-stable-2.4}
```
fidencio, line 30: Is this something we will have to do on every branch bump?

.ci/lib.sh
```diff
@@ -12,7 +12,7 @@ export KATA_ETC_CONFIG_PATH="/etc/kata-containers/configuration.toml"
 
 export katacontainers_repo=${katacontainers_repo:="github.com/kata-containers/kata-containers"}
 export katacontainers_repo_dir="${GOPATH}/src/${katacontainers_repo}"
-export kata_default_branch="${kata_default_branch:-main}"
+export kata_default_branch="${kata_default_branch:-stable-2.4}"
```
fidencio, line 15: Same question.

.ci/static-checks.sh
```diff
@@ -40,7 +40,7 @@ typeset -r arch_func_regex="_arch_specific$"
 repo=""
 specific_branch="false"
 force="false"
-branch=${branch:-main}
+branch=${branch:-stable-2.4}
```
fidencio, line 43: Same question, as I'm looking for a way to have done, maybe, by the script called on the jenkins side.

fidencio (commented):
lgtm, but I'd like to understand better whether we can avoid manually changing the default branch every time we branch.

.ci/ci-fast-return.sh
wainersm: I don't have an answer right now, I will have to look at the scripts and see if there is a way to read branch value from the same source.

Just to be clear, if we drop those changes than the QEMU problem is still fixed by the other commit (https://github.com/kata-containers/tests/pull/4807/commits/fc20fff6bb4d9489668420cabbf41b64c1b254ca).

wainersm: done.

wainersm:
@fidencio based on the failed static check https://github.com/kata-containers/tests/runs/6581583405?check_suite_focus=true I suspect we will need the other commit that I dropped

fidencio:
> @fidencio based on the failed static check https://github.com/kata-containers/tests/runs/6581583405?check_suite_focus=true I suspect we will need the other commit that I dropped

Ouch, you're right, we need those. :-/
I thought those were fixed by a PR raised by @snir911 a long time ago, that would ensure we're running the tests against the correct branch.

snir911:
I remember something vaguely, but it seems it fail because of an actual broken link, a, i missing something?

snir911:
@wainersm if the commits that changed the files are only doc changes and has no conflicts i'd cherry pick them and add to this PR, otherwise, I'll go with 2.
(although I'm not strongly against the 2 option directly, i'll be fine with that too :P)

wainersm:
> @wainersm if the commits that changed the files are only doc changes and has no conflicts i'd cherry pick them and add to this PR, otherwise, I'll go with 2. (although I'm not strongly against the 2 option directly, i'll be fine with that too :P)

Follow are the changes, they re-organize the test files. It is not a bug fix nor exactly a new feature...

https://github.com/kata-containers/tests/pull/4751
https://github.com/kata-containers/tests/pull/4752
https://github.com/kata-containers/tests/pull/4758

Let me cherry-pick them all and see what happens.

snir911 (approved):
OK, nevermind, LGTM, just fix the link

wainersm:
Updated this PR with a new commit to fix the links in the README.md file.

fidencio (approved):
lgtm, thanks @wainersm!

PR #4843, opened 2022-06-13
metrics: Update cassandra to support block volumes

### LINKED ISSUE #4842: metrics: update cassandra test to support block volumes

GabyCT:
This PR updates the cassandra kuberentes test to support block volumes.

Fixes #4842


metrics/disk/cassandra_kubernetes/cassandra.sh
```diff
@@ -125,21 +125,51 @@ function cassandra_start() {
 	wait_time=20
  	sleep_time=2
 
+	vol_capacity="500M"
```
amshinde, line 128: Is this capacity enough for the cassandra workload? If I recall the stateful set was using at least 1G with three replicas. 
I would bump this to 3G atleast,

metrics/disk/cassandra_kubernetes/cassandra.sh
```diff
@@ -125,21 +125,51 @@ function cassandra_start() {
...
 	kubectl create -f "${SCRIPT_PATH}/runtimeclass_workloads/cassandra-service.yaml"
 
 	# Check service
 	kubectl get svc | grep "$service_name"
 
-	# Create local volumes
-	kubectl create -f "${SCRIPT_PATH}/runtimeclass_workloads/local-volumes.yaml"
-
-	# Create statefulset
-	kubectl create -f "${SCRIPT_PATH}/runtimeclass_workloads/cassandra-statefulset.yaml"
```
amshinde, line 138: Why is the stateful set not being used anymore?

metrics/disk/cassandra_kubernetes/volume/local-storage.yaml
```diff
@@ -0,0 +1,6 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: local-storage
```
amshinde, line 4: Can we change the name here to block-local-storage to be clear and similarly update at all other places. 
My idea is we could augment the test in the near future both to test block as well as virtio-fs.
For testing virtio-fs we could use another storage class setup using host volumes. What do you think?

metrics/disk/cassandra_kubernetes/cassandra.sh
GabyCT: @amshinde yes it is being used see https://github.com/kata-containers/tests/blob/9822556eb79afac3f8f1ca4bbe169850dc3b10f6/metrics/disk/cassandra_kubernetes/cassandra.sh#L165

metrics/disk/cassandra_kubernetes/cassandra.sh
GabyCT: @amshinde changes applied

metrics/disk/cassandra_kubernetes/volume/local-storage.yaml
GabyCT: @amshinde changes applied

amshinde (approved):

dborquez (approved):
LGTM  thanks @GabyCT

PR #4881, opened 2022-06-23
DO-NOT-MERGE: Adjust tests to a newer version of crictl

fidencio:


stevenhorsman:
@fidencio - IIRC I think we were using this PR when we were trying to debug the issues when bumping to containerd 1.6.6, so I think we can close it out now?

Issue #4894, opened 2022-06-30
CC | confidential: Adjust switch_image_service_offload()


fidencio:
Due to the changes introduced as part of
https://github.com/kata-containers/kata-containers/pull/4565, we should
adjust the regular expression used for switching the configuration on.

PR #4895, opened 2022-06-30
CC | confidential: Adjust switch_image_service_offload()

fidencio:
Due to the changes introduced as part of
https://github.com/kata-containers/kata-containers/pull/4565, we should
adjust the regular expression used for switching the configuration on.

Fixes: #4894

### DIFF
--- a/integration/confidential/lib.sh
+++ b/integration/confidential/lib.sh
@@ -29,11 +29,11 @@ switch_image_service_offload() {
 
 	case "$1" in
 		"on")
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):
The change looks good. ~Does the commit need a `Depends-on: github.com/kata-containers/kata-containers#4565` in order to pass the tests as otherwise I'm not sure the regex will match if the `hash` comment is still there?~ Ignore this - I've just seen that you've done the depends on in the kata-containers PR, so as long as that is merged and then this tested we are all good!

Issue #4896, opened 2022-06-30
ci: Set SEV TEE_TYPE


Alex-Carter01:
As a part of #4861, SEV CI tests will want to specify their TEE_TYPE, and have everything configured. This will start the process of configuring kata based on the choice of SEV. This allows the use of the new kernel build added as a part of #4838

I intend to make a followup PR with a case in install_runtime.sh which will make changes to the kata-config file based similarly on TEE_TYPE being set to the SEV option.

PR #4838, opened 2022-06-09
ci: add sev build option for kernel

Alex-Carter01:
Adding another case to the CI's kernel builder for sev.

Depends on https://github.com/kata-containers/kata-containers/pull/4180 this PR in kata for additions to the kernel builder for efi_secret which is needed for SEV.

Fixes #4837
Signed-off-by: Alex Carter [alex.carter@ibm.com](mailto:alex.carter@ibm.com)

@fitzthum @larrydewey @ryansavino

PR #4897, opened 2022-06-30
ci: set SEV TEE_TYPE

Alex-Carter01:
SEV tests need kata to be configured based on the selected TEE TYPE. This change ensures that the correct kernel build is made based on previous functionality from #4838

Fixes: #4896

Signed-off-by: Alex Carter <alex.carter@ibm.com>

### DIFF
--- a/.ci/install_kata.sh
+++ b/.ci/install_kata.sh
@@ -28,6 +28,10 @@ if [ "${TEE_TYPE:-}" == "tdx" ]; then
 	KATA_BUILD_QEMU_TYPE="${KATA_BUILD_QEMU_TYPE:-tdx}"
 fi
 
+if [ "${TEE_TYPE:-}" == "sev" ]; then
+	KATA_BUILD_KERNEL_TYPE=sev
+fi
+
 echo "Install Kata Containers Image"
 echo "rust image is default for Kata 2.0"
 "${cidir}/install_kata_image.sh" "${tag}"


GabyCT (approved):

Alex-Carter01:
Updated just to add information to the commit message

.ci/install_kata.sh
```diff
@@ -28,6 +28,10 @@ if [ "${TEE_TYPE:-}" == "tdx" ]; then
 	KATA_BUILD_QEMU_TYPE="${KATA_BUILD_QEMU_TYPE:-tdx}"
 fi
 
+if [ "${TEE_TYPE:-}" == "sev" ]; then
+	KATA_BUILD_KERNEL_TYPE="${KATA_BUILD_KERNEL_TYPE:-sev}"
```
amshinde, line 32: Wouldn't you want this to be `KATA_BUILD_KERNEL_TYPE=sev`  instead of using the previous value if set?

Alex-Carter01:
Switched to just sev per @amshinde 's suggestion. In full transparency my reasoning was to follow the lead of tdx to let someone select a different kernel build if they really want. But as much as I think about it I can't think of a case where someone would want SEV without the appropriate kernel.

fidencio (approved):
lgtm, thanks @Alex-Carter01!

PR #4919, opened 2022-07-11
build: Enable QEMU cache for s390x

BbolroC:
As per the confirmation that `build-static-qemu.sh` from `kata-container`
works for s390x, it makes sense to turn on caching for the architecture.
The corresponding Jenkins job has been created and tested:
http://jenkins.katacontainers.io/job/kata-containers-2.0-qemu-s390x/

Fixes: #4920 

### DIFF
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -40,8 +40,8 @@ update_cache() {
 
 build_static_qemu() {
 	info "building static QEMU"
-	# only x86_64 is supported for building static QEMU
-	[ "$ARCH" != "x86_64" ] && return 1
+	# only x86_64 and s390x are supported for building static QEMU
+	[ "$ARCH" != "x86_64" ] && [ "$ARCH" != "s390x" ] && return 1
 
 	(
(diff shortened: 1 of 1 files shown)


BbolroC:
@GabyCT Could you review this if you have time? It would be grateful if you let me know what is further needed to achieve the goal outside this PR.

GabyCT:
@BbolroC first you need to create a jenkins job to have the qemu-cache please see http://jenkins.katacontainers.io/view/JJB%20managed/job/kata-containers-main-qemu-x86_64/, in order to test that you PR is working you should test it by using the jenkins job and see that is working properly

BbolroC:
Hi, @GabyCT . Just one small question regarding the Jenkins job: do I have to create another job named `kata-containers-main-qemu-s390x` even if the source code checks different name like https://github.com/kata-containers/tests/blob/main/.ci/install_qemu.sh#L213 (see that it has a term `2.0` instead of `main`)

GabyCT (approved):

fidencio (approved):
lgtm, thanks @BbolroC!

Issue #4925, opened 2022-07-13
Make test into several sub-tests to run them individually


BbolroC:
**Which feature do you think can be improved?**

The test running time for `s390x` was around 1h 15m. It is quite long compared to that of `x86_64` (around 35m). We've tried to reduce this time by introducing caching for qemu (#4919) and kernel image, but only 5 mins are saved (the running time now is reduced to around 1h 10m).

In [jenkins_job_build.sh](https://github.com/kata-containers/tests/blob/main/.ci/jenkins_job_build.sh#L173), we could divide the test into 3 chunks like:

- Unit test: only for `s390x` and `aarch64` (measured running time: 15 mins)
- Main test: for all
- Snap test: for all, but `x86_64` (measured running time: 10 mins)

If we make them separate, triggered individually by a parameter, run them in parallel, it would bring us another 25-min time saving.

The measured running time, when the 3 parts are run individually, are as follows:

- Unit test: 42 mins (http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-s390x-PR/986/)
- Main test: 43 mins (http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-s390x-PR/987/)
- Snap test: 36 mins (http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-s390x-PR/988/)

**How can it be improved?**

The test in [jenkins_job_build.sh](https://github.com/kata-containers/tests/blob/main/.ci/jenkins_job_build.sh#L173) could be divided into three parts above-mentioned and triggered individually by a parameter. 

**Additional Information**

Anything else to add?

**Before raising this enhancement request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?

PR #4926, opened 2022-07-13
build: Make test into several sub-tests

BbolroC:
This PR is to make the current test divided into 3 parts and triggered individually via a parameter.
The code changes do not break any existing CI configurations. Please look at the issue below for the detailed explanation. 

Fixes: #4925

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -23,6 +23,22 @@ WORKSPACE=${WORKSPACE:-}
 BAREMETAL=${BAREMETAL:-false}
 TMPDIR=${TMPDIR:-}
 
+usage() {
+	cat <<EOF
+Usage: jenkins_job_build.sh REPOSITORY [OPTION]...
+
+REPOSITORY
+	repository name: github.com/kata-containers/(kata-containers|tests)
+
+OPTION:
+	-t, --type      test type: (all|unit|main|snap) (default: all)
+
+	Examples:
+	jenkins_job_build.sh "github.com/kata-containers/tests"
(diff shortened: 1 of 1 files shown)


liubin (approved):

.ci/jenkins_job_build.sh
```diff
@@ -31,11 +47,47 @@ fi
...
+	option=$1
+	case ${option} in
+	-t|--type)
+		if [[ $# -lt 2 ]]; then
+			echo "Missing parameter after ${option}" >&2
+			exit 1
+		fi
+		shift
+		parameter=$1
+		case ${parameter} in
```
liubin, line 78: Nit: Since https://github.com/kata-containers/tests/pull/4926/files#diff-2b38dc854ae065a444de231fcf9436e6954ee0b9c71052edda190c1c0a72f5e6R290-R302 is checking the `test_type`, it's safe to assign `$parameter` to `test_type` directly.

BbolroC: Thanks for the comment. I will reflect this on the code. :wink:

BbolroC: Ah, but in this case, this check could function as an early catch for a wrong parameter, which does not make the script go through `set.up` until the error is identified. What do you think?

BbolroC:
@Amulyam24 it looks there is a problem on the slave:

```
[kubelet-check] Initial timeout of 40s passed.
[kubelet-check] It seems like the kubelet isn't running or healthy.
```

This has also happened to #4931 since the last evening. It would be grateful if you could have a look at the machine.

bergwolf (approved):

PR #4932, opened 2022-07-13
ci: Allow commits to start with 'Merge' instead of subsystem

### LINKED ISSUE #4929: 'Check Subsystem' fails for merge commits
Currently the 'Commit Message Check' for a PR fails when there is a merge commit included - this is because the merge commit has no subsystem type in the commit message which causes failure.
You can see an example of this [here](https://github.com/kata-containers/tests/runs/7324203259?check_suite_focus=true) in the latest PR I opened to merge main into the CCv0 branch:

bookinabox:
Allows merges not down via PRs not to be flagged by the commit checker.

Lines that start with 'Merge ' will be accepted alongside commits with a
subsystem of the form 'subsystem: '.

Fixes #4929

Signed-off-by: Derek Lee <derlee@redhat.com>

### DIFF
--- a/.github/workflows/commit-message-check.yaml
+++ b/.github/workflows/commit-message-check.yaml
@@ -93,6 +93,6 @@ jobs:
       uses: tim-actions/commit-message-checker-with-regex@v0.3.1
       with:
         commits: ${{ steps.get-pr-commits.outputs.commits }}
-        pattern: '^[\s\t]*[^:\s\t]+[\s\t]*:'
+        pattern: '^[\s\t]*[^:\s\t]+[\s\t]*:|^Merge\s\t*'
         error: 'Failed to find subsystem in subject'
         post_error: ${{ env.error_msg }}


snir911:
@stevenhorsman can you TAL, it was mentioned in the issue you fixed that for the kata-container repo

fidencio:
> @fidencio - I'm not sure if you have a view on whether we add this fix into `main` vs `CCv0`?

I think we can have this into main, although it'll never be used there.
Any fork, and I sincerely hope we lower the number of forks rather than increasing them, can simply take advantage of it then.

fidencio:
> > @fidencio - I'm not sure if you have a view on whether we add this fix into `main` vs `CCv0`?
> 
> I think we can have this into main, although it'll never be used there. Any fork, and I sincerely hope we lower the number of forks rather than increasing them, can simply take advantage of it then.

Hmm. Scratch that, Steve pointed out that we only have that rule for the `CCv0` branch for now.  So, let's follow what we already have and have a similar check as part of the `CCv0`, for the sake of consistency.

stevenhorsman:
Based on fidencio's comments I'll try and re-target this PR at `CCv0` and hope that is doesn’t confuse the automation too much 🤞

bookinabox:
Thanks for checking it out!

PR #4942, opened 2022-07-16
runk: add tests for runk

### LINKED ISSUE #4882: Add tests cases for runk
Add tests for runk as an OCI runtime.

cyyzero:
Use runk as the default OCI runtime of containerd to validate runk's
commands. Tests includes creating containers, exec a process, kill and
delete containers.

Fixes #4882

Signed-off-by: Chen Yiyang <cyyzero@qq.com>

### DIFF
--- a/integration/containerd/runk/runk-tests.sh
+++ b/integration/containerd/runk/runk-tests.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Kata Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# This test will validate runk with containerd
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o errtrace
+
+dir_path=$(dirname "$0")
+source /etc/os-release || source /usr/lib/os-release
+source "${dir_path}/../../../lib/common.bash"
+source "${dir_path}/../../../.ci/lib.sh"
+source "${dir_path}/../../../metrics/lib/common.bash"
+
+KATACONTAINERS_REPO=${katacontainers_repo:="github.com/kata-containers/kata-containers"}
+RUNK_SRC_PATH="${GOPATH}/src/${KATACONTAINERS_REPO}/src/tools/runk"
+RUNK_BIN_PATH="/usr/local/bin/runk"
+TEST_IMAGE="docker.io/library/busybox:latest"
+CONTAINER_ID="id1"
+PID_FILE="${CONTAINER_ID}.pid"
+
+# runk can't work well on cgroup V2 environment now, so we temporarily skip this test
+if [ $(stat -f --format %T /sys/fs/cgroup) == "cgroup2fs" ]; then
+    echo "runk can't work well on cgroup V2 environment now, so we temporarily skip this test"
+    exit 0
+fi
+
+setup() {
+    # can't find cargo when make runk below, so we install rust to make runk build correctly
+    "${dir_path}/../../../.ci/install_rust.sh" && source "$HOME/.cargo/env"
+    echo "restart containerd service"
+    sudo systemctl restart containerd
+    echo "pull container image"
+    check_images ${TEST_IMAGE}
+}
+
+install_runk() {
+    echo "Install runk"
+    pushd ${RUNK_SRC_PATH}
+    make
+    sudo make install
+    popd
+}
+
+test_runk() {
+    echo "start container with runk"
+    sudo ctr run --pid-file ${PID_FILE} --rm -d --runc-binary ${RUNK_BIN_PATH} ${TEST_IMAGE} ${CONTAINER_ID}
+    read CID PID STATUS <<< $(sudo ctr t ls | grep ${CONTAINER_ID})
+    [ ${PID} == $(cat ${PID_FILE}) ] || die "pid is not consistent"
+    [ ${STATUS} == "RUNNING" ] || die "contianer status is not RUNNING"
+
+    echo "exec process in a container"
+    sudo ctr t exec --exec-id id1 ${CONTAINER_ID} sh -c "echo hello > /tmp/foo"
+    [ "hello" == "$(sudo ctr t exec --exec-id id1 ${CONTAINER_ID} cat /tmp/foo)" ] || die "exec process failed"
+
+    echo "test ps command"
+    sudo ctr t exec --detach --exec-id id1 ${CONTAINER_ID} sh
+    # one line is the titles, and the other 2 lines are porcess info
+    [ "3" == "$(sudo ctr t ps ${CONTAINER_ID} | wc -l)" ] || die "ps command failed"
+
+    echo "kill the container and poll until it is stopped"
+    sudo ctr t kill --signal SIGKILL --all ${CONTAINER_ID}
+    # poll for a while until the task receives signal and exit
+    local cmd='[ "STOPPED" == "$(sudo ctr t ls | grep ${CONTAINER_ID} | awk "{print \$3}")" ]'
+    waitForProcess 10 1 "${cmd}" || die "failed to kill task"
+
+    echo "check the container is stopped, and delete it"
+    # there is only title line of ps command
+    [ "1" == "$(sudo ctr t ps ${CONTAINER_ID} | wc -l)" ] || die "kill command failed"
+    sudo ctr t rm ${CONTAINER_ID} || die "failed to delete task"
+    [ -z "$(sudo ctr t ls | grep ${CONTAINER_ID})" ] || die "failed to delete task"
+    sudo ctr c rm ${CONTAINER_ID} || die "failed to delete container"
+}
+
+clean_up() {
+    rm -f ${PID_FILE}
+}
+
+setup
+install_runk
+test_runk
+clean_up
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -65,6 +65,8 @@ case "${CI_JOB}" in
 			echo "INFO: Running Confidential Container tests"
 			sudo -E PATH="$PATH" CRI_RUNTIME="containerd" bash -c "make cc-containerd"
 		fi
+		echo "INFO: Running runk test"
+		sudo -E PATH="$PATH" CRI_RUNTIME="containerd" bash -c "make runk"
 		;;
 	"CRIO_K8S")
 		echo "INFO: Running kubernetes tests"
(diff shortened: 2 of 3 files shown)


cyyzero:
\test

integration/containerd/runk/runk-tests.sh
```diff
@@ -0,0 +1,61 @@
...
+setup() {
+    echo "restart containerd service"
+    sudo systemctl restart containerd
+    echo "pull container image"
+    sudo ctr images pull ${TEST_IMAGE}
+}
+
+test_runk() {
+    echo "start container with runk"
+    sudo ctr run --pid-file ${CONTAINER_ID}.pid --rm -d  --runc-binary ${RUNK_BIN_PATH} ${TEST_IMAGE} ${CONTAINER_ID}
```
liubin, line 37: It's better to configure containerd to use runk like https://github.com/kata-containers/kata-containers/blob/main/src/tools/runk/README.md#configure-containerd-to-use-runk and run
`ctr run --runtime runk ...`

integration/containerd/runk/runk-tests.sh
```diff
@@ -0,0 +1,61 @@
...
+    read CID PID STATUS <<< $(sudo ctr t ls | grep ${CONTAINER_ID})
+    [ ${PID} == $(cat ${CONTAINER_ID}.pid) ] || die "pid is not consistent"
+    [ ${STATUS} == "RUNNING" ] || die "contianer status is not RUNNING"
+
+    echo "exec process in a container"
+    sudo ctr t exec -d --exec-id id1  ${CONTAINER_ID} sh -c "while true; do echo hello; sleep 1; done"
```
liubin, line 43: It's better to run exec and get some results. For example:
```bash
ctr exec touch /tmp/foo
ctr exec ls /tmp/foo
```
create one file in the first exec and check it existence in the second exec.

integration/containerd/runk/runk-tests.sh
cyyzero: It seems that `ctr` and `nerdctl` ignores [plugins."io.containerd.grpc.v1.cri"] configuration because they do not use the CRI API. https://github.com/containerd/containerd/discussions/6162#discussioncomment-2032800

cyyzero:
@liubin I have rebased this branch. Please approve and run tests again.

integration/containerd/runk/runk-tests.sh
```diff
@@ -0,0 +1,77 @@
...
+    echo "Install runk"
+    pushd ${RUNK_SRC_PATH}
+    make
+    sudo make install
+    popd
+}
+
+test_runk() {
+    echo "start container with runk"
+    sudo ctr run --pid-file ${PID_FILE} --rm -d  --runc-binary ${RUNK_BIN_PATH} ${TEST_IMAGE} ${CONTAINER_ID}
```
ManaSugi, line 52: Nit: please remove an extra space between `-d` and `--runc-binary`.

cyyzero:
Maybe it is caused by race condition,  since handling signal is asynced. @ManaSugi 
But how to avoid it? Sleep for a while and then check the status?

ManaSugi:
@cyyzero Yes, it seems just race condition, please add sleep for a while (maybe about 5 seconds)  with the comment and then check the status.

cyyzero:
@liubin Please approve and run tests

cyyzero:
`jenkins-ci-ARM-ubuntu-18-04` and `jenkins-metrics-ubuntu-20-04` failed before entering runk tests. Please rerun them. @ManaSugi @liubin

cyyzero:
@liubin @ManaSugi  It is able to pass all the tests now.  Do you have any other suggestions for runk tests?

ManaSugi (approved):

integration/containerd/runk/runk-tests.sh
```diff
@@ -0,0 +1,72 @@
...
+    [ ${STATUS} == "RUNNING" ] || die "contianer status is not RUNNING"
+
+    echo "exec process in a container"
+    sudo ctr t exec --exec-id id1 ${CONTAINER_ID} sh -c "echo hello > /tmp/foo"
+    [ "hello" == "$(sudo ctr t exec --exec-id id1 ${CONTAINER_ID} cat /tmp/foo)" ] || die "exec process failed"
+
+    echo "kill and delete container"
+    sudo ctr t kill --signal SIGKILL --all ${CONTAINER_ID}
+    # sleep for a while to make sure the task is killed
+    sleep 5s
```
liubin, line 57: `ctr` has a global option `--timeout`, that could prevent time wasting by `sleep 5s`

integration/containerd/runk/runk-tests.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+#
+# Copyright (c) 2021-2022 Kata Contributors
```
liubin, line 3: `2021-2022` -> `2022`

integration/containerd/runk/runk-tests.sh
cyyzero: `sudo ctr t kill --signal SIGKILL --all ${CONTAINER_ID}` to `sudo ctr --timeout 5s t kill --signal SIGKILL --all ${CONTAINER_ID}`?

GabyCT: any reason of not using https://github.com/kata-containers/tests/blob/main/lib/common.bash#L221 @cyyzero ?

cyyzero: @GabyCT Thanks for reminding me of this function! I will use it.

integration/containerd/runk/runk-tests.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo "Install runk"
+    pushd ${RUNK_SRC_PATH}
+    make
+    sudo make install
+    popd
+}
+
+test_runk() {
+    echo "start container with runk"
+    sudo ctr run --pid-file ${PID_FILE} --rm -d --runc-binary ${RUNK_BIN_PATH} ${TEST_IMAGE} ${CONTAINER_ID}
```
GabyCT, line 45: any reason why you are putting runc (--runc-binary) and not kata? or why not io.containerd.runc.v2?

integration/containerd/runk/runk-tests.sh
```diff
@@ -0,0 +1,72 @@
...
+CONTAINER_ID="id1"
+PID_FILE="${CONTAINER_ID}.pid"
+
+setup() {
+    # can't find cargo when make runk below, so we install rust to make runk build correctly
+    "${dir_path}/../../../.ci/install_rust.sh" && source "$HOME/.cargo/env"
+    echo "restart containerd service"
+    sudo systemctl restart containerd
+    echo "pull container image"
+    sudo ctr images pull ${TEST_IMAGE}
```
GabyCT, line 32: you should also use https://github.com/kata-containers/tests/blob/b77a49f6bcc4fab38a513fcf7cf9ee1e13c59d40/metrics/lib/common.bash#L67

integration/containerd/runk/runk-tests.sh
cyyzero: Just like runc, [runk](https://github.com/kata-containers/kata-containers/tree/main/src/tools/runk) is a standard OCI container runtime. It is in development and lacks testing. So I let containerd use runk to verify its functionality.

integration/containerd/runk/runk-tests.sh
cyyzero: Ok, I will use it .  But why is it under matrics directory? Maybe moving to global lib is better.

integration/containerd/runk/runk-tests.sh
cyyzero: Only SIGKILL or SIGSTOP can ensure stop the init process in a pid namespace.

According to [man 7 pid_namespaces](https://man7.org/linux/man-pages/man7/pid_namespaces.7.html), only SIGKILL or SIGSTOP can ensure stop the init process in a pid namespace.:

>        Likewise, a process in an ancestor namespace can—subject to the
>        usual permission checks described in [kill(2)](https://man7.org/linux/man-pages/man2/kill.2.html)—send signals to the
>        "init" process of a child PID namespace only if the "init"
>        process has established a handler for that signal.  (Within the
>        handler, the siginfo_t si_pid field described in [sigaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html)
>        will be zero.)  SIGKILL or SIGSTOP are treated exceptionally:
>        these signals are forcibly delivered when sent from an ancestor
>        PID namespace.  Neither of these signals can be caught by the
>        "init" process, and so will result in the usual actions
>        associated with those signals (respectively, terminating and
>        stopping the process).

`Ctr kill` send SIGTERM by default. There is a high probability that the container process( also as the init process in PID namespace) does not establish a handler for SIGTERM. So if `ctr t kill` command (which is a process in an ancestor PID namespace) send SIGTERM, container process will not receive.

In conclusion, here should use `ctr t kill --signal SIGKILL --all`. Is it ok to update clean_env and add these options for `ctr t kill`?  I'm not sure if this will affect kata's tests.

cyyzero:
@GabyCT Branch updated. Here must sends SIGKILL, so I doesn't use `clean_env_ctr`, but instead use `waitForProcess` directly to poll and wait for the container process receiving signal and exiting. 

cc @liubin @ManaSugi

cyyzero:
Some tests failed. The expression in `$cmd` to the right of the equal sign is executed and expanded in advance.  `$cmd` should be a raw string and I fixed it. Please run the test again.

cyyzero:
@GabyCT It can pass all tests now. Please review again.

cyyzero:
Rebase and add tests for `ps` command. @liubin

liubin (approved):

cyyzero:
Test fails on ubuntu 22.04 because `status::get_all_pid` only works on cgroup v1 now.  Temporarily skip tests on cgroup v2 einvitonment.  Please run test again.  @ManaSugi @liubin

liubin:
> Test fails on ubuntu 22.04 because `status::get_all_pid` only works on cgroup v1 now. Temporarily skip tests on cgroup v2 einvitonment. Please run test again. @ManaSugi @liubin

Thank you, @cyyzero.

I created an issue to track this, please see https://github.com/kata-containers/kata-containers/issues/4821.

PR #4961, opened 2022-07-22
ci: add sudo when compile runtime-rs and enable k8s-seccomp.bats

yqleng1987:
add sudo when compile runtime-rs and enable the case: k8s-seccomp.bats
because it's already fixed: https://github.com/kata-containers/kata-containers/pull/4692

Signed-off-by: yqleng <yqleng@linux.alibaba.com>


Tim-Zhang (approved):
LGTM

liubin (approved):

Issue #4972, opened 2022-07-26
Add a exec test case for terminated successfully     


lifupan:
When an exec forked some daemon process which would inherit
    its stdio fds, thus when the exec main process exited, its
    io stream wouldn't terminated, even at this case, we should
    make sure the exec terminated successfully without hang there.

PR #4973, opened 2022-07-26
test: Add a exec test case for terminated successfully

lifupan:
When an exec forked some daemon process which would inherit
its stdio fds, thus when the exec main process exited, its
io stream wouldn't terminated, even at this case, we should
make sure the exec terminated successfully without hang there.

Fixes: #4972

### DIFF
--- a/integration/kubernetes/k8s-exec.bats
+++ b/integration/kubernetes/k8s-exec.bats
@@ -30,6 +30,10 @@ setup() {
 	# kubectl exec -it "$pod_name" -- ls -tl /
 	kubectl exec "$pod_name" -- date
 
+	# Test if the exec cmd exit successfully when the exec's stdout/stderr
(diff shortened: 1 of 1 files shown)


liubin (approved):

fidencio (approved):
lgtm, thanjs @lifupan!

wainersm:
hi @lifupan , would make sense to exec on the container afterwards to also check the `tail` process is still alive?

liubin:
@wainersm give us a good advice, so I will add the DNM label.

fidencio:
As we're in the process (a still long process) of moving tests from this repo to the kata-containers one, we're closing PRs that are either stale for too long.

Please, if you think this has been mistakenly closed, open it agains the `kata-containers` repo as in the coming months this repo will be archived.

PR #4977, opened 2022-07-26
Amd cc sev

ryansavino:


functional/sev/run.sh
```diff
@@ -0,0 +1,441 @@
...
+  sudo -E PATH=$PATH "$@"
+}
+
+build_and_install_kata_runtime_4270() {
+  cd ${WORKING_DIR}
+
+  # Clone kata-containers repo and checkout PR 4270
+  git clone https://github.com/kata-containers/kata-containers.git
+  cd kata-containers
+  gh pr checkout 4270
```
c3d, line 45: Is the intent to replace this with the actual PR being tested?

ryansavino: This checkout of PR 4270 is temporary until it gets merged. It's close to being merged, but until then, wanted to get CI running.

fidencio:
I'm adding the wip label till 4270 gets merged.

functional/sev/run.sh
fidencio: @ryansavino, `gh pr checkout` implies that we have https://github.com/cli/cli installed in machine, which may not be the case.

A simple `git checkout` would be a better way to avoid the dependency on the tool.

fidencio (dismissed):
@ryansavino, in general I'm not exactly following why you don't use the machinery we already have as part of the CI, instead of creating your own scripts to build every single piece of the stack.

We're working on having the CI testing exactly the same components that we ship, and by having your own way to build those components we're actually going on the opposite direction of what we want to achieve.

wainersm:
> @ryansavino, in general I'm not exactly following why you don't use the machinery we already have as part of the CI, instead of creating your own scripts to build every single piece of the stack.
> 

@fidencio I suggested @ryansavino that we could start with the scripts he had although they aren't using the CI machinery and once we had a clean execution then we could do the refactoring. And now we reached that point where the code can be adapted to use the CI stuffs and we will have means to test it.

fidencio:
> > @ryansavino, in general I'm not exactly following why you don't use the machinery we already have as part of the CI, instead of creating your own scripts to build every single piece of the stack.
> 
> @fidencio I suggested @ryansavino that we could start with the scripts he had although they aren't using the CI machinery and once we had a clean execution then we could do the refactoring. And now we reached that point where the code can be adapted to use the CI stuffs and we will have means to test it.

Okay, but then we keep this as a WIP and avoid merging it as it's, and refactor it before getting it in.

ryansavino:
@fidencio to add to what @wainersm said, I'm in the process of integrating those pieces into their respective places in the existing ci. Once #4270 is merged, it will make this possible. Just trying to get a little ahead of that by having some working ci for it.

functional/sev/run.sh
```diff
@@ -0,0 +1,444 @@
...
+
+  # Build and install the CC Kata Containers runtime
+  export GO111MODULE=on
+  cd ${WORKING_DIR}/kata-containers/src/runtime
+  make && esudo make install
+}
+
+build_rootfs() {
+  cd ${WORKING_DIR}/kata-containers/tools/osbuilder/rootfs-builder
+  esudo GOPATH=$GOPATH AGENT_INIT=yes USE_DOCKER=true DOCKER_RUNTIME=runc AA_KBC=offline_sev_kbc SKOPEO_UMOCI=yes ./rootfs.sh ubuntu
```
wainersm, line 55: At the point this script is executed, the whole kata-containers was built and installed. So maybe we should not rebuild the image here....

The Jenkins job will export `CI_JOB=CC_SEV_CRI_CONTAINERD_K8S` and the variables on the block https://github.com/kata-containers/tests/blob/ab84d51927209ab9df8f144b63fbaaec8f8ef6c8/.ci/ci_job_flags.sh#L111 are all exported afterwards. The list of variables are:
```
KATA_BUILD_CC="yes"
CCV0="yes"
UMOCI="yes"
AA_KBC="offline_fs_kbc"
KUBERNETES=yes
TEE_TYPE="sev"
```
Ok, the value of `AA_KBC` is currently `offline_fs_kbc` but in this script it is `offline_sev_kbc`. We probably should adjust this. Something like that:
```
vagrant@ubuntu2004:~/go/src/github.com/kata-containers/tests$ git diff
diff --git a/.ci/ci_job_flags.sh b/.ci/ci_job_flags.sh
index 946704ed..be9c11e9 100755
--- a/.ci/ci_job_flags.sh
+++ b/.ci/ci_job_flags.sh
@@ -117,7 +117,11 @@ case "${CI_JOB}" in
                                export SKOPEO=yes
                        fi
                        if [[ "${CI_JOB}" =~ K8S ]]; then
-                               export AA_KBC="offline_fs_kbc"
+                               if [[ "${CI_JOB}" =~ SEV ]]; then
+                                       export AA_KBC="offline_sev_kbc"
+                               else
+                                       export AA_KBC="offline_fs_kbc"
+                               fi
                                export KUBERNETES=yes
                        fi
                        if [ "${CI_JOB}" == "CC_SEV_CRI_CONTAINERD_K8S" ]; then

```
For sev case, do we need skopeo in the guest image? I thought the plan was to release with image-rs support instead. I am asking to assess if `SKOPEO_UMOCI=yes` should really be exported. Actually it seems `SKOPEO_UMOCI` isn't even used by the build scripts.

functional/sev/run.sh
```diff
@@ -0,0 +1,444 @@
...
+
+esudo tee rootfs-ubuntu/etc/default/agent.cfg <<EOF
```
wainersm, line 57: I understand that this script is rebuilding the guest image just to have this `agent.cfg` file inside. We have a function (`cp_to_guest_img()`) which is used by the tests to copy files to the guest image so that this kind of rebuild is not needed.  This is the function https://github.com/kata-containers/tests/blob/CCv0/lib/common.bash#L255

So you can do the following:

1) place this file somewhere, for example, `functional/sev/fixtures/agent.cfg`
2) source the library file at the begin of this script (`${tests_repo_dir}/lib/common.bash`)
3) Before you execute the sev tests, use `cp_to_guest_img()` to copy the file to the guest image.

See how we do it in a test case for the agent: https://github.com/kata-containers/tests/blob/ab84d51927209ab9df8f144b63fbaaec8f8ef6c8/integration/containerd/confidential/agent_api.bats#L49

functional/sev/run.sh
```diff
@@ -0,0 +1,444 @@
...
+  #esudo ./build-kernel.sh -x sev install
+
+  # Per above, run this to install
+  bzImage="kata-linux-efi-secret-v5.17-rc6-93/arch/x86_64/boot/bzImage"
+  vmlinuz="/usr/share/kata-containers/vmlinuz-efi-secret-v5.17-rc6-93-sev"
+  esudo install -o root -g root -m 0644 -D "${bzImage}" "${vmlinuz}"
+  esudo ln -sf "${vmlinuz}" "/usr/share/kata-containers/vmlinuz-sev.container"
+}
+
+build_initrd() {
```
wainersm, line 125: As I mentioned in my comments to `build_rootfs()`, we should be able to remove this rebuild of the guest image. There is only one thing I'd like to check, here you export `AGENT_INIT=yes` and I guess it is the default already.

@stevenhorsman the guest image for CC is already built with `AGENT_INIT=yes`, doesn't it?

stevenhorsman: In CC at the moment we just use the default, which is `AGENT_INIT=no`, so I guess that would need to go in the ci_job_flags parameters for this job?

wainersm: @ryansavino do we really need the agent as init process? If so, then we will have to build a guest image specifically to SEV

ryansavino: SEV requires initrd image. I have fixes for initrd with systemd in https://github.com/kata-containers/kata-containers/pull/4987. Need this PR to get in to proceed with testing. I also have changes in `.ci/install_kata_image.sh` to accommodate as well.

functional/sev/run.sh
ryansavino: https://github.com/kata-containers/kata-containers/pull/4270 is now merged. This part removed.

functional/sev/run.sh
```diff
@@ -0,0 +1,444 @@
...
+trap cleanup EXIT
+
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+NC='\033[0m'
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+tests_repo_dir="$(realpath "$script_dir/../..")"
+
+export WORKING_DIR="${HOME}/work"
```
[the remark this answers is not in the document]
ryansavino: Removed references to WORKING_DIR variable. No need to clone kata repo anymore as 4270 is now merged. Removed other unneeded environment variable exports.

functional/sev/run.sh
ryansavino: SEV - don't need SKOPEO. Adjusted `ci_job_flags.sh` to export additional SEV params for `KATA_BUILD_KERNEL_TYPE`, `AA_KBC`, and `TEST_INITRD`.

functional/sev/run.sh
ryansavino: Completed. I added a separate method in the `run.sh` to copy the agent.cfg. The one in common.bash is meant to work with dracut filesystem and the image isn't gunzipped. It uses `cat` during the export instead of `zcat` needed to export the gunzipped unbuntu initrd image.

ryansavino:
Add SEV support PR (https://github.com/kata-containers/kata-containers/pull/4270) is needed - Merged
Kernel upgrade to 5.19 PR (https://github.com/kata-containers/kata-containers/pull/4861) is needed - Merged
Initrd fixes for ubuntu systemd (https://github.com/kata-containers/kata-containers/pull/4987) is needed - Pending Review

functional/sev/run.sh
```diff
@@ -0,0 +1,352 @@
...
+  mout=$(esudo sev-snp-measure -v --mode=sev --output-format=base64 \
+    --ovmf="${EDK2_REPO_DIR}/Build/AmdSev/DEBUG_GCC5/FV/OVMF.fd" \
+    --kernel="/usr/share/kata-containers/vmlinuz-sev.container" \
+    --initrd="/usr/share/kata-containers/kata-containers-sev-initrd.img" \
+    --append="tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k cryptomgr.notests net.ifnames=0 pci=lastbus=0 console=hvc0 console=hvc1 debug panic=1 nr_cpus=1 scsi_mod.scan=none agent.log=debug agent.log=debug initcall_debug agent.config_file=/etc/default/agent.cfg console=ttyAMA0,115200 console=tty highres=off console=ttyS0" \
+  )
+
+  # Trim text and error if value is empty
+  measurement=$(echo ${mout} | sed "s/^\(Calculated SEV guest measurement: \)\(.*\)$/\2/g")
+  if [[ -z "${measurement}" ]]; then >&2 echo "Measurement is invalid"; return 1; fi
```
dubek, line 199: Suggestion: remove the `-v` in line 190, and then you don't need the `sed` in line 198 (without `-v` it'll print just the base64 value).

Also, generally, there's no need for `sudo` when running sev-snp-measure, unless it is need in order to read one of the files (ovmf, kernel, initrd).

functional/sev/run.sh
```diff
@@ -0,0 +1,352 @@
...
+
+  # Add key, keyset and policy with measurement to DB
+  mysql -u${KBS_DB_USER} -p${KBS_DB_PW} -h ${KBS_DB_HOST} -D ${KBS_DB} -e "INSERT INTO secrets VALUES (10, 'key_id1', 'RcHGava52DPvj1uoIk/NVDYlwxi0A6yyIZ8ilhEX3X4=', 10);"
+  mysql -u${KBS_DB_USER} -p${KBS_DB_PW} -h ${KBS_DB_HOST} -D ${KBS_DB} -e "INSERT INTO keysets VALUES (10, 'KEYSET-1', '[\"key_id1\"]', 10);"
+  mysql -u${KBS_DB_USER} -p${KBS_DB_PW} -h ${KBS_DB_HOST} -D ${KBS_DB} -e "INSERT INTO policy VALUES (10, '[\"${measurement}\"]', '[]', 0, 0, '[]', now(), NULL, 1);"
```
dubek, line 204: I think you can do something like this:

```
mysql -u${KBS_DB_USER} -p${KBS_DB_PW} -h ${KBS_DB_HOST} -D ${KBS_DB} <<EOF
  INSERT INTO secrets VALUES (10, 'key_id1', 'RcHGava52DPvj1uoIk/NVDYlwxi0A6yyIZ8ilhEX3X4=', 10);
  INSERT INTO keysets VALUES (10, 'KEYSET-1', '["key_id1"]', 10);
  INSERT INTO policy VALUES (10, '["${measurement}"]', '[]', 0, 0, '[]', now(), NULL, 1);
EOF
```

It might be a bit more readable, and no need to escape `"` chars.

dubek (commented):
few comments (mostly style):

functional/sev/run.sh
```diff
@@ -0,0 +1,352 @@
...
+  RUST_LOG=debug cargo run &
+  sleep 5
+  popd
+}
+
+install_sev_snp_measure() {
+  # Clone and link
+  git clone https://github.com/IBM/sev-snp-measure.git
+  esudo ln -sf "sev-snp-measure/sev-snp-measure.py" "/usr/bin/sev-snp-measure"
+}
```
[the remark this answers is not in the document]
ryansavino: done

functional/sev/run.sh
ryansavino: done

functional/sev/run.sh
ryansavino: done

ryansavino:
I still need to setup the encrypted image on quay.io to avoid running the encrypt and docker push commands.

ryansavino:
Need https://github.com/kata-containers/kata-containers/pull/5010 merged for this.

ryansavino:
Need https://github.com/kata-containers/kata-containers/pull/5013 merged for this.

functional/sev/fixtures/agent.cfg
```diff
@@ -0,0 +1,44 @@
...
+[endpoints]
+allowed = [
+"AddARPNeighborsRequest",
+"AddSwapRequest",
+"CloseStdinRequest",
+"CopyFileRequest",
+"CreateContainerRequest",
+"CreateSandboxRequest",
+"DestroySandboxRequest",
+"ExecProcessRequest",
```
fitzthum, line 16: Do we want to comment this out?

functional/sev/fixtures/agent.cfg
```diff
@@ -0,0 +1,44 @@
...
+"GuestDetailsRequest",
+"ListInterfacesRequest",
+"ListRoutesRequest",
+"MemHotplugByProbeRequest",
+"OnlineCPUMemRequest",
+"PauseContainerRequest",
+"PullImageRequest",
+"ReadStreamRequest",
+"RemoveContainerRequest",
+"ReseedRandomDevRequest",
```
fitzthum, line 28: Also this one?

functional/sev/run.sh
```diff
@@ -0,0 +1,337 @@
...
+main() {
+  mkdir -p test
+  source ~/.profile
+
+  # Verify docker authentication
+  if ! esudo cat /root/.docker/config.json | grep auths >/dev/null; then >&2 echo "ERROR: Not logged into docker"; exit 1; fi
+  esudo docker login
+
+  # kata config and kernel params
+  modify_kata_config
```
fitzthum, line 311: Will we still need this once #5013 is merged?

functional/sev/run.sh
```diff
@@ -0,0 +1,337 @@
...
+
+initrd_add_files() {
+  rootfs_dir="$(mktemp -d)"
+  initrd_path="$(kata-runtime kata-env --json | jq -r .Initrd.Path)"
+  
+  # Extract gzip initrd into temp directory
+  zcat "${initrd_path}" | cpio --extract --preserve-modification-time --make-directories --directory="${rootfs_dir}"
+  
+  # Copy agent.cfg to initrd
+  esudo cp "functional/sev/fixtures/agent.cfg" "${rootfs_dir}/etc/default"
```
fitzthum, line 71: Maybe we should parameterize the agent config path since we also have to add it to the kerne params? Also, is this path in `/etc/default` consistent with the TDX tests?

functional/sev/run.sh
```diff
@@ -0,0 +1,337 @@
...
+  esudo sed -i -e "s|^\(firmware = \"\).*\"|\1${EDK2_REPO_DIR}/Build/AmdSev/DEBUG_GCC5/FV/OVMF.fd\"|g" /etc/kata-containers/configuration.toml
+  
+  # kernel_params
+  esudo sed -i -e 's/kernel_params = "\(.*\)"/kernel_params = " agent.log=debug initcall_debug agent.config_file=\/etc\/default\/agent.cfg console=ttyAMA0,115200 console=tty highres=off console=ttyS0"/g' /etc/kata-containers/configuration.toml
+
+  # confidential_guest, guest_attestation
+  esudo sed -i -e 's/# confidential_guest = true/confidential_guest = true/g' /etc/kata-containers/configuration.toml
+  esudo sed -i -e 's/#guest_pre_attestation = true/guest_pre_attestation = true/g' /etc/kata-containers/configuration.toml
+  esudo sed -i -e 's/#guest_pre_attestation_proxy="localhost:50051"/guest_pre_attestation_proxy = "localhost:44444"/g' /etc/kata-containers/configuration.toml
+  esudo sed -i -e 's/#guest_pre_attestation_keyset="KEYSET-1"/guest_pre_attestation_keyset = "KEYSET-1"/g' /etc/kata-containers/configuration.toml
```
fitzthum, line 53: At some point we should test out setting this via annotation rather than in the config file.

fitzthum (approved):
This looks good. There are still some moving parts, but I think we are on the right track.

.ci/install_kata_kernel.sh
```diff
@@ -55,6 +55,12 @@ build_and_install_kernel() {
 	"${kernel_packaging_dir}/build-kernel.sh" -v "${kernel_version}" ${extra_opts} "setup"
 	"${kernel_packaging_dir}/build-kernel.sh" -v "${kernel_version}" ${extra_opts} "build"
 	sudo -E PATH="$PATH" "${kernel_packaging_dir}/build-kernel.sh" -v "${kernel_version}" ${extra_opts} "install"
+
+	### REMOVE after kata-containers PR #5010 is merged
```
fitzthum, line 59: 5010 is closed so let's update this comment.

.ci/install_kata_kernel.sh
```diff
@@ -55,6 +55,12 @@ build_and_install_kernel() {
...
+	if [ "${kernel_type}" == "sev" ]; then
+		sudo ln -sf "vmlinux-sev.container" "/usr/share/kata-containers/vmlinux.container"
```
fitzthum, line 61: somewhat surprised to see the uncompressed kernel here.

ryansavino: Not actually using it. Temporary hacky workaround for https://github.com/kata-containers/kata-containers/issues/5008.

.ci/install_kata_kernel.sh
ryansavino: Needed to change this to 5013. Updated the code comment.

functional/sev/run.sh
ryansavino: Nope. I'll remove it when 5013 is in.

functional/sev/run.sh
```diff
@@ -0,0 +1,337 @@
...
+  esudo sed -i -e 's/^# *\(enable_debug\).*=.*$/\1 = true/g' /etc/kata-containers/configuration.toml
+  
+  # kernel, initrd, image, firmware
+  esudo sed -i -e 's/kernel = "\(.*\)"/kernel = "\/usr\/share\/kata-containers\/vmlinuz-sev.container"/g' /etc/kata-containers/configuration.toml
+  esudo sed -i -e 's/initrd = "\(.*\)"/initrd = "\/usr\/share\/kata-containers\/kata-containers-initrd.img"/g' /etc/kata-containers/configuration.toml
+  esudo sed -i -e 's/image = "\(.*\)"/#image = ""\ninitrd = "\/usr\/share\/kata-containers\/kata-containers-initrd.img"/g' /etc/kata-containers/configuration.toml
+  esudo sed -i -e "s|^\(firmware = \"\).*\"|\1${EDK2_REPO_DIR}/Build/AmdSev/DEBUG_GCC5/FV/OVMF.fd\"|g" /etc/kata-containers/configuration.toml
+  
+  # kernel_params
+  esudo sed -i -e 's/kernel_params = "\(.*\)"/kernel_params = " agent.log=debug initcall_debug agent.config_file=\/etc\/default\/agent.cfg console=ttyAMA0,115200 console=tty highres=off console=ttyS0"/g' /etc/kata-containers/configuration.toml
```
fitzthum, line 47: I'm not sure we need all the extra console stuff anymore.

ryansavino: This will get removed with 5013. I'll also remove them from the kernel params in the `sev-snp-measure` call when 5013 is in.

functional/sev/run.sh
ryansavino: Annotation meaning the `Makefile` variables for the sev config template?

functional/sev/run.sh
ryansavino: @fidencio are you guys putting the agent.cfg in a different place. I was looking through the codebase and was having trouble finding where you guys are doing this.

functional/sev/fixtures/agent.cfg
ryansavino: done

functional/sev/fixtures/agent.cfg
ryansavino: done

functional/sev/run.sh
fitzthum: I think these options can be set via a Kubernetes annotation, which for some fields (like `keyset`) is how a user will probably set them. Keyset and proxy should be changed at the scope of the pod rather than the scope of the node. So at some point we might want to adjust the testing to account for this.

ryansavino:
Okay, I think this one is pretty much good to go pending these PR merges:
https://github.com/kata-containers/kata-containers/pull/4987
https://github.com/kata-containers/kata-containers/pull/5013

Jenkins job is passing:
http://jenkins.katacontainers.io/job/wainer-CCv0-ubuntu-20.04-x86_64-CC_SEV_CRI_CONTAINERD_K8S/89/console

I've removed attestation agent image creation and encryption, and I'm pulling the encryption key and ssh key from the quay.io docker image labels. I have a separate script for creating the encrypted image that I can share/add somewhere if desired.

Let me know if anyone wants me to change/fix something.

functional/sev/run.sh
fidencio: This part should be done via the configuration file, for values that we know to be the default ones, in the qemu-sev.conf

fitzthum:
So it seems like we are building the components via the test script for the first release rather than using https://github.com/kata-containers/tests/pull/5048 and the `kata-cc` bundle. 

Is that correct @ryansavino @wainersm

wainersm:
> So it seems like we are building the components via the test script for the first release rather than using #5048 and the `kata-cc` bundle.
> 
> Is that correct @ryansavino @wainersm

That's correct @fitzthum . But notice #5048 was reverted in #5102 because it breaks the TDX CI; it turned out that the kata-deploy build scripts don't handle all the cases were we need to use a proxy (the TDX CI is behind a firewall).

.ci/install_kata_kernel.sh
wainersm: @ryansavino  now that PR 5013 is merged can we remove this?

.ci/install_runtime.sh
```diff
@@ -118,7 +118,7 @@ case "${KATA_HYPERVISOR}" in
 	"qemu")
 		if [ "$TEE_TYPE" == "tdx" ]; then
 			enable_hypervisor_config "${PKGDEFAULTSDIR}/configuration-qemu-tdx.toml"
-		else
+		else ### Add SEV after kata-containers 5013 merged
```
wainersm: @ryansavino likewise, can we add the case for sev now?

functional/sev/run.sh
```diff
@@ -66,6 +66,11 @@ initrd_add_files() {
   # Copy agent-config.toml to initrd
   esudo cp "functional/sev/fixtures/agent-config.toml" "${rootfs_dir}/etc"
 
+  # REMOVE after kata-containers #4987 merged - getty hang issue fix and kata agent systemd symlinks
```
wainersm: Likewise, https://github.com/kata-containers/tests/issues/4987 is merged.

.ci/install_kata_kernel.sh
```diff
@@ -56,11 +56,6 @@ build_and_install_kernel() {
 	"${kernel_packaging_dir}/build-kernel.sh" -v "${kernel_version}" ${extra_opts} "build"
 	sudo -E PATH="$PATH" "${kernel_packaging_dir}/build-kernel.sh" -v "${kernel_version}" ${extra_opts} "install"
 
-	### REMOVE after kata-containers PR #5013 is merged
```
wainersm: ah, ok, will be squashed.

functional/sev/run.sh
wainersm: @fitzthum all the guest_pre_attestation_* and sev_cert_chain can be passed via pod annotation?

wainersm:
Hi @ryansavino 

> Okay, I think this one is pretty much good to go pending these PR merges: [kata-containers/kata-containers#4987](https://github.com/kata-containers/kata-containers/pull/4987) [kata-containers/kata-containers#5013](https://github.com/kata-containers/kata-containers/pull/5013)
> 
> Jenkins job is passing: http://jenkins.katacontainers.io/job/wainer-CCv0-ubuntu-20.04-x86_64-CC_SEV_CRI_CONTAINERD_K8S/89/console
> 
> I've removed attestation agent image creation and encryption, and I'm pulling the encryption key and ssh key from the quay.io docker image labels. I have a separate script for creating the encrypted image that I can share/add somewhere if desired.
> 
> Let me know if anyone wants me to change/fix something.

IMO it is ready to be merged. Any improvements can be made upon this base which seems solid enough.

hey, thanks for fighting to get it done!

wainersm (approved):

ryansavino:
Thanks @wainersm. Unfortunately, there's a new bug that's causing the SEV kernel build to fail:
https://github.com/kata-containers/kata-containers/pull/5095

It needs to get merged from `main` to `CCv0` as well.

I want to get successful jenkins build before merging. I'll update when everything is working. Hopefully Tues.

wainersm:
@ryansavino ready to be merged?

ryansavino:
> @ryansavino ready to be merged?

Jenkins job is passing: http://jenkins.katacontainers.io/job/wainer-CCv0-ubuntu-20.04-x86_64-CC_SEV_CRI_CONTAINERD_K8S/98

However, if we can get https://github.com/kata-containers/kata-containers/pull/5024 merged, I can remove the agent config file from this PR. @wainersm do you want to merge it now anyways? I can always clean it up later.

ryansavino:
Yep

PR #4978, opened 2022-07-26
agent-ctl: Install agent-ctl in install_runtime.sh

### LINKED ISSUE #4602: Install agent-ctl on CI
Currently `agent-ctl` is not installed on CI environments. However, the tool can be used on tests to determine conditions, for example, whether the agent support seccomp before running a seccomp test (see https://github.com/kata-containers/tests/issues/4303). With the emerging confidential containers specific features (see https://github.com/kata-containers/ci/pull/455#issuecomment-1072413330), it is likely more tests will be relying on `agent-ctl`. Therefore, let's build and install it on CI.

bookinabox:
Installs agent-ctl in install_runtime.sh which allows better debugging
in the ci environment.

Fixes #4602

Signed-off-by: Derek Lee <derlee@redhat.com>

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -14,7 +14,6 @@ cidir=$(dirname "$0")
 
 source "${cidir}/lib.sh"
 source /etc/os-release || source /usr/lib/os-release
-KATACONTAINERS_REPO=${katacontainers_repo:="github.com/kata-containers/kata-containers"}
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 KATA_EXPERIMENTAL_FEATURES="${KATA_EXPERIMENTAL_FEATURES:-}"
 MACHINETYPE="${MACHINETYPE:-q35}"
@@ -45,16 +44,16 @@ export SHAREDIR=${PREFIX}/share
 OPENSHIFT_CI="${OPENSHIFT_CI:-false}"
 
 runtime_config_path="${SYSCONFDIR}/kata-containers/configuration.toml"
-runtime_src_path="${GOPATH}/src/${KATACONTAINERS_REPO}/src/runtime"
+runtime_src_path="${katacontainers_repo_dir}/src/runtime"
+agent_ctl_path="${katacontainers_repo_dir}/src/tools/agent-ctl"
 
 PKGDEFAULTSDIR="${DESTDIR}${SHAREDIR}/defaults/kata-containers"
 NEW_RUNTIME_CONFIG="${PKGDEFAULTSDIR}/configuration.toml"
 # Note: This will also install the config file.
 
+clone_katacontainers_repo
+
 build_install_shim_v2(){
-	if [ ! -d "$runtime_src_path" ]; then
-		go get "$KATACONTAINERS_REPO"
-	fi
 	pushd "$runtime_src_path"
 	make
 	sudo -E PATH=$PATH make install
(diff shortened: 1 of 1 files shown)


.ci/install_runtime.sh
```diff
@@ -63,6 +66,19 @@ build_install_shim_v2(){
 
 build_install_shim_v2
 
+build_install_agent_ctl(){
+	if [ ! -d "$agent_ctl_path" ]; then
```
wainersm, line 70: You should use the `clone_katacontainers_repo()`  method from `.ci/lib.sh` (already sourced). Then use the already exported `katacontainers_repo_dir` variable to reference the path to the kata-containers repo.

.ci/install_runtime.sh
```diff
@@ -12,6 +12,8 @@ set -o errtrace
 
 cidir=$(dirname "$0")
 
+bash "${cidir}/install_rust.sh" && source "$HOME/.cargo/env"
```
wainersm, line 15: I would call the rust install script from `build_install_agent_ctl()`.  Here it is loosely in between variables declarations...

.ci/install_runtime.sh
```diff
@@ -46,6 +48,7 @@ OPENSHIFT_CI="${OPENSHIFT_CI:-false}"
 
 runtime_config_path="${SYSCONFDIR}/kata-containers/configuration.toml"
 runtime_src_path="${GOPATH}/src/${KATACONTAINERS_REPO}/src/runtime"
+agent_ctl_path="${GOPATH}/src/${KATACONTAINERS_REPO}/src/tools/agent-ctl"
```
wainersm, line 51: use `katacontainers_repo_dir` variable. The `runtime_src_path` and other variables on this file are doing wrong.

wainersm:
Thanks for working on this @bookinabox ! I left just a few comments.

wainersm:
All the other CI jobs except for jenkins-ci-ppc64le-ubuntu-20-04 are passing, which is good. I remember @fidencio commenting somewhere that job was unstable recently. Let me re-trigger it.

wainersm (approved):

.ci/install_runtime.sh
```diff
@@ -63,6 +62,17 @@ build_install_shim_v2(){
 
 build_install_shim_v2
 
+build_install_agent_ctl(){
+	bash "${cidir}/install_rust.sh" && source "$HOME/.cargo/env"
```
fidencio, line 66: AFAIR we're not building any rust component on the host side, but rather building them inside a container.
Would be possible to do the same here?

bookinabox: I'm not super familiar with the project still, and I'm a bit confused. the agent-ctl runs on the host side, right?

fidencio: @bookinabox, oops, my bad, let me take a step back here.

It runs on the host side, but what I was afraid of was this being the only reason we end up installing rust in the CI machines, just to build the `agent-ctl` tool.  If that was the case, I'd suggest we rather build it inside a container (which would have rust installed there) and just install the binary in the correct location.

However, it seems we're already installing rust on the Ci machine as we build `virtiofsd` there (for non-x86_64 machines), thus I think this one is fine to be merged as it's.

Does the explanation make sense?

fidencio (approved):
lgtm, thanks @bookinabox!

.ci/install_runtime.sh
bookinabox: yes, thanks!

PR #4996, opened 2022-08-04
common: Improve ctr cleanup

### LINKED ISSUE #4995: improve ctr cleanup function

GabyCT:
This PR improves the ctr cleanup function as it was seen than when we
are using RunC the current cleanup is not able to delete properly
the containers as we got an error saying that we can't delet
non stopped containers. This PR fixes that issue as it was tested
both on RunC and Kata and is working properly.

Fixes #4995

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -234,6 +234,7 @@ clean_env_ctr()
 	for i in "${containers[@]}"; do
 		task="$(sudo ctr task ls | grep -E "\<${i}\>" | cut -f1 -d" " || true)"
 		[ -n "${task}" ] && sudo ctr task kill "${task}"
+		[ -n "${task}" ] && sudo ctr tasks kill -s SIGKILL "${task}"
 	done
 
 	# do not stop if the command fails, it will be evaluated by waitForProcess


fidencio (approved):
lgtm, thanks @GabyCT!

lib/common.bash
```diff
@@ -234,6 +234,7 @@ clean_env_ctr()
 	for i in "${containers[@]}"; do
 		task="$(sudo ctr task ls | grep -E "\<${i}\>" | cut -f1 -d" " || true)"
 		[ -n "${task}" ] && sudo ctr task kill "${task}"
+		sudo ctr tasks kill -s SIGKILL "${task}"
```
liubin, line 237: Should add a pre-check `[ -n ${taks} ] `like the line above?

GabyCT: @liubin changes applied, thanks

lib/common.bash
```diff
@@ -234,6 +234,7 @@ clean_env_ctr()
 	for i in "${containers[@]}"; do
 		task="$(sudo ctr task ls | grep -E "\<${i}\>" | cut -f1 -d" " || true)"
 		[ -n "${task}" ] && sudo ctr task kill "${task}"
+		[ -n "${task}" ] && sudo ctr tasks kill -s SIGKILL "${task}"
```
cmaf, line 237: @GabyCT, we have both lines to first try to cleanly kill it, and then force?

GabyCT: @cmaf yes it seems that we will need to do that as even if we kill them some containers are left and we will need to force to kill them, I saw some issues of this in the containerd repo

cmaf (approved):

Issue #5022, opened 2022-08-12
fw/arm64: UEFI image installation slow down the Arm CI test


jongwu:
It takes at least 5min that build UEFI from source. We can save the UEFI image as backup in the first build and copy it in the following installation.

PR #5023, opened 2022-08-12
fw/arm64: speed up installation of UEFI image

jongwu:
Building UEFI image from source take at least 5 min. We can build it
once and save it as backup then copy the UEFI image there to cancel the
build in the following UEFI image installation.

Fixes: #5022


fidencio:
@jongwu, another possible approach would be to cache the UEFI as we do for the kernel / QEMU, and download the cached binary as part of the CI.

As the ARM CI is running on baremetal machines, your approach will do, but if we ever happen to have ARM VMs running, having it "downloadable" would be better.

jongwu:
> @jongwu, another possible approach would be to cache the UEFI as we do for the kernel / QEMU, and download the cached binary as part of the CI.
> 
> As the ARM CI is running on baremetal machines, your approach will do, but if we ever happen to have ARM VMs running, having it "downloadable" would be better.

Yes, cache UEFI at somewhere of net site is better. But I have nowhere to place it.

jongwu:
Help TAL @fidencio @GabyCT. Thanks!

GabyCT (approved):

fidencio (approved):
lgtm, thanks @jongwu!

PR #5032, opened 2022-08-16
fc: fix uncompressed dir path

bergwolf:
In the latest FC releases, the asset we download and decompress
has the following syntax:

release-vX.X.X-ARCH/firecracker-vX.X.X-ARCH

until v0.25 there was no ARCH in the folder name. Since we are updating
the supported FC version, this patch fixes the script that installs the
firecracker binary and the jailer.

Signed-off-by: Anastassios Nanos <ananos@nubificus.co.uk>


fidencio (approved):
lgtm, thanks @bergwolf!

GabyCT (approved):

Issue #5034, opened 2022-08-16
move nydus tests after simpler stability checks


bergwolf:
Nydus test is more complex and relies on third-party components. Let's move it to after simpler stability checks so that it makes life easier to diagnose failures.

PR #5035, opened 2022-08-16
ci: move nydus test to after stability

bergwolf:
Nydus test is more complext and relies on third party components.
Let's move it to after simpler stability checks so that it makes life
easier to diagnose failures.

Fixes: #5034


fidencio (approved):
lgtm, thanks @bergwolf!

GabyCT (approved):

Issue #5028, opened 2022-08-15
CC | Add a CI for testing the CCv0 branch with devmapper


fidencio:
We should make sure that we face no regressions when using the CCv0 content with the devmapper snapshotter.

Right now we're only testing with shared file systems (VirtioFS, to be more explicit).

PR #5046, opened 2022-08-19
CC | Add flags for testing the CCv0 branch using devmapper

fidencio:
Right now we've been testing the CCv0 branch using VirtioFS.  However, as devmapper is most likely the option to be used with Confidential Containers, and as I hit issues with it already, let's ensure we have jobs testing our work using the device mapper snapshotter as well.

Fixes: #5028


fidencio:
/test

This is to ensure that I didn't break any of the existing tests.

fidencio:
/fidencio-test-devmapper

Triggering this as I've added some **test** jobs for devmapper.  I'm expecting those to break as devmapper is **not** working on my environment with `service_offload = true`.  I have a patch for this, which I will submit shortly after having the CI working.

Issue #5071, opened 2022-08-26
Skip test from kata/cri-o CI that use the create_runtime_with_allowed_annotation helper function


littlejawa:
The failure on https://github.com/cri-o/cri-o/pull/6123 showed we still have an issue with one test, and need to skip it until we solve it.

Root cause is described here: https://github.com/cri-o/cri-o/issues/6175.

PR #5072, opened 2022-08-26
Skip additional test from shm_size.bats

littlejawa:
This test was passing, but the test actually checks for failure,
and we were failing for the wrong reasons.

Also adding links to all tests related to the same root cause.

Fixes: #5071


fidencio (approved):
lgtm, thanks @littlejawa!

GabyCT (approved):

PR #5076, opened 2022-08-26
metrics: Remove pvc in cassandra metrics test

### LINKED ISSUE #5075: metrics: delete PVC in cassandra test

GabyCT:
This PR adds the removal for PVC in the Cassandra metrics test, this
will avoid the issue that when we are removing the PVC it gets stuck
in "terminating" status for a while and will allow the final
unmount from the node so that PVC can be properly delete it.

FIxes #5075


dborquez (approved):
lgtm, thank yoiu @GabyCT

Issue #5079, opened 2022-08-26
CC | Adapt cc-kubernetes tests to work on TEE based CIs


fidencio:
Both `agent_image.bats` and `agent_image_encrypted.bats` tests start by cleaning up the kernel parameters used in the configuration file, which is a big "no-no" for the TEEs configurations.

TDX, at least, requires `disable_tdx_filter` to be passed to as a kernel parameter, otherwise the VM won't even boot.

I'm not sure about SEV / SEV-SNP case, but I'd guess they also need at least something passed.

This is the very first issue that I hit.  There may be more, but let me report this one to start tracking those ...

PR #5080, opened 2022-08-26
confidential: Adjust Kubernetes confidential jobs for TDX capable CIs

fidencio:
This patch set is needed in order to get the current `cc-kubernetes` jobs passing in the TDX capable CI.

---

k8s: Make confidential tests aware of the proxy

Right now the confidential tests are not aware of the proxy, which leads
to failures on the TDX CI, as this CI runs inside Intel network.

Fixes: #5082

---

k8s: confidential: Use `kubernetes_create_cc_pod`

Instead of duplicating the code for creating and checking whether a
pod's been created, let's take advantage of the already present
`kubernetes_create_cc_pod` function.

---

k8s: confidential: Fix identation in `kubernetes_create_cc_pod`

We're using tabs instead of spaces in this file.

---

k8s: confidential: Increase the wait timeout to 60s

Let's double the wait timeout as I've noticed, in the internal TDX CI,
that pulling the image inside the guest may take longer than expected
due to our proxies.

In order to avoid random failures due to this, let's simply double the
amount of time and this will not have any side-effect on the other jobs.

---

k8s: confidential: Remove wrong documentation

`kubernetes_wait_cc_pod_be_ready` doesn't receive a third argument.

---

k8s: confidential: Respect original kernel_params

Let's make sure the original kernel parameters are respected, as those
are very much important for the TEE bases CIs, considering some of those
options are a must in order to have the VMs booting up.

Fixes: #5079

---

### DIFF
--- a/integration/kubernetes/confidential/agent_image.bats
+++ b/integration/kubernetes/confidential/agent_image.bats
@@ -49,6 +49,24 @@ setup() {
 	clear_kernel_params
 	add_kernel_params \
 		"agent.container_policy_file=/etc/containers/quay_verification/quay_policy.json"
+
+	# In case the tests run behind a firewall where images needed to be fetched
+	# through a proxy.
+	local https_proxy="${HTTPS_PROXY:-${https_proxy:-}}"
+	if [ -n "$https_proxy" ]; then
+		echo "Enable agent https proxy"
+		add_kernel_params "agent.https_proxy=$https_proxy"
+
+		local local_dns=$(grep nameserver /etc/resolv.conf \
+			/run/systemd/resolve/resolv.conf  2>/dev/null \
+			|grep -v "127.0.0.53" | cut -d " " -f 2 | head -n 1)
+		local new_file="${BATS_FILE_TMPDIR}/$(basename ${pod_config})"
+		echo "New pod configuration with local dns: $new_file"
+		cp -f "${pod_config}" "${new_file}"
+		pod_config="$new_file"
+		sed -i -e 's/8.8.8.8/'${local_dns}'/' "${pod_config}"
+		cat "$pod_config"
+	fi
 	
 	copy_files_to_guest
 }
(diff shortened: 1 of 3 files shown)


integration/kubernetes/confidential/agent_image.bats
```diff
@@ -49,6 +49,20 @@ setup() {
 	clear_kernel_params
 	add_kernel_params \
 		"agent.container_policy_file=/etc/containers/quay_verification/quay_policy.json"
+	if [ -n "$https_proxy" ]; then
```
wainersm, line 54: I remember the first time @arronwy introduced the handler for https proxying, the url could be passed via `HTTPS_PROXY` (uppercase) or `https_proxy` (lowercase). See https://github.com/kata-containers/tests/blob/CCv0/integration/containerd/confidential/agent_api.bats#L22

I am saying that just to be sure you don't need such as handling here as well.

Apart from that, I think we should refactor the code to avoid those boilerplate. In a future PR, I mean.

integration/kubernetes/confidential/lib.sh
```diff
@@ -150,7 +150,7 @@ assert_pod_fail() {
 
 setup_decryption_files_in_guest() {
     local rootfs_agent_config="/etc/agent-config.toml"
-    sudo -E AA_KBC_PARAMS="offline_fs_kbc::null" envsubst < ${katacontainers_repo_dir}/docs/how-to/data/confidential-agent-config.toml.in | sudo tee ${rootfs_agent_config}
+    sudo -E AA_KBC_PARAMS="offline_fs_kbc::null" HTTPS_PROXY="${https_proxy}" envsubst < ${katacontainers_repo_dir}/docs/how-to/data/confidential-agent-config.toml.in | sudo tee ${rootfs_agent_config}
```
wainersm, line 152: please use `${https_proxy:-}` to avoid bash complaining about unset variable...

wainersm (approved):
@fidencio left a few comments. In general it is good to go. thanks!

integration/kubernetes/confidential/lib.sh
fidencio: Done!

integration/kubernetes/confidential/agent_image.bats
fidencio: Done!

fidencio:
`jenkins-ci-ubuntu-20.04-x86_64-cc_skopeo_cri_containerd_k8s_devmapper` passed, it's just now updated:
http://jenkins.katacontainers.io/job/tests-CCv0-ubuntu-20.04-x86_64-CC_SKOPEO_CRI_CONTAINERD_K8S-PR/88/

PR #5081, opened 2022-08-26
CCv0 | kubernetes/lib: fix regex in retrieve_sandbox_id()

### LINKED ISSUE #5077: CC | Adapt the tests so that they can be ran from the Operator CI too

I am working on a CI for the CC operator, and the idea is to re-use the cc tests from kata containers. I glanced at some tests and they will need to be changed slightly, for instance, don't configure containerd because it is already done by the operator.

Use that issue to track those kind of changes.

Refers to https://github.com/confidential-containers/operator/issues/47

wainersm:
In retrieve_sandbox_id() the output of `ps -ef` is parsed, the regex looks for
occurencies of "id". Currently the regex expect an extra space after the
id number which can lead to not getting the id.

This commit changed the regex used to a more accurate one.

Fixes https://github.com/kata-containers/tests/issues/5077

### DIFF
--- a/integration/kubernetes/confidential/lib.sh
+++ b/integration/kubernetes/confidential/lib.sh
@@ -97,7 +97,7 @@ kubernetes_create_cc_pod() {
 # Retrieve the sandbox ID 
 #
 retrieve_sandbox_id() {
-	sandbox_id=$(ps -ef | grep containerd-shim-kata-v2 | egrep -o "id [^,][^,].* " | awk '{print $2}')
+	sandbox_id=$(ps -ef | grep containerd-shim-kata-v2 | egrep -o "\s\-id [a-z0-9]+" | awk '{print $2}')
 }
 
 # Check out the doc repo if required


wainersm:
Hi @GeorginaKin @Megan-Wright ! To be honest I didn't understand the original regex `egrep -o "id [^,][^,].* "`. So could you please review that PR?

littlejawa:
Hey @wainersm,
I'm not sure I understand.
The existing code `egrep -o "id [^,][^,].*"`, should match the string "id" followed by a space, then additional characters. So the "id" in "confidential" should not be reported by it.
I tried with your example output, and I get only the expected ID.

Your code is more specific, so I'm fine with it. I just don't get why it fixes something.
Unless there are other situations where the output contains "id [something]" ?

wainersm:
> Hey @wainersm, I'm not sure I understand. The existing code `egrep -o "id [^,][^,].*"`, should match the string "id" followed by a space, then additional characters. So the "id" in "confidential" should not be reported by it. I tried with your example output, and I get only the expected ID.
> 
> Your code is more specific, so I'm fine with it. I just don't get why it fixes something. Unless there are other situations where the output contains "id [something]" ?

Hi @littlejawa , you are right but I swear the extra 'id' were messing with the parsing on my environment. I will reproduce again and determine for sure what is the problem.

Hey, thanks for the review!

wainersm:
Hi @littlejawa @GeorginaKin ! I managed to reproduce the problem. It turns out to be an extra whitespace on the end of the regex. See:
```
$ ps -ef | grep containerd-shim-kata-v2
root       42429       1  0 14:29 ?        00:00:00 /opt/confidential-containers/bin/containerd-shim-kata-v2 -namespace k8s.io -address /run/containerd/containerd.sock -publish-binary /opt/confidential-containers/bin/containerd -id 2d561a3f51b3bbfb0b45f7ba05d4454407f8681bb492175391d7292e654252d4
root       43012   42665  0 14:34 pts/0    00:00:00 grep --color=auto containerd-shim-kata-v2
$ ps -ef | grep containerd-shim-kata-v2 | egrep -o "id [^,][^,].* "
$ ps -ef | grep containerd-shim-kata-v2 | egrep -o "id [^,][^,].*"
id 2d561a3f51b3bbfb0b45f7ba05d4454407f8681bb492175391d7292e654252d4
```
The original regex is `"id [^,][^,].* "`. I changed to `"id [^,][^,].*"` and it then got the id.

However on this PR I am suggesting `"\s\-id [a-z0-9]+"` that I believe is more accurate. If you agree I will use that regex; just need to adjust the commit message.

littlejawa (approved):
LGTM
Thanks!

stevenhorsman:
So I think that the regex got copied into the tests repo from the `ccv0.sh` script and is a combination of me, Fabiano (https://github.com/kata-containers/kata-containers/commit/5fdbdaafd3e53588c8f8a3bdce4bca2b18b7816e) and probably some doc I read once. I agree that it's not very intuitive as it is and for most of the 'real' scenarios I think the sandbox ID is a hex string and would match your regex, but I think in some of the general Kata tests where we specify a hard-coded id we use a mix of upper and lower case characters and hyphens, so we might just need to be aware of making this too specific. It should be fine for it's current usage though.

stevenhorsman:
> > So I think that the regex got copied into the tests repo from the `ccv0.sh` script and is a combination of me, Fabiano ([kata-containers/kata-containers@5fdbdaa](https://github.com/kata-containers/kata-containers/commit/5fdbdaafd3e53588c8f8a3bdce4bca2b18b7816e)) and probably some doc I read once. I agree that it's not very intuitive as it is and for most of the 'real' scenarios I think the sandbox ID is a hex string and would match your regex, but I think in some of the general Kata tests where we specify a hard-coded id we use a mix of upper and lower case characters and hyphens, so we might just need to be aware of making this too specific. It should be fine for it's current usage though.
> 
> hmmm I thought the sandbox id were always generated by the runtime as a hex string as you mentioned. @stevenhorsman do you think that removing the extra space, i.e. from `"id [^,][^,].* "` to `"\s\-id [^,][^,].*"` would work out on most cases?

Hey Wainer, I'm sorry that I confused things. I threw out the possibility of some of the hard-coded sandboxID labels that we have in `crictl` pod-config.yaml coming through, but after getting a chance to actually test it this morning I've found that this `uid` is different from the `id` of the pod that we are trying to grab here, which seems to always be generated as a hex string, so I think `\s\-id [a-f0-9]+` (or potentially even `\s\-id [[:xdigit:]]+` if all our environments are posix compliant) should be good enough to match and is clearer than the previous code.

I don't see a reason why the trailing space is important, but Fabiano might as I think it was added in his commit?

stevenhorsman (approved):
LGTM. Thanks

Issue #5091, opened 2022-08-31
Ubuntu 22.04 CI is failing because `python` is a package that's not available


fidencio:
```
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python2-minimal python2 dh-python 2to3 python-is-python3
```

GabyCT:
Closing this as it seems that is not an issue

PR #5092, opened 2022-08-31
ubuntu: Always install python3

fidencio:
We're facing the following issue as part of the CI:
```
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python2-minimal python2 dh-python 2to3 python-is-python3
```

Considering that python3 is good enough for everything we need, let's
stick to it.

Fixes: #5091


stevenhorsman (approved):
LGTM. Thanks!

fidencio:
Aha, this one is not actually needed as part of the release, we just need a backport of another one that introduced the check, thus I'm closing it.

PR #5099, opened 2022-08-31
metrics: Remove persistent volume in Cassandra

### LINKED ISSUE #5098: remove persistent volume in cassandra

GabyCT:
This PR removes the persistent volume that we have in our
Cassandra metrics test.

Fixes #5098


fidencio (approved):
lgtm, thanks @GabyCT!

Out of curiosity, why we don't need the PV?

dborquez (approved):
lgtm

GabyCT:
In the case of Cassandra doesn't like shared storage and it's essential to have low latency random reads and good sequential writes at the same time. This kind of requirements can be only satisfied wit local and persistent storage solutions.

Issue #5105, opened 2022-09-02
arm64: enable virtio-mem test


jongwu:
virtio-mem has supported for current kernel on arm64, so should enable the test

PR #5106, opened 2022-09-02
integration: enable virtio-mem test for arm64

jongwu:
The current kernel 5.19 has supported virtio-mem on arm64. Therefore
enable the test.

Fixes: #5105


fidencio (approved):
lgtm, thanks @jongwu!

jongwu:
Sorry @fidencio, I find that Virtio-mem is not supported for arm64 on QEMU v6.2 but on v7.0. We should upgrade QEMU to v7.0 then go back to enable this test on arm64.

Issue #5107, opened 2022-09-02
CC | test: Encrypted image pull test can't be run by a non-root user


stevenhorsman:
## Description

I'm trying to follow the CCv0 instructions of [Using Kubernetes to create a Kata confidential containers pod from the encrypted ssh demo sample image](https://github.com/kata-containers/kata-containers/blob/CCv0/docs/how-to/how-to-build-and-test-ccv0.md#using-kubernetes-to-create-a-kata-confidential-containers-pod-from-the-encrypted-ssh-demo-sample-image) as a non-root user, but due to the change in https://github.com/kata-containers/tests/commit/97664005126c69b424eab18b695287bf15a53773 is doesn't work any longer.
When I (and the script) run `kata-runtime kata-env` as a non-root user I get
`open /etc/kata-containers/configuration.toml: permission denied` which is why we added the sudo call in the first place. In the  commit the comment says:
>  We don’t need to run `kata-runtime kata-env` as privileged user

so there might be something else that is missing, but otherwise I think we need to add `sudo` back in, albeit with `-E PATH=$PATH` to try and help with the problem that Wainer was addressing

stevenhorsman:
Closed with #5108

PR #5108, opened 2022-09-02
ci: Add sudo's to kata-runtime commands in CC code

stevenhorsman:
In order to allow the CC how-to dev instructions to run as a non-root user
we need to run kata-runtime as sudo, but we should pass through the
PATH to ensure sudo can access it.

Fixes: #5107

### DIFF
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -270,7 +270,7 @@ cp_to_guest_img() {
 	rootfs_dir="$(mktemp -d)"
 
 	# Open the original initrd/image, inject the agent file
-	local image_path="$(kata-runtime kata-env --json | jq -r .Image.Path)"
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @stevenhorsman!

wainersm:
hi @stevenhorsman , recently I removed one of those `sudo` calls because it was failing with the operator CI. I think I was misguided by the fact that the installation via operator allow a non-root user to run `kata-runtime kata-env` (the config file is read-only `-rw-r--r--` for others). Should the operator secure that file?

I didn't test yet but I believe your changes won't break the operator CI again as this you are passing `PATH`.

wainersm (approved):
Thanks @stevenhorsman !

PR #5116, opened 2022-09-07
CC: PoC of supporting signature verification with image-rs & the Offline filesystem KBC

stevenhorsman:


sameo (approved):

stevenhorsman:
Now https://github.com/kata-containers/kata-containers/pull/5138 has merged, I've updated the commit to remove the dependency on that. I think my kata-containers branch was fully up-to-date, so it shouldn't make a difference, but I want to /test to double check that.

stevenhorsman:
@wainersm - if you aren't on a plane I wondered if you wanted the chance to review this PR before I merge it?

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -140,7 +140,7 @@ assert_logs_contain() {
 teardown() {
 	# Print the logs and cleanup resources.
 	echo "-- Kata logs:"
-	sudo journalctl -xe -t kata --since "$start_date"
+	sudo journalctl -xe -t kata --since "$start_date" -n 2500
```
wainersm: side note: it has printed the msgs from kata-env that usually is not what we want to. In the future I should send a patch to filter out their message. is it a good idea?

wainersm (approved):
It looks great @stevenhorsman ! One more step towards skopeo removal \o/

wainersm:
@stevenhorsman I didn't merge because I didn't understand if you want to run the tests again or not.

integration/kubernetes/confidential/agent_image.bats
stevenhorsman: Yeah - that's a good idea. It's definitely a fine balance between printing enough to debug and adding too much!

stevenhorsman:
> @stevenhorsman I didn't merge because I didn't understand if you want to run the tests again or not.

I've re-run these since the kata-containers side was merged and it all passed, so I think we are good to merge now. Thanks for the review Wainer!

Issue #5137, opened 2022-09-21
CC: Only add `agent.container_policy_file` kernel_params if skopeo used


stevenhorsman:
Currently our test code always adds `agent.container_policy_file` to the kata kernel_params in the signature verification tests. However now we support this feature with image-rs, we can just set this option if skopeo is used and remove another difference from the default set-up and in the measurement

stevenhorsman:
Closed with #5138 merging.

PR #5138, opened 2022-09-21
ci: Add skopeo guard to config change

stevenhorsman:
- To try and minimise the kernel_params changes that we do, only add the `agent.container_policy_file` if we need it for skopeo. This means it shouldn't trigger in the operator CI

Fixes: #5137


wainersm (approved):
Thanks for working on that @stevenhorsman !

snir911 (approved):

PR #5152, opened 2022-09-27
ci: Fix skopeo set-up

stevenhorsman:
Set the rootfs "etc/containers/" directory explicitly


stevenhorsman:
It looks like this wasn't the issue with the skopeo tests

Issue #5156, opened 2022-09-28
CI: cri-o CI failing on renamed / added default runtime tests


littlejawa:
The cri-o CI testing the kata runtime is failing on some configuration tests related to default runtime being used.

These tests must be skipped in the kata CI, which uses a different runtime anyway.

[This PR](https://github.com/cri-o/cri-o/pull/6257) rename an existing test, and adds another one, both failing with kata. We need to skip both.

PR #5157, opened 2022-09-28
Skip additional tests on cri-o CI

littlejawa:
One of those tests is already skipped, but was renamed. The other is similar. In both cases, the integration test is checking the behaviour of cri-o with default runtimes, which is irrelevant for kata, where the runtime is different anyway.
Those tests need to be skipped.

Fixes: #5156


littlejawa:
@wainersm @GabyCT 
May I ask you a review? There is a PR on cri-o blocked by this.

bergwolf (approved):

gkurz (approved):
LGTM. Thanks @littlejawa !

Issue #5185, opened 2022-10-07
static-checks: https://www.intel.com/ being marked as invalid


stevenhorsman:
If there is an update to the README.md then the documentation static checks trigger and the link to intel.com is failing:
`ERROR: found HTTP error status codes for URL https://www.intel.com/ (403)` [e.g.](https://github.com/kata-containers/kata-containers/actions/runs/3204152170/jobs/5235107667).

After some manual testing I think it's a regression introduce in #4883 as `-A "$user_agent"` was removed from the `curl` request in this PR:
```
$ curl -sIL -H "Accept-Encoding: zstd, br, gzip, deflate" https://www.intel.com
HTTP/2 403
```
vs
```
$ curl -sIL -A $user_agent -H "Accept-Encoding: zstd, br, gzip, deflate" https://www.intel.com
HTTP/2 301
```

I'm not sure if it was a deliberate change to remove the user agent, so have asked Derek, but in the mean time I'd like to try fixing it.

PR #5186, opened 2022-10-07
ci: Add user agent back into curl request

stevenhorsman:
Add user agent back into curl request after it was removed in #4883

Fixes: #5185

### DIFF
--- a/.ci/static-checks.sh
+++ b/.ci/static-checks.sh
@@ -500,7 +500,7 @@ check_url()
 		curl_args+=("-u ${GITHUB_USER}:${GITHUB_TOKEN}")
 	fi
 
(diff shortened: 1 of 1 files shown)


stevenhorsman:
I have a test PR I used to reproduce this: https://github.com/kata-containers/kata-containers/pull/5371. I'm not 100% sure how we test it before it is merged though as I think we need to trigger the static checks on the kata-containers repo and the static-checks don't use the Depends-on IIRC.

liubin:
Thank you @stevenhorsman.

I manually tested it with `curl` command, and this should work.

So I'd like to merge this one asap and fix it if it does not work.

liubin (approved):

Issue #5228, opened 2022-11-01
CC: Update tests to reflect signature verification configuration options


stevenhorsman:
Do the tests updates required for https://github.com/kata-containers/kata-containers/issues/4888

Notably the acceptance criteria is below, but those scenarios are largely covered, but just need the set-up working.

## Acceptance Criteria

_**Scenario**: Ensure that signature validation isn't required with default settings_
**Given** I install kata CC (longer term via the operator, but short term building it locally) without any extra signature verification configuration
**When** I create a pod from a non-signed container image
**Then** The pull image works and the pod can start

_**Scenario**: Ensure the kata-configuration.toml setting is passed through to image-rs and a non-signed image is rejected_
**Given** I install kata CC (longer term via the operator, but short term building it locally) without any signature verification configuration and set `enable_signature_verfication=true` (or the final equivalent) in kata-configuration.toml 
**When** I create a pod from a non-signed container image
**Then** The pull image fails with a message that reflects that signature verification is required

_**Scenario**: Ensure the kata-configuration.toml setting is passed through to image-rs and a signed image is not rejected_
**Given** I install kata CC (longer term via the operator, but short term building it locally) and set `enable_signature_verfication=true` (or the final equivalent) in kata-configuration.toml and provide a valid signature verification configuration to the attestation agent/kbc
**When** I create a pod from a matching signed container image
**Then** The pull image works and the pod can start

stevenhorsman:
Closed with #5229 merging

PR #5229, opened 2022-11-01
ci: Set-up sig verification config for tests

stevenhorsman:
Update tests that need the signature verification to be enabled, by removing the default kernel_param that disables it

Depends-on: github.com/kata-containers/kata-containers#5552
Fixes: #5228

### DIFF
--- a/integration/confidential/lib.sh
+++ b/integration/confidential/lib.sh
@@ -116,6 +116,24 @@ clear_kernel_params() {
 		"$RUNTIME_CONFIG_PATH"
 }
 
+# Remove parameters in the 'kernel_params' property on kata's configuration.toml
+#
+# Parameters:
+#	$1 - parameter name
+#
+# Environment variables:
+#	RUNTIME_CONFIG_PATH - path to kata's configuration.toml. If it is not
+#			      export then it will figure out the path via
+#			      `kata-runtime env` and export its value.
+#
+remove_kernel_param() {
+	local param_name="${1}"
+	load_runtime_config_path
+
+	sudo sed -i "/kernel_params = /s/$param_name=[^[:space:]\"]*//g" \
+		"$RUNTIME_CONFIG_PATH"
+}
+
 # Enable the agent console so that one can open a shell with the guest VM.
 #
 # Environment variables:
@@ -222,6 +240,10 @@ setup_common_signature_files_in_guest() {
 }
 
 setup_offline_fs_kbc_signature_files_in_guest() {
+	# Enable signature verification via kata-configuration by removing the param that disables it
+	remove_kernel_param "agent.enable_signature_verification"
+
+	# Set-up required files in guest image
 	setup_common_signature_files_in_guest
 	add_kernel_params "agent.aa_kbc_params=offline_fs_kbc::null"
 	cp_to_guest_img "etc" "${SHARED_FIXTURES_DIR}/offline-fs-kbc/aa-offline_fs_kbc-resources.json"
--- a/functional/sev/run.sh
+++ b/functional/sev/run.sh
@@ -59,7 +59,7 @@ run_kbs() {
 calculate_measurement_and_add_to_kbs() {
   local kernel_path="$(kata-runtime kata-env --json | jq -r .Kernel.Path)"
   local initrd_path="$(kata-runtime kata-env --json | jq -r .Initrd.Path)"
-  local append="tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k cryptomgr.notests net.ifnames=0 pci=lastbus=0 console=hvc0 console=hvc1 quiet panic=1 nr_cpus=1 agent.config_file=/etc/agent-config.toml"
+  local append="tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k cryptomgr.notests net.ifnames=0 pci=lastbus=0 console=hvc0 console=hvc1 quiet panic=1 nr_cpus=1 agent.config_file=/etc/agent-config.toml agent.enable_signature_verification=false"
 
   # Generate digest from sev-snp-measure output - this also inserts measurement values inside OVMF image
   measurement=$(~/.local/bin/sev-snp-measure --mode=sev --output-format=base64 \
--- a/integration/containerd/confidential/agent_image.bats
+++ b/integration/containerd/confidential/agent_image.bats
@@ -31,10 +31,8 @@ setup() {
 	assert_pod_fail
 }
 
-@test "$test_tag Test can pull an unencrypted image inside the guest" {
+@test "$test_tag Test can pull an unencrypted image inside the guest without signature config" {
 	local container_config="${FIXTURES_DIR}/container-config.yaml"
-
-	setup_signature_files
 
 	create_test_pod
 


GabyCT (approved):

stevenhorsman:
The failing tests seem to be working locally, so I've added some debug in to see if I can work out what I've done wrong /test

stevenhorsman:
@wainersm - my sed injection to remove the parameters seems to be going wrong in the CI (but somehow working locally), so if you have some cycles to take a look and see if you can recommend a better solution, that would be greatly appreciated as this change is blocking some new 0.2 function tests from going in :(

stevenhorsman:
> @wainersm - my sed injection to remove the parameters seems to be going wrong in the CI (but somehow working locally), so if you have some cycles to take a look and see if you can recommend a better solution, that would be greatly appreciated as this change is blocking some new 0.2 function tests from going in :(

I've worked out the problem. We are losing the trailing quote ending the `kernel_params` section, so I'll try and fix the regex to sort this.

stevenhorsman:
The latest issue was the agent api restriction test's agent config toml file override the signature verification setting and causing the failure, so I've added the setting there until we fix this issue.

integration/containerd/confidential/fixtures/agent-configuration-no-exec.toml
```diff
@@ -44,3 +44,6 @@ allowed = [
         "WaitProcessRequest",
         "WriteStreamRequest"
 ]
+
+# TODO - we can remove this once this file doesn't override cmdline parameters
+enable_signature_verification = false
```
arronwy, line 49: I not sure whether we need add `enable_signature_verification` at the beginning of the config file? I tried to add `https_proxy` before, and did not take effect when added at the bottom.

stevenhorsman: Thanks for the tip - I'll try updating that. And we should probably look into it more later!

fidencio (approved):
lgtm, thanks @stevenhorsman!

stevenhorsman:
The SEV test are failing because they are using the embedded the agent-config.toml in their rootfs, so without Tobin’s PR they will end up with signature verification turned on.
I can add enable_signature_verification = false to their sample agent config, they don’t work like the rest of the tests and add/modify the agent-config.toml as required, but just use the one in rootfs.sh, so if I update that to fix their test then it will go into the kata-payload for the release. I think we might need to re-look at these tests to understand why there is so much duplicated/slight different code there and refactor to have a common code base to help avoid this sort of thing in future \cc @ryansavino

integration/containerd/confidential/fixtures/agent-configuration-no-exec.toml
```diff
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+
```
fidencio, line 6: This empty line shouldn't be here, but it's not worth removing unless there's some other changes to be done.

stevenhorsman: Yeah - bad undo-ing of my temp code :(

stevenhorsman:
@ryansavino @Alex-Carter01 @fitzthum - we need some help with the SEV tests. I can see from the job log that the sec config has the setting to ignore signature verification: `kernel_params = "agent.config_file=/etc/agent-config.toml  agent.enable_signature_verification=false`, but I'm not sure if you change it elsewhere, or the new setting needs adding to the measurement of something in https://github.com/kata-containers/tests/blob/83617961791b4fffa0274a83a14ec7be43ad83b5/integration/kubernetes/confidential/sev.bats#L178 ?? As there is no log output I can't debug it. Note this is a hard-blocker with new 0.2 features backed up behind it.

integration/confidential/lib.sh
```diff
@@ -222,6 +240,10 @@ setup_common_signature_files_in_guest() {
 }
 
 setup_offline_fs_kbc_signature_files_in_guest() {
+	# Enable signature verification via kata-configuration by removing the param that disables it
+	remove_kernel_param "agent.enable_signature_verification"
```
fitzthum, line 244: `enable_signature_verification` disables signature verification?

fidencio: `enable_signature_verification` **enables** signatyre verification, if set to `true`, if set to `false` signature verification won't be enabled.

fitzthum: Yeah this comment just struck me as a bit confusing.

fitzthum:
Looking into the tests rn. I think we will need to have a real think about both how we deal with the configs for different platforms/tests and how we turn on and off signature verification. It's a bit all over the place right now.

fitzthum:
Ok, @stevenhorsman @ryansavino I think it is probably a missing parameter in the measurement. The file Steve linked above is the Operator test, which is smart enough to handle additional kernel params, but the Kata test has the kernel parameters hard-coded. It looks like we need to update this line https://github.com/kata-containers/tests/blob/CCv0/functional/sev/run.sh#L62 

Where is the signature_verification parameter added? Isn't the default to not require signatures?

fidencio:
> Where is the signature_verification parameter added? Isn't the default to not require signatures?

https://github.com/kata-containers/kata-containers/pull/5552/files

fitzthum:
> https://github.com/kata-containers/kata-containers/pull/5552/files

Yeah so that change to the Makefile will break the kata-sev tests. It's fairly easy to update the test to include the new param, but I thought signature verification was going to be disabled by default. See https://github.com/kata-containers/tests/issues/5228. If it is then we would sidestep this problem completely.

fidencio:
> Yeah so that change to the Makefile will break the kata-sev tests. It's fairly easy to update the test to include the new param, but I thought signature verification was going to be disabled by default. See #5228. If it is then we would sidestep this problem completely.

That's **not** what's been discussed in https://github.com/kata-containers/kata-containers/issues/4888.

There it's been suggested, without any pushback, to have it enabled by default as it's a security feature.

stevenhorsman:
> > https://github.com/kata-containers/kata-containers/pull/5552/files
> 
> Yeah so that change to the Makefile will break the kata-sev tests. It's fairly easy to update the test to include the new param, but I thought signature verification was going to be disabled by default. See #5228. If it is then we would sidestep this problem completely.

My original approach was to have signature verification off by default, people wanted to have the default to be enabled as it is a security feature (see https://github.com/kata-containers/kata-containers/issues/4888 for some of the discussion) and image-rs are planning to change their default to enable it if it isn't set as well, so we'll line up with them. Our compromise was to add a default parameter to kata-configuration to turn it off for backwards compatibility.

If the test update is easy can someone please provide it? Is it added it to `append_default` like I mentioned earlier?

integration/containerd/confidential/fixtures/agent-configuration-no-exec.toml
stevenhorsman: Just an update - now the ability for the kernel_params to override agent config is in I've been able to remove this all together :)

fidencio:
Considering SEV tests are not required, can we have a estimation on when it'd be okay to merge it in case we have no help to adapt these tests?  It'd help us to unblock other pieces, and folks working on the SEV side would be able to take the appropriate time to re-work the tests, in case it's too complicated.

Mind, I just did something quite similar for the Cloud Hypervisor CI, which was a required one and was blocking PRs to be merged.

fitzthum:
Hm yeah I missed the discussion on https://github.com/kata-containers/kata-containers/issues/4888

I don't think it's a good idea to default to verifying signatures in this release given that we don't really support that on TDX or SEV. I think it would make more sense to default to no for now and then change it once signature verification is more mature. I understand that it's usually good to have defaults be more secure, but at least this flag is measured.

It's pretty easy to update the kata test, but we will also need to update the kata-cc payload to make sure that the SEV config file deployed via the operator contains the flag to disable signatures. That's not too hard either, but we don't have tons of time. I will take a look at the other thread, but still my feeling is that we should just default to not verifying for now.

fitzthum:
See #5235, which should sync up the kata SEV test.

fidencio:
/test

I've updated this PR, bringing in @fitzthum proposed changes from #5235

fitzthum (approved):

PR #5244, opened 2022-11-08
confidential/lib: allow to disable configure_cc_containerd()

### LINKED ISSUE #5077: CC | Adapt the tests so that they can be ran from the Operator CI too

wainersm:
The tests for confidential containers were implemented so that changes for the containerd configuration file are required. For example, the needed to pass the `cri_handler`. However, when those tests are ran from the operator CI, the containerd is already proper configured. Moreover, changing containerd's configuration can break it completely.

This commit changed the logic of configure_cc_containerd() so that if `TESTS_CONFIGURE_CC_CONTAINERD=no` is exported then the function simply does nothing.

Fixes #5077

### DIFF
--- a/integration/confidential/lib.sh
+++ b/integration/confidential/lib.sh
@@ -173,9 +173,18 @@ disable_full_debug() {
 # Parameters:
 #	$1 - (Optional) file path to where save the current containerd's config.toml
 #
+# Environment variables:
+#	TESTS_CONFIGURE_CC_CONTAINERD - if set to 'no' then this function
+#					become bogus.
+#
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @wainersm!

This is the case for the Operator Ci, for instance.

Issue #5263, opened 2022-11-17
Update the tests to using k8s v1.25.x


fidencio:
https://github.com/kata-containers/kata-containers/issues/5681 is updating k8s, and we should update the tests accordingly.

PR #5264, opened 2022-11-17
Adjust the tests code to k8s v1.25.x

fidencio:
versions: Update flannel to its latest release

Let's bump flannel to its v0.20.1 release as we're bumping k8s to its
v1.25.x, and the version of flannel being used relies on
PodSecurityPolicy, which is now deprecated.

---

integration: Adapt kubernetes/init.sh to k8s v1.25.x

We can drop the NoSchedule taint as it's not used with v1.25.x, while
we use the `control-plane` label instead.

Fixes: #5263

---

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -156,7 +156,7 @@ configure_network() {
 			info "Restart the CRI-O service due to $issue"
 			sudo systemctl restart crio
 		fi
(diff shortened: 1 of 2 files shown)


stevenhorsman:
@fidencio - given we are on flannel 0.21 now and doing our Kubernetes tests on AKS, I think we can close this PR?

Issue #5256, opened 2022-11-15
CC: add e2e tests for IBM SE on CI


BbolroC:
**Which feature do you think can be improved?**

Similar to #4977, we would like to establish e2e tests for IBM secure execution (SE) to verify the functionality for TEE-enabled CC. 

We are currently waiting for a public test machine (SE-enabled) on IBM cloud. In the mean time, as an initial step, we would like to set up a CI job without SE, but with an encrypted image of a test pod on the current Jenkins slaves for s390x. 
When the machine is available (hopefully in December, 2022), we will extend the tests to embrace SE.

**How can it be improved?**

1. Declare `CC_SE_CRI_CONTAINERD_K8S` for `CI_JOB` at  https://github.com/kata-containers/tests/blob/CCv0/.ci/ci_job_flags.sh#L108
2. Add a condition `CC_CRI_CONTAINERD_K8S_SE_QEMU` below 
https://github.com/kata-containers/tests/blob/CCv0/.ci/run.sh#L77
3. Add a test call to `make`  target `cc-se-containerd`
https://github.com/kata-containers/tests/blob/CCv0/.ci/run.sh#L77
4. Place a test script at `functional/se` like https://github.com/kata-containers/tests/tree/CCv0/functional/sev
5. Check if any modification is needed at building a rootfs image
https://github.com/kata-containers/tests/blob/CCv0/.ci/install_kata_image.sh#L51
6. Check if a special kernel for SE is needed
https://github.com/kata-containers/tests/blob/CCv0/.ci/install_kata_kernel.sh#L66
7. Add additional condition for SE for configuration placement under https://github.com/kata-containers/tests/blob/CCv0/.ci/install_runtime.sh#L134

I would like to hear any feedback from you: @stevenhorsman and @hbrueckner

fidencio:
> As mentioned above I don't think we want to use these. The TDX CC tests are probably a better model to use, but I don' think they are ready quite yet.

I second that.

BbolroC:
> https://github.com/kata-containers/tests/blob/005e1c7322c8fbe50b13e7606c2c819c7a72fedf/.ci/run.sh#L81
> 
> to call the `cc-kubernetes` test targets - the sev tests don't match the structure, or test cases of the rest of our CC testing, so I'd want us to invest in doing the SE tests correctly and build on the current tests cases, and just change the preparation if nescessary.

[...]

PR #5262, opened 2022-11-17
Authenticated registry tests

stevenhorsman:
Add the three new test scenarios described in the issue for containerd w/ crictl and kubernetes 

Fixes: #5209
Depends-on: github.com/kata-containers/kata-containers#5680

PR #5282, opened 2022-11-24
CC: add e2e tests for IBM SE on CI

BbolroC:
This is to add 2 e2e test targets for IBM SE on CI. One could find a detailed discussion at the issue below.

Info: this is an initial draft which makes the tests only run on a Z machine without SE functionality. (SE capable machines not yet prepared in public), which leads to no implementation for building a secured image, handling a public key and preparing a SE specific configuration file for the runtime.

Fixes: #5256


hbrueckner (approved):
@BbolroC Awesome! Thanks a lot!

/lgtm

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -104,6 +108,7 @@ assert_logs_contain() {
 
 @test "$test_tag Test can launch pod with measured boot enabled" {
 	switch_measured_rootfs_verity_scheme dm-verity
+	remove_kernel_param "agent.container_policy_file"
```
hbrueckner, line 111: Hi @BbolroC,
I think it it would be good for others with some insights this change (as discussed offline).

BbolroC: Thanks for the comment. In the setup, the following kernel parameter is added if `SKOPEO` is set:

```
if [ "${SKOPEO:-}" = "yes" ]; then
    add_kernel_params \
	"agent.container_policy_file=/etc/containers/quay_verification/quay_policy.json"
fi
```

which drives an agent to run a skopeo command with `--policy /etc/containers/quay_verification/quay_policy.json`, but in this test case, this does not exist because it is not copied into guest. And in this test, `dm-verity` is involved, so we would get an corruption error if the policy file is copied.

It was the best to me to remove a kernel parameter `agent.container_policy_file` before launching a pod. I look forward to any better solutions for this.

stevenhorsman: Hey Choi, it is probably worth noting at this point that the skopeo tests are pretty flaky, so we've turned them off for other platforms as the path is effectively deprecated and replaced by image-rs, so we should be tidying up this soon-ish anyway and removing that config option.

stevenhorsman:
\cc @wainersm

integration/kubernetes/confidential/agent_image.bats
BbolroC: yep, thanks for sharing. :wink:

stevenhorsman: Sorry, should have added that this might explain why we've not seen failures with the other jobs as we are running the skopeo tests. It might be worth considering not bothering with the skopeo variant for s390x and just having crictl and kubernetes (with kubernetes being the priority by a fair way), but sorry if you've already done lots of the skopeo adaptations already!

BbolroC: Ah, it is very kind of you. It was good to learn about skopeo and how the test works with a signing mechanism. worth spending. :wink:

GabyCT (approved):

BbolroC:
Sorry for lots of test triggers. I was creating a dedicated test job on Jenkins for this. All has been set. This is the last trigger.

/test-s390x-cc

BbolroC:
a CI job: http://jenkins.katacontainers.io/job/kata-containers-CCv0-ubuntu-20.04-s390x-CC_CRI_CONTAINERD_K8S-PR

BbolroC:
> I wonder if it is worth removing the skopeo job flags from this PR if we aren't going to create. matching job (which I'd agree with)?

Yep, it makes sense. I will remove it when I get back to my desk. Thanks!

stevenhorsman (approved):
LGTM - thanks @BbolroC

BbolroC:
The 2 CI job failures for s390x are attributed to the issue for image-rs (see https://github.com/kata-containers/tests/pull/5262#issuecomment-1329096806).

.ci/ci_job_flags.sh
```diff
@@ -148,22 +148,25 @@ case "${CI_JOB}" in
...
-	export KATA_BUILD_KERNEL_TYPE="tdx"
-	export KATA_BUILD_QEMU_TYPE="tdx"
-	export KUBERNETES="yes"
-	export AA_KBC="offline_fs_kbc"
-	export KUBERNETES=yes
+	if [[ "${CI_JOB}" =~ TDX ]]; then
+		export TEE_TYPE="tdx"
+		export KATA_BUILD_KERNEL_TYPE="tdx"
+		export KATA_BUILD_QEMU_TYPE="tdx"
+	elif [[ "${CI_JOB}" =~ _SE_ ]]; then
```
wainersm, line 167: Hi @BbolroC ! Sorry for taking a long time to review your PR.

I'm glad more people know how to extend the CI and tests :)

My first question (and perhaps the only, I am still going to look the remain of this PR) is: IBM SE doesn't need a specific kernel, it does?    (Item 6 in https://github.com/kata-containers/tests/issues/5256)

wainersm: Same question for the hypervisor (QEMU).

BbolroC: yes, afaik, but I think @hbrueckner could give you a clearer answer for this.

wainersm:
@BbolroC apart from the kernel/qemu question, this PR looks good to go. Thanks again for your contribution!

.ci/ci_job_flags.sh
hbrueckner: Hi @wainersm ,

> My first question (and perhaps the only, I am still going to look the remain of this PR) is: IBM SE doesn't need a specific kernel, it does? (Item 6 in #5256)

No specific kernel or qemu is required. Secure Execution support is available since kernel v5.7 and qemu v5.1.

stevenhorsman:
From the s390x cc job log we can see that the authenticated registry pull worked:
```
19:10:44 # Dec 05 19:08:49 kata-ci-worker-03 kata[1125257]: time="2022-12-05T19:08:49.7113512Z" level=debug msg="reading guest console" console-protocol=unix console-url=/run/vc/vm/7b7331432b6a0dbe3fb32970db74f8aa2d19694d626573615e09e1bb726fb35e/console.sock name=containerd-shim-v2 pid=1125257 sandbox=7b7331432b6a0dbe3fb32970db74f8aa2d19694d626573615e09e1bb726fb35e source=virtcontainers subsystem=sandbox vmconsole="{\"msg\":\"pull and unpack image \\\"confidential-containers-auth_nginx\\\", with image-rs succeeded \",\"level\":\"INFO\",\"ts\":\"2022-12-05T19:08:49.699054565Z\",\"name\":\"kata-agent\",\"subsystem\":\"rpc\",\"pid\":\"165\",\"version\":\"0.1.0\",\"source\":\"agent\"}"
```
but the failure was on the create of the container. This is to be expected as the test image is only amd64. https://github.com/kata-containers/tests/pull/5292 attempts to make the test image multi-arch, but that is failing as I think it needs code from here, so my suggestion is we merged this as an improvement on the current code, with just the one failing test and then I'll rebase my PR on it and hopefully we'll have them all working. If this isn't achievable in the short term then I can make that test either architecture specific, modify the success criteria, or similar.

Issue #5283, opened 2022-11-25
Remove round_post9_test.go and round_pre10_test.go


liubin:
Hi @GabyCT 

We have two files:
- cmd/checkmetrics/round_post9_test.go
- cmd/checkmetrics/round_pre10_test.go

which are used to use `math.Round` after go 1.10.

Now we are using go 1.19, so we can drop these to files and use `math.Round` replace `Round`.

PR #5284, opened 2022-11-25
remove files that for different golang versions

liubin:
We have two files:

- cmd/checkmetrics/round_post9_test.go
- cmd/checkmetrics/round_pre10_test.go

which are used to provide a Round method.

After golang 1.10, we can use  math.Round directly.

Fixes: #5283


GabyCT (approved):

cmaf (approved):
thanks @liubin

Issue #5295, opened 2022-12-01
CI: CNI plugins gone during re-creation of k8s cluster on baremetal


BbolroC:
# Description of problem

During a k8s e2e test, sometimes, it reaches the re-creation of the cluster due to timeout of getting a network ready (flannel daemonset) like:

```
serviceaccount/flannel created
configmap/kube-flannel-cfg created
daemonset.apps/kube-flannel-ds created

[init.sh:160] INFO: Wait for Flannel pods to show up
INFO: Environment initialization failed. Clean up and try again.

[cleanup_env.sh:31] INFO: Reset Kubernetes
[reset] Reading configuration from the cluster...
```

But, in the subsequent initialization of the cluster, kubelet fails to get a pod in the namespace `kube-system` ready like:

```
NAMESPACE     NAME                                        READY   STATUS              RESTARTS   AGE
kube-system   coredns-64897985d-544pg                     0/1     ContainerCreating   0          2m7s
kube-system   coredns-64897985d-7tz55                     0/1     ContainerCreating   0          2m7s
kube-system   etcd-kata-ci-worker-03                      1/1     Running             1          2m23s
kube-system   kube-apiserver-kata-ci-worker-03            1/1     Running             1          2m22s
kube-system   kube-controller-manager-kata-ci-worker-03   1/1     Running             1          2m22s
kube-system   kube-flannel-ds-4x8s5                       1/1     Running             0          2m7s
kube-system   kube-proxy-zjxtk                            1/1     Running             0          2m7s
kube-system   kube-scheduler-kata-ci-worker-03            1/1     Running             1          2m22s
```

It is because the cni plugin files (e.g. flannel) at `/opt/cni/bin` have been removed during the clean-up triggered by the aforementioned failure:

```
Warning  FailedCreatePodSandBox  2m3s  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "a76b89b5db56f3dcb65933c5848447344eadc0fd61109873d9af3e4c6909456d": plugin type="flannel" failed (add): failed to find plugin "flannel" in path [/opt/cni/bin]
```

# Expected result

The plugin files at `/opt/cni/bin`  are not removed during the clean-up. The cluster is successfully created. 

# Actual result

The files are removed and the cluster is not created, which leads to a test failure.

PR #5290, opened 2022-11-30
CI: Relax timeout threshold for flannel

BbolroC:
The timeout threshold for flannel for s390x is a little bit tight. This is to relax it to 60 secs.

Fixes: #5288

Signed-off-by: Hyounggyu Choi <Hyounggyu.Choi@ibm.com>
(cherry picked from commit dff44db4f17867b453a0f2914085194c322796c2)

PR #5297, opened 2022-12-01
WIP CI: Restore removed cni plugin files from backup

BbolroC:
I would like to explain what brings me come to this code change to fix the issue. 

1. Is the issue reproducible?

&rightarrow; Yes, when we set the timeout from the current 60 secs to 10 secs ([init.sh#L161](https://github.com/kata-containers/tests/blob/main/integration/kubernetes/init.sh#L161)), the test reaches the timeout at the 1st try and the follow-up initialization of the cluster fails due to the issue of interest.

2. Are the plugin files alive after a separate run of [init.sh](https://github.com/kata-containers/tests/blob/main/integration/kubernetes/init.sh) and [cleanup_env.sh](https://github.com/kata-containers/tests/blob/main/integration/kubernetes/cleanup_env.sh) in the case where no timeout occurs? 

&rightarrow; Yes, plugin files are alive if a cluster is successfully created (running `init.sh`) and cleaned up (running `cleanup_env.sh`) separately.

3. When are the files removed?

&rightarrow; When the scripts are called from [jenkins_job_build.sh](https://github.com/kata-containers/tests/blob/main/.ci/jenkins_job_build.sh) in a single run (no manual intervention in-between). 
&rightarrow; The files are removed while the `containerd` service is stopped in the cleanup (due to the timeout for flannel). Even they are deleted when the cleanup is called while a cluster is fully functioning like:

```
NAMESPACE     NAME                                        READY   STATUS    RESTARTS   AGE
kube-system   coredns-64897985d-bdgdk                     1/1     Running   0          63s
kube-system   coredns-64897985d-m98pd                     1/1     Running   0          63s
kube-system   etcd-kata-ci-worker-01                      1/1     Running   0          87s
kube-system   kube-apiserver-kata-ci-worker-01            1/1     Running   0          87s
kube-system   kube-controller-manager-kata-ci-worker-01   1/1     Running   0          87s
kube-system   kube-flannel-ds-hjm6f                       1/1     Running   0          64s
kube-system   kube-proxy-xfgjb                            1/1     Running   0          64s
kube-system   kube-scheduler-kata-ci-worker-01            1/1     Running   0          87s
```

The code change is just to make backup for the cni plugins before the cluster creation and restore them if `flannel` is not found after the `containerd` service starts.

It is hard to verify this PR with the current CI because the timeout for flannel is too generous (60 secs). But it has been verified in the following scenario:

1. 1st run: set the timeout for flannel to 10 secs &rightarrow; cleanup is triggered
2. 2nd run: set the timeout to 60 secs &rightarrow; the test should end up with success if the plugins are in place ([test log](http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-s390x-main-PR/1309/consoleText)). Otherwise, it hangs at the initialization phase and fails ([test log](http://jenkins.katacontainers.io/job/kata-containers-2.0-tests-ubuntu-s390x-main-PR/1307/consoleText)).

If you are interested at how the code changes for the scenario look like, see at https://gist.github.com/BbolroC/e4828d8f2b06ab35504867d19ab22155

Fixes: #5295

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -274,6 +274,10 @@ start_cri_runtime_service() {
 
 	sudo systemctl status "${cri}" --no-pager || \
 		die "Unable to start the ${cri} service"
+	if [ ! -f /opt/cni/bin/flannel ]; then
+		info "Plugin \"flannel\" not found, restoring it from backup"
+		tar xvfz /opt/plugins-backup.tar.gz -C /
+	fi
 }
 
 main() {
--- a/integration/kubernetes/run_kubernetes_tests.sh
+++ b/integration/kubernetes/run_kubernetes_tests.sh
@@ -125,6 +125,16 @@ fi
 
 pushd "$kubernetes_dir"
 info "Initialize the test environment"
+plugin_dir="/opt/cni/bin"
+if [ ! -f "${plugin_dir}/flannel" ]; then
+	die "Cannot start testing due to lack of plugin \"flannel\""
+fi
+if [ $(ls "${plugin_dir}" | wc -w) -gt 1 ]; then
+	info "Making backup for cni plugins..."
+	tar cvfz /opt/plugins-backup.tar.gz "${plugin_dir}"
+else
+	die "No plugins, but flannel"
+fi
 wait_init_retry="30"
 if ! bash ./init.sh; then
 	info "Environment initialization failed. Clean up and try again."


bergwolf:
[jenkins_job_build.sh](https://github.com/kata-containers/tests/blob/main/.ci/jenkins_job_build.sh) is supposed to set up the test environment, run tests, and clean up the environment, is it? It appears to me that the problem is why kubernetes is still running while containerd is stopped and cni files are removed.

liubin:
I have one question similar to @bergwolf: who deleted the CNI while the CI job is running?

I see this code, the clean_up will also be called before running the test, but I don't think one node can run two jobs at the same time.

https://github.com/kata-containers/tests/blob/619eb600e2e3e11b01f9aa966d3c1f35240fec9e/.ci/jenkins_job_build.sh#L133-L144

And as another try, can you increase the timeout waiting for the flannel process?

https://github.com/kata-containers/tests/blob/619eb600e2e3e11b01f9aa966d3c1f35240fec9e/integration/kubernetes/init.sh#L168-L170

And in the [init.sh#L168-L181](https://github.com/kata-containers/tests/blob/619eb600e2e3e11b01f9aa966d3c1f35240fec9e/integration/kubernetes/init.sh#L168-L181), we are nothing checking the process(by `waitForProcess`) and pod (by `kubectl wait`), I think we can delete the first wait check.

BbolroC:
> [jenkins_job_build.sh](https://github.com/kata-containers/tests/blob/main/.ci/jenkins_job_build.sh) is supposed to set up the test environment, run tests, and clean up the environment, is it? It appears to me that the problem is why kubernetes is still running while containerd is stopped and cni files are removed.

Hi, Thanks for the question. As you follow the test code, the 1st initialization is made at https://github.com/kata-containers/tests/blob/main/integration/kubernetes/run_kubernetes_tests.sh#L129. If there is a failure during the initialization, the code reaches the cleanup at https://github.com/kata-containers/tests/blob/main/integration/kubernetes/run_kubernetes_tests.sh#L131 thans to https://github.com/kata-containers/tests/blob/main/integration/kubernetes/init.sh#L8, then it waits 30 secs for another try. But the plugin files are removed while a containerd is stopped during the cleanup at https://github.com/kata-containers/tests/blob/main/integration/kubernetes/run_kubernetes_tests.sh#L131. 

This removal does not happen normally (I've confirmed) when we manually run `setup.sh` &rightarrow; `init.sh` (whether something goes wrong or not) &rightarrow; `cleanup_env.sh`. It only occurs in a single run from `jenkins_job_build.sh`.

Kubernetes is not running. It is perfectly cleaned up and re-initialized. The point 1 through 3 above is experiments I've made to find a cause. I will answer to @liubin 's question when your question is resolved. :wink: Thanks.

BbolroC:
> I have one question similar to @bergwolf: who deleted the CNI while the CI job is running?
> 
> I see this code, the clean_up will also be called before running the test, but I don't think one node can run two jobs at the same time.
> 
Yes, you are right. Two job are not running at the same time. The cleanup before the test does not removes the plugin files. The files are removed only when the cleanup is called due to the timeout for flannel while a cluster is running.
> 
> And as another try, can you increase the timeout waiting for the flannel process?

I have done it already at https://github.com/kata-containers/tests/pull/5290/files. The issue would hardly happen at the moment. But I've made this PR just to cover that rare case. 

> https://github.com/kata-containers/tests/blob/619eb600e2e3e11b01f9aa966d3c1f35240fec9e/integration/kubernetes/init.sh#L168-L170
> 
> And in the [init.sh#L168-L181](https://github.com/kata-containers/tests/blob/619eb600e2e3e11b01f9aa966d3c1f35240fec9e/integration/kubernetes/init.sh#L168-L181), we are nothing checking the process(by `waitForProcess`) and pod (by `kubectl wait`), I think we can delete the first wait check.

As you see the code, the second wait assume that there is a pod for the flannel daemonset at [init.sh#L167](https://github.com/kata-containers/tests/blob/main/integration/kubernetes/init.sh#L167). That is why we introduced the 1st wait.

integration/kubernetes/init.sh
```diff
@@ -274,6 +274,10 @@ start_cri_runtime_service() {
 
 	sudo systemctl status "${cri}" --no-pager || \
 		die "Unable to start the ${cri} service"
+	if [ ! -f /opt/cni/bin/flannel ]; then
```
liubin, line 277: Hi @BbolroC, if the issue only occurs in s390x, so is it better to add an $ARCH check and run this logic only for s390x?

BbolroC: @fidencio told me it happened during the TDX CI tests as well. I think it is common across arches.

bergwolf: I think we can restrict it to bear metal test environment instead of just s390x.

bergwolf:
The flannel cni files are removed here, https://github.com/kata-containers/tests/blob/main/integration/kubernetes/cleanup_bare_metal_env.sh#L28-L32

Since .kube and other cni files/configs are removed as well. I wonder if we need to fix the workflow or just the flannel files.

BbolroC:
> The flannel cni files are removed here, https://github.com/kata-containers/tests/blob/main/integration/kubernetes/cleanup_bare_metal_env.sh#L28-L32
> 
> Since .kube and other cni files/configs are removed as well. I wonder if we need to fix the workflow or just the flannel files.

Hup, you are right. I have made this PR by assuming that a flag `BAREMETAL` was false, which is not true, because the test machine is KVM-based. I have to update this PR in a way of fixing a workflow. Thanks a lot for pointing this out.

BbolroC:
I will raise another PR with a new workflow because it turns out this started based on the wrong assumption.

Issue #5304, opened 2022-12-02
CC | Don't install TDVF twice


fidencio:
We've been calling `install_tdvf.sh` from both [install_qemu.sh](https://github.com/kata-containers/tests/blob/CCv0/.ci/install_qemu.sh#L76) and [install_kata.sh](https://github.com/kata-containers/tests/blob/8a0fbf72c390b8dc18df61a9e93042c3fec13e2d/.ci/install_kata.sh#L70).

PR #5305, opened 2022-12-02
CC | Avoid building TDVF twice

fidencio:
TDVF is already installed as part of the install_kata.sh

Fixes: #5304


wainersm (approved):
Good catch @fidencio ! Thx!

Issue #5314, opened 2022-12-07
CC: use the default configuration for runtime for kata-containers CI jobs


BbolroC:
# Description of problem

The kata-runtime refers to a default configuration (`/etc/kata-containers/configuration.toml` or `/usr/share/defaults/kata-containers/configuration.toml`) when a flag `KATA_BUILD_CC` is true and an argument to `ci_entry_point.sh` includes `github.com/kata-containers`. This leads to the following error for the relevant CI jobs (see [one](http://jenkins.katacontainers.io/job/kata-containers-CCv0-ubuntu-20.04-s390x-CC_CRI_CONTAINERD_K8S-PR/34/consoleText) of the logs):

```
Cannot find usable config file (config file "/etc/kata-containers/configuration.toml" unresolvable: file /etc/kata-containers/configuration.toml does not exist, config file "/usr/share/defaults/kata-containers/configuration.toml" unresolvable: file /usr/share/defaults/kata-containers/configuration.toml does not exist)
[install_kata.sh:86] ERROR: kata-runtime kata-env
```

It should be `/opt/confidential-containers/share/defaults/kata-containers/configuration.toml` for CC.

PR #5315, opened 2022-12-07
CC: defer configuring runtime config settings for kata-container CI

BbolroC:
This is to remove `config-settings.go` at `jenkins_job_build.sh` and defer generating it at runtime installation for kata-container CI jobs on non-x86_64 architectures.

Fixes: #5314

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -210,6 +210,10 @@ if [ "${METRICS_CI}" = "false" ]; then
 		# If not a PR, we are testing on stable or master branch.
 		[ -z "$pr_number" ] && specific_branch="true"
 		"${ci_dir_name}/static-checks.sh" --only-arch "$kata_repo" "$specific_branch"
+		# It is too early to configure defaultRuntimeConfiguration for CC
+		# The file generation should be deferred at runtime installation with
+		# /opt/confidential-containers/share/defaults/kata-containers/configuration.toml
+		rm -f src/runtime/pkg/katautils/config-settings.go
 	fi
 fi
 


BbolroC:
This is a remedy for the test failures for http://jenkins.katacontainers.io/job/kata-containers-CCv0-ubuntu-20.04-s390x-CC_CRI_CONTAINERD_K8S-PR/

Thanks for your help @stevenhorsman. I look forward to feedback from @fidencio and @GabyCT.

cc: @hbrueckner

.ci/jenkins_job_build.sh
```diff
@@ -210,6 +210,12 @@ if [ "${METRICS_CI}" = "false" ]; then
 		# If not a PR, we are testing on stable or master branch.
 		[ -z "$pr_number" ] && specific_branch="true"
 		"${ci_dir_name}/static-checks.sh" --only-arch "$kata_repo" "$specific_branch"
+		# It is too early to configure defaultRuntimeConfiguration for CC
+		# The file generation should be deferred at runtime installation with
+		# /opt/confidential-containers/share/defaults/kata-containers/configuration.toml
+		if [ "$KATA_BUILD_CC" == "yes" ]; then
+			rm -f src/runtime pkg/katautils/config-settings.go
```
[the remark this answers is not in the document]
BbolroC: typo;; kkk thanks for pointing it out. :wink:

.ci/jenkins_job_build.sh
```diff
@@ -210,6 +210,12 @@ if [ "${METRICS_CI}" = "false" ]; then
 		# If not a PR, we are testing on stable or master branch.
 		[ -z "$pr_number" ] && specific_branch="true"
 		"${ci_dir_name}/static-checks.sh" --only-arch "$kata_repo" "$specific_branch"
+		# It is too early to configure defaultRuntimeConfiguration for CC
+		# The file generation should be deferred at runtime installation with
+		# /opt/confidential-containers/share/defaults/kata-containers/configuration.toml
+		if [ "$KATA_BUILD_CC" == "yes" ]; then
```
stevenhorsman, line 216: This isn't necessarily a suggestion, but I wanted to open the discussion to see what people think:
- Is there an argument that we remove the conditional and always delete (or attempt to) the `src/runtime/pkg/katautils/config-settings.go` after the static-checks. At the moment it only impacts CC builds, but it was a bit of a bear trap and difficulty debugging, so I don't want someone else to hit that incase there is another scenario that we change the default config location. As this doesn't impact x86 jobs it's more easy to over look.

Another left field suggestion is how far are we away from being able to run the s390x static checks on GHA in which case we can remove the early creation of this?

BbolroC: Here are my two cents even if I have little knowledge why this code block is placed in this way:

1. I agree with removing the condition if it does not break the existing CI jobs.
2. Regarding the static checks, I could make it separate and run it on GHA for s390x (putting in my backlog :wink:), but not sure if we could remove this after it. I am wondering how other architectures are planning to introduce a GHA CI job for this.

cc: @Amulyam24 @ryansavino

fidencio: @stevenhorsman, this will hit the other CIs by the moment we start using the kata-deploy scripts as part of the `main` branch.  So, yep, I'd be in favour of removing the file after running the static-checks (as the first step).

Then, a second step would be to use the GHA to run the static checks for all the arches.  AFAIR GitHub does provide ARM machines (so, no need to have an external runner), and we can rely on the external s390x one.  That'd be my dream. :-)

BbolroC:
Why don't we merge this unless there is something to discuss further?

.ci/jenkins_job_build.sh
stevenhorsman: Hey Choi - sorry to un-resolve this discussion, but I'm not sure we got to a conclusion, unless I've missed something. You stated:
> I agree with removing the conditional if it does not break the existing CI jobs.

Have we been able to test that as my (biased) interpretation is that this approach is supported by all three of us?

BbolroC: > this will hit the other CIs by the moment we start using the kata-deploy scripts as part of the main branch. 

I thought that the conditional is needed for other CIs, isn't it? Otherwise, we could try another run without it. Sorry, I am not good at understanding something clearly. :wink:

stevenhorsman: Ok, sorry - it is as likely that I'm misinterpreting, so I'll be clearer here:
Can we remove the `if [ "$KATA_BUILD_CC" == "yes" ]; then` condition from this PR as I think that will resolve the issue for all the CIs just here rather than deferring it until we hit it later? I don't see a down side to this, but obviously we'll need to re-run the tests to double check.

stevenhorsman: Just seen your re-push. Thanks!

BbolroC:
Note: we cannot verify the change because any CI tests triggered by this PR does not reach the removal of the file.

stevenhorsman:
> Note: we cannot verify the change because any CI tests (non-x86, here only s390x) triggered by this PR does not reach the removal of the file.

Sorry, I'm probably missing something obvious. If we run the s390x CC kata-containers PR test on a kc PR that depends on this (like updating https://github.com/kata-containers/kata-containers/pull/4588/commits) would that not reach this code?

stevenhorsman (approved):
LGTM. Thanks!

Issue #5318, opened 2022-12-12
CC | Don't install / build agent-ctl for CC


fidencio:
CC doesn't use agent-ctl for anything and its build time is quite expensive for us.

PR #5319, opened 2022-12-12
CC | install_runtime: Don't install agent-ctl for CC

fidencio:
CC doesn't use agent-ctl for anything and its build time is quite expensive for us.

Fixes: #5318

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -98,7 +98,9 @@ build_install_agent_ctl(){
 	popd
 }
 
-build_install_agent_ctl
+if [ "${KATA_BUILD_CC}" == "no" ]; then
+	build_install_agent_ctl
+fi
 
 experimental_qemu="${experimental_qemu:-false}"
 


wainersm:
> @BbolroC, mind to take a look at the s390x CIs? @UnmeshDeodhar, @ryansavino, mind to take a look at the sev CI?

The SEV one failed because:

```
07:56:17 functional/sev/run.sh: line 187: cd: simple-kbs: No such file or directory
```

I didn't look at functional/sev/run.sh but maybe it is not cloning the simple-kbs properly.

wainersm (approved):
@fidencio I had this plan to use agent-ctl to test the Agent API restriction config but I never found time to work on it. I didn't know it was time consuming its build. Thanks @fidencio !

fidencio:
Thanks, @wainersm.  If we end up using the agent-ctl we'll work on caching it, but as we don't use it right now we better not install it at all.

I'm adding the `do-not-merge` label so we ensure that the SEV CI will pass.

Issue #5320, opened 2022-12-13
SEV functional test failing due to unnecessary cargo env sourcing


ryansavino:
# Description of problem

SEV functional test failing.

# Expected result

Expected test pass.

# Actual result

```
20:11:16 INFO: Running Confidential Containers tests for AMD SEV
20:11:16 bash functional/sev/run.sh
20:11:16 functional/sev/run.sh: line 196: /home/jenkins/.cargo/env: No such file or directory
20:11:16 ERROR Encountered with Exit Code: 1
```

PR #5321, opened 2022-12-13
Sev cargo sourcing

ryansavino:
sev: added rust installation on host

Needed to build sevctl

Fixes: [#5320](https://github.com/kata-containers/tests/issues/5320)

Signed-Off-By: Ryan Savino <ryan.savino@amd.com>

### DIFF
--- a/functional/sev/run.sh
+++ b/functional/sev/run.sh
@@ -193,7 +193,9 @@ cleanup() {
 }
 
 main() {
-  source "$HOME/.cargo/env"
+  # Rust required to build sevctl
+  "${tests_repo_dir}/.ci/install_rust.sh" && source "$HOME/.cargo/env"
+  
   mkdir -p test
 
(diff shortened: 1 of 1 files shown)


ryansavino:
Forgive the `size/huge` label and review request. Forgot to set the upstream branch to `CCv0` instead of `main`

stevenhorsman (approved):
This looks okay to me, but I still want to see the sev tests reworked to follow the pattern of the rest of the integration tests as this should resolve any sev special processing like this and I don't think we will get SEV support merged into `main` without that.

fidencio:
So, it seems the issue this PR is solving is related to https://github.com/kata-containers/tests/pull/5319, which skips the installation of agent-ctl (not used anywhere in our CC work) and ends up also skipping the rust installation on the host side.

Nice catch!

PR #5344, opened 2022-12-17
ci:add k8s-block-volume.bats

Tim-0731-Hzt:
add k8s-block-volume.bats

Fixes:#5343

Depends-on:github.com/https://github.com/kata-containers/kata-containers/pull/5441

Signed-off-by: Zhongtao Hu <zhongtaohu.tim@linux.alibaba.com>

### DIFF
--- a/.ci/aarch64/install_rom_aarch64.sh
+++ b/.ci/aarch64/install_rom_aarch64.sh
@@ -123,6 +123,19 @@ clean_up()
 	sudo rm -rf "${EDK2_WORKSPACE}"
 }
 
+enable_pflash_in_config()
+{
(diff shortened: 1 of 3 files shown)


integration/kubernetes/run_kubernetes_tests.sh
```diff
@@ -95,6 +95,7 @@ else
                 "k8s-shared-volume.bats" \
                 "k8s-volume.bats" \
                 "k8s-nginx-connectivity.bats" \
+		"k8s-block-volume.bats" \
```
littlejawa, line 98: I think you can remove it from the comment above too ?

Tim-0731-Hzt: fixed

Issue #5348, opened 2022-12-20
containerd logs printed twice


liubin:
In [cri-containerd test](integration/containerd/cri/integration-tests.sh), if the test cases failed, the containerd log will be printed twice:

Once is in containerd's `script/test/cri-integration.sh`:

https://github.com/containerd/containerd/blob/9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6/script/test/cri-integration.sh#L52-L53

And the second time is in `integration/containerd/cri/integration-tests.sh`:

https://github.com/kata-containers/tests/blob/89d0e1757d987964d4e142fd2516ec0eff79d17a/integration/containerd/cri/integration-tests.sh#L155-L163

We can remove the second one.

PR #5351, opened 2022-12-20
cri-containerd: avoid printing containerd log twice

liubin:
After calling containerd's integration test script, then trap the error and print out containerd's log if the remain tests failed.

Fixes: #5348


fidencio (approved):
lgtm, thanks @liubin!

Tim-Zhang (approved):

Issue #5360, opened 2023-01-04
SEV: Can't get image labels from encrypted image


fitzthum:
Currently we try to `docker pull` the encrypted image to get the encryption key and ssh key that are stored in the labels. Unfortunately `docker pull` fails with encrypted images. We should switch to pulling the unencrypted image (which has the same labels) for this purpose.

CC: @ryansavino 

Example failure: http://jenkins.katacontainers.io/job/tests-CCv0-ubuntu-20.04_sev-x86_64-CC_SEV_CRI_CONTAINERD_K8S-PR/197/console

fitzthum:
Fixed.

PR #5361, opened 2023-01-04
SEV: Fix pulling image labels

fitzthum:
Docker fails to pull encrypted image to get labels. Use the unencrypted image (which has the same labels) instead.

Fixes: #5360

Signed-off-by: Tobin Feldman-Fitzthum <tobin@ibm.com>

cc: @ryansavino

### DIFF
--- a/functional/sev/run.sh
+++ b/functional/sev/run.sh
@@ -71,7 +71,7 @@ calculate_measurement_and_add_to_kbs() {
   if [[ -z "${measurement}" ]]; then >&2 echo "Measurement is invalid"; return 1; fi
 
   # Get encryption key from docker image label
-  enc_key=$(esudo docker inspect quay.io/kata-containers/encrypted-image-tests:unencrypted \
+  enc_key=$(esudo docker inspect ghcr.io/fitzthum/encrypted-image-tests:unencrypted \
     | jq -r '.[0].Config.Labels.enc_key')
 
   # Add key, keyset and policy with measurement to DB
@@ -113,7 +113,7 @@ spec:
       runtimeClassName: kata
       containers:
       - name: encrypted-image-tests
-        image: quay.io/kata-containers/encrypted-image-tests:encrypted
+        image: ghcr.io/fitzthum/encrypted-image-tests:encrypted
         imagePullPolicy: Always
 EOF
 
@@ -141,7 +141,7 @@ EOF
   pod_ip=$(esudo kubectl get pod -o wide | grep encrypted-image-tests | awk '{print $6;}')
   
   # Get ssh key from docker image label and save to file
-  esudo docker inspect quay.io/kata-containers/encrypted-image-tests:unencrypted \
+  esudo docker inspect ghcr.io/fitzthum/encrypted-image-tests:unencrypted \
     | jq -r '.[0].Config.Labels.ssh_key' \
     | sed "s|\(-----BEGIN OPENSSH PRIVATE KEY-----\)|\1\n|g" \
     | sed "s|\(-----END OPENSSH PRIVATE KEY-----\)|\n\1|g" \
@@ -203,7 +203,7 @@ main() {
   pip install sev-snp-measure
 
   # Pull unencrypted docker image to get labels
-  esudo docker pull quay.io/kata-containers/encrypted-image-tests:unencrypted
+  esudo docker pull ghcr.io/fitzthum/encrypted-image-tests:unencrypted
 
   # Copy agent-config.toml to initrd image
   initrd_add_files
--- a/functional/sev/run.sh
+++ b/functional/sev/run.sh
@@ -26,6 +26,24 @@ esudo() {
   sudo -E PATH=$PATH "$@"
 }
 
+install_containerd_fork() {
+  local containerd_repo="https://github.com/confidential-containers/containerd"
+  local containerd_tag="v1.6.6.0"
+  local build_dir="$(mktemp -d -t containerd-XXXXXXXXXX)/containerd"
+
+  esudo apt-get install libbtrfs-dev
+
+  git clone -b "$containerd_tag" "$containerd_repo" "$build_dir"
+  pushd "$build_dir"
+  make
+  sudo make install
+  popd
+
+  rm -rf "$build_dir"
+  esudo systemctl restart containerd
+
+}
+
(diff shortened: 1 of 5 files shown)


functional/sev/run.sh
```diff
@@ -71,7 +71,7 @@ calculate_measurement_and_add_to_kbs() {
   if [[ -z "${measurement}" ]]; then >&2 echo "Measurement is invalid"; return 1; fi
 
   # Get encryption key from docker image label
-  enc_key=$(esudo docker inspect quay.io/kata-containers/encrypted-image-tests:encrypted \
+  enc_key=$(esudo docker inspect quay.io/kata-containers/encrypted-image-tests:unencrypted \
```
fidencio, line 74: Do we actually need to get the encryption key, add it to the DB, if we're dealing with an unencrypted image?

I mean, can't we just get rid of these steps if the tests are *not* using an encrypted image?

fidencio (commented):
@fitzthum, I don't fully understand the motivation of the PR.   Aren't the current tests passing with encrypted images?  Is this change specific to SEV-ES / SEV-SNP?

fidencio:
> @fitzthum, I don't fully understand the motivation of the PR. Aren't the current tests passing with encrypted images? Is this change specific to SEV-ES / SEV-SNP?

Oh, I checked the issue now and it points to an error.  This error was not seen before because those tests were not running?  Or because there was a change that caused those tests to fail?

functional/sev/run.sh
fitzthum: Yeah slightly confusing but this test is for an encrypted image. This docker pull doesn't affect the image that we actually pull inside the guest. We just download the image because it has the ssh key and the encryption key in the labels and we need those for the test.

fidencio (approved):
lgtm, assuming that the SEV tests are green!
thanks @fitzthum!

fitzthum:
@fidencio

> Oh, I checked the issue now and it points to an error. This error was not seen before because those tests were not running? Or because there was a change that caused those tests to fail?

Yeah this is fallout from the encrypted test image being overwritten with an unencrypted image. We could get the labels from the unencrypted image, but once we switched to the encrypted image this stuff broke. There might be a few other related fixups.

fidencio:
> @fidencio
> 
> > Oh, I checked the issue now and it points to an error. This error was not seen before because those tests were not running? Or because there was a change that caused those tests to fail?
> 
> Yeah this is fallout from the encrypted test image being overwritten with an unencrypted image. We could get the labels from the unencrypted image, but once we switched to the encrypted image this stuff broke. There might be a few other related fixups.

Yeah, there may be :-)

```
16:33:38 Events:
16:33:38   Type     Reason     Age               From               Message
16:33:38   ----     ------     ----              ----               -------
16:33:38   Normal   Scheduled  21s               default-scheduler  Successfully assigned default/encrypted-image-tests-6dcdbcbc84-6d8j7 to amd-coco-ci-ubuntu2004-001
16:33:38   Warning  Failed     16s               kubelet            Failed to pull image "quay.io/kata-containers/encrypted-image-tests:encrypted": rpc error: code = Unknown desc = failed to pull and unpack image "quay.io/kata-containers/encrypted-image-tests:encrypted": failed to unpack image on snapshotter overlayfs: failed to extract layer sha256:c29210396b2d76e99506b81bd93e95f67c7920987d6ea1b2675aca3f30a4137b: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount4045135272: archive/tar: invalid tar header: unknown
16:33:38   Normal   BackOff    15s               kubelet            Back-off pulling image "quay.io/kata-containers/encrypted-image-tests:encrypted"
16:33:38   Warning  Failed     15s               kubelet            Error: ImagePullBackOff
16:33:38   Normal   Pulling    4s (x2 over 17s)  kubelet            Pulling image "quay.io/kata-containers/encrypted-image-tests:encrypted"
16:33:38   Warning  Failed     4s (x2 over 16s)  kubelet            Error: ErrImagePull
16:33:38   Warning  Failed     4s                kubelet            Failed to pull image "quay.io/kata-containers/encrypted-image-tests:encrypted": rpc error: code = Unknown desc = failed to pull and unpack image "quay.io/kata-containers/encrypted-image-tests:encrypted": failed to unpack image on snapshotter overlayfs: failed to extract layer sha256:c29210396b2d76e99506b81bd93e95f67c7920987d6ea1b2675aca3f30a4137b: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount178191664: archive/tar: invalid tar header: unknown
```

fitzthum:
So this appears to be a somewhat more serious issue involving an incompatibility between image-rs and quay with encrypted images. It seems like quay only supports encrypted images with schema version 1, but image-rs does not support schema version 1. I'm still trying to confirm all this but it seems like image-rs was never tested with encrypted quay images. I am not yet sure about the best way to resolve, bu there are a couple of options.

fitzthum:
> One easy option would be to store the image in dockerhub, WDYT?

Yeah in the short term we will probably need to switch to some other container registry. Dockerhub seems to work but there is a concern about rate limiting. I am looking around a bit to see if there is a better option. Depending on how that goes I may add a temporary commit to this PR that switches to dockerhub just so we can see if here are more issues.

fitzthum:
Ok, I have some test images in ghcr that seem promising. I added a commit to point to those images instead of the quay ones. I think there is no pull limit for ghcr. Currently the images are just attached to my github account. I can probably move them to the CoCo org, but let's see if this actually fixes anything first.

fitzthum:
Ok, it looks like we have another issue after switching to ghcr. It seems ocicrypt is trying to use the default keyfile rather than trying to use the keyprovider. I am looking into it.

fitzthum:
Looks like it's https://github.com/kata-containers/tests/issues/5367. I am working on a fix.

fitzthum:
Ok, I think I have finally gotten to the bottom of these issues. Ignore my sketchy fixups for now. There are a few more problems that I will describe on Monday.

fitzthum:
Ok, operator tests should be fixed now. Although I have noticed that sometimes the tests run before the runtimeclasses are ready. We might need to extend a timeout somewhere. 

Not sure about the kata CI yet, but we're making some progress there too.

Issue #5362, opened 2023-01-05
Kata / cri-o job is broken


littlejawa:
# Description of problem

The CI job running on cri-o repository to test kata with cri-o is broken.
The root cause is an update of CNI, causing a renaming of the config files.

See https://github.com/cri-o/cri-o/issues/6481

PR #5266, opened 2022-11-17
Update CRI-O CI to be using the release-1.26 branch

fidencio:
cri-o: Remove hack to enforce golang 1.18+

This is not needed anymore as we've bumped the golang version in our CI
to use golang 1.19.2.

---

cri-o: Update to the release-1.26 when using main

Let's make sure we're testing against the latest commits from the CRI-O
side.

Fixes: #5265

PR #5363, opened 2023-01-05
Fix CNI config file extension for cri-o testing.

littlejawa:
The extension for CNI config file has been modified in recent versions of cri-o, from "conf" to "conflist".

Fixes: #5362

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -332,7 +332,10 @@ main() {
 		if [ ! -d "${cni_directory}" ]; then
 			sudo mkdir -p "${cni_directory}"
 		fi
-		sudo cp "${crio_repository_path}/contrib/cni/10-crio-bridge.conf" "${cni_directory}"
+		# The config file for cni in cri-o v1.26 and higher has the extension "conflist"
+		# Using a wildcard here to keep support for both names.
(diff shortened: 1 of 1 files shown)


c3d (approved):
Looks good enough to me

integration/kubernetes/init.sh
```diff
@@ -332,7 +332,10 @@ main() {
 		if [ ! -d "${cni_directory}" ]; then
 			sudo mkdir -p "${cni_directory}"
 		fi
-		sudo cp "${crio_repository_path}/contrib/cni/10-crio-bridge.conf" "${cni_directory}"
+		# The config file for cni in cri-o in newer versions has the extension "conflist"
```
amarlearning, line 335: Instead of saying newer version, can we specify the exact version i.e. `1.26`

littlejawa: Thanks. Fixed

fidencio (approved):
lgtm, thanks @littlejawa!

fidencio:
@littlejawa, you need to also update the CRI-O version we're using, which means also update the k8s version we're using.

littlejawa:
>  you need to also update the CRI-O version we're using, which means also update the k8s version we're using.

@fidencio - see https://github.com/kata-containers/kata-containers/pull/5989, hope I'm getting it right.

How about #5266 - we should merge it too?

PR #5386, opened 2023-01-13
SEV: Remove redundant containerd installation

fitzthum:
Let's see if removing the redundant containerd installation breaks stuff.

@ryansavino @fidencio @stevenhorsman


fidencio (approved):
lgtm, thanks @fitzthum and @ryansavino!

Issue #5414, opened 2023-01-27
sev.bats - Use RUNTIMECLASS instead of OPERATOR env variable


ryansavino:
# Description of problem

sev.bats uses an environment variable OPERATOR=true to determine how to set the runtime class. This was done to distinguish between `kata-qemu` for the general CI and `kata-qemu-sev` for the operator CI.

Per @wainersm's guidance, this should be done instead using the setting of the variable RUNTIMECLASS.

PR #5416, opened 2023-01-27
sev: Use RUNTIMECLASS instead of OPERATOR env variable

ryansavino:
Change sev.bats to use env variable RUNTIMECLASS instead of OPERATOR=true to determine how to set the runtime class.

Fixes: #5414

Signed-Off-By: Ryan Savino <ryan.savino@amd.com>


wainersm (approved):
Thanks @ryansavino !

fitzthum (approved):
Yeah this seems much more direct. Fundamentally we are just trying to set the right runtime class and we already have a variable for that.

amshinde (approved):

Issue #5429, opened 2023-02-02
CC: Enable e2e tests for IBM Z secure execution (SE)


BbolroC:
**Which feature do you think can be improved?**

As a follow-up for #5282 where the kubernetes e2e test was instrumented for s390x, we need to run the test on a IBM's TEE environment (secure execution).

PR #5430, opened 2023-02-02
CC: Enable e2e tests for IBM Z secure execution (SE)

BbolroC:
This is to run e2e tests differently based on TEE support for s390x. If the environment is detected, a secure image is built. Otherwise, kernel+rootfs is used for the test.

Fixes: #5429

### DIFF
--- a/integration/confidential/lib.sh
+++ b/integration/confidential/lib.sh
@@ -55,6 +55,13 @@ switch_image_service_offload() {
 #			      `kata-runtime env` and export its value.
 #
 switch_measured_rootfs_verity_scheme() {
+	# In the case of IBM Secure Execution for Linux, the ultravisor (trusted IBM zSystems CPU firmware),
+	# before bootstrapping, performs integrity checks by the measurements in the integrity-protected
+	# Secure Execution header which are calculated while a secure image is built based on kernel, cmdline, and initrd.
+	if [ "${TEE_TYPE}" == "se" ] && [ "$1" == "dm-verity" ]; then
+		skip "test for IBM zSystems & LinuxONE SE"
+	fi
+
 	# Load the RUNTIME_CONFIG_PATH variable.
 	load_runtime_config_path
 
@@ -85,6 +92,12 @@ add_kernel_params() {
 
 	sudo sed -i -e 's#^\(kernel_params\) = "\(.*\)"#\1 = "\2 '"$params"'"#g' \
 		"$RUNTIME_CONFIG_PATH"
+
+	if [ "${TEE_TYPE}" = "se" ]; then
+		local kernel_params=$(sed -n -e 's#^kernel_params = "\(.*\)"#\1#gp' \
+			"$RUNTIME_CONFIG_PATH")
+		build_se_image "${kernel_params}"
+	fi
 }
 
 # Get the 'kernel_params' property on kata's configuration.toml
--- a/lib/common.bash
+++ b/lib/common.bash
@@ -105,6 +105,24 @@ waitForProcess() {
 	return 1
 }
 
+# Build a IBM zSystems & LinuxONE secure execution image
+#
+# Parameters:
+#	$1		- existing kernel parameters to be wrapped in parmfile
+#
+build_se_image() {
+	source "${_script_dir}/../.ci/lib.sh"
+	[ -d "${katacontainers_repo_dir}" ] || clone_katacontainers_repo
+	source "${katacontainers_repo_dir}/ci/lib.sh"
+
+	export HKD_PATH=$(dirname ${CI_HKD_PATH})
+
+	kernel_params="${1:-}"
+	local base_dir="$(dirname $(sudo -E PATH=$PATH kata-runtime kata-env --json | \
+	jq -r .Kernel.Path))"
+	build_secure_image "${kernel_params}" "${base_dir}" "${base_dir}"
+}
+
 # Check if the $1 argument is the name of a 'known'
 # Kata runtime. Of course, the end user can choose any name they
 # want in reality, but this function knows the names of the default
@@ -296,13 +314,17 @@ cp_to_guest_img() {
 	else
 		local initrd_path="$(sudo -E PATH=$PATH kata-runtime kata-env --json | \
 			jq -r .Initrd.Path)"
+		if [ "${TEE_TYPE}" = "se" ]; then
+			initrd_path="$(dirname $(sudo -E PATH=$PATH kata-runtime kata-env --json | \
+			jq -r .Kernel.Path))/kata-containers-initrd.img"
+		fi
 		if [ ! -f "$initrd_path" ]; then
 			echo "Guest initrd and image not found"
 			rm -rf "$rootfs_dir"
 			return 1
 		fi
 
-                if ! cat "${initrd_path}" | cpio --extract --preserve-modification-time \
+                if ! zcat "${initrd_path}" | cpio --extract --preserve-modification-time \
                         --make-directories --directory="${rootfs_dir}"; then
                         echo "Failed to uncompress the image file: $initrd_path"
                         rm -rf "$rootfs_dir"
@@ -337,6 +359,15 @@ cp_to_guest_img() {
 			rm -rf "$rootfs_dir"
 			return 1
 		fi
+	fi
+
+	if [ "${TEE_TYPE}" = "se" ]; then
+		if [ -z "$RUNTIME_CONFIG_PATH" ]; then
+			extract_kata_env
+		fi
+		local kernel_params=$(sed -n -e 's#^kernel_params = "\(.*\)"#\1#gp' \
+			"$RUNTIME_CONFIG_PATH")
+		build_se_image "${kernel_params}"
 	fi
 
 	rm -rf "$rootfs_dir"
--- a/.ci/ci_job_flags.sh
+++ b/.ci/ci_job_flags.sh
@@ -158,7 +158,9 @@ case "${CI_JOB}" in
 		export KATA_BUILD_QEMU_TYPE="tdx"
 		export AA_KBC="eaa_kbc"
 	elif [[ "${CI_JOB}" =~ _SE_ ]]; then
-		export TEE_TYPE="se"
+		if grep -q 'prot_virt=1' /proc/cmdline && grep -Eq '^facilities.* 158 .*' /proc/cpuinfo; then
+			export TEE_TYPE="se"
+		fi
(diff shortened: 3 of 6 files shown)


hbrueckner (approved):
LGTM! Thanks a lot for your work @BbolroC!

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -122,6 +122,11 @@ assert_logs_contain() {
 	assert_pod_fail "$pod_config"
 }
 
+@test "$test_tag Test can pull a unencrypted signed image from a protected registry" {
```
stevenhorsman, line 125: What was the reason for moving this test out of interest?

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -181,6 +181,15 @@ assert_logs_contain() {
 	assert_logs_contain 'Validate image failed: \[PublicKeyVerifier { key: CosignVerificationKey'
 }
 
+# This test should be run first before the following 2 tests
```
stevenhorsman, line 184: I don't like relying on the order of tests for them to pass as it is fragile. I thought that the setup run between tests should have reset the state anyway? If not we need to look at fixing this.

lib/common.bash
```diff
@@ -105,6 +105,61 @@ waitForProcess() {
 	return 1
 }
 
+# Check if the 1st argument version is greater than and equal to 2nd one
```
stevenhorsman, line 108: Can we refactor there methods to call the version in kata-containers (or vice versa) rather than duplicating them?

integration/kubernetes/confidential/agent_image.bats
BbolroC: Yeah, as far as I figured it out, [setup()](https://github.com/kata-containers/tests/blob/f07f559f7ea93798458865bbeeca8173e00f6f47/integration/kubernetes/confidential/agent_image.bats#L75-L94) resets a configuration file, but a root file system for an image/initrd.  In [cp_to_guest_img()](https://github.com/kata-containers/tests/blob/f07f559f7ea93798458865bbeeca8173e00f6f47/lib/common.bash#L274-L343), the image/initrd is modified and used for test, but it is not restored during the setup(), which leads to dependency between the tests. I am not sure if this was intentional. But you are right. We could fix this in a separate PR first.

integration/kubernetes/confidential/agent_image.bats
BbolroC: Reasoning is specified below. :wink:

lib/common.bash
BbolroC: yep, I will refactor them. Thanks for pointing it out. :+1:

integration/kubernetes/confidential/agent_image.bats
stevenhorsman: I guess that leads me to the question of how they worked before you moved them? Is the difference due to the agent-config.toml now being added in the rootfs.sh rather than us doing that during the test time?

stevenhorsman: Interestingly I've just spotted that the crictl test have the follow in their teardown:
https://github.com/kata-containers/tests/blob/f07f559f7ea93798458865bbeeca8173e00f6f47/integration/containerd/confidential/agent_api.bats#L79-L81
so maybe we can just add this to the Kubernetes teardown as well to resolve the problem.
@wainersm - can you recall if there is a difference in the way that the guest image setup works for crictl and Kubernetes tests?

BbolroC:
@stevenhorsman all tests are passed with the original order after having `build_se_image()` called when kernel parameters are configured during `setup()` and `teardown()`. I overlooked an update of the secure image with a new parmfile (including kernel parameters) and only focused on change of rootfs. This explains why we experienced such a weird phenomenon (2 tests cases we looked into only modify kernel parameters, but rootfs. So we were not able to see what we had expected.)

integration/confidential/lib.sh
```diff
@@ -55,6 +55,13 @@ switch_image_service_offload() {
 #			      `kata-runtime env` and export its value.
 #
 switch_measured_rootfs_verity_scheme() {
+	# In the case of IBM Secure Execution for Linux, the ultravisor (trusted IBM zSystems CPU firmware),
```
stevenhorsman, line 58: I'm a bit torn on the check being in `switch_measured_rootfs_verity_scheme`. On the onehand having it in just a single place is good, but I don't know if it would be obvious to anyone adding tests to `agent_image.bats`?

I think it is fine as is, just thinking out loud a bit!

stevenhorsman (approved):
I have a few minor comments/questions, but generally I think it looks good. I just wanted to check what the next steps are:
My understanding is that we don't have a public SE-enabled machine yet, if this is correct, could you add the results of your local run to the PR just so we can see the output? Also is the plan for this to be run behind the IBM firewall but triggered on PRs (like the tdx tests), or just manually ad-hoc until we have a box we can add to jenkins?

Thanks!

integration/confidential/lib.sh
BbolroC: I agree with your thought. (I was thinking it actually :wink:) It would be something to be relocated based on others' feedback. Let's see what others think about this placement.

lib/common.bash
```diff
@@ -10,6 +10,13 @@
 _script_dir=$(dirname "${BASH_SOURCE[0]}")
 source "${_script_dir}/error.sh"
 
+# This is needed for version_greater_than_equal() for tests,
+# which is also used for kata-containers during kata build
+if [ -d "${_script_dir}/../../kata-containers" ]; then
```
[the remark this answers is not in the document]
BbolroC: I am totally for your opinion. I will refactor the code as advised. Thanks a lot!

lib/common.bash
```diff
@@ -339,6 +385,15 @@ cp_to_guest_img() {
...
 
+	if [ "${TEE_TYPE}" = "se" ]; then
+		if [ -z "$RUNTIME_CONFIG_PATH" ]; then
+			extract_kata_env
+		fi
+		local kernel_params=$(sed -n -e 's#^kernel_params = "\(.*\)"#\1#gp' \
+			"$RUNTIME_CONFIG_PATH")
+		build_se_image "${kernel_params}"
+	fi
+
```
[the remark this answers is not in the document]
BbolroC: yes, as far as I know, that is the only way to get the change of rootfs reflected on the se image. But I would like to listen to @hbrueckner 's opinion as well.

hbrueckner:
Hi @stevenhorsman 

> I have a few minor comments/questions, but generally I think it looks good. I just wanted to check what the next steps are: My understanding is that we don't have a public SE-enabled machine yet, if this is correct, could you add the results of your local run to the PR just so we can see the output? 

Providing a public SE-enabled machine is still work in progress. Until it will be available, would it be an option to share results from internal testing to move forward and get it merged? Wdyt?

stevenhorsman:
> Providing a public SE-enabled machine is still work in progress. Until it will be available, would it be an option to share results from internal testing to move forward and get it merged? Wdyt?

As long as the changes don't have any negative impact on the other platforms (so we have green tests as usual) that sounds okay to me

lib/common.bash
```diff
@@ -105,6 +112,41 @@ waitForProcess() {
 	return 1
 }
 
+# Build a IBM zSystems & LinuxONE secure execution image
+#
+# Parameters:
+#	$1		- existing kernel parameters to be wrapped in parmfile
+#
+build_se_image() {
```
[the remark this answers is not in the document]
BbolroC: Yep, I will make them as little different as possible. Thanks.

GabyCT (approved):

BbolroC:
@fidencio @stevenhorsman I've created a Jenkins job which fetches and dumps the log of the internal test. The test runs on a daily basis and the job will be updated accordingly. Here is a result of the first run:
http://jenkins.katacontainers.io/job/kata-containers-CCv0-ubuntu-20.04-s390x-SE-daily/1/consoleText

stevenhorsman:
> @fidencio @stevenhorsman I've created a Jenkins job which fetches and dumps log of the internal test. The test runs on a daily basis and the job will be triggered accordingly. Here is a result of the first run: http://jenkins.katacontainers.io/job/kata-containers-CCv0-ubuntu-20.04-s390x-SE-daily/1/consoleText

This is great to see. A couple of questions:
- I assume it is running the latest 'CCv0' branch each night?
- If there is a failure on the internal tests, with the results be marked red to reflect that?
Thanks!

stevenhorsman:
> Yeah, it is totally viable, but there is an internal issue (flaky) on the test `Test can pull an encrypted image inside the guest with decryption key` regarding:
> 
> ```
> failed to pull manifest Registry error: url https://registry-1.docker.io/v2/katadocker/ccv0-ssh/manifests/s390x,
> envelope: OCI API errors: [OCI API error: You have reached your pull rate limit. You may increase the limit by authenticating
> ```
> 
> It looks there is a company policy for the `docker.io` plan, which it is a bit challenging for me to resolve. It would be great if we could switch the registry to `quay.io`.

I know Intel have hit this as well and I think also AMD, so it's something that is being discussed. There is an issue with quay.io and skopeo having a difference of opinion on schema version IIRC, so I know in the past when Jakob and I tried to push the encrypted image there it didn't work and I think Tobin/Ryan found the same more recently. I think Fabiano might have reached out to Hendrik to see if use using ICR would be possible, but I don't know if that went anywhere.

I think when the KBS URI changes that Ding has proposed go in happen then all the existing images we use will be invalidated, so I hope at that point we can ensure we create proper multiarch images into a container registry that is reliably, but we'll see.

This problem shouldn't block this PR anyway.

BbolroC:
Why don't we merge this unless the test failures for tdx have something to do with this change?

stevenhorsman:
I'm assuming that this has a dependency on https://github.com/kata-containers/kata-containers/pull/6207, so we need to get that merged first?

fidencio:
Also network related issues here. :-/

fidencio:
QEMU TDX CI has passed but didn't report back: http://10.112.240.208:8080/job/tests-CCv0-qemu-TDX-CRI_CONTAINERD_K8S-PR/387/

fidencio:
Cloud Hypervisor CI has passed but didn't report back: http://10.112.240.208:8080/job/tests-CCv0-cloud-hypervisor-TDX-CRI_CONTAINERD_K8S-PR/384/

Issue #5448, opened 2023-02-14
SEV: consolidate sev.bats


UnmeshDeodhar:
Use as many generic functionalities as possible and keep SEV specific functionalities in sev.bats.

PR #5449, opened 2023-02-14
confidential: consolidate sev.bats

UnmeshDeodhar:
Refactor sev.bats Move generic functionality to lib.sh and keep SEV specific changes in sev.bats

Fixes: #5448

Signed-off-by: Unmesh Deodhar <udeodhar@amd.com>

Update sev.bats

### DIFF
--- a/integration/kubernetes/confidential/sev.bats
+++ b/integration/kubernetes/confidential/sev.bats
@@ -73,22 +73,6 @@ configure_containerd() {
   fi
 }
 
-# Wait until the pod is 'Ready'. Fail if it hits the timeout.
-kubernetes_wait_for_pod_ready_state() {
-  local pod_name="${1}"
-  local wait_time="${2:-60}"
-
-  esudo kubectl wait --for=condition=ready pod/$pod_name --timeout=${wait_time}s
-}
-
-# Wait until the pod is 'Deleted'. Fail if it hits the timeout.
-kubernetes_wait_for_pod_delete_state() {
-  local pod_name="${1}"
-  local wait_time="${2:-60}"
-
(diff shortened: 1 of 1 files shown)


UnmeshDeodhar:
Requesting review: @stevenhorsman @wainersm @fidencio

integration/kubernetes/confidential/lib.sh
```diff
@@ -31,6 +31,35 @@ kubernetes_wait_cc_pod_be_ready() {
...
+#
+# Parameters:
+#	$1 - the sandbox ID
+#	$2 - wait time in seconds. Defaults to 120. (optional)
+#
+kubernetes_wait_cc_pod_be_deleted() {
+  local pod_name="${1}"
+  local wait_time="${2:-120}"
+
+  esudo kubectl wait --for=delete pod/$pod_name --timeout=${wait_time}s
```
fidencio, line 44: There's already some functions in `lib/common.bash`, such as `WaitForProcess` that could be used instead of this one.

If we want to add a function to specifically wait for a pod to be deleted, I'd suggest taking the bullet to add this everywhere needed, and also do the same for the case where we wait for a pod to be ready.

fidencio (commented):
@UnmeshDeodhar, I'm not sure if this is exactly what folks had in mind when we asked for consolidating the SEV tests.

What we were mostly looking for is having the same set of tests run on different arches (the common ones shared between non TEEs and TDX) also being executed with SEV, and then start moving the bats from the SEV tests to `integration/kubernetes` and running them considiotnally.

That would help to close the gap between the SEV tests and all the other ones we currently have.

Issue #5467, opened 2023-02-20
arm64/rom: should enable pflash after install it


jongwu:
Once we need install rom for arm64 we should also enable it in configuration.

PR #5468, opened 2023-02-20
uefi/arm64: enable pflash after install it.

jongwu:
Enable pflash in configuration after installation. Also, as nvdimm support for qemu on arm need ACPI support, uefi installation should be set to default when install runtime.

Fixes: #5467

### DIFF
--- a/.ci/aarch64/install_rom_aarch64.sh
+++ b/.ci/aarch64/install_rom_aarch64.sh
@@ -123,6 +123,19 @@ clean_up()
 	sudo rm -rf "${EDK2_WORKSPACE}"
 }
 
+enable_pflash_in_config()
+{
+	runtime_config_prefix=("/etc" "/usr/share/defaults" "/opt/kata/defaults")
+	for con_pre in "${runtime_config_prefix[@]}"
+	do
+		config_path="${con_pre}/kata-containers/configuration.toml"
+		[ -f "${config_path}" ] || continue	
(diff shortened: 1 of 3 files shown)


fidencio:
@jongwu, out of curiosity, how could we make sure those bits are properly set up as part of kata-deploy?

jongwu:
Hi @fidencio -, kata-deploy calls install_runtime.sh and then install_rom_aarch64.sh execute accordingly.  Without this patch rom may not be installed as the ENABLE_ARM64_UEFI is not set.

fidencio:
> Hi @fidencio -, kata-deploy calls install_runtime.sh and then install_rom_aarch64.sh execute accordingly. Without this patch rom may not be installed as the ENABLE_ARM64_UEFI is not set.

Sorry, my question wasn't clear.  Let me re-phrase it.

If you go to the `kata-containers/kata-containers` repo, and call `make kata-tarball`, we'll get the content that is released and that's part of kata-deploy.  Is there something on that workflow that would ensure pflash is installed and enabled?

jongwu:
@fidencio -, Oh, sorry about misunderstanding your words.

I think "make kata-tarball" won't build qemu UEFI on Arm, as the install_rom script locates under tests repo. So, should we add install_rom_aarch64 script under kata-containers/ then build rom in qemu-tarball?

fidencio:
> I think "make kata-tarball" won't build qemu UEFI on Arm, as the install_rom script locates under tests repo. So, should we add install_rom_aarch64 script under kata-containers/ then build rom in qemu-tarball?

I think that would be a good addition to make, yes.

fidencio (approved):
lgtm, thanks @jongwu!

jongwu:
TAL this? @GabyCT , thanks!

GabyCT (approved):

PR #5477, opened 2023-02-21
tests: Link to instead of copying the default config

fidencio:
By doing this we can ensure that even if the containerd or runtime configuration points to a specific path, changes in the default config will be reflected in the specific path.

### DIFF
--- a/integration/nydus/nydus_tests.sh
+++ b/integration/nydus/nydus_tests.sh
@@ -17,6 +17,7 @@ source "${dir_path}/../../.ci/lib.sh"
 source "/etc/os-release" || source "/usr/lib/os-release"
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 
+default_config=""
 need_restore_kata_config=false
 kata_config_backup="/tmp/kata-configuration.toml"
 SYSCONFIG_FILE="/etc/kata-containers/configuration.toml"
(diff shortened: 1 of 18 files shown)


integration/nydus/nydus_tests.sh
```diff
@@ -86,13 +86,13 @@ function config_kata() {
 	sudo mkdir -p /etc/kata-containers
 	if [ -f "$SYSCONFIG_FILE" ]; then
 		need_restore_kata_config=true
-		sudo cp -a "${SYSCONFIG_FILE}" "${kata_config_backup}"
```
wainersm, line 89: @fidencio here it saves the file to a backup that is restored later. Either you preserve this backup mechanism or change the logic to switch the link to the default config.

fidencio: Solved, thanks!

wainersm: It is still different: sudo cp -a" ${SYSCONFIG_FILE}" "${kata_config_backup}"

wainersm (approved):
@fidencio you deserve entering the heaven after changing so many seds

fidencio:
> @fidencio you deserve entering the heaven after changing so many seds

sed++ for helping me to easily change the seds. :-)

integration/nydus/nydus_tests.sh
fidencio: Sorry, my bad, fixed now.

studychao (approved):
LGTM！

fidencio:
Let's drop it for now, as the reason behind doing -- having runtime-rs properly loading ConfigPath -- is already solved.

Issue #5486, opened 2023-02-24
dragonball: skip_footloose not pass on arm64


jongwu:
skip_footloose test can't pass on arm-dragonball

> 15:33:06 1..1
15:33:06 not ok 1 Footloose pod
15:34:37 # (in test file k8s-footloose.bats, line 39)
15:34:37 #   `kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"' failed

link: http://jenkins.katacontainers.io/job/kata-containers-dragonball-ubuntu-Arm-20.04-PR/478/console

PR #5487, opened 2023-02-24
k8s/arm64: skip footloose test for DB

jongwu:
footloose test based on dragonball can't pass on Arm. Skip for now until we fix it.

Fixes: #5486


GabyCT (approved):

studychao (approved):
Thanks!

Issue #5492, opened 2023-02-26
runtime-rs : exclude footloose k8s CI in aarch64


studychao:
Currently, fooloose k8s CI is only solved in x86_64 runtime-rs.
We should exclude it for now in aarch64 in order not to block arm
dragonball ci.

PR #5493, opened 2023-02-26
runtime-rs : exclude footloose k8s CI in aarch64

studychao:
Currently, fooloose k8s CI is only solved in x86_64 runtime-rs. We should exclude it for now in aarch64 in order not to block arm dragonball ci.

fixes: #5492


jongwu:
Hi @studychao -,  I have raised a PR to skip footloose on arm64, see https://github.com/kata-containers/tests/pull/5487

studychao:
Oh, thanks! Your modification makes more sense, I'll close mine.

PR #5496, opened 2023-02-27
metrics: Use sudo with ctr in common.bash

### LINKED ISSUE #5495: update metrics lib common.sh to use sudo with ctr

GabyCT:
This PR uses sudo with ctr in common.bash as part of consistency as in other functions that are using ctr or docker.

Fixes #5495

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -70,7 +70,7 @@ check_images()
 	local img req_images=( "$@" )
 	for img in "${req_images[@]}"; do
 		echo "ctr pull'ing: $img"
-		if ! ctr image pull "$img"; then
+		if ! sudo "${CTR_EXE}" image pull "$img"; then
 			die "Failed to pull image $img"
 		fi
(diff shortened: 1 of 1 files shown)


metrics/lib/common.bash
```diff
@@ -70,7 +70,7 @@ check_images()
 	local img req_images=( "$@" )
 	for img in "${req_images[@]}"; do
 		echo "ctr pull'ing: $img"
-		if ! ctr image pull "$img"; then
+		if ! sudo ctr image pull "$img"; then
```
dborquez, line 73: For uniformity, could you use $CTR_EXE instead of the command itself?

dborquez (changes requested):

metrics/lib/common.bash
GabyCT: @dborquez changes applied

dborquez (approved):
thank you @GabyCT

cmaf (approved):

Issue #5502, opened 2023-03-01
Tests not cleaning up stale artifacts


UnmeshDeodhar:
Tests are not cleaning up stale coco artifacts in /opt/confidential-containers/share/kata-containers

PR #5503, opened 2023-03-01
Remove stale coco artifacts

UnmeshDeodhar:
Remove coco artifacts.

Fixes: #5502

Signed-of-by: Unmesh Deodhar <udeodhar@amd.com>

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -464,6 +464,28 @@ gen_clean_arch() {
 	# make sure cargo and rustup home dir are removed
 	info "remove cargo and rustup home dir"
 	sudo rm -rf ~/.cargo ~/.rustup
+
+	# make sure to clean up linked binaries
+	unlink /opt/confidential-containers/share/ovmf
+	sudo rm -f /usr/share/ovmf
+
+	unlink /opt/confidential-containers/share/td-shim
+	sudo rm -f /usr/share/td-shim
+
+	unlink /opt/confidential-containers/share/tdvf
+	sudo rm -f /usr/share/tdvf
+
+	unlink /opt/confidential-containers/bin/qemu-system-x86_64
+	sudo rm -f /usr/bin/qemu-system-x86_64
+
+	unlink /opt/confidential-containers/bin/qemu-system-x86_64-tdx
+	sudo rm -f /usr/bin/qemu-system-x86_64-tdx
+
+	unlink /opt/kata/libexec/virtiofsd
+	sudo rm -rf /usr/libexec/virtiofsd
+
+	info "remove all coco artifacts"
+	sudo rm -rf /opt/confidential-containers
 }
 
 check_git_version() {


fidencio:
This works well, but I think we should also remove some links we create during the installation.
For instance, if you go to the install scripts, you'll notice we end up linking to binaries / ovmf / and whatnot to /usr/local/bin.  Please, take a look at those scripts and also clean up the scripts we create there.

UnmeshDeodhar:
@fidencio I have added the changes that delete and unlink binaries from /opt/confidential-containers please let me know if this looks okay.

.ci/lib.sh
```diff
@@ -464,6 +464,23 @@ gen_clean_arch() {
 	# make sure cargo and rustup home dir are removed
 	info "remove cargo and rustup home dir"
 	sudo rm -rf ~/.cargo ~/.rustup
+
+	# make sure coco artifacts are removed
+	info "remove coco artifacts"
+	sudo rm -rf /opt/confidential-containers
```
fidencio, line 470: So, here you removed everything from `/opt/confidential-containers` ...

.ci/lib.sh
```diff
@@ -464,6 +464,23 @@ gen_clean_arch() {
...
+
+	# make sure to clean up binaries
+	unlink /opt/confidential-containers/share/ovmf
```
fidencio, line 473: ... and then here you're trying to unlink a dir that was just removed in a few lines above.

As this pattern repeats in all the removals, so I'd suggest to move the `sudo rm -rf /opt/confidential-containers` to be the last part of the cleanup.

.ci/lib.sh
```diff
@@ -464,6 +464,23 @@ gen_clean_arch() {
...
+	sudo rm -f /usr/share/ovmf
+
+	unlink /opt/confidential-containers/share/td-shim
+	sudo rm -f /usr/share/td-shim
+
+	unlink /opt/confidential-containers/share/tdvf
+	sudo rm -f /usr/share/tdvf
+
+	unlink /opt/confidential-containers/bin/qemu-system-x86_64
+	sudo rm -f /usr/bin/qemu-system-x86_64
```
fidencio, line 483: You're missing:
*  `./opt/confidential-containers/bin/qemu-system-x86_64-tdx` which seems to be linked as `/usr/bin/qemu-system-x86_64-tdx`
* `/opt/kata/libexec/virtiofsd`, which is linked as `/usr/libexec/virtiofsd`
`

fidencio (changes requested):
I've left a few comments.

fidencio (approved):
lgtm, thanks @UnmeshDeodhar!

fidencio:
Cloud Hypervisor TDX is green but didn't report back: http://10.112.240.208:8080/job/tests-CCv0-cloud-hypervisor-TDX-CRI_CONTAINERD_K8S-PR/374/console

PR #5518, opened 2023-03-09
ci: replace $TRAVIS with $CI in setup

jepio:
Nothing else relies on `TRAVIS`, `CI` can be checked for the same purpose and is defined by gh-actions.

See: kata-containers/kata-containers#3544


fidencio (approved):
lgtm, thanks @jepio!

PR #5525, opened 2023-03-13
metrics: Fix memory inside container test

### LINKED ISSUE #5524: metrics: fix memory inside container test

GabyCT:
This PR replaces ctr with CTR_EXE which is defined in the metrics lib common.bash as well as it uses sudo to have uniformity across the metrics scripts.

Fixes #5524


snir911 (approved):
LGTM, thanks

cmaf (approved):

PR #5541, opened 2023-03-21
crio: use go 1.20 for the kata/crio CI job

littlejawa:
cri-o has started to use go 1.20 some times ago.
For this job, running from the cri-o repository, we need to use the same version.

NOTE: there is no failure on the main branch *yet*, but I've seen some PR fail on this job because of this, when they use modules requiring this version of go. Whenever those PR get merged on cri-o/main, we will have an issue.

Fixes: #5559


wainersm (approved):
Thanks @littlejawa !

littlejawa:
Closing: The test running on crio repository has been moved to the crio framework. This change is not needed anymore.

Issue #5544, opened 2023-03-21
CI: kata/cri-o job is failing when pulling stress-kata image


littlejawa:
# Description of problem

The cri-o CI job entrypoint is defined in https://github.com/kata-containers/tests/blob/main/.ci/ci_crio_entry_point.sh
It runs the "EXTERNAL_CRIO" job from run.sh.

As part of this job, we run the "make kubernetes" command, which fails during the step "Running within memory constraints"
The failure is caused by an ImagePullBackoff error when trying to get the stress-kata:latest image

```
#   Type     Reason     Age                From               Message
#   ----     ------     ----               ----               -------
#   Normal   Scheduled  90s                default-scheduler  Successfully assigned default/memory-test to fedora35cloudd6d590
#   Normal   Pulling    51s (x3 over 88s)  kubelet            Pulling image "localhost:5000/stress-kata:latest"
#   Warning  Failed     51s (x3 over 88s)  kubelet            Failed to pull image "localhost:5000/stress-kata:latest": rpc error: code = Unknown desc = pinging container registry localhost:5000: Get "http://localhost:5000/v2/": dial tcp [::1]:5000: connect: connection refused
#   Warning  Failed     51s (x3 over 88s)  kubelet            Error: ErrImagePull
#   Normal   BackOff    11s (x5 over 87s)  kubelet            Back-off pulling image "localhost:5000/stress-kata:latest"
#   Warning  Failed     11s (x5 over 87s)  kubelet            Error: ImagePullBackOff

```

Comparing the failing logs with an older, successful build, I can see that the job used to create and pull an image named "localhost/stress-kata:latest" (without the ":5000").
Also the logs show that the stress image was built with podman when it worked, and is now built with docker.

Since this issue has started with the merge of https://github.com/kata-containers/tests/pull/5409, I suspect the change stems from the removal of the line in lib.sh that used to set the "USE_PODMAN=true" flag just before the stress image build for Fedora VMs. Which lead me to try different things (to no avail):
- run the build with "USE_PODMAN=true"
- run the build with "USE_DOCKER=true"
- run the build with "USE_DOCKER=true and USE_PODMAN=false"

I've also tried to force the use of the same stress image name than in the working build, without using podman, but that failed too.

I will continue to investigate, but any help is appreciated.

PR #5551, opened 2023-03-21
cri-o: explicitly setup docker local registry

fidencio:
Recently we've changed the way we build our components in the CI, and with that change we've been forcing the usage of docker (instead of podman) with CRI-O.

This, however, introduced a regression and podman local registries are transparently understood by CRI-O, but the docker one must be explicitly configured.

Fixes: #6510

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -91,6 +91,13 @@ unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.
   "debian" = "docker.io/library/debian"
   # Oracle Linux
   "oraclelinux" = "container-registry.oracle.com/os/oraclelinux"
+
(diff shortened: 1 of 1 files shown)


littlejawa (approved):
Couldn't see it run with the cri-o CI job, as it's not running from this repository.
The rationale and the fix looks good to me though.

littlejawa:
For the record: this is for fixing https://github.com/kata-containers/tests/issues/5544

gkurz (approved):
I don't know if this will address issues reported by @littlejawa in #5544 but it won't make things worse. Maybe just document why we're doing that with a comment and fix the issue number in the changelog ?

LGTM. Thanks @fidencio !

amshinde (approved):

PR #5556, opened 2023-03-23
ci: do not stop containerd when testing with crio

littlejawa:
When starting the cri service, we need to stop/restart it.

We used to stop containerd, assuming it was the one being started anyway, but it is an error: when we test with cri-o, stopping containerd also stops docker, which is needed by cri-o to access its image repository.

Fixes: #5544


littlejawa:
cc @fidencio

cmaf (approved):

PR #5571, opened 2023-03-30
Add statistics on the data collected in boot times

### LINKED ISSUE #5447: Add statistics on the data collected in boot times bench
Currently boot times is able to collect results for N-runs. 
In order to reduce noise (outliers) we propose to complement the data results with some statistics on the data.

**How can it be improved?**
Add extra statistic items to the results as such as:

- Mean
- Std-Dev
- CoV

dborquez:
This patch adds useful statistics to help to analyze collected data as the following:
- size: sample size
- avg: average
- sd: standard deviation
- cov: coefficent of variance

Fixes #5447

### DIFF
--- a/metrics/time/launch_times.sh
+++ b/metrics/time/launch_times.sh
@@ -34,6 +34,10 @@ CALCULATE_KERNEL=
 
 REQUIRED_CMDS=("bc" "awk")
 
+# set the total number of decimal digits after the decimal point 
+# for representing the calculations results
+CALC_SCALE=4
+
 # The date command format we use to ensure we capture the ns timings
 # Note the no-0-padding - 0 padding the results breaks bc in some cases
 DATECMD="date -u +%-s:%-N"
@@ -42,6 +46,16 @@ DATECMD="date -u +%-s:%-N"
 # and does not need to have as large a store of entropy anymore as the value
 # of 256 was found to work well with common cryptographic algorithms
 entropy_level="256"
+
+# Grabs the number of iterations performed
+num_iters=0
+
+# The individual results are stored in an array
+declare -a total_result_ds
+declare -a to_workload_ds
+declare -a in_kernel_ds
+declare -a to_kernel_ds
+declare -a to_quit_ds
 
 check_entropy_level() {
 	retries="10"
@@ -68,7 +82,7 @@ sn_to_ns() {
 
 # convert 'nanoseconds' (since epoch) into a 'float' seconds
 ns_to_s() {
-	printf "%.03f" $(bc <<< "scale=3; $1 / 1000000000")
+	printf "%.0${CALC_SCALE}f" $(bc <<< "scale=$CALC_SCALE; $1 / 1000000000")
 }
 
 run_workload() {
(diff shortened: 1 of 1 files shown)


metrics/time/launch_times.sh
```diff
@@ -68,105 +82,124 @@ sn_to_ns() {
...
-
-		kernel_period=$(echo $kernel_last_line | awk '{print $2}' | tr -d "]")
-
-		# And we can then work out how much time it took to get to the kernel
-		to_kernel_period=$(printf "%0f" $(bc <<<"scale=3; $(ns_to_s $workload_period) - $kernel_period"))
-	else
-		kernel_period="0.0"
-		to_kernel_period="0.0"
-	fi
+    start_time=$($DATECMD)
```
GabyCT, line 89: @dborquez seems like you need to fix the indentation

GabyCT:
@dborquez the script lgtm, however, you also need to modify the values for the limits so it can use the standard deviation, the average for the kata metrics CI

metrics/time/launch_times.sh
```diff
@@ -197,6 +230,125 @@ init () {
...
+    avg=0
+
+    [ -z "$result_arr" ] && die "List of results was not passed to the calc_avg_array() function when trying to calculate the average result".
+
+    sum=$(IFS='+'; echo "scale=4; ${result_arr[*]}" | bc)
+    avg=$(echo "scale=$CALC_SCALE; $sum / $num_iters" | bc)
+    printf "%.0${CALC_SCALE}f" $avg
+}
+
+# Computes the stadard deviation of the data
```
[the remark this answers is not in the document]
dborquez: @GabyCT thanks, I will fix it

metrics/time/launch_times.sh
```diff
@@ -68,105 +82,124 @@ sn_to_ns() {
...
+        # And we can then work out how much time it took to get to the kernel
+        to_kernel_period=$(printf "%0f" $(bc <<<"scale=$CALC_SCALE; $(ns_to_s $workload_period) - $kernel_period"))
+    else
+        kernel_period="0.0"
+        to_kernel_period="0.0"
+    fi
+
+    total_result="$(ns_to_s $total_period)"
+    total_to_workload="$(ns_to_s $workload_period)"
+    total_in_kernel=$kernel_period
```
GabyCT, line 151: @dborquez for the kernel do we need the ns_to_s or not?

metrics/time/launch_times.sh
```diff
@@ -197,6 +230,125 @@ init () {
...
+    sum=$(IFS='+'; echo "scale=4; ${result_arr[*]}" | bc)
+    avg=$(echo "scale=$CALC_SCALE; $sum / $num_iters" | bc)
+    printf "%.0${CALC_SCALE}f" $avg
+}
+
+# Computes the stadard deviation of the data
+calc_sd_array() {
+    data=("$@")
+    sum_sqr_n=0
+    LSCALE=13
```
GabyCT, line 249: @dborquez why we have this value?

metrics/time/launch_times.sh
dborquez: @GabyCT the test is computing the `total_in_kernel `metric without a conversion from ns to s because it takes that value from the dmesg last line: [launch-times.sh](https://github.com/kata-containers/tests/blob/c148ff53096f9f314488017c1d68a8f684ba3e48/metrics/time/launch_times.sh#L147)

metrics/time/launch_times.sh
dborquez: In the middle of calculating a statistic, we should avoid trimming or rounding numbers because the fractions are very small,
but once we have a result, we can limit the number of decimal digits.

metrics/time/launch_times.sh
dborquez: @thank you @GabyCT, now fixed

GabyCT (approved):

chavafg (approved):

Issue #5583, opened 2023-04-03
Suppot resource URI in SEV tests


fitzthum:
The SEV tests for encrypted images need to be updated to work with the resource URI changes. 

This is part of https://github.com/kata-containers/tests/issues/5521

PR #5579, opened 2023-03-31
Signatures: add resources with new URI scheme; PART 2

jepio:
Some extra changes on top and the following PRs:
- #5577
- #5582 
- #5584

PR #5584, opened 2023-04-03
SEV: Make SEV tests work with resource URI

fitzthum:
Switch to container image with resource URI format and modify resource names to use new URI format.

Fixes: #5583


fitzthum:
These changes were merged as part of #5579.

Issue #5622, opened 2023-04-29
Remove `container_restart_test.go.patch` included in cri integration tests


amshinde:
PR [6751](https://github.com/kata-containers/kata-containers/pull/6751) bumps the containerd version to 1.7.0.
CI is failing with error:
```
+ bin/cri-integration.test
# github.com/containerd/containerd/integration [github.com/containerd/containerd/integration.test]
integration/container_restart_test.go:28:3: undefined: pauseImage
make[1]: *** [Makefile:220: bin/cri-integration.test] Error 2
```

This error comes from a test that we have included in this repo that we copy over to the containerd integration test directory for running it along with other integration tests:
https://github.com/kata-containers/tests/blob/a1ddca8eff7d80b25f9feabe96fbb61d2ed06b97/integration/containerd/cri/container_restart_test.go.patch

Inspecting the above error for `undefined: pauseImage`, I found that the container restart test uses the variable `pauseImage` that is no longer defined in the containerd integration tests package. Instead the pause image is now exposed using another api. 
I was going to port `container_restart_test.go` to use the new API, but saw that this test has long been merged into the containerd repo along with further enhancements been made to it.  We should be using that instead and delete the test that we have in our repo.

PR #5623, opened 2023-04-29
cri: Remove container_restart_test.go.patch

amshinde:
See issue #5622 for detailed explanation.
This test has long been upstreamed in the containerd repo. More-over this test no longer works for newer container versions(1.7.0) due to changes made upstream.
We should make use of the upstream test.

Fixes: #5622


cmaf (approved):
thanks @amshinde

fengwang666 (approved):
LGTM. Thanks!

PR #5625, opened 2023-05-05
static-checks: Switch to GET for URL checks

jepio:
Some endpoints block HEAD requests while allowing GET. To prevent false positives from url check, switch to using GET, but still only read the header.

Fixes: #6627


fengwang666 (approved):

fidencio (approved):
lgtm, thanks @jepio!

PR #5626, opened 2023-05-08
ci: Remove uses of procenv

jepio:
Procenv has been regularly hanging in the teardown of vfio test, possibly since upgrading to fedora 37. The procenv output does not appear to be consumed or useful to anyone, so remove all traces of it.

Fixes: kata-containers/kata-containers#6790

### DIFF
--- a/.ci/setup_env_alinux.sh
+++ b/.ci/setup_env_alinux.sh
@@ -66,7 +66,6 @@ declare -A packages=( \
 	[ostree]="ostree-devel" \
 	[metrics_dependencies]="bc jq" \
 	[crudini]="crudini" \
-	[procenv]="procenv" \
 	[haveged]="haveged" \
 	[libsystemd]="systemd-devel" \
 	[redis]="redis" \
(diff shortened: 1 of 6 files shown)


fidencio (approved):
lgtm, thanks @jepio!

fidencio (approved):
lgtm, thanks @jepio!

jodh-intel (approved):
Hi @jepio - Ironically, I added `procenv` to help debug CI failures. In fact, that's why I wrote the tool in the first place! :smile:

It would be useful to understand why it is hanging. I wasn't aware of this problem and cannot reproduce locally on F37. However, since this seems to be causing issues, if you don't have time to debug further, I'm happy for us to remove the tool.

jepio:
> It would be useful to understand why it is hanging. I wasn't aware of this problem and cannot reproduce locally on F37. However, since this seems to be causing issues, if you don't have time to debug further, I'm happy for us to remove the tool.

I see the irony and I am also puzzled as to why it is hanging. I tried to run the Fedora VM test locally and also can't reproduce.

wainersm (approved):
I also agree that ideally we should be debugging and fixing the issue with procenv, but on the other hand it seems everybody lacks time and this issue has blocked PRs. So I think that merge this is the best thing we can do at the moment.

PR #5634, opened 2023-05-10
static-checks: Enable cookie store for curl

jepio:
Some websites <cough>twitter</cough> refuse to return any data without the request including a cookie. Activate cookie processing in curl so that we can retrieve the data and finally get url checks to work.

Fixes: #6627


dborquez (approved):
LGTM, thanks @jepio

fidencio (approved):
lgtm, thanks @jepio!

Issue #5643, opened 2023-05-17
tests: Stop running `make kubernetes` for containerd jobs


fidencio:
Those tests are now covered as part of the GitHub actions we have, and
currently we're duplicating those (and also spending the resources
twice).

Let's stop running those for the cases we already cover with GitHub
actions, and update those little by little as we progress on the GHA
effort.

PR #5644, opened 2023-05-17
tests: Sstop running `make kubernetes` with containerd

fidencio:
Those tests are now covered as part of the GitHub actions we have, and currently we're duplicating those (and also spending the resources twice).

Let's stop running those for the cases we already cover with GitHub actions, and update those little by little as we progress on the GHA effort.

Fixes: #5643


GabyCT (approved):

stevenhorsman (approved):
LGTM. Thanks

PR #5653, opened 2023-05-22
ci: Fetch k8s gpg key from dl.k8s.io

jepio:
The gpg key hosting at packages.cloud.google.com is known to fail regularly, and it is recommended[1] to use a different mirror for the gpg key. Switch to fetching apt gpg key from there, to reduce test failures.


fidencio (approved):
lgtm, thanks @jepio!

fidencio (approved):
lgtm, thanks @jepio!

PR #5681, opened 2023-06-06
ci: Switch to http kubernetes mirror

jepio:
SEV tests regularly fail with this error message:

  Err:5 https://packages.cloud.google.com/apt kubernetes-xenial-unstable Release
    Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: 142.250.189.174 443]

It is likely, that the issue is caused by a TLS MITM proxy, that does network scanning. If we switch to the http version of the mirror, we might have more luck. This is safe to do because the mirror is verified using gpg signatures.

Fixes: #7039


stevenhorsman (approved):
LGTM. Thanks @jepio!

cmaf (approved):

GabyCT (approved):

Issue #5694, opened 2023-06-23
memory footprint inside container is failing


dborquez:
The test collects different memory metrics on each run, for a variable number of runs.
The test computes the average of the results but when one or more of the runs fails, the calculated average dosn't discart that failed result, and generates an incorrect avg value which causes the CI metrics to fail.

CI metrics error example:
http://jenkins.katacontainers.io/job/kata-containers-2.0-metrics-ubuntu-20-04-PR/4869/console

PR #5695, opened 2023-06-23
metrics: Fix non valid results on memory usage inside container

dborquez:
This PR make the metrics test repeats failed runs,
up to three times.
Removes csv output format, as it is not needed.

Fixes: #5694

### DIFF
--- a/.ci/run_metrics_PR_ci.sh
+++ b/.ci/run_metrics_PR_ci.sh
@@ -34,10 +34,7 @@ source "${SCRIPT_DIR}/../metrics/lib/common.bash"
 
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 
-# Density paramter to set the number of iterations of the test
-DENSITY_FORMAT_RESULTS="${DENSITY_FORMAT_RESULTS:-json}"
-
-# Density parameters to select between 'csv' and 'json' results format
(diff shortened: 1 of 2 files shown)


GabyCT (approved):

.ci/run_metrics_PR_ci.sh
```diff
@@ -34,10 +34,7 @@ source "${SCRIPT_DIR}/../metrics/lib/common.bash"
 
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 
-# Density paramter to set the number of iterations of the test
-DENSITY_FORMAT_RESULTS="${DENSITY_FORMAT_RESULTS:-json}"
-
-# Density parameters to select between 'csv' and 'json' results format
+# Num of repetitions for the 'memory inside container' metrics check.
 DENSITY_TEST_REPETITIONS="${DENSITY_TEST_REPETITIONS:-1}"
```
chavafg, line 38: not exactly related to this change, but realized that you are expecting `.ci/run_metrics_PR.sh` to be called with one argument, but actually it is being called with no arguments, check: https://github.com/kata-containers/tests/blob/main/.ci/run.sh#L136

I think that as this is a variable needed for a specific test (and this scripts launches several tests), it would be better to not get the value from an argument of this script.

fidencio (approved):
lgtm, thanks @dborquez

Issue #5724, opened 2023-07-19
license: avoid checking license for *.dtd file


studychao:
Since .dtd is used in Dragonball for aarch64 unit test and .dtd is a binary file so it could not be added apache 2.0 license line of code.

So we should avoid checking license for *.dtd file in our CI.

PR #5725, opened 2023-07-19
license: avoid checking license for *.dtd file

studychao:
Since .dtd is used in Dragonball for aarch64 unit test and .dtd is a binary file so it could not be added apache 2.0 license line of code.

So we should avoid checking license for *.dtd file in our CI.

fixes: #5724


fidencio (approved):
lgtm, thanks @studychao!

quanweiZhou (approved):

Issue #5726, opened 2023-07-21
vfio: Fix default port assignment with an explicit value in the TOML


zvonkok:
Fix default port assignment with an explicit value in the TOML

It was implicit before and now the runtime will bail out if we did not set it explicitly.
Update the configuration and set it. This will make the intent more clear.

PR #5727, opened 2023-07-21
vfio: Fix default port assignment

zvonkok:
Fix default port assignment with an explicit value in the TOML It was implicit before and now the runtime will bail out if we did not set it explicitly. Update the configuration and set it. This will make the intent more clear.

Fixes: #5726


BbolroC (approved):
LGTM, thanks @zvonkok !

fidencio (approved):
lgtm, assuming green tests!

PR #5729, opened 2023-07-21
vfio: Extend the current VFIO test for GPUs

zvonkok:
The current VFIO test only works with a virtualized network device.
Extend the VFIO tests so that "any" device that is passed-through can be tested.


zvonkok:
Added cold_plug_vfio test besides hot_plug_vfio test.

BbolroC (approved):
LGTM, thanks @zvonkok !

functional/vfio/run.sh
```diff
@@ -189,12 +210,23 @@ setup_configuration_file() {
...
 			# Make sure we have set hot_plug_vfio to a reasonable value
-			sudo sed -i -e 's|^#hot_plug_vfio =.*$|hot_plug_vfio = "bridge-port"|' -e 's|^hot_plug_vfio = .*$|hot_plug_vfio = "bridge-port"|' "${kata_config_file}"
+			sed -i -e 's|^#hot_plug_vfio =.*$|hot_plug_vfio = "bridge-port"|' -e 's|^hot_plug_vfio =.*$|hot_plug_vfio = "bridge-port"|' "${kata_config_file}"
 		else
 			warn "Variable machine_type only applies to qemu. It will be ignored"
 		fi
 	fi
 
+        if [ "${VFIO_HOTPLUG}" != "bridge-port" ]; then
+		sed -i -e "s|^#hot_plug_vfio =.*$|hot_plug_vfio = \"${VFIO_HOTPLUG}\"|"  -e "s|^hot_plug_vfio =.*$|hot_plug_vfio = \"${VFIO_HOTPLUG}\"|" "${kata_config_file}"
```
jepio, line 220: Would this work?
```suggestion
		sed -i -e "s|^#?hot_plug_vfio =.*$|hot_plug_vfio = \"${VFIO_HOTPLUG}\"|" "${kata_config_file}"
```

functional/vfio/run.sh
```diff
@@ -189,12 +210,23 @@ setup_configuration_file() {
...
+                cat "${kata_config_file}" | grep -v '#' | grep -v '^$'
+        fi
+
+        if [ "${VFIO_COLDPLUG}" != "bridge-port" ]; then
+		sed -i -e "s|^#cold_plug_vfio =.*$|cold_plug_vfio = \"${VFIO_COLDPLUG}\"|"  -e "s|^cold_plug_vfio =.*$|cold_plug_vfio = \"${VFIO_COLDPLUG}\"|" "${kata_config_file}"
```
jepio, line 225: Would this work?

```suggestion
		sed -i -e "s|^#?cold_plug_vfio =.*$|cold_plug_vfio = \"${VFIO_COLDPLUG}\"|" "${kata_config_file}"
```

functional/vfio/run.sh
```diff
@@ -189,12 +210,23 @@ setup_configuration_file() {
...
-			sudo sed -i -e 's|^#hot_plug_vfio =.*$|hot_plug_vfio = "bridge-port"|' -e 's|^hot_plug_vfio = .*$|hot_plug_vfio = "bridge-port"|' "${kata_config_file}"
+			sed -i -e 's|^#hot_plug_vfio =.*$|hot_plug_vfio = "bridge-port"|' -e 's|^hot_plug_vfio =.*$|hot_plug_vfio = "bridge-port"|' "${kata_config_file}"
 		else
 			warn "Variable machine_type only applies to qemu. It will be ignored"
 		fi
 	fi
 
+        if [ "${VFIO_HOTPLUG}" != "bridge-port" ]; then
+		sed -i -e "s|^#hot_plug_vfio =.*$|hot_plug_vfio = \"${VFIO_HOTPLUG}\"|"  -e "s|^hot_plug_vfio =.*$|hot_plug_vfio = \"${VFIO_HOTPLUG}\"|" "${kata_config_file}"
+                cat "${kata_config_file}" | grep -v '#' | grep -v '^$'
```
jepio, line 221: can we get rid of the print here and on line 226 and leave just the one at the end of `setup_configuration_file`?

functional/vfio/run.sh
```diff
@@ -25,20 +25,34 @@ HYPERVISOR=
 MACHINE_TYPE=
 IMAGE_TYPE=
 
+# Option to choose an alternative PCI device for the VFIO test
+VFIO_PCI_CLASS=${VFIO_PCI_CLASS:-"Ethernet controller"}
+VFIO_PCI_NAME=${VFIO_PCI_NAME:-"Virtio.*network device"}
+VFIO_CHECK_GUEST_KERNEL=${VFIO_CHECK_GUEST_KERNEL:-"ip a | grep \"eth\" || die \"Missing VFIO network interface\""}
+VFIO_HOTPLUG=${VFIO_HOTPLUG:-"bridge-port"}
+VFIO_COLDPLUG=${VFIO_COLDPLUG:-"bridge-port"}
```
jepio, line 33: Won't this result in us always configuring both `hot_plug_vfio` and `cold_plug_vfio`? What will be the result of that?

functional/vfio/run.sh
```diff
@@ -288,7 +322,7 @@ main() {
 	#
 	# Get the device ready on the host
 	#
-	setup_configuration_file
+	#setup_configuration_file
```
jepio, line 325: remove this line

functional/vfio/run.sh
zvonkok: Yes.

functional/vfio/run.sh
zvonkok: Should work if we have a space or not between `#` and the keyword.

functional/vfio/run.sh
zvonkok: Should work if we have a space or not between # and the keyword.

functional/vfio/run.sh
zvonkok: The shim will fail saying you've set both which is not working, good catch.

zvonkok: Added the correct defaults PTAL

zvonkok:
@GabyCT PTAL Thanks!

functional/vfio/run.sh
```diff
@@ -189,12 +210,21 @@ setup_configuration_file() {
...
 
+        if [ "${VFIO_HOTPLUG}" != "bridge-port" ]; then
+		sed -i -e "s|^#?hot_plug_vfio =.*$|hot_plug_vfio = \"${VFIO_HOTPLUG}\"|"  -e "s|^hot_plug_vfio =.*$|hot_plug_vfio = \"${VFIO_HOTPLUG}\"|" "${kata_config_file}"
+        fi
+
+        if [ "${VFIO_COLDPLUG}" != "bridge-port" ]; then
+		sed -i -e "s|^#?cold_plug_vfio =.*$|cold_plug_vfio = \"${VFIO_COLDPLUG}\"|"  -e "s|^cold_plug_vfio =.*$|cold_plug_vfio = \"${VFIO_COLDPLUG}\"|" "${kata_config_file}"
+        fi
+
+
```
jepio, line 227: Turns out this pattern
```
^#?hot_plug_vfio =.*$
```
doesn't work unless you pass `-E` to `sed. The coldplug test seems to have been skipped (judging from the jenkins log).

So how about you make it: `#*.*hot_plug_vfio.*`, and you can really collapse both cases:
```bash
$ cat test
#hot_plug_vfio = bla
# hot_plug_vfio = bla
hot_plug_vfio = bla
$ sed -e 's|^#*.*hot_plug_vfio.*|hot_plug_vfio = noport|' test
hot_plug_vfio = noport
hot_plug_vfio = noport
hot_plug_vfio = noport
```

zvonkok: Ack makes sense, given this test file
```
hot_plug_vfio = "bridge-psdfasdfaort"
hot_plug_vfio="blub-blub"
hot_plug_vfio= "blbe"
hot_plug_vfio ="blfasdf"

# hot_plug_vfio = "bridge-psdfasdfaort"
# hot_plug_vfio="blub-blub"
# hot_plug_vfio= "blbe"
# hot_plug_vfio ="blfasdf"


#hot_plug_vfio = "bridge-psdfasdfaort"
#hot_plug_vfio="blub-blub"
#hot_plug_vfio= "blbe"
#hot_plug_vfio ="blfasdf"
```
and 
```
sed -e 's|^#*.*hot_plug_vfio.*|hot_plug_vfio = "no-port"|'
```
we get 
```
hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"

hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"


hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"
hot_plug_vfio = "no-port"
```

zvonkok:
@jepio if the config variable is not available it will be not set, updating the sed command ...

jepio:
> @jepio if the config variable is not available it will be not set, updating the sed command ...

I think there is some logic error here:
```
        if [ "${VFIO_COLDPLUG}" != "bridge-port" ]; then
```
wouldn't you want this to check against `no-port`? Same for `VFIO_HOTPLUG`.

zvonkok:
@jepio if you look at the config file generated there is no `cold_plug_vfio` variable to update. 

```
10:59:50 hot_plug_vfio = "no-port"
```
The cold-plug test-case is started without the `cold_plug_vfio=....` set.

zvonkok:
Need to update the logic, a VFIO test can have both hot-plug and cold-plug set to the same value. One may want to test 
```
hot_plug_vfio="root-port"
```
first and then 
```
cold_plug_vfio="root-port"
```

zvonkok:
So essentially we only need one VFIO_PLUG setting since we test hot and cold plug in one run.

zvonkok:
Yep, we need an "simple" runtime fix before we can merge this.

zvonkok:
cold-plug vfio_mode=guest-kernel works as expected, 
cold-plug vfio_mode=vfio misses the check for the VFIO control file.

zvonkok:
During testing why it fails for `cold_plug` I found a bug in the `hot_plug` implementation. It depends in which order you supplied the vfio-control-device and the vfio-device-group. 
`1st vfio-control-device 2nd vfio-device-group -> works`
`2st vfio-device-group  2nd vfio-control-group -> breaks`

PR #5755, opened 2023-08-25
tests/kubernetes: Correctly run cleanup_bare_metal_env.sh

jepio:
Correct the quoting so that we run the script correctly. It currently fails with:

bash: .../tests/integration/kubernetes/cleanup_bare_metal_env.sh false: No such file or directory

Fixes: #5525


fidencio (approved):
lgtm, thanks @jepio!

Issue #5767, opened 2023-09-08
CCv0 | tests: And test_common.sh library for confidential Kubernetes functions


stevenhorsman:
In the containerd tests we have https://github.com/kata-containers/tests/blob/CCv0/integration/containerd/confidential/tests_common.sh that is a library of common test functions so we don't need to duplicate them in many places. With the new kubernetes pull_with_snapshotter  tests we need something similar to avoid a lot of copy-paste.

stevenhorsman:
Closed with #5768 merging

PR #5768, opened 2023-09-08
tests: Refactor out common functions

stevenhorsman:
- Refactor out common functions into test_common.sh library. This libarary can be used in other bats test files and is prep for the pull_image with snapshotter tests
- Pattern similar to that used in containerd tests

Fixes: #5767


arronwy (approved):
Thanks @stevenhorsman LGTM!

PR #5784, opened 2023-10-16
C cv0 test workarounds

stevenhorsman:


stevenhorsman:
I don't think we need this any more

PR #5785, opened 2023-10-19
CC | tests: Set io.containerd.cri.runtime-handler annotation

fidencio:
The CRI PullImage does not have a "runtime class" input, and this is being discussed at the kubernetes level on:
* https://github.com/kubernetes/enhancements/pull/4217
* https://github.com/kubernetes/kubernetes/pull/118907

For now, we're forced to pass the specific annotation to containerd.

### DIFF
--- a/integration/kubernetes/confidential/agent_image.bats
+++ b/integration/kubernetes/confidential/agent_image.bats
@@ -164,7 +164,10 @@ setup() {
 
 	assert_pod_fail "${pod_config}"
 
-	assert_logs_contain "containerd" 'failed to resolve reference \\"quay.io/kata-containers/confidential-containers-auth:test\\": pulling from host quay.io failed with status code \[manifests test\]: 401 UNAUTHORIZED'
+	# The error will be different depending on the containerd version being used.
+	# The former assert is the one coming from containerd 1.6.x, while the latter is the cone coming from 1.7.x
+	assert_logs_contain "containerd" 'failed to resolve reference \\"quay.io/kata-containers/confidential-containers-auth:test\\": pulling from host quay.io failed with status code \[manifests test\]: 401 UNAUTHORIZED' || \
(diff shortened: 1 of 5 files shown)


stevenhorsman (approved):
LGTM. Thanks!

integration/kubernetes/confidential/agent_image.bats
```diff
@@ -164,7 +164,10 @@ setup() {
 
 	assert_pod_fail "${pod_config}"
 
-	assert_logs_contain "containerd" 'failed to resolve reference \\"quay.io/kata-containers/confidential-containers-auth:test\\": pulling from host quay.io failed with status code \[manifests test\]: 401 UNAUTHORIZED'
+	# The error will be different depending on the containerd version being used.
+	# The former assert is the one coming from containerd 1.6.x, while the latter is the cone coming from 1.7.x
```
portersrc, line 168: nit: "one"

portersrc, line 168: nit: "one"

integration/kubernetes/confidential/lib.sh
```diff
@@ -67,13 +67,15 @@ retrieve_sandbox_id() {
 
 # Check out the doc repo if required
 checkout_doc_repo_dir() {
-    local doc_repo=github.com/confidential-containers/documentation
+    local doc_repo=github.com/confidential-containers/confidential-containers
```
portersrc, line 70: good catch

portersrc (commented):
lgtm :+1:

integration/kubernetes/confidential/agent_image.bats
fidencio: As you marked it as a nit, I will save some trees and avoid re-running the CIs as they take a at least an hour, thanks for the review though

Issue #5787, opened 2023-10-20
Intermittent nydus error complaining about already existing key


ryansavino:
This error pops up intermittently in the CI (potentially user space) when the pods are launched:

```
Error: failed to create containerd container: create instance 1726: object with key "1726" already exists: unknown
```

This appears to be affecting multiple host types.

PR #5788, opened 2023-10-20
nydus: add var directory cleanup

ryansavino:
Fixes: #5787

### DIFF
--- a/integration/kubernetes/cleanup_env.sh
+++ b/integration/kubernetes/cleanup_env.sh
@@ -37,10 +37,17 @@ main () {
 	[ "${container_engine}" == "docker" ] && restart_docker_service
 	registry_server_teardown
 
+	# Remove created nydus snapshots
+	for i in `sudo ctr -n k8s.io snapshot --snapshotter nydus list | grep -v KEY | cut -d' ' -f1 || true`; do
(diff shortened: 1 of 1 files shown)


integration/kubernetes/cleanup_env.sh
```diff
@@ -41,6 +41,7 @@ main () {
 	sudo systemctl stop "${CRI_RUNTIME}"
 
 	sudo kill -9 $(pidof "containerd-nydus-grpc") || true
+	sudo rm -rf "/var/lib/containerd-nydus" || true
```
fidencio, line 44: @ryansavino, as part of the cleanup, I'd also add:
```
for i in `sudo ctr -n k8s.io snapshot --snapshotter nydus list | grep -v KEY | cut -d' ' -f1`; do sudo ctr -n k8s.io snapshot --snapshotter nydus rm $i; done
```

ryansavino: Thanks! I have added this.

ryansavino: Sure thing. Also, I moved the nydus snapshot cleanup to right before the CRI_RUNTIME stop call. In case this might cause any issues.

fidencio (approved):
lgtm, thanks @ryansavino!

Issue #5790, opened 2023-11-06
ci: test: k8s: agent_image rootfs check


stevenhorsman:
In the kubernetes agent_image test we currently have a check:
```
echo "Check the image was not pulled in the host"
	local pod_id=$(kubectl get pods -o jsonpath='{.items..metadata.name}')
	retrieve_sandbox_id
	rootfs=($(find /run/kata-containers/shared/sandboxes/${sandbox_id}/shared \
		-name rootfs))
	[ ${#rootfs[@]} -eq 1 ]
```
to ensure that the image hasn't been pulled onto the host. The reason that the check is for a single rootfs is that we found that the pause image was always pulled on the host, presumably due to it being needed to create the pod sandbox.

With the introduction of the nydus-snapshotter code we've found that on some systems (SE and TDX) it appears to be in a different location with nydus-snapshotter, so check for 1, or 0. See an issue at https://github.com/kata-containers/tests/issues/5781 to track this.

We don't have time to understand this fully now, so we just want the tests to pass and check that we don't have both the pause and main pod container image pulled

stevenhorsman:
Closed with #5791 merging

Issue #5781, opened 2023-10-12
CC: newly pulled pause image by snapshotter stored in an unexpected location 


BbolroC:
# Description of problem

With a config `IMAGE_OFFLOAD_TO_GUEST=yes` and `FORKED_CONTAINERD=no`, a pod creation under IBM Z SE is sometimes stuck in a `CreateContainerError` state with the following error:

```
Error: failed to create containerd container: create instance 697: object with key "697" already exists: unknown
```

It is a known issue with an upstream containerd `v1.6.8` (https://github.com/kata-containers/tests/pull/5775#issuecomment-1750968483). A quick remedy would be to remove a `pause` image and get the snapshotter to pull the image. But the newly pulled image is stored in an unexpected location (originally `/run/kata-containers/shared/sandboxes/${sandbox_id}/shared` is expected) as follows: 

```
# ls -lah /run/kata-containers/shared/sandboxes/a322d916b5dc547d1dce178d31b13091418793a9675a8aa006fcfecd49f8bbc1/shared
total 16K
drwxr-x--- 3 root root 160 Oct 12 11:04 .
drwx------ 5 root root 100 Oct 12 11:04 ..
-rw-r--r-- 1 root root 103 Oct 12 11:04 a322d916b5dc547d1dce178d31b13091418793a9675a8aa006fcfecd49f8bbc1-e9967091f9448d8a-resolv.conf
-rw-r--r-- 1 root root  11 Oct 12 11:04 efde0bf9b12e2e127bdb007f58e4dfb893d990fc64b8063f9594c1c1753c06ce-44e4e6f3b60b2926-hostname
-rw-r--r-- 1 root root 103 Oct 12 11:04 efde0bf9b12e2e127bdb007f58e4dfb893d990fc64b8063f9594c1c1753c06ce-4c6bb0d5b7fc98ff-resolv.conf
-rw-rw-rw- 1 root root   0 Oct 12 11:04 efde0bf9b12e2e127bdb007f58e4dfb893d990fc64b8063f9594c1c1753c06ce-83476f850307d009-termination-log
-rw-r--r-- 1 root root 205 Oct 12 11:04 efde0bf9b12e2e127bdb007f58e4dfb893d990fc64b8063f9594c1c1753c06ce-844b44105b991bcd-hosts
[...]

PR #5791, opened 2023-11-06
ci: test: k8s: agent_image rootfs check

stevenhorsman:
In the kubernetes agent_image test we currently have a check:
```
echo "Check the image was not pulled in the host"
	local pod_id=$(kubectl get pods -o jsonpath='{.items..metadata.name}')
	retrieve_sandbox_id
	rootfs=($(find /run/kata-containers/shared/sandboxes/${sandbox_id}/shared \
		-name rootfs))
	[ ${#rootfs[@]} -eq 1 ]
```
to ensure that the image hasn't been pulled onto the host. The reason that the check is for a single rootfs is that we found that the pause image was always pulled on the host, presumably due to it being needed to create the pod sandbox.

With the introduction of the nydus-snapshotter code we've found that on some systems (SE and TDX) it appears to be in a different location with nydus-snapshotter, so check for 1, or 0. See an issue at https://github.com/kata-containers/tests/issues/5781 to track this.

We don't have time to understand this fully now, so we just want the tests to pass and check that we don't have both the pause and test pod container image pulled, so set the check to pass if there are 1, or 0 rootfs' found in /run/kata-containers/shared/sandboxes/

Fixes: #5790

### DIFF
--- a/integration/kubernetes/confidential/agent_image.bats
+++ b/integration/kubernetes/confidential/agent_image.bats
@@ -61,10 +61,6 @@ setup() {
 }
 
 @test "$test_tag Test can pull an unencrypted image inside the guest" {
-	# See an issue at https://github.com/kata-containers/tests/issues/5781
-	if [ "${TEE_TYPE}" = "se" ]; then
-		skip "test until the containerd is updated to v1.7 for IBM Z Secure Execution"
-	fi
 	create_test_pod
 
 	echo "Check the image was not pulled in the host"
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @stevenhorsman!

BbolroC (approved):
lgtm, thanks @stevenhorsman !

fidencio:
> /test-tdx

Don't bother about this, I don't think we'll be luck enough to have the artefacts built. :-/
We'll have to test it on the operator side.

PR #5818, opened 2024-05-30
ci: adding yq-shim to support v3 and v4

beraldoleal:
Some tests uses yq installed by this repository but other tests might use yq installed by the kata-containers repository.

We are pushing yq to v4 within the main kata repository and this shim will handle only basic syntax of yq v3 and v4 during the transition period.


beraldoleal:
@wainersm the failed jenkins jobs here are the same as #5819 which is doing nothing. Are those jobs flaky and safe to ignore for this PR?

wainersm:
> @wainersm, maybe I'm missing something... but the failed jenkins jobs here are the same as #5819 which is doing nothing. Are those jobs flaky and safe to ignore for this PR?

I think it's safe; let me ask one more fix: add the `Fixes` in your commit.

wainersm (approved):

